Issue #184, opened 2018-04-18
Log level panel needs a delayed refresh to pickup changes to effective levels


febbraiod:
From ATM: We have an issue I recall from hoist-sencha where you need an additional delayed refresh to pick up the "effective" level...Confusing currently where it seems like your change didn't work, or maybe you don't understand how it works.

lbwexler:
I fixed this in HoistCore -- the log level changes are now synchronous.  This seemed like a more appropriate fix, and was not too hard to make.
https://github.com/exhi/hoist-core/commit/dc387e885bea14b0443

PR #187, opened 2018-04-18
Reload log levels panel 500ms after rec update, Fixes: #184

febbraiod:
Opening this for discussion. I don't think this is the correct solution. This seems like such a specific need for LogLevel panel, that adding this prop to all RestStores is overkill.

After I came to this conclusion I wanted to apply a fix using an event that would fire after saveRecordInternalAsync resolved, but firing/handling such an event turned out to be less straight forward than I imagined. 

Rather than burning a bunch of time trying to make that work, I figured I'd wait to see what you guys might have to say about it.

### DIFF
--- a/admin/tabs/logging/LogLevelPanel.js
+++ b/admin/tabs/logging/LogLevelPanel.js
@@ -16,6 +16,7 @@ export class LogLevelPanel extends Component {
 
     store = new RestStore({
         url: 'rest/logLevelAdmin',
+        reloadOnUpdate: 500,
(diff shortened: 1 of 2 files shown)


lbwexler:
Don -- Thanks for kicking off this discussion, and identifying a real need here in terms of functionality and the need for an event based mechanism or better mobx support for this kind of thing. 

 I am going to go ahead and close this pull request and delete the branch now that we have identified the direction we want to go in.

PR #200, opened 2018-04-24
HoistField observes and reacts to changes to the external value.

### LINKED ISSUE #188: TextField doesn't clear display value when model prop is set to null.
When using a TextField in 'bound' mode, setting the bound property on the model to null does not immediately clear the field's display. Instead, one must focus the TextField before the displayed value is cleared.

Note this is note the case for the other field types I tried (NumberField, DayField) - these fields clear their display immediately upon changing the bound value.

I haven't tested yet if this applies only to setting the bound prop to null, or to setting the bound model property in general.

TomTirapani:
Previously, the external value was only checked for changes on focus. Fixes #188


lbwexler:
👍

PR #202, opened 2018-04-25
Data view

lbwexler:
Creating pull request for review purposes

### DIFF
--- a/admin/tabs/preferences/PreferencePanel.js
+++ b/admin/tabs/preferences/PreferencePanel.js
@@ -40,6 +40,7 @@ export class PreferencePanel extends Component {
             {
                 name: 'local',
                 type: 'bool',
+                defaultValue: false,
                 required: true
             },
             {
@@ -73,7 +74,7 @@ export class PreferencePanel extends Component {
         editors: [
             {field: 'name'},
             {field: 'type'},
-            {field: 'defaultValue'},
+            {field: 'defaultValue', type: 'boolSelect'},
             {field: 'local'},
             {field: 'notes', type: 'textarea'},
             {field: 'lastUpdated'},
--- a/admin/tabs/preferences/UserPreferencePanel.js
+++ b/admin/tabs/preferences/UserPreferencePanel.js
@@ -65,7 +65,7 @@ export class UserPreferencePanel extends Component {
         editors: [
             {field: 'name'},
             {field: 'username'},
-            {field: 'userValue'},
+            {field: 'userValue', type: 'boolSelect'},
             {field: 'lastUpdated'},
             {field: 'lastUpdatedBy'}
         ]
--- a/cmp/form/CheckField.js
+++ b/cmp/form/CheckField.js
@@ -32,7 +32,7 @@ export class CheckField extends HoistField {
         const {text, style} = this.props;
 
         return checkbox({
-            checked: this.renderValue,
+            checked: !!this.renderValue,
             onChange: this.onChange,
             style: {...style, marginBottom: '0px', marginRight: '0px'},
             label: label(text),
--- a/cmp/form/NumberField.js
+++ b/cmp/form/NumberField.js
@@ -48,11 +48,12 @@ export class NumberField extends HoistField {
     }
 
     toExternal(val) {
-        return Number.parseFloat(val);
+        const ret = Number.parseFloat(val);
+        return isFinite(ret) ? ret : null;
     }
 
     toInternal(val) {
-        return val ? val.toString() : '';
+        return val != null ? val.toString() : '';
     }
 }
 export const numberField = elemFactory(NumberField);
--- a/cmp/form/TextAreaField.js
+++ b/cmp/form/TextAreaField.js
@@ -14,10 +14,11 @@ import {HoistField} from './HoistField';
  * A Text Area Field
  *
  * @prop rest, see general properties for HoistField
- * 
- * @prop autoFocus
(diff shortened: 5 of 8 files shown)


lbwexler:
Couple of general comments here:

+ I think we should create a new minimal DataView and DataViewModel, that supports just the API we want to support.  It can *contain* a grid, but the fact that it uses Grid underneath for its implementation should be hidden from applications.   We are only really using the grid for the virtual paging it provides.

+ Seems like sorting *might* be done by just sorting the data directly before passing to grid, rather than adding a bunch of hidden columns.  I guess that is an implementation detail, but it seems to me, the simpler we keep the grid the better.

+ Is there a way to hide the grid headers without doing that clunky override?

+ The way the the rest params are documented seems confusing i.e. '{[]}'.  Is that correct?  Its an object, not an array.

lbwexler:
Hi -- Sorry for the delay in giving these comments.  Happy to discuss in further detail, but I guess my main concern is that it seems like this API is too coupled to Grid/GridModel.  Specifically, weird that the the DataView has to expose a "grid model" as a public property.

I was thinking that the API for DataViewModel would simply reveal the actual  itemFactory/store/sort/  which represent its API.   The DataView could just *contain* a Local GridModel/Grid combination to do its rendering, and sync it as needed to the DataViewModel.

dataView/DataView.js
```diff
@@ -0,0 +1,51 @@
...
+import {baseCol} from 'hoist/columns/Core';
+
+@hoistComponent()
+class DataView extends Component {
+
+    render() {
+        const {store, selection, contextMenuFn, itemFactory} = this.model,
+            {rowCls, rowHeight} = this.props;
+
+        return grid({
```
lbwexler: Can we set this grid model up in the constructor?  We don't expect to handle changes for any of these props?
Props are passed to component constructor and recreating in render seems risky -- if we are ever forced to re-render this component, we throw out all of our state!

dataView/DataView.js
```diff
@@ -0,0 +1,51 @@
...
+                store,
+                selection,
+                contextMenuFn,
+                columns: [
+                    baseCol({
+                        field: 'id',
+                        flex: 1,
+                        cellRendererFramework: (
+                            class extends Component {
+                                render() {
```
lbwexler: Its a pain, but actually think we need to implement a refresh() method here as well on this Ad-hoc component, or we get warnings.    It can just return false.  Search for cellRendererFramework.   

Idea: Maybe we can make this a column enhancement where we automatically create this class if the user just passes a renderFunction with the name renderElement?

dataView/DataViewModel.js
```diff
@@ -0,0 +1,61 @@
...
+
+    // Immutable public properties
+    itemFactory = null;
+    store = null;
+    selection = null;
+    contextMenuFn = null;
+
+    static defaultContextMenu = () => {
+        return new GridContextMenu([
+            'copy',
```
lbwexler: Weird for this thing to be using a "GridContextMenu"

grid/GridModel.js
```diff
@@ -41,6 +39,7 @@ export class GridModel {
     /**
      * @param {BaseStore} store - store containing the data for the grid.
      * @param {Object[]} columns - collection of column specifications.
+     * @param {StoreSelectionModel} selection - optional selection model to use
```
lbwexler: Let's start using the new syntax for optional arguments

dataView/DataViewModel.js
TomTirapani: Yeah, I thought the same - but the GridContextMenu has all built in support for 'requireRecords' etc, so seemed a bit wasteful to make DataViewContextMenu that would be basically be identical

dataView/DataView.js
TomTirapani: Good catch, will do. Also, good idea about renderFunction in the column definition, although this is a slightly different use-case (we are passing the entire record, not the specified column field)

dataView/DataView.js
```diff
@@ -0,0 +1,52 @@
...
+    render() {
+        const {rowCls, rowHeight} = this.props;
+        return grid({
+            model: this._gridModel,
+            gridOptions: {
+                headerHeight: 0,
+                rowClass: rowCls,
+                // Note: Setting a fixed row height is required for now.
+                // Ag-grid 17.1.x will introduce an 'autoHeight' property that will size rows according to content
+                rowHeight: rowHeight
```
amcclain: Hi - we recently took 17.1.0 into develop - does that get us this feature, or is it still on the planning board for them?

TomTirapani: Oh, I must have missed that - let me try

dataView/DataView.js
```diff
@@ -0,0 +1,50 @@
...
+            store,
+            selection,
+            contextMenuFn,
+            columns: [
+                baseCol({
+                    field: 'id',
+                    flex: 1,
+                    autoHeight: true,
+                    renderElement: (props) => {
+                        return itemFactory({record: props.data});
```
lbwexler: use of "props" here for the parameter name is a bit confusing.  This is really the grid record being rendered, not the props of the data view or constructor this code sits in.  

Also, think if we go ahead with this idea of a BaseCol enhancement, the name of the property should be something like  or 'elementRenderer' or 'elementCellRenderer'.   This is really analogous to a cellRenerder, which is an existing API object in AG grid.

PR #204, opened 2018-04-25
Async combo field

TomTirapani:
Added AsyncComboField, a variation on ComboField which populates its items via a `queryAsync` function. The primary use case for this would be querying a remote server and returning a list of options.

Also added the option of overriding the default item renderer for fields which use it.

### DIFF
--- a/cmp/form/dropdown/AsyncComboField.js
+++ b/cmp/form/dropdown/AsyncComboField.js
@@ -9,7 +9,7 @@ import {hoistComponent, elemFactory} from 'hoist/core';
 import {observable, setter} from 'hoist/mobx';
 import {Classes, suggest} from 'hoist/kit/blueprint';
 
-import {HoistField} from './HoistField';
+import {BaseDropdownField} from './BaseDropdownField';
 
 /**
  * ComboBox Field, which populates its items via an async function (e.g. querying a remote server)
@@ -22,7 +22,7 @@ import {HoistField} from './HoistField';
  * @prop itemRenderer, optional custom itemRenderer, a function that receives (item, itemProps)
  */
 @hoistComponent()
-export class AsyncComboField extends HoistField {
+export class AsyncComboField extends BaseDropdownField {
     @observable.ref @setter items = [];
 
     static defaultProps = {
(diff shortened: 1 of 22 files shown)


cmp/form/dropdown/AsyncComboField.js
```diff
@@ -0,0 +1,85 @@
...
+ * @prop rest, see properties for HoistField
+ *
+ * @prop queryAsync, async function that receives (string query)
+ *          Should return a collection of form [{value: string, label: string}, ...] or [val, val, ...]
+ * @prop placeholder, text to display when control is empty
+ * @prop itemRenderer, optional custom itemRenderer, a function that receives (item, itemProps)
+ */
+@hoistComponent()
+export class AsyncComboField extends BaseDropdownField {
+    @observable.ref @setter items = [];
```
lbwexler: Call this options, and normalize before setting, not at render time.

TomTirapani:
Example usage:

```
comboField({
    model: model,
    field: 'value',
    queryAsync: (query) => {
        return XH.fetchJson({
            url: '...',
            params: {query: query}
        }).then(data => {
            return data;
        });
    }
});
```

cmp/form/dropdown/AsyncComboField.js
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @prop rest, see properties for HoistField
+ *
+ * @prop queryAsync, async function that receives (string query)
+ *          Should return a collection of form [{value: string, label: string}, ...] or [val, val, ...]
+ * @prop placeholder, text to display when control is empty
+ * @prop itemRenderer, optional custom itemRenderer, a function that receives (item, itemProps)
+ */
+@hoistComponent()
+export class AsyncComboField extends BaseDropdownField {
```
lbwexler: Not sure about the name of this component.    Its not just that its one time async loading, but it dynamically filters on every keystroke.

cmp/form/dropdown/AsyncComboField.js
```diff
@@ -0,0 +1,85 @@
...
+
+    static defaultProps = {
+        placeholder: 'Search'
+    }
+
+    delegateProps = ['className', 'style', 'placeholder', 'disabled'];
+
+    constructor(props) {
+        super(props);
+        this.addAutoRun(() => this.onQueryChange(), {delay: 100});
```
lbwexler: should the delay be configurable?

cmp/form/dropdown/AsyncComboField.js
```diff
@@ -0,0 +1,85 @@
...
+        const {style, width, itemRenderer, disabled} = this.props;
+
+        const options = this.normalizeOptions(this.items),
+            value = this.renderValue;
+
+        return suggest({
+            popoverProps: {popoverClassName: Classes.MINIMAL},
+            $items: options,
+            onItemSelect: this.onItemSelect,
+            itemRenderer: itemRenderer || this.defaultItemRenderer,
```
lbwexler: reading an overridden itemRenderer from props can probably be built into the shared getter

PR #212, opened 2018-04-26
Make JsonField configurable in parallel with codemirror

febbraiod:


cmp/form/JsonField.js
```diff
@@ -27,17 +27,27 @@ import './JsonField.css';
...
  * @prop rest, see general properties for HoistField
+ *      Accepts props documented below as well as any properties supported by the CodeMirror api,
+ *      with the exception of 'value', which is managed by HoistField.
+ *      Provided CodeMirror props will override any editor properties preconfigured here.
+ *      See: https://codemirror.net/doc/manual.html#api_configuration for details.
  *
  * @prop width, width of field, in pixels
  * @prop height, width of field, in pixels
- * @prop lineWrapping, Whether field should scroll or wrap for long lines. Defaults to false
+ * @prop disabled, Whether the field is readonly, Defaults to false
```
febbraiod: This prop is technically documented in HoistField, but we are pulling it out and setting on the codemirror editor's readOnly prop so I thought it might be clearer to include here.

cmp/form/JsonField.js
```diff
@@ -27,17 +27,27 @@ import './JsonField.css';
...
  */
 @hoistComponent()
 export class JsonField extends HoistField {
 
     editor = null;
     taCmp = null;
 
+    static defaultProps = {
+        width: null,
```
febbraiod: Looking at the code mirror source it appears that an undefined height or width will be handled the same as null in the setSize method, but in the docs it suggests null. Including here allows us to follow the docs, be more self documenting, and avoids having to use 'let' in the implementation to allow conversation as we were doing before.

https://github.com/codemirror/CodeMirror/blob/2c741bd6742678f67e812307b2f78db4f4566d9d/src/edit/methods.js#L392
cm.setSize(width: number|string, height: number|string)
Programmatically set the size of the editor (overriding the applicable CSS rules). width and height can be either numbers (interpreted as pixels) or CSS units ("100%", for example). You can pass null for either of them to indicate that that dimension should not be changed.

lbwexler:
Hi: 

So I think I had in mind a distinct bucket to hold all of the code mirror props.  

The ticket mentions modeling this after Grid and how it passes options to AG Grid (via 'gridOptions').    Could we analogously pass the code mirror options here in an "editorOptions" bucket?  This would avoid the pretty difficult job of teasing them out of the already large list of properties on JsonField.

Assuming the replacement of @prop to proptypes under way in another branch.  If not, it could happen here.

cmp/form/JsonField.js
```diff
@@ -57,10 +66,11 @@ export class JsonField extends HoistField {
...
 
     createJsonEditor(taCmp) {
+        const {height, width, disabled, value, onChange, onCommit, model, field, style, className, ...editorProps} = this.props;
+
         const editorSpec = {
             theme: this.darkTheme ? 'dracula' : 'default',
             mode: 'application/json',
-            lineWrapping: this.props.lineWrapping || false,
             lineNumbers: true,
             autoCloseBrackets: true,
```
lbwexler: Shouldn't lineNumbers and autoCloseBrackets also go with ...editorProps?

cmp/form/JsonField.js
```diff
@@ -22,19 +23,49 @@ import 'codemirror/addon/lint/lint.js';
...
+        /** height of field, in pixels */
+        height: PT.number,
+        /**
+         * Configuration object with any properties supported by the CodeMirror api.
+         * See: https://codemirror.net/doc/manual.html#api_configuration for details.
+         * */
+        editorProps: PT.object
+    };
+
+    static defaultProps = {
```
febbraiod: I think we have good reason to use defaultProps here. As mentioned in the previous PR:

Looking at the code mirror source it appears that an undefined height or width will be handled the same as null in the setSize method, but in the docs it suggests null. Including as defaultProps allows us to follow the docs, be more self documenting, and avoids having to use 'let' in the implementation to allow conversation as we were doing before.

https://github.com/codemirror/CodeMirror/blob/2c741bd6742678f67e812307b2f78db4f4566d9d/src/edit/methods.js#L392

cm.setSize(width: number|string, height: number|string)
Programmatically set the size of the editor (overriding the applicable CSS rules). width and height can be either numbers (interpreted as pixels) or CSS units ("100%", for example). You can pass null for either of them to indicate that that dimension should not be changed.

cmp/form/JsonField.js
```diff
@@ -57,41 +88,25 @@ export class JsonField extends HoistField {
...
+        const taDom = ReactDOM.findDOMNode(taCmp),
             editor = codemirror.fromTextArea(taDom, editorSpec);
 
         editor.on('change', this.handleEditorChange);
         editor.on('focus',  this.onFocus);
         editor.on('blur',  this.onBlur);
         editor.on('keyup',  this.onKeyUp);
 
-        let {height, width} = props;
-        if (!(isUndefined(height) && isUndefined(width))) {
```
febbraiod: defaultProps also allows us to not do this business. Unless there is a reason to jump through these hoops, the new code is much easier to read. (But perhaps we WANT undefined sometimes, or perhaps we want to set size when one is null and the other is undefined? Or both are null? Perhaps Colin could shed some light here?).

PR #216, opened 2018-04-26
PropType check for menu item should be 'oneOfType'

febbraiod:
Saw the context menu throwing when testing an unrelated change.

We were using oneOf but:

"PropTypes.oneOf is a simple enum of literal values. It means that if you’re going to pass a value to this component, it better be one of these exact values."

Came across this:
https://github.com/facebook/react/issues/1919
https://jaketrent.com/post/react-oneof-vs-oneoftype/


amcclain:
Good catch thank you.  PropTypes seem like they will hopefully be helpful and worth it, but clearly we need to be using them correctly!

PR #219, opened 2018-04-27
Drop years from visits chart labels, Fixes #217

### LINKED ISSUE #217: Drop year from Admin > Activity > Daily Visitors chart
Given that common usage will be for very recent past - and that if date filters are adjusted it's pretty clear what years are involved - think we can drop the year and avoid the wrapping.

screen shot 2018-04-26 at 9 11 34 pm

febbraiod:
Not 100% sure about this. First ways I tried look like Month-Year to me, not necessarily Month-Day (See below).

I went with D-MMM so: 8 May, 18 Jul, 28 Aug...ect. is this too exotic for the US?

screen shot 2018-04-27 at 9 42 49 am
screen shot 2018-04-27 at 9 46 35 am


amcclain:
Looks good (much better) to me

PR #220, opened 2018-04-27
spellCheck attr support added to text field component

cnrudd:


cmp/form/TextField.js
```diff
@@ -15,15 +15,16 @@ import {HoistField} from './HoistField';
...
  *
- * @prop autoFocus
- * @prop type, 'text' or 'password'
- * @prop placeholder, text to display when control is empty
- * @prop width, width of field, in pixels
+ * @prop {boolean} [autoFocus=false]
+ * @prop {string } [type] - 'text' or 'password'
+ * @prop {string} [placeholder] - text to display when control is empty
+ * @prop {number} [width] - width of field, in pixels
+ * @prop {boolean} [spellCheck=true]
```
lbwexler: False makes much more sense to me as a default here.  I don't know where I have ever wanted spell checking in any hoist application until now.

cnrudd: After discussing with Lee, I'll make the default be:
spellcheck = false for text fields, and true for textareas fields.

And we'll see how that feels.

amcclain:
Hi - why the asymmetry, where one control has the attribute in its delegate props and the other does not?

PR #222, opened 2018-04-30
Grid column chooser

johnbc:
1st pass on gridColumnChooser


amcclain:
Great thanks for opening this up - eager to take a look today and will do so.

amcclain:
Hi again - took a quick look to provide any immediate feedback I can - discussed with Lee prior and he will also take a look as soon as he has a chance as we're eager to get this in.

My first and primary thought is that we probably want the ColumnChooserModel to interact with the GridModel and use that to read / manipulate columns. The GridModel is our class to hold and process any state we want to define/load/adjust, then it gets flushed into the grid itself. That should avoid the chooser model needing to use those nested calls to locate various parts of the native ag-grid API, and also give us more options / hooks to ensure it can provide whatever data something like the chooser needs in a nicely normalized form.  Does that make sense?

Regarding the column properties, we should be to use our [fileColFactory](https://github.com/exhi/hoist-react/blob/develop/columns/Utils.js) method to default and document any column props the framework expects. We have it as a (not yet documented but soon) standard that we run all column definitions through this method, even if only via `baseCol()` so these defaults are applied.

An example there would be to ensure headerName gets set.  That happens internally in ag-grid based on the `field` prop, but the resulting value isn't then in the model (and also doesn't seem to be easily accessible via the grid API).  We could default it within `fileColFactory` to ensure it's always set on the definition (suggest using `lodash.startCase()`).  

Minor note - for the extended properties we are adding like `excludeFromChooser`, I am not sure if we should preface them with `xh` as we did in Sencha world to indicate "this is not something you will find in the ag-grid docs" or if we want to let them go unprefixed but document here.  We can discuss all together.

Finally re. launching from the default context menu - that's great, would probably default `enableColumnChooser` to true. I think we will also want another convenience button of some kind that can be linked to a given grid and show its chooser so we can place in more visible locations like a toolbar.

With that will give you / Lee a chance to review and see what you think - especially re. the models - and happy to talk more anytime.

amcclain:
Sorry - know I said I would leave it but also just thinking - we should think about / account for the chooser model picking up changes to the columns that it didn't make itself.  Right now I believe it gets initialized once along with the grid and then assumes its state stays in sync because it's the only thing changing columns.

If it's instead bound to the GridModel it can be reactive to the observable columns there.

johnbc:
re interacting with through  the GridModel. Completely agree.

fielColFactory etc. Great.

Re. the xh prefix. I think it is extraneous.  I don;t think we want to distinguish between property names that happen to match those of the underlying implementation, and those which do not. Maybe I am missing something?

Default to enabledColumnChooser: true -- sure i'll just have to be carefull that LRChooser's grids disable it :)

I see what I can do towards a pre-packaged button.

amcclain:
Ah good point re. the default - I think you are correct that it's likely best to require opt-in given all the contexts in which grids can be used (and all cols should always be shown, etc).  👍 

Also am happy to proceed without tacking on more xh's everywhere. I think we are taking sufficient control of these APIs across the board that you kinda need to know that things might or might not map directly onto the underlying libraries.

johnbc:
I am not longer so sure about  default col def properties. It seems that these get send directly to the agGrid, which warns about properties it is not familiar with. This seems like a nice thing for it to do, but now it feels our model isn't align with the agGridl model.

It would be nice if we had a way on the LeftRightChooser to set the headings of the left and right sides.  Right now we have fairly generic headings, but would be nice override for specific uses cases. I also didn't see a way to change the 'No records found...' message which appears when one of the L/R pickers is empty.

amcclain:
Hi - worked on this a bit more to handle the custom props we want for the chooser (and other things) a bit differently so we can keep them around in the column config but avoid passing them to ag-grid.  Also shortened the name up, added a convenience button component (see new gridColumnChooser branch in toolbox) and a few other tweaks.  

I think we can / should merge this so we can get it in there, look at saving state, and keep iterating as we go along.  Happy to review / discuss anything tomorrow!

PR #233, opened 2018-05-01
When typeField changes, clear dependent fields' values. Fixes #225

### LINKED ISSUE #225: RestForm should clear values when datatype changes in "add" mode

febbraiod:


rest/RestControl.js
```diff
@@ -175,6 +176,12 @@ export class RestControl extends Component {
         const model = this.model;
         return !model.value && ['lastUpdatedBy', 'lastUpdated'].includes(model.field.name);
     }
+
+    onTypeFieldChange = () => {
+        const model = this.model;
+        model.parent.clearTypeFields(model.field.name);
```
febbraiod: Is this odd? I think that the restFormModel is the right place to do the heavy lifting here. However, on one hand, I think it might be weird to have this cmp reach through to call the clearTypeFields method on it's model's parent.

On the other hand, this operation is very specific to this control that is a typeField. Do we want to have a method on the RestControlModel that is only ever called by this field?

PR #235, opened 2018-05-01
Public method for setting field values en masse via RestFormModel, Fixes #225

### LINKED ISSUE #225: RestForm should clear values when datatype changes in "add" mode

febbraiod:
+ Allow rest editor to pass onChange handler
+ RestForm clears values when datatype changes in "add" mode


lbwexler:
Don and I just reviewed a more surgical solution to this within RestFormModel.  Closing.

PR #236, opened 2018-05-01
track service updates

cnrudd:
Removed exception property from the service.
Added doc, and added support for msg & messages.

In removing the exception property from the service, I was obliged to remove it from the promise's use of track.  The assumption being that we will just not log promise exceptions at all.  That may be the wrong assumption.

### DIFF
--- a/admin/tabs/monitor/MonitorResultsModel.js
+++ b/admin/tabs/monitor/MonitorResultsModel.js
@@ -69,7 +69,9 @@ export class MonitorResultsModel {
         return XH
             .fetchJson({url: 'monitorAdmin/forceRunAllMonitors'})
             .then(
-                ToastManager.show({
+                // Wrapping the toast call in a function causes it to run only
+                // if the fetchJson was successful.
+                () => ToastManager.show({
                     message: 'Request received. Results will be generated shortly.'
                 })
             )
--- a/admin/tabs/monitor/MonitorResultsModel.js
+++ b/admin/tabs/monitor/MonitorResultsModel.js
@@ -69,8 +69,6 @@ export class MonitorResultsModel {
         return XH
             .fetchJson({url: 'monitorAdmin/forceRunAllMonitors'})
             .then(
-                // Wrapping the toast call in a function causes it to run only
-                // if the fetchJson was successful.
                 () => ToastManager.show({
                     message: 'Request received. Results will be generated shortly.'
                 })
--- a/promise/Promise.js
+++ b/promise/Promise.js
@@ -134,7 +134,6 @@ Object.assign(Promise.prototype, {
         const startTime = Date.now();
 
         return this.catch(e => {
-            trackCfg.exception = e;
             throw e;
         }).finally(() => {
             trackCfg.elapsed = Date.now() - startTime;
--- a/promise/Promise.js
+++ b/promise/Promise.js
@@ -131,11 +131,13 @@ Object.assign(Promise.prototype, {
     track(trackCfg) {
         if (!trackCfg) return this;
 
+        if (typeof trackCfg === 'string') {
+            trackCfg = {msg: trackCfg};
+        }
+
         const startTime = Date.now();
 
-        return this.catch(e => {
-            throw e;
-        }).finally(() => {
+        return this.finally(() => {
             trackCfg.elapsed = Date.now() - startTime;
             XH.track(trackCfg);
         });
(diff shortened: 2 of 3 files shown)


promise/Promise.js
```diff
@@ -134,7 +134,6 @@ Object.assign(Promise.prototype, {
         const startTime = Date.now();
 
         return this.catch(e => {
-            trackCfg.exception = e;
             throw e;
```
amcclain: No need to catch and immediately re-throw here, although this does now raise the interesting point of what the behavior should be when you have a track statement on a promise that throws.

If you have a catchDefault (or other customized error handling that delegates to handleException) you can get a client error report.  But it does seem problematic to track as if nothing went wrong (which is what this will do).

svc/TrackService.js
```diff
@@ -10,8 +10,27 @@ import {stripTags} from 'hoist/utils/HtmlUtils';
 
 export class TrackService extends BaseService {
 
+    /**
```
amcclain: Would also like to have a class-level comment above, even though this single method clearly encapsulates much of the functionality.  Still worth it to take a few sentences and explain what user tracking is in general.  Feel free to take anything useful from hoist-core readme.  See ConfigService.js and PrefService.js for some examples (although those class-level comments are certainly longer than we would need for TrackService).

Part of what's driving this in my mind is that the JS classes might well be the first and primary source for documentation that a non-core Hoist developer will have readily available (in the form of the downloaded package).

svc/TrackService.js
```diff
@@ -10,8 +10,27 @@ import {stripTags} from 'hoist/utils/HtmlUtils';
...
+     * Create a Track Log entry.
+     * Client metadata is set automatically by the server's parsing of request headers.
+     *
+     * @param {Object|string} options - if a string, it will become the msg value.
```
amcclain: The multiple types get wrapped in parens `{(Object|string)}` (yeah...it's a lot of punctuation)

svc/TrackService.js
```diff
@@ -10,8 +10,27 @@ import {stripTags} from 'hoist/utils/HtmlUtils';
...
+     * @param {string} [options.msg] - user-supplied message. Required if options is an object.
+     * @param {string} [options.message] - same as options.msg.  Just supporting msg & message. Use either, not both.
```
amcclain: Would avoid use of "user" here as that implies an end-user (ala client errors, where we do have a user-supplied message). Might suggest "Short description of the activity being tracked."

I would also vote for only including one of these (which is our favorite again? 😄) in the doc string. Could note in that comment "Can be passed as `msg` for backwards compatibility") Then don't surround that one in brackets. (It's technically optional in that you can pass the entire thing a string, but I think it makes sense to document it as a required prop of the object form....)

What happens if this doesn't end up getting passed a message? Does the server throw?

svc/TrackService.js
```diff
@@ -10,8 +10,27 @@ import {stripTags} from 'hoist/utils/HtmlUtils';
...
+     * @param {string} [options.category] - user-supplied category.
+     * @param {Object|Array} [options.data] - user-supplied data collection.
+     * @param {number} [options.elapsed] - time in milliseconds some activity took.
+     * @param {string} [options.severity] - importance flag, such as: OK|WARN|EMERGENCY
+     *                 (errors should be tracked by the ErrorTrackingService, not sent
+     *                 in this TrackService).
```
amcclain: I'm sorry really not trying to drive everyone crazy, but we've been going with the standard where param docs that need to wrap to multiple lines just get indented three tabs (twelve spaces) from the left margin, so they end up under the "m" in param.

That way if you have multiple param comments that wrap they don't end up with different left-edges due to the potentially variable length of the {type} declarations.  And in this case then you can end up with just two lines overall for this param.

svc/TrackService.js
cnrudd: Added class level comment.

svc/TrackService.js
cnrudd: fixed.

svc/TrackService.js
cnrudd: fixed.

admin/tabs/monitor/MonitorResultsModel.js
```diff
@@ -69,7 +69,9 @@ export class MonitorResultsModel {
         return XH
             .fetchJson({url: 'monitorAdmin/forceRunAllMonitors'})
             .then(
-                ToastManager.show({
+                // Wrapping the toast call in a function causes it to run only
+                // if the fetchJson was successful.
+                () => ToastManager.show({
```
amcclain: Good catch ensuring that this `then` takes an arrow function, but that is the standard form for these promise chains - it was just wrong before!  So no need for this comment here as it makes it seem like something unusual / in-particular is happening.

PR #237, opened 2018-05-02
Add propTypes and class-level doc comments to hoist/cmp. Fixes #211

### LINKED ISSUE #211: Doc bombing in cmp package.
Our cmp directory is home to most of our public API components.

Let's start with this directory, and ensure that all props for these components are documented with PropType, rather than.

Let's also place at least a  very minimal class comment on each class.

febbraiod:


cmp/leftRightChooser/impl/ItemRenderer.js
```diff
@@ -4,15 +4,26 @@
...
 
 /**
  * Render items in the LeftRightChooser.
+ *
+ * This is an implementation class private to Hoist
+ * @private
  */
 export class ItemRenderer extends Component {
 
+    static propTypes = {
```
febbraiod: Not sure we need this, the props here are handled by ag-grid. Don't think it hurts anything either.

cmp/LoadMask.js
```diff
@@ -15,20 +16,17 @@ import './LoadMask.scss';
...
     BACKGROUND = 'rgba(0,0,0, 0.25)';
 
-    static defaultProps = {
-        isDisplayed: false,
-        model: null,
-        inline: true
+    static propTypes = {
+        isDisplayed: PT.bool,
+        model: PT.object,
+        inline: PT.bool
```
amcclain: isDisplayed is certainly self evident, but I would add a comment on model and inline. Something like: "PromiseModel instance that will show this mask whenever pending" and whatever exactly inline is.... 😄

cmp/RelativeTimestamp.js
```diff
@@ -39,13 +40,17 @@ const defaultOptions = {
...
-
 @hoistComponent()
 class RelativeTimestamp extends Component {
+
+    static propTypes = {
+        /** Date object that will be used as reference, can also be specified in milliseconds*/
+        timestamp: PT.oneOfType([PT.instanceOf(Date), PT.number]),
+        /** See getRelativeTimestamp} options */
+        options: PT.object
+    };
```
amcclain: I like the fancy typing on timestamp to express both acceptable types. Stray character in the options comment there.

cmp/RelativeTimestamp.js
```diff
@@ -39,13 +40,17 @@ const defaultOptions = {
 
 /**
  * A RelativeTimestamp component - It displays time relative to `now`
```
amcclain: Could improve this a bit - maybe "Displays the approximate amount of time between a given timestamp and the present moment in a friendly, human readable form, automatically updating itself on a regular interval to stay current."

cmp/RelativeTimestamp.js
```diff
@@ -83,14 +88,14 @@ export const relativeTimestamp = elemFactory(RelativeTimestamp);
...
- *          nowString, string - String used as display property when timestamp is within the nowEpsilon interval.
- *          emptyResult, string - String to be used when timestamp is undefined
+ * @param {(Date|int)} timestamp - Date object or milliseconds that will be used as reference for this component
+ * @param {Object} [options]
+ * @param {boolean} [options.allowFuture] - Allow dates greater than Date.now()
+ * @param {string} [options.futureSuffix] - Appended to future timestamps
+ * @param {string} [options.pastSuffix] - Appended to past timestamps
+ * @param {integer} [options.nowEpsilon] - Interval (in seconds) that will serve as threshold for the nowString.
+ * @param {string} [options.nowString] - Used as display property when timestamp is within the nowEpsilon interval.
+ * @param {string} [options.emptyResult] - Used when timestamp is undefined
```
amcclain: I would say "returned" vs. "used" / "used as display property"

cmp/StoreFilterField.js
```diff
@@ -51,7 +57,7 @@ export class StoreFilterField extends Component {
         let searchTerm = escapeRegExp(this.value);
 
         let filter = null;
-        if (searchTerm && fields.length) {
+        if (searchTerm && fields && fields.length) {
```
amcclain: We should note fields as required above and avoid this check.  Configuring it without fields seems like an error to me.

cmp/ToastManager.js
```diff
@@ -22,10 +22,10 @@ export const ToastManager = {
      * Show a 'Toast' message.
      *
      * @param {string} message - the message to show in the toast.
-     * @param {element} icon - icon to be displayed
-     * @param {number} timeout - time in milliseconds to display the message.
-     * @param {string} intent - the Blueprint intent.
-     * @param {string} position - position in viewport to display. See Blueprint Position enum. 
+     * @param {element} [icon] - icon to be displayed
```
amcclain: Is this `Element` as it's a class vs. a primitive? (Still not entirely sure of the exact nature of some of these things)

cmp/clipboard/ClipboardMenuItem.js
```diff
@@ -6,16 +6,27 @@
...
  */
 class ClipboardMenuItem extends Component {
 
-    static defaultProps = {
+    static propTypes = {
+        icon: PT.element,
+        text: PT.string,
+        cls: PT.string,
+        successMessage: PT.string,
+        style: PT.object
```
amcclain: Of these, successMessage seems to me to be the only one that's unique to this component and not something you would get on a button. I can also see including icon and text to help make it clear what the two primary props are you will almost certainly be specifying, but cls and style are generally on everything.  Does that make sense / seem consistent enough?

cmp/contextmenu/ContextMenuItem.js
```diff
@@ -20,14 +20,14 @@ export class ContextMenuItem {
...
-     *          The prepareFn is a callback that is triggered before each time the menuitem is shown.
-     *          It can be used to modify the menuitem based on the record / selection.
+     * @param {Object} [icon] - optional icon to be displayed.
+     * @param {function} [action] - Executed when the user clicks the menuitem.
+     * @param {Object[]} [items] - child menu items.
+     * @param {boolean} [disabled] - true to disable this item.
+     * @param {boolean} [hidden] - true to hide this item.
+     * @param {function} [prepareFn] - function of the form (item, record, selection) => {}
+     *      The prepareFn is a callback that is triggered before each time the menuitem is shown.
+     *      It can be used to modify the menuitem based on the record / selection.
```
amcclain: There is a specific way to document the signature for callbacks, but not sure if we're drinking that Kool Aid. http://usejsdoc.org/tags-param.html#callback-functions  @lbwexler ?

cmp/contextmenu/StoreContextMenu.js
```diff
@@ -16,9 +16,8 @@ export class StoreContextMenu {
     items = [];
 
     /**
-     * @param {Object[]} items - collection of StoreContextMenuItems, configs to
-     *          create them, or Strings.  If a String, value can be '-' for a
-     *          seperator, or a key for a native AG Grid menu item.
+     * @param {Object[]} items - collection of StoreContextMenuItems, configs to create them, or Strings.
+     *      If a String, value can be '-' for a separator, or a key for a native AG Grid menu item.
```
amcclain: See isn't that lovely?

cmp/feedback/FeedbackDialog.js
```diff
@@ -12,7 +12,7 @@ import {textAreaField, toolbar} from 'hoist/cmp';
 
 /**
  * A simple dialog component to collect user feedback from directly within the application.
- * @see FeedbackService
+ * See FeedbackService
```
amcclain: No `@see`?  I was finding that IntelliJ was typically able to figure these out and provide the ctrl+click option, so I was kind of happy to use them, especially  when we had a case like this one where that's all we planned to say on the line.

cmp/form/DayField.js
```diff
@@ -5,29 +5,34 @@
...
  */
 @hoistComponent()
 export class DayField extends HoistField {
 
-    static defaultProps = {
-        popoverPosition: 'auto'
-    }
+    static propTypes = {
+        /** Width of field */
+        width: PT.number,
```
amcclain: This is another one that seems like a lower-level thing and not specific to this particular field.

cmp/form/HoistField.js
```diff
@@ -32,20 +33,26 @@ import {observable, setter, computed} from 'hoist/mobx';
...
+        /** handler to fire when value changes */
+        onChange: PT.func,
+        /** handler to fire when value is committed */
+        onCommit: PT.func,
+        /** model to bind to */
+        model: PT.object,
+        /** name of property in model to bind to */
+        field: PT.string,
+        /** is control disabled */
+        disabled: PT.bool,
```
amcclain: For the callbacks I would communicate briefly that they get passed the new value. Could skip disabled as self-evident. Would pref caps.

cmp/form/Label.js
```diff
@@ -5,21 +5,25 @@
...
- * @prop width, width of field, in pixels
  */
 @hoistComponent()
 export class Label extends HoistField {
 
+    static propTypes = {
+        children: PT.node,
+        style: PT.object,
+        className: PT.string,
+        width: PT.number
```
amcclain: Same question re. "standard" props that apply to almost all components, unless there is something special about the way they are processed in a given comp that merits a comment. 

I am not positive I have the right mental model about some of these things so happy to discuss. Also not sure how this all interacts with the "delegateProps" concept, which I just discussed with Lee earlier today (and still need to review).

cmp/form/dropdown/BaseDropdownField.js
```diff
@@ -17,6 +18,11 @@ import {HoistField} from '../HoistField';
  */
 export class BaseDropdownField extends HoistField {
 
+    static propTypes = {
+        /** Text to display when control is empty */
+        placeholder: PT.string
```
amcclain: E.g. these dropdown fields also could take a width, right?  (To make the point that I don't think we need it on the others, not suggesting we add it here.)

cmp/form/dropdown/ComboField.js
```diff
@@ -5,23 +5,29 @@
...
 @hoistComponent()
 export class ComboField extends BaseDropdownField {
 
+    static propTypes = {
+        /** Collection of form [{value: string, label: string}, ...] or [val, val, ...] */
+        options: PT.arrayOf(PT.oneOfType([PT.object, PT.string])).isRequired,
+        /** Text to display when control is empty */
+        placeholder: PT.string,
+        /** Optional custom itemRenderer, a function that receives (item, itemProps) */
+        itemRenderer: PT.func
```
amcclain: Unclear to me what item / itemProps are in this context. (I have a TODO for myself to catch up on the various combobox-like-things APIs.)

cmp/leftRightChooser/LeftRightChooserFilter.js
```diff
@@ -16,12 +17,14 @@ import {escapeRegExp} from 'lodash';
  */
 @hoistComponent()
 class LeftRightChooserFilter extends Component {
-    @setter @observable value = '';
 
-    static defaultProps = {
-        fields: []
+    static propTypes = {
+        /** width of field, in pixels */
+        fields: PT.arrayOf(PT.string)
```
amcclain: Nope 😄 
But this is also a pure implementation class, so I don't know that we need the propTypes here.

cmp/leftRightChooser/impl/ChooserToolbar.js
```diff
@@ -13,6 +13,9 @@ import {Icon} from 'hoist/icon';
 
 /**
  * A Toolbar for the LeftRightChooser.
+ *
+ * This is an implementation class private to Hoist
+ * @private
```
amcclain: I would be fine with just the first line and the annotation (and no "this is an impl..." boilerplate)

cmp/leftRightChooser/impl/ItemRenderer.js
amcclain: Would not include as it's an impl class and I think *not* having the propTypes is a useful signal to that effect.

cmp/resizable/ResizableModel.js
```diff
@@ -24,6 +24,12 @@ export class ResizableModel {
 
     /**
      * Construct this object.
```
amcclain: We can lose this intro line on any constructors

cmp/LoadMask.js
```diff
@@ -44,7 +42,7 @@ export class LoadMask extends Component {
                 style: {backgroundColor: this.BACKGROUND}
             },
             usePortal: !inline,
-            item: inline ? this.getInlineChild() : this.getViewportChild()
+            item: inline !== false ? this.getInlineChild() : this.getViewportChild()
```
lbwexler: like this way of doing the defaulting on line 45, but you are using the old way on line 44!  Seems like you need a single local variable here for this!

cmp/ToastManager.js
lbwexler: Well, its a common, and very specific thing in the API, and it would be nice to have a specific name for it.  Strictly speaking its not an instance of class Element, but I think we would do well to refer to it as 'Element'.

cmp/clipboard/ClipboardMenuItem.js
lbwexler: agree we need to leave style and cls off

cmp/contextmenu/ContextMenuItem.js
lbwexler: Would like to skip.  A function is always in some sense a "callback", and we have the function type and fn suffix already.

cmp/feedback/FeedbackDialog.js
lbwexler: Agree its useful, and cool even if it doesn't link in tools.

febbraiod: OK. I was actually using them in a few places as well, but thought I was drinking too much kool-Aid. (Was using with @link as well is THAT too much? 

@see {@link http://usejsdoc.org/tags-see.html}

cmp/form/HoistField.js
lbwexler: Are these prop types actually checked on the subclass instances?  Or are they overridden by the specification of propTypes on the subclass?   The latter seems more likely to me.
 Would be good to know....

cmp/leftRightChooser/LeftRightChooserFilter.js
lbwexler: Actually, think Gil had in mind that this was going to be part of the API as an auxillary object.  Apps have to add it, and put it somewhere in the layout.

cmp/leftRightChooser/impl/ChooserToolbar.js
lbwexler: totally agree with this.  Need to keep these private docs boilerplate to a minimum.

cmp/resizable/Resizable.js
```diff
@@ -84,12 +79,12 @@ export class Resizable extends Component {
             items.push(this.renderCollapsedChild());
         }
 
-        if (isCollapsible) {
+        if (isCollapsible !== false) {
             const collapser = this.getCollapser();
             items = (isContentFirst ? [...items, collapser] : [collapser, ...items]);
```
lbwexler: If we are going to use the getter pattern, the default needs to be in the getter!

cmp/resizable/impl/Dragger.js
```diff
@@ -11,12 +11,16 @@ import {elemFactory, hoistComponent} from 'hoist/core';
...
+/** This is an implementation class private to Hoist
+ * @private
+ */
 @hoistComponent()
 export class Dragger extends Component {
 
     resizeState = null;
 
     static propTypes = {
+        /** Position of resize border in relation to resizable component */
```
lbwexler: probably ok to keep the prop types on this impl class since they are already here, but ditch the obvious comment

cmp/tab/TabPaneModel.js
```diff
@@ -27,8 +27,8 @@ export class TabPaneModel {
 
     /**
      * @param {string} id - unique ID, used for generating routes.
-     * @param {string} name - display name for the tab.
-     * @param {Component} - component (or React node) to be displayed within the tab.
+     * @param {string} [name] - display name for the tab.
+     * @param {Component} component - component (or React node) to be displayed within the tab.
```
lbwexler: This is actually a lie.  Its the class itself, not an instance of it.   See the code below where it is used:

Should be something like
@param {Object} componentClass - class of React Component to be displayed within the tab.

cmp/form/dropdown/SelectField.js
```diff
@@ -13,15 +14,20 @@ import {BaseDropdownField} from './BaseDropdownField';
...
 @hoistComponent()
 export class SelectField extends BaseDropdownField {
 
+    static propTypes = {
+        /** Collection of form [{value: string, label: string}, ...] or [val, val, ...] */
+        options: PT.arrayOf(PT.oneOfType([PT.object, PT.string])).isRequired,
+        /** Text to display when control is empty */
+        placeholder: PT.string,
+        /** Optional custom itemRenderer, a function that receives (item, itemProps) */
+        itemRenderer: PT.func
```
febbraiod: rename optionRenderer see Lee's note on comboField. I agree this is clearer.

cmp/resizable/impl/Dragger.js
febbraiod: Maybe this was dense of me, but when I first glanced at this class (without reading any further than the prop) I thought the 'side' could have meant the side of the viewport we would be dragging the component out from. i.e. top wouldn't mean the drag bar is on top of the cmp, but that you'd drag down from the top of the window.

cmp/form/HoistField.js
febbraiod: It appears that they are overridden by the subclass. Passed a className prop to a dayField as a int. The propType check on Hoist field did not pick it up. I had to add the PropType definition for className to dayField specifically to catch it.

lbwexler:
yes, of course.  Perhaps we just want to have a single block of these on
HoistField?  Where we indicate the optional ones that only apply to certain
controls at the bottom, with the controls they apply to?

cmp/form/HoistField.js
febbraiod: From Lee: Perhaps we just want to have a single block of these on
HoistField? Where we indicate the optional ones that only apply to certain
controls at the bottom, with the controls they apply to?

febbraiod: I think that sounds great in terms of functionality and avoids having to redefined the ones common to HoistField if we want the check on the subclasses. 

However, it would take away the convenience of see subclass specific props at the top of that class (unless we replace the doc string which we were moving away from)

lbwexler:
Something has to give here.  We need to revisit HoistField in a number of
ways.  Let's leave them as is (but in a consistent state, with all shared
ones on HoistField only) and address later.

cmp/form/HoistField.js
febbraiod: From Lee:

Something has to give here. We need to revisit HoistField in a number of
ways. Let's leave them as is (but in a consistent state, with all shared
ones on HoistField only) and address later.

cmp/leftRightChooser/LeftRightChooserFilter.js
amcclain: Ah I see - in particular I see now that this field is not included in the LR chooser component itself as I thought it was. Let's see how it shakes out - I can understand how different uses of the component might/might not need/want the filter or a toolbar or otherplace to put it.

cmp/form/dropdown/ComboField.js
febbraiod: As I indicated in a comment below, optionsRender is clearer in terms of what this is. However, on reflection itemRenderer is what it is called in the blueprint docs, and itemProps is what gets passed by them. Perhaps a link to the blueprint docs is the better answer?

With regard to the propTypes for the mixed array. I couldn't find an example of this and it's not documented but I tested passing a mixed array of strings and objects and we had no trouble. Alternatively the docs state that you can pass a custom validation to arrayOf if we feel more comfortable with that.

cmp/leftRightChooser/LeftRightChooserFilter.js
febbraiod: No definitely not. Sorry will correct as I am going through.

cmp/form/dropdown/ComboField.js
amcclain: I would use `options` as we're taking over the API to some extent here, and using `options` for the data elements themselves (which I still think is a good idea).  Where we can provide the relevant info in a succinct form w/o linking to other docs, we should aim to do so (especially for a comp like this that will be very common).

Sounds like we are OK in terms of the array typing. I would hesitate to go further into a custom array validator at this point / for this use case.

Issue #230, opened 2018-05-01
Standardize on msg vs. message (or support both)


amcclain:
This has been a perennial source of confusion / bugs for me for our handful of APIs where we take a config with a "message" parameter.  

E.g. ExceptionHandler takes `message`, FeedbackService `msg`, server-side tracking/error/feedback APIs use `msg`, ToastManager and MessageModel `message`, TrackService `msg`.

Given that msg appears to be baked further into the server, it might be most expedient to standardize on that, but I do think it could be worth it to provide for a fallback on the client-side APIs.

amcclain:
The specific issue I ran into right now was TrackService taking msg, so if we had one change to make I would vote for accepting either there, or switching to message and updating any current code.  That could be done as part of #209

lbwexler:
atm and I both think 'message' makes the most sense here.

Let's just support msg silently, where necessary.  Best not to document that.

cnrudd:
ok, will make sure message is supported everywhere.

cnrudd:
We have decided that we will not accept `msg` in the JS API.  I have removed all instances of it.  The change is in the develop branch.

PR #239, opened 2018-05-02
Standardize on msg vs. message (or support both) (#230)

cnrudd:
These are the changes to the JS in hoist react.   Are you guys thinking you want the groovy server coder in hoist core to also support message & msg?

### DIFF
--- a/app/impl/ExceptionDialogModel.js
+++ b/app/impl/ExceptionDialogModel.js
@@ -33,7 +33,7 @@ export class ExceptionDialogModel {
         const svc = XH.errorTrackingService,
             {exception, userMessage, options} = this;
         if (svc.isReady) {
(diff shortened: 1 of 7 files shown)


amcclain:
Hi - no, not looking to change the server, as Hoist itself is the only expected caller there so we don't need to worry about backwards compatibility and therefore there's no need to complicate the API.  But great to be getting this longstanding irritant cleaned up on the client.

Not sure what my thoughts are re. adding the fallback in all cases - e.g. FeedbackService where again the only expected caller is our own FeedbackDialog component. With our standard use of the destructured arguments the fallback ends up a bit more "in your face" than I expected. I would be up for either and defer to you two there.

amcclain:
The message is loud and clear now.

PR #240, opened 2018-05-03
Panel component

gportela85:
https://github.com/exhi/hoist-react/issues/234

### DIFF
--- a/cmp/panel/impl/PanelHeader.js
+++ b/cmp/panel/impl/PanelHeader.js
@@ -1,0 +1,47 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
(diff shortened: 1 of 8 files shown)


gportela85:
@amcclain Panel with `title`, `headerItems`, `topToolbar` and `bottomToolbar`
screen shot 2018-05-03 at 5 39 54 pm

cmp/panel/impl/PanelHeader.js
```diff
@@ -25,21 +25,15 @@ class PanelHeader extends Component {
...
-                            flex: 1,
-                            cls: 'xh-header-title',
-                            item: title
-                        }),
-                        ...headerItems
-                    ]
-                }) : null
+                    item: title
+                }) : null,
+                ...headerItems
```
amcclain: I made this change here because I didn't want the headerItems to go away if you didn't specify a string title but I didn't test it, nor did I have a chance yet to look at e.g. how do things behave w.r.t. wrapping which is something we should consider.

PR #241, opened 2018-05-03
Generate versions of formatters that return React Elements.

TomTirapani:


format/FormatNumber.js
```diff
@@ -276,4 +276,12 @@ export const numberRenderer = createRenderer(fmtNumber),
...
\ No newline at end of file
+    percentRenderer = createRenderer(fmtPercent);
+
+export const fmtNumberEl = createElFormatter(fmtNumber),
+    fmtThousandsEl = createElFormatter(fmtThousands),
+    fmtMillionsEl = createElFormatter(fmtMillions),
+    fmtBillionsEl = createElFormatter(fmtBillions),
+    fmtQuantityEl = createElFormatter(fmtQuantity),
+    fmtPriceEl = createElFormatter(fmtPrice),
+    fmtPercentEl = createElFormatter(fmtPercent);
```
amcclain: Am trying to think if I have any other suggestions for the names here.  Was thinking just `fooEl` to match `fooRenderer` - that sounds OK with e.g. `dateTimeEl` but less so with `millionsEl`....

Also kinda tempted to just drop any qualifier and use `dateTime` or `millions` directly...  Given that you are always importing from a format package it doesn't seem entirely crazy...  

Another option would be `dateTimeFormatter` or `millionsFormatter`.  Seems more readable / natural in some ways, although long.

So I can see why you landed where you did.  The names are a bit awkward but also probably not that common.

Was that helpful? 😄

format/FormatUtils.js
```diff
@@ -27,6 +28,18 @@ export function createRenderer(formatter) {
     };
 }
 
+/**
+ * Generate a formatter that returns a React Element.
+ *
+ * @param {function} [formatter] - an existing formatter method.
```
amcclain: No brackets around formatter here as it's non-optional.

format/FormatUtils.js
```diff
@@ -27,6 +28,18 @@ export function createRenderer(formatter) {
...
+ */
+export function createElFormatter(formatter) {
+    return function(v, opts) {
+        const html = {__html: formatter(v, opts)};
+        return span({dangerouslySetInnerHTML: html});
```
amcclain: Was thinking again as to whether or not we should have any reservations here about using dangerously...  

One argument we can make is that this is displaying the output of our formatters, which currently take primitives and return strings that might have markup, but where the markup is generated by us and can't contain random generated stuff.

Realizing that one exception to that is our support for `tipFn` could end up injecting some additional content into the markup from another source.  But that should probably be escaped anyway - maybe we want to do that for good measure.

amcclain:
@TomTirapani if you have a chance to look at HTML escaping the output of tipFn in the general formatter code, that would be great.  It seems to me that it's a reasonable bit of hardening to do regardless of how it is used.

@lbwexler checking to see if you have any input on the names of the formatter variants that return element factories.  I don't have any new ideas, and am OK going with what Tom has here which I think is clear.

format/FormatNumber.js
lbwexler: Two things:


+ I think the 'fooRenderer' convention itself is targeting Ext and could be reworked.
+ How about 'quantityFmt' and 'quantityFmtEl'?  Like the idea of not splitting the modifier around the variable name?

Maybe a quick discussion of this is in order?

format/FormatUtils.js
lbwexler: Doesn't the application data within the *markup* present the risk? (Not sure we need to address this immediately -- don't see how this any worse that what we currently do in sencha hoist, regarding inserting application data in to the Dom

TomTirapani:
I'm looking at this again with the context of tipFn, and thinking it might be better to rethink this whole approach.

It's only really the `fmtSpan` function that's causing us issues, for both tipFn and the renderers etc. I was thinking perhaps we can change this to return a span react element rather than html string, and then we should be able to use the renderers everywhere in a react app without needing to use `dangerouslySetInnerHTML`.

The only problem this presents is with ag-grid, which expects a html string. We could put something in our column factory that handles converting react elements to static html for this purpose (`react-dom` has a `renderToStaticMarkup` method). Another option would be to use a default `cellRendererFramework`, and we always use elements in ag-grid.

In this way, we're using elements by default, and switching to html string only for ag-grid (which is the exception).

I haven't got all the details figured out, but wanted to get your thoughts. I think it's probably more correct, and will help us get out of this mindset of concatenating html strings that's left over from Ext.

lbwexler:
Well if we could figure out how to write a truly efficient cellRendererFramework with elements that might be the best solution.  I wonder if AG Grid is turning the strings back into elements under the hood anyhow.  That is, I wonder how deep the React implementation goes in AG Grid?

TomTirapani:
Took a second stab at this, going with the 'react-first' approach outlined above:

https://github.com/exhi/hoist-react/pull/268
https://github.com/exhi/piq-react/pull/9

PR #248, opened 2018-05-04
Always delete known custom colProps before passing to grid. Fixes #244

### LINKED ISSUE #244: Need a clean way to null out Hoist-added grid colDef props
If you have a columnFactory defined like this:
```
const sampleCol = colFactory({
  headerName: 'Sample',
  field: 'sampleField',
  fixedWidth: 100
});
```
where you typically want the column to be 100px wide, but you also want to use it in one special grid where you want it to flex, you might try:

```
sampleCol({flex: 1, fixedWidth: null})
```

But you will start getting a console warning "invalid colDef property 'fixedWidth'" because the code within `fileColFactory` is not cleaning up your fixedWidth config.

We should ensure we always clear out custom configs even if they are falsey or an unexpected value (i.e. for `align`). Should also take a sec to think about any possible meaning for zeros or if there are other override scenarios we should account for.

febbraiod:


columns/Utils.js
```diff
@@ -26,25 +27,21 @@ export function fileColFactory(fileVals = {}) {
...
-                delete colProps.align;
             }
 
             if (colProps.align === 'right') {
                 colProps.headerClass.push('xh-column-header-align-right');
                 colProps.cellClass.push('xh-align-right');
-                delete colProps.align;
             }
 
             if (colProps.flex) {
```
amcclain: Might as well handle flex as we do fixedWidth with the `isNumber` check - if just for symmetry here.  (As discussed both flex:0 and fixedWidth:0 are weird choices, but we're not considering them to be entirely invalid.)

That's my only comment - I think the use of the lodash omit works well here.

febbraiod: Done.

PR #249, opened 2018-05-04
Suppress grid empty text entirely, Fixes #242

### LINKED ISSUE #242: Provide way to suppress grid empty text entirely
Passing `overlayNoRowsTemplate: ''` does not override the default template - presumably the empty string is considered falsey in terms of how the defaulting works (which caught me a bit by surprise).

It seems reasonable to have a way to not show any empty text.

febbraiod:
Opening for discussion. Implemented on activity grid but don't necessarily think it needs this.


amcclain:
I am still torn as to what is on the model API and what is on the component API. Since this is so visual / display-oriented, I see it as a prop of the component vs. its model.

Would also like to clearly relate specifying no empty text with specifying custom empty text. If we had our own prop for the custom empty text we could accomplish that by processing a falsey value as you would expect (and show nothing).

But this gets to the really central question of whether or not we're gonna add top-level props to Grid, if the gridOptions props are all ag-grid props, and how they relate.  Let's discuss that next week and try to come up with a quick decision?

lbwexler:
Discussed with Anselm -- we think in the case of *grid* this does belong on the model.  We may want to change it on the fly in code, so it represents grid *state*.  

Let's rename it 'emptyText' and have it default to null.

lbwexler:
We'll see if we can expose it as an observable, and hence make it dynamically settable.  Not immediately needed....

grid/GridModel.js
```diff
@@ -42,8 +42,10 @@ export class GridModel {
      * @param {BaseStore} store - store containing the data for the grid.
      * @param {Object[]} columns - collection of column specifications.
      * @param {StoreSelectionModel} [selection] - selection model to use
+     * @param {boolean} [emptyText] - empty text to display if grid has no records. Can be valid HTML.
```
amcclain: Great change and cleanup, thank you!  Only tweak is to declare this a string in the doc comment.

PR #253, opened 2018-05-04
Mask for disabled / inactive components, Fixes #246

### LINKED ISSUE #246: Mask for disabled / inactive components
Variant on LoadMask, but without the spinner. 

Could be integrated in forthcoming Panel component via a convenience `disabled` prop there, but also might be useful in other standalone cases for other containers, etc.

febbraiod:


cmp/mask/DisabledMask.js
```diff
@@ -0,0 +1,49 @@
...
+
+import './Mask.scss';
+
+/**
+ * Simple Mask for Disabled or Inactive components.
+ */
+@hoistComponent()
+export class DisabledMask extends Component {
+
+    BACKGROUND = 'rgba(0,0,0, 0.60)';
```
febbraiod: Quite a bit darker than the loadMask. I found that, without the spinner, 25% wasn't enough to clearly tell me the component wasn't active. I have a demo branch in PIQ-React with a button to temporarily disable the positions grid:

https://github.com/exhi/piq-react/tree/demoDisabledMask

febbraiod:
Had a thought or two about this over the weekend closing to tweak.

cmp/mask/ComponentMask.js
```diff
@@ -0,0 +1,56 @@
...
+            usePortal: false,
+            item: frame({
+                width: '100%',
+                height: '100%',
+                alignItems: 'center',
+                justifyContent: 'center',
+                style: {
+                    textAlign: 'center'
+                },
+                items: displayText || null
```
febbraiod: Need to explicitly set to null here? Might undefined accomplish the same?

cmp/mask/ComponentMask.js
```diff
@@ -0,0 +1,54 @@
...
+        const {isDisplayed, text} = this.props;
+        if (!isDisplayed) return null;
+
+        return overlay({
+            cls: 'xh-mask',
+            autoFocus: false,
+            isOpen: true,
+            canEscapeKeyClose: false,
+            backdropProps: {
+                style: {backgroundColor: this.BACKGROUND}
```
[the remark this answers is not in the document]
febbraiod: Spoke to Gil a bit about his changes. We should still be able to set this color via CSS. But should  keep the backdrop as with the addition of Classes.OVERLAY_SCROLL_CONTAINER it will expand to fill it's container and will simply the code thats doing that now.

Will circle back to this a little later on.

cmp/mask/Mask.js
```diff
@@ -0,0 +1,50 @@
...
+            cls: `xh-mask ${Classes.OVERLAY_SCROLL_CONTAINER}`,
+            autoFocus: false,
+            isOpen: true,
+            canEscapeKeyClose: false,
+            usePortal: false,
+            item: box({
+                cls: 'xh-mask-body',
+                alignItems: 'center',
+                justifyContent: 'center',
+                style: {
```
febbraiod: Can be done away with if we don't intend to have text that would span any wider than the component we're masking. But this will make the text display in a more pleasant way if the text should need to wrap, eg for a mask on a skinny widget.

febbraiod:
Additionally, the padding on .xh-mask-body cause the overlay's child box (the text container) to render an 10px square even if the text prop has not been provided.

PR #256, opened 2018-05-07
Dropdown fields support a requireSelection prop, Fixes #223

### LINKED ISSUE #223: Dropdown fields should support a requireSelection prop to disallow inputting new values
https://github.com/exhi/hoist-react/commit/b620d027ef2ea9485e7f95f5fcb01b7ac1c4d347#comments

febbraiod:


rest/RestControl.js
```diff
@@ -85,6 +85,7 @@ export class RestControl extends Component {
             model,
             field: 'value',
             options,
+            requireSelection: true,
```
febbraiod: This is test code. Figure out how we want to default.

admin/tabs/logging/LogLevelPanel.js
```diff
@@ -47,7 +47,7 @@ export class LogLevelPanel extends Component {
         ],
         editors: [
             {field: 'name'},
-            {field: 'level'}
+            {field: 'level', type: 'lookupSelect'}
```
febbraiod: Waiting on decision regarding log level lookups null handling to decide exactly how to handle this control.

If we decide to leave well enough alone and not remove 'None' as a lookup we don't have to do this here. We can just set this field to lookupStrict and get the SelectField automatically. If we remove this somewhat confusingly named lookup, we should NOT set lookupStrict(as we would be adding a null option). In which case, we need this flag to get a selectField rather than a comboField.

PR #257, opened 2018-05-07
Reenable nullable REST comboFields

febbraiod:
Not sure if this was taken out because of the redundant 'None' option on log levels, but it seems to me that combo fields for rest forms should have the option to be nullable.

Log Level issue dealt with here:

https://github.com/exhi/hoist-core/pull/27


febbraiod:
Similar change is part of: https://github.com/exhi/hoist-react/pull/256

Might render this PR redundant.

PR #264, opened 2018-05-08
Standard Component for Resizable.collapsedDisplay, Fixes #254

### LINKED ISSUE #254: Standard Component for Resizable.collapsedDisplay
We might end up integrating resizable further with Panel, etc. but in the meantime could still be a useful clean-up to have a standard component that could be passed as Resizable's collapsedDisplay - could take an icon and title string, similar to PanelHeader.

febbraiod:


amcclain:
I would like to avoid the overloading of toolbar here, as toolbars might have styling / get extra functionality that's unrelated to this placeholder's job of being a reasonably-sized box with an icon and title. 

(Also - did we test for components that collapse to the left or right?)

Re. building into Panel - can we accomplish that by simply having Panel implement `renderCollapsed` and return just a PanelHeader with its title and icon?  (Would still need to handle the "vertical" orientation but probably could do with some light modifications to the header comp.)

Note in that case I don't _think_ we would want to include any Panel `headerItems` in the placeholder - but we could use that same prop to support an expand button ala Ext...

amcclain:
Sorry @lbwexler, I was referring to use of the toolbar comp in the current diff, not anything related to the Panel proposal.  I would be all for having convenient access here in Panel.

PR #268, opened 2018-05-09
Support returning both react elements and html strings in all formatters, by passing 'asElement' flag

TomTirapani:
+ Added `toHtml` to return html string from a react element
+ Removed `linkTag` and `imgTag` utils. We should use `a` and `img` element factories instead.


lbwexler:
This is really great.  Is that all there is to it?

TomTirapani:
Yeah - once we made the nested spans in the formatters elements rather than html strings, it all sort of fell into place.

Check the PIQ P/R for an example of the changes an app would have to make - namely, using `cellRendererFramework` rather than `cellRenderer` (valueFormatter can stay the same for dates where we don't expect span elements), and using the `toHtml` method for situations where we really need html strings (in this case, chart formatters)

TomTirapani:
I've updated this so that formatters support generating both react elements and html strings

PR #270, opened 2018-05-09
Support null and non-json responses in FetchService. Fixes #266

### LINKED ISSUE #266: FetchService.fetchJson throws exception on 204 No Content response
Encountered this making a DELETE request to an endpoint which returns 204 on success. See https://github.com/github/fetch/issues/268 for related info.

In this scenario I would expect to just receive `null` as the promise value in my calling code.

```
SyntaxError: Unexpected end of JSON input
    at FetchService._callee$ (FetchService.js:69)
    at tryCatch (runtime.js:62)
    at Generator.invoke [as _invoke] (runtime.js:296)
    at Generator.prototype.(anonymous function) [as next] (webpack-internal:///./node_modules/regenerator-runtime/runtime.js:114:21)
    at step (FetchService.js:9)
    at eval (FetchService.js:9)
```

haynesjm42:
* Extract general purpose fetch() method from fetchJson()
* Handle 204 responses in fetchJson()


amcclain:
In hoist-sencha we decoded/returned JSON [based on the content-type](https://github.com/exhi/hoist-sencha/blob/670420e9b54ace10aa0c089f8642319999cbe477/grails-app/assets/src/xh/all/svc/AjaxService.js#L70) - would it be better to do that here so we don't have a new method (and need to bomb a bunch of code)?  It's also the case that the vast majority of calls are for JSON content, so would be nice to keep the shorter function name for the main use case.

Another alternative if the content type proved unworkable for some reason would be to add an option to the method - e.g. `decodeJSON` - that defaults to true but is available to turn off.

amcclain:
Sorry missed that key / obvious point - was rushing

haynesjm42:
Updated the docs. I think I followed the new format correctly but please let me know if I missed anything!

lbwexler:
👍 Great documentation!

PR #271, opened 2018-05-09
Add eager loading options for tabs and lookups, Fixes #238

### LINKED ISSUE #238: Add/Delete Preference bugs
These are low priority as discussed with Lee:

After creating a new default Preference, if you switch to the User Preference panel to add a corresponding record, the new preference is not available in the User Preference add form without a hard refresh. By the same token recently deleted ones will still be there.

Less importantly, when deleting a default Preference from hoist admin, if you go to the User Preference panel the records for corresponding user preferences are still in the grid but, if you delete one the server will throw because those records were automatically deleted on the server. This is also true in Hoist-Sencha.

febbraiod:
+Eager load user prefs panel


rest/data/RestStore.js
```diff
@@ -20,16 +20,22 @@ export class RestStore extends UrlStore {
 
     _lookupsLoaded = false;
 
-    constructor({dataRoot = 'data', ...rest}) {
+    /**
+     * @param {string} [dataRoot] - Name of root node for records in returned data
+     * @param {boolean} [eagerLoadLookups] - Whether lookups should be loaded each time loadAsync is called
```
lbwexler: how about  'reloadLookupsOnLoad'?  Seems slightly clearer, and gets to the point of *re*loading.

cmp/tab/TabContainerModel.js
```diff
@@ -87,6 +87,8 @@ export class TabContainerModel {
         
         this.selectedId = child ? id : children[0].id;
 
+        if (child.eagerLoad) this.setLastRefreshRequest(Date.now());
```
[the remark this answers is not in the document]
febbraiod: Yeah I noticed that too. I suppose we could call the child component's loadAsync directly, but that seemed wrong too as the tab pane's doc strings indicate:

* Reloads the active tab whenever its lastLoaded is out of date with its parent's 'lastRefreshRequest'.
 and
 * Contained components that load data/state from the server should implement loadAsync(), but
 * generally leave the calling of that method to this component

cmp/tab/TabContainerModel.js
```diff
@@ -88,6 +88,8 @@ export class TabContainerModel {
         
         this.selectedId = child ? id : children[0].id;
 
+        if (child.reloadOnShow) this.setLastRefreshRequest(Date.now());
```
febbraiod: Comment was covered up by rename, but still relevant:

lbwexler 25 minutes ago  Owner
This seems wrong to me. If any single child pane with this option is selected -- all panes in the container that are subsequently shown will get an extra refresh as well.

  @febbraiod
febbraiod 17 minutes ago  Member
Yeah I noticed that too. I suppose we could call the child component's loadAsync directly, but that seemed wrong too as the tab pane's doc strings indicate:

Reloads the active tab whenever its lastLoaded is out of date with its parent's 'lastRefreshRequest'.
and
Contained components that load data/state from the server should implement loadAsync(), but
generally leave the calling of that method to this component

PR #273, opened 2018-05-09
For exceptions, default logOnServer to showAsError value

febbraiod:
This change was made in Hoist Sencha last week to solve a client app problem. We should be consistant here.


lbwexler:
Great catch Don .  👍

PR #275, opened 2018-05-09
Hoist Core Refactor

lbwexler:
+ Reorganize core to a three decorator triad (HoistModel, HoistService, HoistComponent)
+ Add lifecycle for destruction, especially in models.  Cleanup auto runs and event handlers.
+ Add formal 'mixins' : 'Reactive'/'EventTarget'
+ Provide general class utils for mixing in functionality to components.
+ Cleanup ensureServiceReadyAsync() => initAsync()
+ Provide basic methods loadAsync() and loadData() on grids (that trampoline to underlying store)
+ Add new method XH.renderApp() to DRY app boot sequence,


amcclain:
Awesome - huge amount of work and I think really helps to clarify our core objects and get us to a much better place around lifecycle management, etc. etc.

PR #278, opened 2018-05-10
NumberField enhancements, Fixes #269

### LINKED ISSUE #269: NumberField enhancements
Looking to super-charge NumberField for our applications. Blueprint has [an example](http://blueprintjs.com/docs/v2/#core/components/forms/numeric-input) with some extended features, but notes they are not part of the core field.

+ Support shorthand entry - 2.4m, 5k, -123b - this is the main one.
+ Render value with commas inserted to help avoid 10x errors. When editing, could not display commas or handle them gracefully (i.e. allow them to be entered but ignore them when ultimately evaluating the input).

BP example also shows support for scientific notation and inline arithmetic. I do _not_ think these make sense for our use cases. 

Will think / suggestions welcome on any other additional behaviors that would be appropriate to consider.

febbraiod:


cmp/form/NumberField.js
```diff
@@ -24,22 +26,31 @@ export class NumberField extends HoistField {
...
-        max: PT.number
+        max: PT.number,
+        /** Number of decimal places to allow on field's value, defaults to 4*/
+        precision: PT.number,
+        /** Set to true to automatically fill in zeros in accord with precision */
+        zeroPad: PT.bool,
+        /** Constrain input to numeric characters, should be set to false for advanced input evaluation */
+        allowNumericCharactersOnly: PT.bool,
+        /** Whether to display large values with commas */
+        displayWithDelimiters: PT.bool
```
febbraiod: Not super sure about the defaults I chose for these new props. I have gonna accustom to the admin use case and am not really sure about what the most commonly desired behavior will be. 

Pretty sure zeroPad should be false.
Precision is defaulted to 4 as that is what we have done for client apps in the past.
As for the commas, and shorthand eval really not sure.

febbraiod:
Reviewed with Colin. Closing for a tweak or two

febbraiod:
Reviewed with Lee, closing for tweaks and to avoid double review work

lbwexler:
Don and I have gone through this, and I am happy to merge.  @amcclain any comments?  Do the names seem reasonable to you?

cmp/form/NumberField.js
```diff
@@ -24,22 +25,37 @@ export class NumberField extends HoistField {
...
+        max: PT.number,
+        /** Number of decimal places to allow on field's value, defaults to 4 */
+        precision: PT.number,
+        /** Allow/automatically fill in trailing zeros in accord with precision, defaults to false */
+        zeroPad: PT.bool,
+        /** Set to true for advanced input evaluation, defaults to false.
+            Inputs suffixed with k, m, or b will be calculated as thousands, millions, or billions respectively */
+        supportShorthandUnits: PT.bool,
+        /** Whether to display large values with commas */
+        displayWithDelimiters: PT.bool,
```
amcclain: Kinda wonky name on this one, especially as it might be interpreted as affecting the decimal point or some other kind of delimiter.  Would `displayWithCommas` or `useCommas` or `addCommas` be better?

cmp/form/NumberField.js
```diff
@@ -24,22 +25,37 @@ export class NumberField extends HoistField {
...
         /** maximum value */
-        max: PT.number
+        max: PT.number,
+        /** Number of decimal places to allow on field's value, defaults to 4 */
+        precision: PT.number,
+        /** Allow/automatically fill in trailing zeros in accord with precision, defaults to false */
+        zeroPad: PT.bool,
+        /** Set to true for advanced input evaluation, defaults to false.
+            Inputs suffixed with k, m, or b will be calculated as thousands, millions, or billions respectively */
+        supportShorthandUnits: PT.bool,
```
amcclain: Would `enable` be slightly more standard / common than `support`?

cmp/form/NumberField.js
```diff
@@ -24,22 +25,37 @@ export class NumberField extends HoistField {
         /** minimum value */
         min: PT.number,
         /** maximum value */
-        max: PT.number
+        max: PT.number,
+        /** Number of decimal places to allow on field's value, defaults to 4 */
+        precision: PT.number,
```
amcclain: `maxPrecision` ?  Don't feel strongly, I do think it's implied, but that does seem marginally more correct.

cmp/form/NumberField.js
```diff
@@ -52,13 +68,44 @@ export class NumberField extends HoistField {
...
+        value = this.parseValue(value);
+        return isNaN(value) ? null : value;
     }
 
-    toInternal(val) {
-        return val != null ? val.toString() : '';
+    toInternal(value) {
+        if (value == null) return '';
+        const props = this.props,
+            precision = props.precision != null ? props.precision : 4,
```
amcclain: Trying to decide if it makes sense to have 4 as a default here, versus some larger value (16) or no limit. When accepting user entry it seems more aggressive to truncate precision entered by the user than it does in the corresponding case of formatting a number for display.

That said, we have the prop available, so a high precision use case can deliberately up it, so I am happy to leave four as a likely most-useful-default.

PR #279, opened 2018-05-11
LoginPanel - Login Message

gportela85:
https://github.com/exhi/hoist-react/issues/263

### DIFF
--- a/app/impl/LoginPanel.js
+++ b/app/impl/LoginPanel.js
@@ -63,7 +63,7 @@ export class LoginPanel extends Component {
                             cls: 'xh-login__warning'
                         }),
                         loginMessage ? box({
-                            cls: 'xh-login-message',
+                            cls: 'xh-login__message',
                             item: loginMessage
                         }) : null
                     ],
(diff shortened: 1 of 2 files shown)


amcclain:
Hi - thanks for this, and apologies as I know I said it looked good when you sent the screenshot via Slack... but I think it's worth avoiding the use of the special callout component here, along with its blueprint-specific CSS class, etc.  While BP is definitely available for use, we're looking to somewhat minimize the number of direct usages of relatively uncommon comps when they don't add obvious value.

It would be simpler to just insert the message as a node above the form fields right in the panel.  (That way it also gets wrapped in the border, etc.)  We could put it in a div w/an `xh-login-message` class and give it some padding or a light background. 

That make sense?  Thanks again.

gportela85:
Removed Blueprint callout
Better to leave the message at the top or bottom? 
screen shot 2018-05-11 at 8 56 23 am
screen shot 2018-05-11 at 8 58 07 am

lbwexler:
I vote for the bottom for sure!

PR #281, opened 2018-05-11
Make dropdown field options mobx observable

TomTirapani:
+ Add setOptions method

### DIFF
--- a/cmp/form/dropdown/BaseDropdownField.js
+++ b/cmp/form/dropdown/BaseDropdownField.js
@@ -8,6 +8,7 @@
 import {PropTypes as PT} from 'prop-types';
 import {isObject, find} from 'lodash';
 import {menuItem} from 'hoist/kit/blueprint';
+import {observable, action} from 'hoist/mobx';
 
 import {HoistField} from '../HoistField';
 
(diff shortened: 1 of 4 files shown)


lbwexler:
Tom -- can you give a little color on this?  Where are you using it/needing it? 

Couple of questions: 

+ Can we allow the raw "options" to be changed dynamically by the owning component?  Think that would be the standard react way to make this dynamic, rather than the setter.

+ Do we need to distinguish in the API between 'raw' options and 'normalized' options?  Seems weird to receive options from the containing component, and immediately re-publish them with same name.

Let's discuss?  Feel like I need more of the context that you have on this one.

-Lee

TomTirapani:
Hi Lee,

The context is this issue created by Anselm: https://github.com/exhi/hoist-react/issues/280

Before the options were dynamic, we were normalising in the constructor so we could be sure each option was a object. I wanted to maintain this normalise once approach with the setter. In order to allow the raw options to be changed directly without the setter, I think we'd have to normalise just-in-time in the render function. 

But I think this is a better approach all round - the options can be changed without the setter, and we wouldn't need to distinguish between 'raw' and 'normalised' options. The `normalizeOptions` method seems fast enough that we can call it just for rendering, and leave the observable options untouched.

lbwexler:
Tom and I discussed this at length and Tom has a better approach in a new pull request.  Basically if we set the component "options" in the props we give it, best to alter the options in the props we give it as well.  With the approach in this PR, we would need a ref to the component.

PR #283, opened 2018-05-14
LoadMask code enhancement / styling

gportela85:
https://github.com/exhi/hoist-react/issues/259


gportela85:
@amcclain Cleaned up the LoadMask code quite a bit, code rid of some extra wrapping containers and applied the semi-transparent background to it.
Tested with PIQ and Toolbox.

PR #285, opened 2018-05-14
Revise admin app to use panels where appropriate, Fixes #274

### LINKED ISSUE #274: Revise admin app to use panels where appropriate

febbraiod:


rest/RestGrid.js
```diff
@@ -23,15 +23,13 @@ export class RestGrid extends Component {
             {extraToolbarItems} = this.props,
             gridOptions = defaults(this.props.gridOptions, {onRowDoubleClicked: this.onRowDoubleClicked});
         return fragment(
-            vframe(
-                restGridToolbar({model, extraToolbarItems}),
-                frame(
```
febbraiod: Not sure why we were wrapping this grid in a frame. Our grid component already renders it's contents within a frame (a div with a flex: auto property). I can't think of a reason why we'd want this double nesting of frames so I removed. Looks the same to me but wanted to leave a comment in case there was a purpose here that I missed.

amcclain:
Looks great, thank you.  

Not sure how I feel about the remaining tabs where we have a simple vframe holding a toolbar and a main content panel - services, caches, users (might be more).  Panel ends up adding code / nesting here, but at the same time the consistency would be nice.  I could go either way but wanted to raise the point before merging.  

Thanks again.

febbraiod:
Totally agree with consistency regardless of a couple extra lines of code. I got it in my head that these were rest grids, but not so. Have a couple minor notes from Lee as well.

admin/App.js
```diff
@@ -20,15 +20,13 @@ export class App extends Component {
...
-        return vframe({
-            items: [
-                this.renderNavBar(),
-                frame({
-                    cls: 'xh-admin-app-frame',
-                    item: tabContainer({model: XH.appModel.tabs})
-                })
-            ]
+
+        return panel({
```
febbraiod: The entire app really looks like a panel too, any reason not to use here?

lbwexler:
I think consistency is the key, but also having the hook to insert additional toolbar related styling and features.  We may not have it now, but nice to have the hook for it later.

On a related note -- can we introduce a synonym for topToolbar/bottomToolbar as 'tbar' and 'bbar'.  Would be nice to have something that is less of a mouthful and in fact the same charachter length as 'item'.  Would also just be happy to bomb the long variants immediately out of existence.

amcclain:
Great - I had come close to suggesting bbar / tbar when we first released the component. Looping back on it now, it seems like an obvious choice and I am happy we are taking it.

PR #287, opened 2018-05-14
Component baseCls

gportela85:
https://github.com/exhi/hoist-react/issues/252

### DIFF
--- a/app/LockoutPanel.js
+++ b/app/LockoutPanel.js
@@ -26,17 +26,18 @@ export class LockoutPanel extends Component {
     render() {
         const msg = this.props.message || 'Access Denied';
 
-        return viewport(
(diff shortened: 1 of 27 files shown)


gportela85:
This solutions allows multiple baseCls to be merged, example:
Panel has `baseCls: 'xh-panel'`
if we create a `Foo` component: 

    class Foo extends Component {
        render() {
            return panel({
                baseCls: 'xh-foo',
                title: 'Foo',
                item: 'Bar'
            });
        }
    }

Classes will be "xh-panel xh-foo".

If we instantiate with a cls:
`foo({ cls: 'xh-foo-instance' });`

Classes will be: `xh-panel xh-foo xh-foo-instance`

PR #290, opened 2018-05-15
Make agApi observable

TomTirapani:
+ Rename 'gridOptions' > 'agOptions'
+ Rename 'gridApi' > 'agApi'


amcclain:
Excellent thank you. I like `agOptions` - happens to work very nicely with `ag` but I could imagine extending the theme with e.g. `hc` for  highcharts, `bp` for blueprint, etc. if we wanted to give that a try.

PR #293, opened 2018-05-16
Promote styles in elem, layout divs are HoistComponents

febbraiod:


febbraiod:
This won't work as it stands. Will put a style object on a component that won't always make it to the DOM. Closing until further notice.

PR #296, opened 2018-05-18
Add 'layoutSupport' option to HoistComponent

lbwexler:


lbwexler:
Went back and forth about all the naming here for a long time, so happy to discuss how I landed, and take any suggestions and make changes.

Also, would like to discuss the choice of defaulting layoutSupport to 'false'.  I think its safer.  I think there will be many more collisions and unexpected behavior if we enable globally for HoistComponent, and we really don't need to.  To use it properly, you do need to think about it, so don't think its the worst thing to make it more explicit.

PR #303, opened 2018-05-21
LeftRightChooser out of sync

gportela85:
https://github.com/exhi/hoist-react/issues/299

### DIFF
--- a/cmp/leftrightchooser/LeftRightChooserModel.js
+++ b/cmp/leftrightchooser/LeftRightChooserModel.js
@@ -158,12 +158,14 @@ export class LeftRightChooserModel {
     syncSelection() {
         const leftSel = this.leftModel.selection,
             rightSel = this.rightModel.selection,
+            leftSingle = leftSel.singleRecord,
+            rightSingle = rightSel.singleRecord,
             lastSelectedSide = this._lastSelectedSide;
 
-        if (leftSel.singleRecord && lastSelectedSide !== 'left') {
+        if (leftSingle && lastSelectedSide !== 'left') {
             this._lastSelectedSide = 'left';
             rightSel.clear();
-        } else if (rightSel.singleRecord && lastSelectedSide !== 'right') {
+        } else if (rightSingle && lastSelectedSide !== 'right') {
             this._lastSelectedSide = 'right';
             leftSel.clear();
         }


cmp/leftrightchooser/LeftRightChooserModel.js
```diff
@@ -158,12 +158,14 @@ export class LeftRightChooserModel {
     syncSelection() {
         const leftSel = this.leftModel.selection,
             rightSel = this.rightModel.selection,
+            leftSingle = leftSel.singleRecord,
```
gportela85: GridModel#selection is no longer observable, so we need to reference the computed value directly force syncSelection to run automatically

amcclain: This is another example of mobx autoruns being tricky at times to reason about.  I don't think the issue is some change to GridModel.selection - that's a selection model and not something that changes itself directly. 

The underlying issue is more that we always want to be looking at both sides of the selection at all times, but the deliberately "lazy" nature of mobx is preventing that here due to the conditional logic.

Starting from scratch with no selection, a click on the left side triggers the first branch of the conditional. The second condition is not evaluated, so at that point the right selection is not being watched as the autorun ran without hitting it. So when we then click on the right side, nothing changes - leftSingle didn't change, and rightSingle wasn't being watched.

Moving the variable access up above the conditional fixes the issue, but it's not obvious that's important. You can easily imagine someone coming along in a day or a year and thinking "oh these vars are only accessed once, let me save two lines of code and put them inline."  That would be totally valid JS, but break the function as Mobx depends on the side effect of them being accessed to be watched.

@TomTirapani has been flagging this as a reason why reactions are often better / easier to reason about. I just reviewed with @lbwexler and he's putting what we think will be an even clearer syntax into place around registering reactions and being very explicit about what you are watching and what code will definitely then run when any watched observable changes.  Stay tuned for that very shortly!

amcclain: edit - if `selection` was previously the actual selection itself, vs. a model object, that would have always been watching both sides - I'm guessing that's likely what happened and changed. See #306 which looks to remove this source of confusion.

lbwexler: Ah -- I see the problem now -- its that _lastSelectedSide can change (and is changed in both branches)  and it is *not* observable.  For a method like this to work, all the mutable properties mush be observable...

Tricky indeed.

I am closing this PR -- I have fixed using the new addReaction syntax in 

https://github.com/exhi/hoist-react/pull/313

PR #304, opened 2018-05-21
 Match in-app var names to their xh counterparts

febbraiod:


amcclain:
Yikes thanks for catching these!  👍

PR #308, opened 2018-05-22
AppModel should validate that user has basic access to client UI

cnrudd:


app/AppContainer.js
```diff
@@ -110,5 +115,14 @@ export class AppContainer extends Component {
...
         XH.handleException(e, {requireReload: true});
     }
+
+
+    //------------------------
+    // Implementation
+    //------------------------
+    unauthorizedMessage() {
+        const role = XH.appModel.requireRole;
+        return span(`Permission "${role}" is required to use this application.`, <br />, 'Please request this permission from the Help Desk.');
```
amcclain: I don't think we need to bake in concept of Help Desk at this level - apps might well have their own support list or procedure.  Let's leave that off and if we need support for a more customized message in the future we can decide how we wish to implement.

core/XH.js
```diff
@@ -282,6 +287,11 @@ class XhModel {
         );
     }
 
+    userHasRequiredRole(usersRoles) {
+        if (this.appModel.requireRole == 'OPEN_TO_ALL') return true;
+        return usersRoles.includes(this.appModel.requireRole);
```
amcclain: There is `XH.getUser().hasRole()` which we might wish to use here.

app/AppContainer.js
```diff
@@ -65,6 +66,10 @@ export class AppContainer extends Component {
                 return viewport(loadMask({isDisplayed: true}));
             case LoadState.LOGIN_REQUIRED:
                 return loginPanel();
+            case LoadState.UNAUTHORIZED:
```
lbwexler: Let's call this 'ACCESS_DENIED' to link up to the the 'XH.accessDeniedMessage'  (or vice-versa, just need to be consistent.)

app/HoistAppModel.js
```diff
@@ -41,6 +41,13 @@ export function HoistAppModel(C) {
             get() {return false}
         },
 
+        /**
+         * Override in apps appModel to check if user has required role.
```
lbwexler: Let's change the signature of this so that it actually gets passed the user object as the first argument.  Will just make things much simpler and self documenting as to what an application should typically do with this method.

message could be  'Access control not yet implemented for this application.  See AppModel.checkAccess()'

Also, can we update the comment?

app/LockoutPanel.js
```diff
@@ -19,20 +20,26 @@ import './LockoutPanel.scss';
...
-                item: msg
-            })
-        });
+        return viewport(
+            vframe(
+                impersonationBar(),
+                filler(),
+                box({
+                    cls: 'xh-lockout-panel',
+                    justifyContent: 'center',
```
lbwexler: Can probably add all the styles here to the css class for consistency

app/AppContainer.js
```diff
@@ -122,19 +122,19 @@ export class AppContainer extends Component {
...
+            vspacer(10),
+            `
+                You are logged in as ${user.username} 
+                and have the roles [${user.roles.join(', ') || '--'}].
+            `,
+            vspacer(20),
+            logoutButton({
+                text: 'Logout',
+                omit: !XH.appModel.enableLogout
             })
```
amcclain: @cnrudd  and @lbwexler - I didn't see a way that we would need to be refreshing the user dynamically here after returning this message element, hence the tweak to simplify and put the user object itself in a const.  Let me know if I am missing something here!

I also thought it worth using our `omit` support and `vspacer` elem.

PR #312, opened 2018-05-22
Set intent to danger if numberField fails min/max validation

febbraiod:


lbwexler:
I am not sure about this change.  Seems like we want a more general story around validation and display of validation state in hoist controls.  This seems pretty specific and somewhat arbitrary.

How does this differ from just entering a non-number?  Would love to discuss more.

gportela85:
Just wanted to add some notes here, I feel like HoistField should have a validation method and each component extending HoistField should have it's own implementation of the validator.

lbwexler:
Gil -- yes, something like that is what I was thinking.    Please feel free to start sketch out a design in this ticket, if you have any ideas?   Maybe a list of the template methods, what they return, how an app would use them to modify validation, and how a control writer would use them to display validation?

Note:  I do *not* want to start coding this without a plan.  This is a tricky thing to make useful to application -- very easy to build something that is unusable....

lbwexler:
Closing this in favor of https://github.com/exhi/hoist-react/issues/317

PR #316, opened 2018-05-23
HoistField - commitOnChange option

gportela85:
I ran into a couple of cases where having the field being committed on change made more sense than waiting for blur.
So I thought it would be better to have a config for it.


amcclain:
Hi @lbwexler Gil and I discussed this - would be happy to review anytime

cmp/form/dropdown/BaseDropdownField.js
```diff
@@ -25,7 +25,9 @@ export class BaseDropdownField extends HoistField {
 
     static defaultProps = {
         placeholder: 'Select'
-    }
+    };
+
+    commitOnChange = true;
```
lbwexler: seems weird that we need to write this prop directly on the component, and override it there.  Is there no better way to set a default prop for these hoist fields to be different from the inherited value?

gportela85:
@lbwexler This change will work with all `HoistFields` except fields extending `BaseDropDownField`. Manually adding the option to `Text` / `TextArea` / `Number` etc.. seems like a lot of boilerplate. Do you think a better alternative is to just document that `commitOnChange: false` is not supported with `BaseDropDownField` ?

amcclain:
Note that toolbox has this config in use checked in on master.  Noting that here so we are sure to resolve this any which way

cmp/form/dropdown/BaseDropdownField.js
```diff
@@ -19,14 +20,21 @@ import {HoistField} from '../HoistField';
...
     static defaultProps = {
-        placeholder: 'Select'
-    }
+        placeholder: 'Select',
+        commitOnChange: true
+    };
 
+    constructor(props) {
+        super(props);
+        throwIf(!props.commitOnChange, 'A commitOnChange value of false not implemented on BaseDropdownField.');
```
amcclain: @gportela85 / @lbwexler - I am very sorry I missed this, but what is going on here with commitOnChange being forced to true for BaseDropdownField??? 

This is a badly breaking change for one of our client apps I was just prepping for release where we have a QueryComboField that allows users to type in a company ticker. The combo provides suggestions from their prior entries, but doesn't requireSelection - in fact often a user will be entering a new ticker as the lookup values suggested are only for convenience and rely on the user having entered them before.

It is important that this field not flush its value until the user takes a "commit" action such as blurring, pressing enter, or selecting a complete value from the suggest list. Now it fires a change event on every keystroke, resulting in a service call that throws an exception due to bogus / partial input. 

This was exactly the anti-pattern we were looking to _prevent_ with the entire commit system, and why we wanted commitOnChange to be a relatively rare opt-in prop. We cannot assume every change to the contents of a  BaseDropdownField superclass field is a selection as these allow partial type-ahead input.

Let me know if I am missing something here - we will need to block a release until we can resolve.

Issue #315, opened 2018-05-23
Show specific exception when grails server is unavailable at page load time


lbwexler:
Right now, we just throw a wonky exception related to the particular ajax call that failed --  the initial authentication.

Would be good to recognize the actual HTTP error from the request (vs. failed auth) and let the user clearly know that the server is unavailable.

amcclain:
Yes thank you for filing!  Especially since in the SSO case we report a "failed to SSO" message which makes it seem very much like something is misconfigured with auth in particular, vs. the entire back-end not being available.  

Would vote for message like "Unable to contact the UI server."  Not sure if there is too much else we can put in there of use......

PR #318, opened 2018-05-24
Implementation of server unavailable handling. (#315)

cnrudd:


svc/FetchService.js
```diff
@@ -78,9 +78,17 @@ export class FetchService {
...
 
-        const ret = await fetch(url, opts);
-        if (!ret.ok) throw Exception.requestError(opts, ret);
-        return ret;
+        try {
+            const ret = await fetch(url, opts);
+            if (ret && !ret.ok) throw Exception.requestError(opts, ret);
+            return ret;
+        } catch (e) {
+            if (e == 'TypeError: Failed to fetch') {
```
cnrudd: The value of 'e' here seems special to me, and may vary among browsers.  This is the value in Chrome.  'e' was undefined in the console.   The value of e could only be discovered by doing console.log(e).

svc/FetchService.js
```diff
@@ -78,9 +78,17 @@ export class FetchService {
...
 
-        const ret = await fetch(url, opts);
-        if (!ret.ok) throw Exception.requestError(opts, ret);
-        return ret;
+        try {
+            const ret = await fetch(url, opts);
+            if (ret && !ret.ok) throw Exception.requestError(opts, ret);
+            return ret;
+        } catch (e) {
+            if (e.message == 'Failed to fetch') {
```
cnrudd: e has the following properties:
            console.log(e.name)  = TypeError
            console.log(e.message)  = Failed to fetch
            console.log(e.status) undefined
            console.log(e.stack) TypeError: Failed  to fetch

It seemed save to test on 'message'.   But message's value may be browser specific.  this is the value on Chrome.

exception/Exception.js
```diff
@@ -83,6 +83,14 @@ export class Exception {
         return this.createInternal(defaults, {});
     }
 
+    static serverUnavailable(requestOptions) {
+        return this.createInternal({
+            name: 'Server Unavailable',
+            message: 'The application server is not available.',
```
cnrudd: Do we want to say more?  like:  "If your internet connection is good and this condition persists, please contact ...

svc/FetchService.js
cnrudd: This thread: https://github.com/github/fetch/issues/310  suggests the error msg may vary by fetch implementation.

PR #326, opened 2018-05-26
Fixed LogViewer tail mode

gportela85:
https://github.com/exhi/hoist-react/issues/325


admin/tabs/logging/viewer/LogViewerDisplay.js
```diff
@@ -69,12 +70,21 @@ class LogViewerDisplay extends Component {
...
+        } else if (idx === 0) {
+            return this.firstRow.ref;
         }
+
+        return undefined;
+    }
+
+    syncTail() {
+        const {tail} = this.model,
+            rowElem = this[tail ? 'lastRow' : 'firstRow'].value;
```
lbwexler: Doesn't this lose the scroll when you refresh and are not in tail mode?

PR #331, opened 2018-05-30
Enable grid state, Fixes #314

### LINKED ISSUE #314: Grid - Component State
Building off of the recent merge of column chooser, next step with Hoist-level grid customizations includes support for component state to save user column choices, ordering, and sorting.

I propose an initial goal of general feature parity with the hoist-sencha implementation, ideally re-using the same server-side support in hoist-core with minimal or no modifications. We can push on further than this as we go along to support additional features, but restoring support for basic state is a top priority.

Don - if you could review the [hoist-sencha implementation](https://github.com/exhi/hoist-sencha/blob/develop/grails-app/assets/src/xh/desktop/grid/plugin/State.js) that would be an excellent starting point. We can discuss if we want certain features (e.g. width tracking) to be in or out of scope. Main thing is to grok the complexities around restoring saved state in a safe manner that can cope with changes to columns in the code vs. potentially stale user state.

febbraiod:
Opening for review, contains test/demo code that must be removed before merge (5/30/18)


admin/tabs/configs/ConfigPanel.js
```diff
@@ -157,11 +160,17 @@ export class ConfigPanel extends Component {
...
+        return [
+            button({
+                icon: Icon.diff(),
+                text: 'Compare w/ Remote',
+                onClick: this.onDifferBtnClick
+            }),
+            colChooserButton({gridModel}),
+            button({text: 'GroupBy Type', onClick: () => gridModel.setGroupBy('valueType')}),
+            button({text: 'Clear Grouping', onClick: () => gridModel.setGroupBy(null)})
+        ];
```
febbraiod: All code in config panel is for Demo purposes, remove before merge.

cmp/grid/GridStateModel.js
```diff
@@ -0,0 +1,196 @@
...
+
+        return start(() => {
+            this._resetting = true;
+            this.loadState(defaultState);
+        }).then(() => {
+            this.resetState(this.getStateKey());
+        }).finally(() =>{
+            this._resetting = false;
+        });
+    }
```
febbraiod: Not sure if this method needs to be async. It is in Hoist-Sencha, but I think that maybe for a db based override to loading/reseting state. Not sure if we want to prepare for that now or keep it simpler for the moment.

cmp/grid/GridStateModel.js
```diff
@@ -0,0 +1,196 @@
...
+        const sortBy = this.state.sortBy;
+        if (sortBy) this.parent.setSortBy(sortBy);
+    }
+
+    //--------------------------
+    // Helper
+    //--------------------------
+    saveStateChange = debounce(function() {
+        this.saveState(this.getStateKey(), this.state);
+    }, 5 * SECONDS);
```
febbraiod: In Hoist-Sencha we check for the state object and whether or not we are resetting. I think we should always have a state object (even if it's empty).

As far as resetting, I think if we keep the asynchronous state reset we should check for it, but if we make that operation synchronous it's probably unnecessary.

cmp/grid/GridModel.js
```diff
@@ -131,7 +137,12 @@ export class GridModel {
 
     @action
     setGroupBy(field) {
-        const cols = this.columns;
+        const cols = this.columns,
+            groupCol = find(cols, {field});
+
+        // If we have an invalid groupBy field do not set
+        // Let a falsey field through to mean 'no grouping'
+        if (field && !groupCol) return;
```
febbraiod: Is this what we want in terms of behavior? Seems nice to have a way to reset the grouping, but I could also see 'false' meaning "don't change the grouping".

The latter might be helpful if we ever decide to save grouping as part of state. In which case, if we didn't have a groupBy field saved we'd still get the default grouping if we call this. We could also just NOT call this from the stateModel when loading state if we have no groupBy saved. (Perhaps I'm getting ahead of myself?)

PR #340, opened 2018-06-01
DataView emptyText

gportela85:
https://github.com/exhi/hoist-react/issues/332

### DIFF
--- a/cmp/dataview/DataViewModel.js
+++ b/cmp/dataview/DataViewModel.js
@@ -37,7 +37,7 @@ export class DataViewModel {
      * @param {BaseStore} store - store containing the data for the dataview.
      * @param {(StoreSelectionModel|Object|String)} [selModel] - selection model to use,
      *      config to create one, or 'mode' property for a selection model.
-     * @param {string} emptyText - empty text to display if DataView has no records. Can be valid HTML.
+     * @param {string} [emptyText] - empty text to display if DataView has no records. Can be valid HTML.
      * @param {function} [contextMenuFn] - closure returning a StoreContextMenu().
      */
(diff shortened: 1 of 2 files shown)


cmp/dataview/DataViewModel.js
```diff
@@ -32,23 +32,26 @@ export class DataViewModel {
...
     /**
+     * @param {function} [contextMenuFn] - closure returning a StoreContextMenu().
+     * @param {string} emptyText - empty text to display if DataView has no records. Can be valid HTML.
      * @param {function} itemFactory - elemFactory for the component used to render each item.
      *      Will receive record via its props.
-     * @param {BaseStore} store - store containing the data for the dataview.
      * @param {(StoreSelectionModel|Object|String)} [selModel] - selection model to use,
      *      config to create one, or 'mode' property for a selection model.
-     * @param {function} [contextMenuFn] - closure returning a StoreContextMenu().
+     * @param {BaseStore} store - store containing the data for the dataview.
```
lbwexler: This change of course looks fine, but why are we reordering the arguments in the doc string?  Seems like our practice has been that the most important, required arguments go first, and the more secondary (and often optional) arguments follow.  Are you suggesting we go to an alphabetical model?

gportela85: Exactly..  for components with a very small number of configs order them alphabetically, and components with a lot of configs we can pin the most important at the top.
What do you think?

lbwexler: I think we need one rule or another -- how do you determine "a lot of configs".  Let's stick with the existing practice of ordering by importance.

PR #346, opened 2018-06-05
Tweak flex prop for charts, Fixes #342

### LINKED ISSUE #342: Fix Chart component auto-sizing
When 3 charts are aligned side by side in an hframe, the expectation is that all 3 charts share the total width equally.  I find instead that the 3 charts are rendered with different widths; 1st one ~ 600px, 2nd and 3rd quite a bit smaller, with large gaps between the charts.

A possible cause of this issue is that the highcharts lib sets its chart's size by looking for a predefined 'width' css value on the chart's containing div.   When it doesn't find this, it defaults the chart's size to 600px wide. 

In an hframe flexbox with 3 chart children, the chart children widths are not specified in code, since the hframe is itself expected to fill an arbitrary width, and the 3 chart children are expected to divide that width equally into 3.

We might look into whether one these libs has already solved this problem:
https://www.npmjs.com/package/highcharts-react-official
https://www.npmjs.com/package/react-highcharts
https://www.npmjs.com/package/react-jsx-highcharts

febbraiod:


febbraiod:
I have a demo set up on the line chart tab in toolbox. Currently showing different ways to override the default:

https://github.com/exhi/toolbox/tree/chartFlexDemo

If the charts are allowed to get the default the layout, it will give us three evenly spaced charts.

PR #347, opened 2018-06-05
 Provide token parsing for storeContextMenus, Fixes #300

### LINKED ISSUE #300: Grid context menu - string token support for column chooser option
In cases where you are not using the default grid context menu, would like easy way (via string token) to request an item for the column chooser.  Should respect enableColChooser flag on GridModel.

febbraiod:
*Specifically for colChooser right now


febbraiod:
Note: This will be a breaking change for any apps using storeContextMenus as the constructor takes and object now.

PR #358, opened 2018-06-07
Fix requiredSelection behavior in BaseComboField, Fixes #353, Fixes #354

### LINKED ISSUE #353: Empty ComboField throws when requireSelection:true and option selected
Can reproduce in Toolbox forms tab:

+ Add `requireSelection: true` to the "State" comboField
+ Type a character in empty-to-start field to pop up list of options
+ Choose one - exception thrown to console

### LINKED ISSUE #354: Combofield commits interim entry when pressing enter or clicking to select a suggested choice
Can replicate in Toolbox Forms example with "State" combo.

+ This is currently a ComboField with default of `requireSelection:false` 
+ Type e.g. h to show pop-up suggestion of Hawaii
+ Press enter to select that state while carefully looking at the output on the right that shows the values.  
+ You will see "h" flash briefly then "HI".
+ Can see the same when clicking an option.

Field should not commit that intermediate value if you are selecting a displayed option. Realize this is likely tricky due to commit handling on blur/enter but for these fields hoping we can do better if we know the popup is shown.

febbraiod:


cmp/form/dropdown/BaseComboField.js
```diff
@@ -31,20 +36,18 @@ export class BaseComboField extends BaseDropdownField {
 
     forceSelectionToOptionOrRevert() {
         const {internalOptions, internalValue} = this;
+        if (!internalValue) return;
```
febbraiod: doCommit will fire onBlur and call this method. If this happens before an internalValue has been set e.g. when choosing an option from the dropdown this method would throw.

cmp/form/dropdown/BaseComboField.js
```diff
@@ -22,6 +22,11 @@ export class BaseComboField extends BaseDropdownField {
         }
     }
 
+    onItemSelect = (val) => {
+        this.noteValueChange(val.value);
+        super.doCommit();
```
febbraiod: Override to prevent the call to the local doCommit method. If we are selecting from the menu, we want to skip the forceSelectionToOptionOrRevert check which will attempt to compare the options label to this set value.

cmp/form/dropdown/BaseComboField.js
```diff
@@ -31,20 +36,18 @@ export class BaseComboField extends BaseDropdownField {
 
     forceSelectionToOptionOrRevert() {
         const {internalOptions, internalValue} = this;
+        if (!internalValue) return;
 
-        // 0) We have a match, nothing to be done
-        if (find(internalOptions, {label: internalValue})) {
```
febbraiod: Removing this step. If our option labels differ from our values, our intervalValue might be a label match here, but we don't want that to remain the value of the field. We'd want to noteValueChanged(match.value) as below.

amcclain:
Does this also end up fixing #354 ?  It appears to - is it clear why?

Apologies, this is a pretty complicated bit of code all around (the general field system + dropdown support) and I don't have a good grasp of it - hard to really get it in my head while multitasking. But I would love to take this if you are confident about the changes and how it relates to both tickets.

febbraiod:
I didn't cross reference. Will test and try to work out if/why it does.

febbraiod:
Closing in favor of Tom's change

PR #361, opened 2018-06-07
Hide credentials on error

gportela85:
https://github.com/exhi/hoist-react/issues/355

### DIFF
--- a/exception/ExceptionHandler.js
+++ b/exception/ExceptionHandler.js
@@ -70,7 +70,7 @@ export class ExceptionHandler {
         if (!requestOptions.params) return;
 
         params.forEach(param => {
-            delete requestOptions.params[param];
+            requestOptions.params[param] = '******';
         });
     }
 
--- a/exception/ExceptionHandler.js
+++ b/exception/ExceptionHandler.js
@@ -46,8 +46,8 @@ export class ExceptionHandler {
 
         options = this.parseOptions(exception, options);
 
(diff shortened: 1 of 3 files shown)


svc/FetchService.js
```diff
@@ -82,6 +82,10 @@ export class FetchService {
         try {
```
lbwexler: Don't think we want to complicate our general fetch service with knowledge about this very specific handling of this one type of fetch. 

The most general solution would be to look to sanitize any param with the name "password" in our exception handling code.  As a stopgap,  we might just have explicit exception handling code in LoginPanel.js, rather than using .catchDefault()

lbwexler:
Another general option here Gil is to add an argument to the exception handling code of the following form 

obfuscateParams: ['password'].

We could even default it to include 'password', but might let applications add other values.  Could be useful for trading data....

exception/ExceptionHandler.js
```diff
@@ -58,6 +64,16 @@ export class ExceptionHandler {
     //--------------------------------
     // Implementation
     //--------------------------------
+    static hideParams(requestOptions, params) {
+        delete requestOptions.body;
+
+        if (!requestOptions.params) return;
+
+        params.forEach(param => {
+            delete requestOptions.params[param];
```
lbwexler: would it be better if we just set the parameter to ‘*****’
otherwise somebody troubleshooting might think its just missing

PR #365, opened 2018-06-08
Combofield fixes, Fixes #353, Fixes #354

### LINKED ISSUE #353: Empty ComboField throws when requireSelection:true and option selected

### LINKED ISSUE #354: Combofield commits interim entry when pressing enter or clicking to select a suggested choice

TomTirapani:
+ Only attempt to commit onBlur and onKeyPress if !requireSelection. Allows us to remove label matching logic. Fixes #353.
+ Debounce commits. Display internalValue during wait to mask for user. Fixes #354.

### DIFF
--- a/cmp/form/dropdown/BaseComboField.js
+++ b/cmp/form/dropdown/BaseComboField.js
@@ -5,7 +5,7 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
-import {observable, setter, computed} from '@xh/hoist/mobx';
+import {observable, computed, action} from '@xh/hoist/mobx';
 import {debounce} from 'lodash';
 
 import {BaseDropdownField} from './BaseDropdownField';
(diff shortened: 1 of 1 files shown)


amcclain:
Thanks - added some comments and merging now.

As per my expanded comment on the blur handling, I see better now some of the issues @febbraiod was raising around not committing on blur and how a user might well expect that blur to auto-select a value from the suggestion list. Let's loop back with @haynesjm42 and take a look at some UI ideas he has around making it more clear when a user wishes to create a new custom value for a !requireSelection field.

PR #368, opened 2018-06-11
TabContainer changes to support detached 'switcher'. Fixes #320

### LINKED ISSUE #320: NavContainer Component
Propose a new NavContainer/NavModel pair that could replace / augment the current TabContainer/TabModel, but with a bit more flexibility in terms of the UI for switching between children.

TabContainer/Model provide several key features we want to use as part of our core in-app navigation, both for the primary sections/modules of an app and for smaller components where a tab-style approach is needed. These include lazy instantiation of children, an API for managed refreshing, and routing support. 

The downside to TabContainer is that it is currently coupled to a Blueprint `tabs` component that renders the tab-bar (switching UI) right alongside the container itself (as you would expect from a TabContainer). For the standard horizontal tabs, this takes up a meaningful slice of vertical real-estate. 
For more efficient use of space and flexibility, it would be better if we could have a container with the same underling capabilities but then use an alternate UI / component for switching the tabs. 

Specifically, propose we also add a NavControl component that could be bound to the same NavModel as a NavContainer but be placed elsewhere in the layout.  The component itself could reuse the tab-bar part of the BP UI or take another approach (i.e. segmented buttons), and it could be placed wherever convenient - e.g. in the top application toolbar alongside the title and other controls. 

NavContainer could be used w/o a NavControl if you had some use-case where navigation was _only_ done via routes or other programmatic means.

We could continue to support a TabContainer component that uses the same NavModel class, or is a NavContainer that installs the standard tab-bar UI in the normal place.  Or this could be an option on NavContainer.  (I would also be fine keeping the name TabContainer for this entire thing.)

Note that there is some special handling in TabContainer to support nested tabs - this is used in the admin client and we would want to consider and support this use case.

haynesjm42:


cmp/tab/TabPane.js
```diff
@@ -39,7 +39,6 @@ export class TabPane extends Component {
 
         return frame({
             display: isActive ? 'flex' : 'none',
-            margin: 4,
```
haynesjm42: From what I could tell, this was always getting blown away by some blueprint css with an `!important` flag setting the margin to 0

lbwexler:
John and I discussed -- he'll be building out examples of TabContainer in 'detached' and 'standard' mode in toolbox.  Should be helpful for development and review, and we'll have it ready to go for documentation.

cmp/tab/TabContainer.js
```diff
@@ -20,33 +20,36 @@ import './Tabs.scss';
 export class TabContainer extends Component {
 
     render() {
-        const {id, children, selectedId, vertical} = this.model;
+        const {tabPosition, vertical, selectedId, children} = this.model,
+            switcherBefore = tabPosition === 'left' || tabPosition === 'top',
```
lbwexler: Support for right/bottom is a cool enhancement we are getting!

lbwexler:
Hi:  I just want to say that I find this really cool and elegant.  Its a perfect us of the mobx driven model to synchronize these two pieces seamlessly.  

+ Is the idea that in your Apollo example, you would just put the "detached" TabSwitcher in a Navbar with some other stuff?

+ Can we enhance the class comments here, especially of the TabContainerModel and TabSwitcher?  Would be nice to discuss this.

+ Would be cool in toolbox to show TabContainer used with a little ComboBox as its switcher.  Also might help flesh out any issues with the API.

lbwexler:
Looks great.  Think this will be a case where we really want to document examples of how to use this in 'detached' mode -- this can be developed further in toolbox and elsewhere...

PR #369, opened 2018-06-11
Set sort properties on agColDefs, Fixes #324 

### LINKED ISSUE #324: Grid sort lost on setColumns
I noticed this while working on grid state. However, the issue was preexisting. If you have a grid with a column chooser this is easy to see. When showing/hiding columns on a grid, the chooser sets the columns with their adjusted 'hide' props. If your grid was sorted, it no longer is. Same effect when applying agGrid's column order to 'xh' columns prop on the grid model.

febbraiod:


febbraiod:
+Use reactions instead of autoruns for sync methods

PR #376, opened 2018-06-12
Added commitOnChange to SliderField

gportela85:


cmp/form/SliderField.js
```diff
@@ -24,6 +25,8 @@ import {HoistField} from './HoistField';
 export class SliderField extends HoistField {
```
lbwexler: Sorry -- Just noticed this.  Seems very weird to me that just one HoistField would have layoutSupport.

gportela85: Yes, by the default the slider always takes 100% of the room available and applying styles to it has no effect. So I had to wrap it with a box to make it work, otherwise the labels would get cut off. Maybe it's blueprintjs issue.
Spent sometime with @TomTirapani  on this problem this morning

lbwexler: yes, but could we do that workaround internally to the control?  Why does the workaround require it to implement LayoutSupport

lbwexler:
Merging this -- my issues with the layoutSupport of this are unrelated to the PR.  Sorry to conflate the two.

Issue #307, opened 2018-05-21
Hoist AppBar component for standardized application header


amcclain:
Propose a lightweight component to wrap Blueprint's NavBar and further standardize aspects of an application's top header toolbar.

This could be a good point to extend with some additional behavior - e.g. drop-down menus - as we go along. Start simple and get a hook in vs. continuing to copy ad-hoc direct usage of BP component across multiple apps.

Proposed props:
* icon - Icon element - optional - comp could auto-apply 2x size
* title - string - optional, defaults to XH.appName - comp wraps in header cls
* leftItems - node [] - optional, added to left side of bar after above
* rightItems - node [] - optional, added to right side of bar before below
* adminButton - bool - default true adds adminButton - already shows for admins only
* themeButton - bool - default true adds switcher Button
* logoutButton - bool - default true adds logoutButton - already shows if appModel.enableLogout
* refreshButton - bool - we could revisit concept of AppModel firing globalRefresh event with new event support now added by decorators.

Open to any other ideas re. the standard buttons in particular.

PR #378, opened 2018-06-13
AppBar Component

haynesjm42:
#307 

I left the built-in refresh button out since there isn't yet a consistent way to refresh an app, seems like that is a different story and the standard button can be added to AppBar once that mechanism is decided on.

I also updated the Admin app to put the TabSwitcher in the AppBar.

See https://github.com/exhi/toolbox/pull/18 for toolbox usage of the new component

### DIFF
--- a/admin/App.js
+++ b/admin/App.js
@@ -6,15 +6,14 @@
  */
 
 import {Component} from 'react';
-import {navbar, navbarGroup, navbarHeading, button} from '@xh/hoist/kit/blueprint';
-import {XH, HoistComponent} from '@xh/hoist/core';
+import {button} from '@xh/hoist/kit/blueprint';
+import {HoistComponent, XH} from '@xh/hoist/core';
 import {lockoutPanel} from '@xh/hoist/app';
-import {tabContainer} from '@xh/hoist/cmp/tab';
+import {tabContainer, tabSwitcher} from '@xh/hoist/cmp/tab';
 import {frame, panel} from '@xh/hoist/cmp/layout';
-import {logoutButton, themeToggleButton, refreshButton} from '@xh/hoist/cmp/button';
+import {refreshButton} from '@xh/hoist/cmp/button';
 import {Icon} from '@xh/hoist/icon';
 import {appBar} from '@xh/hoist/cmp/appbar';
-import {tabSwitcher} from '@xh/hoist/cmp/tab';
 
 import './App.scss';
 
@@ -44,7 +43,7 @@ export class App extends Component {
             leftItems: [
                 tabSwitcher({
                     model: this.model.tabs,
-                    large: false,
+                    large: false
                 })
             ],
             rightItems: [
--- a/admin/App.js
+++ b/admin/App.js
@@ -41,10 +41,7 @@ export class App extends Component {
             icon: Icon.gears({size: '2x'}),
             title: `${XH.appName} Admin`,
             leftItems: [
-                tabSwitcher({
-                    model: this.model.tabs,
-                    large: false
-                })
+                tabSwitcher({model: this.model.tabs})
             ],
             rightItems: [
                 button({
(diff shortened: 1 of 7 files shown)


amcclain:
Thanks John - looking forwards to checking this out shortly.  I was thinking I would take a run through the current hoist-react apps and migrate them to a remote switcher, but I think it makes more sense to see if we can take an even deeper standardization / cleanup with this component.  Will post any comments or questions shortly.

admin/App.js
```diff
@@ -36,49 +37,44 @@ export class App extends Component {
...
-            ]
+            ],
+            adminButton: false
         });
     }
 
     onContactClick = () => {
         window.open('https://xh.io/contact');
-    }
+    };
```
amcclain: Have wanted to ask what we wanted as our standard here for semicolons after class properties. I think adding them is most consistent with our general standards. 

As per  https://github.com/tc39/proposal-class-public-fields/issues/25 it seems like it's generally more correct and avoids relying on ASI.

I took a quick look at our eslint rules and not entirely sure why it's not flagging them - https://github.com/exhi/hoist-dev-utils/blob/82d45641a1f9016c720e5e3a73a19322b4520e3f/packages/eslint-config/index.js#L83

Anyway - we should file a separate ticket to bomb the code and standardize if we go this way, and ideally get eslint to flag.

cmp/tab/TabSwitcher.js
```diff
@@ -39,7 +39,8 @@ export class TabSwitcher extends Component {
             onChange: this.onTabChange,
             large: !vertical,
             selectedTabId: selectedId,
-            items: children.map(({id, name}) => tab({id, title: name}))
+            items: children.map(({id, name}) => tab({id, title: name})),
+            ...this.props
```
amcclain: This we should pull the model off of the props to avoid passing that through to the underlying Blueprint tabs component (in case it decided to complain about an unknown prop or, worse, declare its own model prop with a different meaning).

cmp/appbar/AppBar.js
```diff
@@ -0,0 +1,73 @@
...
+        /** Items to be added to the left side of the AppBar, immediately after the title. */
+        leftItems: PT.node,
+        /** Items to be added to the right side of the AppBar, before the standard buttons. */
+        rightItems: PT.node,
+        /** Set to true to hide the Launch Admin button. */
+        hideAdminButton: PT.bool,
+        /** Set to true to hide the Theme Toggle button. */
+        hideThemeButton: PT.bool,
+        /** Set to true to hide the Logout button. */
+        hideLogoutButton: PT.bool
```
amcclain: We could call out in this comment that the button won't render for apps that don't support logout (as specific in their AppModel), to avoid confusion for these SSO-enabled apps.

admin/App.js
```diff
@@ -36,49 +37,44 @@ export class App extends Component {
...
-                        navbarHeading(`${XH.appName} Admin`)
-                    ]
+        return appBar({
+            icon: Icon.gears({size: '2x'}),
+            title: `${XH.appName} Admin`,
+            leftItems: [
+                tabSwitcher({
+                    model: this.model.tabs,
+                    large: false
+                })
```
amcclain: The one feature question that's been in my mind re. this comp is if it should also bundle up the creation of a tabSwitcher - could support a `tabContainerModel` prop that, if given, creates this automatically.  That might be a useful convenience and help to further standardize.  (Also looking at the need for `large: false` here which is a bit unfortunate if we do want to use this pattern frequently.)

cmp/tab/TabSwitcher.js
haynesjm42: Nice catch!

admin/App.js
haynesjm42: Yeah was thinking the same, makes sense to standardize on the main app nav switcher I think.

Re: `large: false` on the `tabSwitcher`, I'd prefer that we not default to large tabs in `tabSwitcher`. I kept it in there when I did the TabContainer re-work. Seems like it was mostly there because the top-level tabs were typically used for the app nav, with potentially 1 level of nesting using vertical tabs, so the larger horizontal tabs made sense. Any objection to just taking large tabs for horizontal out of `tabSwitcher`?

amcclain:
Merging this now to include it in 7.0 release this evening or tomorrow.  Lee I am sorry if this rushes a second review - of course we can loop back but I think the updates John made are ones we all generally agreed upon.

PR #383, opened 2018-06-14
Cleaner 4XX & 5XX error presentation to user.

cnrudd:


exception/Exception.js
```diff
@@ -15,37 +15,32 @@ import {isString} from 'lodash';
  */
 export class Exception {
 
+    /**
+     * Create and get back a Javascript Error object
+     * @param {(Object)} cfg - properties to add to the Error object
```
amcclain: Did we mean this to be `{(Object|string)}`?

exception/ExceptionHandler.js
```diff
@@ -115,4 +117,10 @@ export class ExceptionHandler {
     static sessionExpired(exception) {
         return exception && exception.httpStatus === 401;
     }
+
+    static cleanStack(exception) {
+        // statuses of 0, 4XX, 5XX are server errors, so the javascript stack
+        // is irrelevant and potentially misleading
+        if (/^[045]/.test(exception.httpStatus)) delete exception.stack;
```
amcclain: That's great - I had been noticing this and am glad we cleaned it up here.  (Assuming that in general we weren't getting useful info here in terms of what code path triggered the call to the server.)

exception/Exception.js
cnrudd: Yes. I've corrected this in the develop branch.

exception/ExceptionHandler.js
cnrudd: Correct.  The stack was starting with the line that threw the error as a response to the bad server response.

exception/Exception.js
amcclain: Thanks much!

PR #386, opened 2018-06-15
appBar separator extra margin

gportela85:
https://github.com/exhi/hoist-react/issues/384

### DIFF
--- a/cmp/appbar/AppBar.scss
+++ b/cmp/appbar/AppBar.scss
@@ -11,15 +11,4 @@
     font-size: var(--xh-navbar-title-font-size-px);
     margin: 0 var(--xh-pad-px);
   }
-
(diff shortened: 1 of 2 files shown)


gportela85:
Fixed and tested using the `appBar` branch in the toolbox repo.

screen shot 2018-06-14 at 11 07 48 pm

haynesjm42:
@amcclain I thought about adding the separator between `rightItems` and the standard buttons as well, but decided to leave it up to whoever is setting up the appBar since I expect the most common use-case for `rightItems` is to add additional buttons like our standard ones, and I wouldn't want them separated visually.

I considered adding something like an `appButtons` prop so you could add buttons in that standard place for your app, while also adding other items right-aligned to the bar, and in that case would make sense to install a separator between them.

PR #398, opened 2018-06-19
Config differ improvements, Fixes #380, Fixes #381, Fixes #392

### LINKED ISSUE #380: Admin config differ requires "/api" to be appended to URL
In a deployed app the grails endpoints are all exposed under /api via nginx proxy config.

Would like user of config differ and endpoints encoded into appInstances config to not have to know / depend on this.

### LINKED ISSUE #381: Admin config differ "Accept Remote" button broken in popup
Parent model method expects array.  Note this was changed in e409ea68bf100f1998b532570777712ef2c0fed4

### LINKED ISSUE #392: Admin config differ generates spurious diffs for JSON
Not positive on this but it seems like the differ generates JSON diffs when I would not expect them.  Want to see if this is more strict somehow than it should be - or thinking it would be better for JSON configs to ignore whitespace entirely.

febbraiod:
Fixes "Accept Remote" button in popup
Admin config differ requires added "/api" URL in a deployed app
Improved JSON diffing
Triggers a reload of underlying config panel on apply remote

### DIFF
--- a/admin/tabs/configs/differ/ConfigDiffer.js
+++ b/admin/tabs/configs/differ/ConfigDiffer.js
@@ -47,7 +47,7 @@ export class ConfigDiffer extends Component {
                     placeholder: 'https://remote-host/',
                     field: 'remoteHost',
                     width: 200,
-                    options: XH.getConf('xhAppInstances')
+                    options: XH.getConf('xhAppInstances').filter(it => it != window.location.origin)
                 }),
                 button({
                     text: 'Load Diff',
--- a/admin/tabs/configs/differ/ConfigDifferModel.js
+++ b/admin/tabs/configs/differ/ConfigDifferModel.js
@@ -84,12 +84,14 @@ export class ConfigDifferModel  {
     }
 
     async loadAsync() {
-        const remoteHost = `${trimEnd(this.remoteHost, '/') + (XH.baseUrl == '/api/' ? '/api' : '/')}`;
+        const remoteHost = trimEnd(this.remoteHost, '/'),
+            apiAffix = XH.baseUrl[0] == '/' ? XH.baseUrl : '/',
+            remoteBaseUrl = remoteHost + apiAffix;
 
         try {
             const resp = await Promise.all([
                 XH.fetchJson({url: XH.baseUrl + 'configDiffAdmin/configs'}),
-                XH.fetchJson({url: remoteHost + 'configDiffAdmin/configs'})
+                XH.fetchJson({url: remoteBaseUrl + 'configDiffAdmin/configs'})
             ]).linkTo(XH.appLoadModel);
             this.processResponse(resp);
         } catch (e) {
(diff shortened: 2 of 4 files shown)


admin/tabs/configs/differ/ConfigDifferModel.js
```diff
@@ -82,10 +84,12 @@ export class ConfigDifferModel  {
     }
 
     async loadAsync() {
+        const remoteHost = `${trimEnd(this.remoteHost, '/') + (XH.baseUrl == '/api/' ? '/api' : '/')}`;
```
febbraiod: I'm dubious about this. Not sure if nginx will ever change how it changes the backend urls but this seems pretty hard coded. That to one side, I would rather we had a XH.isDeployed (or some variation) to check for this switch which would at least make the code more semantic.

Even better a XH.apiSuffix which could be set to an empty string in development.

admin/tabs/configs/differ/ConfigDiffer.js
```diff
@@ -47,7 +47,7 @@ export class ConfigDiffer extends Component {
                     placeholder: 'https://remote-host/',
                     field: 'remoteHost',
                     width: 200,
-                    options: XH.getConf('xhAppInstances')
+                    options: XH.getConf('xhAppInstances').filter(it => it != window.location.origin)
```
febbraiod: This will not properly filter out the localhost:8080 in dev (since origin will == 'http://localhost:3000'), but I don't think development is where we are really concerned about this behavior

admin/tabs/configs/differ/ConfigDifferModel.js
amcclain: Realizing this is tricky to get really correct as it relates to the remote host and knowing what its baseUrl configuration (and therefore proxy configuration) might be, which isn't easily knowable.

However I think it makes a lot of sense to assume that a remote host has the same baseUrl value as the current host - i.e. that they share the same convention - so checking XH.baseUrl here does make sense to me and seems expedient.

Rather than hard-code the check to look for /api/ however, can we instead check if it's a relative URL and then append it, whatever it is?  That way if for some reason you had a collection of app instances that had decided they didn't want to use our /api/ convention and wanted to standardize on something like /grailsBackEnd/ this code could still work.

amcclain:
Didn't run locally yet to test it out but thank you for getting this in - reading through the code looks good with the one comment above.  Will be great to have this merged shortly.

admin/tabs/configs/differ/ConfigDifferModel.js
febbraiod: Great. I had a comment here to a similar effect yesterday (which i deleted after review with Lee). In addition, this line prompted the idea for XH.isLocalDevelopment. I guess it wouldn't be used here under this plan, but still might be useful to have in general?

lbwexler:
This is a great set of improvements.
We can (and will) continue to refine how we generate paths.

PR #407, opened 2018-06-21
Refactor of TabContainer and related routing (Fixes #391)

### LINKED ISSUE #391: TabContainer should have layoutSupport enabled
2 Pieces to this

1) TabContainer should have layoutSupport:true
Useful in general as per our standard for core Hoist comps, but in particular we will often have top-level app tabContainers wanting a flex:1 in the overall layout. Should look for any current apps nesting a top-level tab container in a frame to get this effect and clean that up.

2) switcherPosition should be moved from TabContainerModel to being a prop on TabContainer. 
This has no effect on the state -- its purely a layout related concept.

haynesjm42:
https://github.com/exhi/hoist-react/issues/391

Added `layoutSupport: true` with the standard default behavior I saw in other usages to TabContainer.

Moved `switcherPosition` to be a prop of TabContainer instead of coming from TabContainerModel. One complication from this was how to specify the position for nested containers, so I added `componentProps` to TabContainerModel as a mechanism for passing props through to child containers. Very open to other suggestions for how to handle this case..

See https://github.com/exhi/toolbox/pull/19 for associated toolbox app changes


cmp/tab/TabSwitcher.js
```diff
@@ -17,23 +17,32 @@ import {omit} from 'lodash';
...
     static propTypes = {
         /** Model to switch tabs on. Should be shared with a TabContainer. */
-        model: PT.instanceOf(TabContainerModel)
+        model: PT.instanceOf(TabContainerModel),
+        /** The position relative to the TabContainer this switcher is controlling. Defaults to 'top'. */
+        orientation: PT.oneOf(['top', 'bottom', 'left', 'right'])
+    };
+
+    static defaultProps = {
+        orientation: 'top'
```
lbwexler: We haven't been using defaultProps in general, because they are not flexible enough to handle alot of the dynamic defaults that we do.  Instead we have been generally implementing the defaults in code in the render method.  We should discuss this further -- obviously what you have here is pretty succinct

admin/App.js
```diff
@@ -156,7 +163,9 @@ export class App {
             },
             {
                 id: 'preferences',
-                switcherPosition: 'left',
+                componentProps: {
```
lbwexler: I see why we need this now -- am going to look a bit more carefully -- can we specify the children as an array of functions that would produce elements and  let us be more general?

I am looking at this today and will report back

cmp/tab/TabSwitcher.js
amcclain: Agreed good topic for discussion - I have a branch with a bunch of HoistField class related standardizations / cleanups that I need to get back to.  There we had some pretty heavy use of defaultProps and IIRC I continued in that direction....

admin/App.js
amcclain: Would love to discuss this some more.  Realizing that the abstraction here of child components-to-be into configs for the model (vs. declared children) might be one of the more surprising aspects of the overall component. 

I wonder if we could still achieve lazy-instantiation while keeping the parent / children relationships defined within the component tree - would that be more standard?

cmp/tab/TabSwitcher.js
haynesjm42: I saw defaultProps used elsewhere and thought it made sense here. The non-react-component parallel would be initializing class member variables when declaring, which I would still do even if some members needed to be initialized dynamically in the ctor. I would also raise the question of whether render is the best place to handle dynamic defaults instead of the ctor, unless the default value needs to be dynamic in such a way where if it wasn't explicitly provided through props, it is dependent on another prop which may change after the component is instantiated (though that smells like an anti-pattern)

lbwexler:
This PR is now ready for review.   Have renamed to reflect the extensive API changes to both routing and TabContainer.

PR #408, opened 2018-06-21
Log viewer tail refresh

gportela85:
https://github.com/exhi/hoist-react/issues/330

### DIFF
--- a/admin/tabs/logging/viewer/LogViewerDisplay.js
+++ b/admin/tabs/logging/viewer/LogViewerDisplay.js
@@ -58,8 +58,8 @@ class LogViewerDisplay extends Component {
     timedRefreshLines = () => {
         const {model, lastRow} = this,
             {tabPaneModel, tail} = model,
-            rect = lastRow.value.getBoundingClientRect(),
-            outOfView = rect.bottom > window.innerHeight;
+            rect = lastRow.value && lastRow.value.getBoundingClientRect(),
+            outOfView = !rect || (rect.bottom > window.innerHeight);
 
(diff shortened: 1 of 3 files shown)


admin/tabs/logging/viewer/LogViewerDisplay.js
```diff
@@ -47,6 +55,17 @@ class LogViewerDisplay extends Component {
         });
     }
 
+    timedRefreshLines = () => {
```
lbwexler: Total nit -- use method here.  Reserving these fat arrow properties for event handlers in render methods.

admin/tabs/logging/viewer/LogViewerDisplay.js
```diff
@@ -47,6 +55,17 @@ class LogViewerDisplay extends Component {
...
+        const {model, lastRow} = this,
+            {tabPaneModel, tail} = model,
+            rect = lastRow.value.getBoundingClientRect(),
+            outOfView = rect.bottom > window.innerHeight;
```
lbwexler: can't lastRow be null if the file is empty (or just hasn't been rendered).  We should make this null save.

lbwexler:
Had one remaining change -- regarding null safety 

As discussed, this whole pattern of having to pass around and consult tabPaneModel to see if we are 'visible' seems wonky, but we have it in a number of places.  Could we establish a concept of visibility that HoistComponent could easily read from context!?  TabPanes could then write to that, so that all children could see whether they were globally visible or not....  This might also help us to cleanup our story around TabContainer itself.

amcclain:
Sweet thank you guys.

PR #411, opened 2018-06-22
More consistent use of throwIf to eval exception conditions

amcclain:


amcclain:
Also note the TODO on the LocalStorageService commit - should we bite the bullet and swap that to use `XH.appCode` now?

PR #414, opened 2018-06-25
Build FeedbackDialog and FeedbackDialogModel into XH, Fixes #379

### LINKED ISSUE #379: Build FeedbackDialog and FeedbackDialogModel into XH
It seems weird that these are provided by Hoist, and the underlying service they rely on is provided by Hoist, but we don't wire them into XH by default.  

Seems like if we think its a core part of the framework, there ought to be a one-line way for any app to display these.  e.g. XH.showFeedbackDialog().

febbraiod:


lbwexler:
Also streamlined the rendering of the 'AboutDialog' and 'VersionDialog'  to avoid doing any work when we know they are not being shown.

PR #423, opened 2018-06-30
ColChooser enhancements

gportela85:
https://github.com/exhi/hoist-react/issues/415

### DIFF
--- a/cmp/grid/ColChooser.js
+++ b/cmp/grid/ColChooser.js
@@ -39,7 +39,7 @@ export class ColChooser extends Component {
                         onClick: this.restoreDefaults
                     }),
                     toolbarSep({
-                        hidden: !gridModel.stateModel
+                        omit: !gridModel.stateModel
                     }),
                     button({
                         text: 'Cancel',
--- a/cmp/grid/ColChooser.js
+++ b/cmp/grid/ColChooser.js
@@ -9,7 +9,7 @@ import {dialog} from '@xh/hoist/kit/blueprint';
 import {HoistComponent, elemFactory} from '@xh/hoist/core';
 import {filler} from '@xh/hoist/cmp/layout';
 import {toolbar} from '@xh/hoist/cmp/toolbar';
-import {leftRightChooser} from '@xh/hoist/cmp/leftrightchooser';
+import {leftRightChooser, leftRightChooserFilter} from '@xh/hoist/cmp/leftrightchooser';
 import {button} from '@xh/hoist/cmp/button';
 import {Icon} from '@xh/hoist/icon';
 
@@ -30,6 +30,7 @@ export class ColChooser extends Component {
             items: [
                 leftRightChooser({model: lrModel, height: 300}),
                 toolbar(
+                    leftRightChooserFilter({model: lrModel, fields: ['text'], anyMatch: true}),
                     button({
                         text: 'Reset',
                         icon: Icon.undo({cls: 'xh-red'}),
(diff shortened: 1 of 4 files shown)


cmp/leftrightchooser/LeftRightChooserFilter.js
```diff
@@ -49,17 +51,21 @@ class LeftRightChooserFilter extends Component {
...
     runFilter() {
-        const fields = this.props.fields || [],
+        const {fields = [], anyMatch} = this.props,
             searchTerm = escapeRegExp(this.value);
 
         const filter = (raw) => {
             return fields.some(f => {
                 const fieldVal = !!searchTerm && raw[f];
-                return ((fieldVal && new RegExp(`(^|\\\\W)${searchTerm}`, 'ig').test(fieldVal)) || !fieldVal);
+                return ((fieldVal && new RegExp(`${anyMatch ? '' : '(^|\\\\W)'}${searchTerm}`, 'ig').test(fieldVal)) || !fieldVal);
```
amcclain: I would suggest we break this up into a few lines. It's pretty long, and with a variable-driven regex in the midst of it it's a lot to grok!

cmp/leftrightchooser/LeftRightChooserFilter.js
```diff
@@ -49,17 +51,21 @@ class LeftRightChooserFilter extends Component {
...
             });
         };
 
         this.model.setDisplayFilter(filter);
     }
+
+    componentWillUnmount() {
+        this.onClearClick();
```
amcclain: Good / interesting catch - assuming that without this a filter remained "stuck" in place at the model level - is that assumption correct?  (Just reading the code online here on Github at the moment.)

For clarity, I would rename the `onClearClick` handler as it is now now longer tightly bound to a clicking interaction.

cmp/grid/ColChooserModel.js
```diff
@@ -20,7 +20,8 @@ export class ColChooserModel {
         this.gridModel = gridModel;
         this.lrModel = new LeftRightChooserModel({
             leftTitle: 'Available Columns',
-            rightTitle: 'Displayed Columns'
+            rightTitle: 'Displayed Columns',
+            leftSortBy: 'text'
```
amcclain: That was handy! 😄

cmp/grid/ColChooser.js
```diff
@@ -30,6 +30,7 @@ export class ColChooser extends Component {
             items: [
                 leftRightChooser({model: lrModel, height: 300}),
                 toolbar(
+                    leftRightChooserFilter({model: lrModel, fields: ['text'], anyMatch: true}),
```
amcclain: A bit worried that the reset button might now be interpreted as "reset the filter" vs. "reset my custom columns" - especially as we don't now do any confirm on reset. (Note, do not think that we want to add a confirm at this point - they can be overkill sometimes.)

Would try moving Reset to the other side of the filler, next to the other buttons. Does that look natural?

cmp/leftrightchooser/LeftRightChooserFilter.js
```diff
@@ -49,17 +51,21 @@ class LeftRightChooserFilter extends Component {
     }
     
     runFilter() {
-        const fields = this.props.fields || [],
+        const {fields = [], anyMatch} = this.props,
```
amcclain: I know this is not new code, but what would the point be of having an empty fields array? That seems like it's an invalid usage.

Would suggest we mark the array as required in the propTypes - is there a way to specify required + non-empty for arrays?

PR #427, opened 2018-07-03
IdleService

gportela85:
https://github.com/exhi/hoist-react/issues/132


lbwexler:
Was looking for a default display of this (app overrideable)  more like what we have in Hoist Sencha.
See App.showIdleWindow()

This looks like a start, but really need that bit in order to merge things.   

Also, the Hoist Sencha version had the ability to change this per user via preference.  Seems like for consistency we really just want to mirror *everything* we have in Hoist Sencha.

TomTirapani:
Took a look at this.

I like how the app can specify any `idleComponent`, allowing us to override the alert to do something different / more elaborate should the app call for it.

We're missing a few features that we need to reach parity with Hoist Sencha:
- A way for Apps to disable idle detection generally.
- A way to disable per user using preferences.
- We should stop any timer based tasks when the app becomes idle. I think we may be able to use `Timer.cancelAll()` for this.
- We should also fire an `appSuspended` event, so that other parts of the app can react if necessary. Alternatively, rather than an event, this could be an observable on App itself.

See https://github.com/exhi/hoist-sencha/blob/develop/grails-app/assets/src/xh/all/svc/IdleService.js for how these features are implemented in Hoist Sencha

app/impl/IdleWrapper.js
```diff
@@ -0,0 +1,22 @@
...
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+import {Component} from 'react';
+import {XH, HoistComponent, elemFactory} from '@xh/hoist/core';
+
+@HoistComponent()
```
lbwexler: Sorry -- you made those last changes so quickly, I hadn't finished my comments. Don't think we need this wrapper component.  This null check can be just made inline in the main renderer.

core/HoistApp.js
```diff
@@ -62,6 +63,21 @@ export function HoistApp(C) {
             };
         },
 
+        /**
+         * App must implement this method with appropriate logic to display a message when the app
+         * has been idle for the amount of minutes defined by the `xhxhIdleTimeoutMins` config.
```
lbwexler: spelling mistake here....

core/HoistApp.js
```diff
@@ -62,6 +63,21 @@ export function HoistApp(C) {
             };
         },
 
+        /**
+         * App must implement this method with appropriate logic to display a message when the app
```
lbwexler: 'may override' instead of 'must implement'

core/HoistApp.js
```diff
@@ -62,6 +63,21 @@ export function HoistApp(C) {
...
+         * has been idle for the amount of minutes defined by the `xhxhIdleTimeoutMins` config.
+         * @param {Function} callback - A callback function to be handled by the showIdleDialog
+         */
+        showIdleDialog(callback) {
```
lbwexler: perhaps this should be named 'renderIdleDialog()' if its job is to return a react element that can be rendered.

svc/IdleService.js
```diff
@@ -0,0 +1,66 @@
...
+
+    @action
+    timeout() {
+        this.task.cancel();
+        this.destroyAppListener();
+        Timer.cancelAll();
+
+        window.dispatchEvent(new CustomEvent('appSuspended'));
+        XH.idleComponent = XH.app.showIdleDialog(() => {
+            window.location.reload();
```
lbwexler: Why is the flag for reload (e.g. reload(true))  missing??  This is clearly in the sencha source, and we spent a long time troubleshooting problems related to that...

svc/IdleService.js
```diff
@@ -0,0 +1,66 @@
...
+        window.removeEventListener('scroll', this.resetTimer, true);
+    }
+
+    @action
+    timeout() {
+        this.task.cancel();
+        this.destroyAppListener();
+        Timer.cancelAll();
+
+        window.dispatchEvent(new CustomEvent('appSuspended'));
```
lbwexler: We don't use window.dispatchEvent, or anything like window events.  Realize we are trying to mimic the sencha implementation here, but not sure we need this.

svc/IdleService.js
gportela85: I used the initial idleService PR (https://github.com/exhi/hoist/pull/250/files) as source for these changes so I would be able to take compare the whole implementation instead of just looking at the final code in `IdleService.js`. That's why some things were missing like using an app config or user config to disable the service. I'll add the flag to disable cache.

lbwexler:
yes, that's always a risk with looking at a single pull request -- gotta
look at the latest and greatest of the code in question,

lbwexler:
I refactored this a bit, and it led to a larger refactor to create a new notion of AppState, which includes AppState.SUSPENDED.

Also simplified the implementation of IdleService and its timers, and also the way in which applications override the default component.

Also some unrelated file moves from '/app' package to '/impl'.

cr: atm (partially, IdleService review still pending)

PR #433, opened 2018-07-06
CSS variables for Blueprint intents.

### LINKED ISSUE #404: Toast intent color should match theme
Similar to how we handle Blueprint button intents and map their colors onto our theme CSS vars. Currently e.g. the success intent is a noticeably "off" shade of green.

### LINKED ISSUE #420: Expose dedicated CSS variables for Blueprint intents
Provide finer-grained control over the colors used for BP button / toast intents, and also choose more muted colors all around for these so they aren't quite so eye-popping / Fisher Price colors.

Note we already define rules related to button intents here, but aren't yet using dedicated CSS vars: https://github.com/exhi/hoist-react/blob/a1cc48324ef49cab1b44301ddb2e18d40098cb1a/kit/blueprint/styles.scss#L121.

Also note related issue #404 which we could grab at the same time.

TomTirapani:
+ Expose dedicated CSS variables for Blueprint intents. Fixes #420
+ Use these vars for Toasts. Fixes #404

### DIFF
--- a/kit/blueprint/styles.scss
+++ b/kit/blueprint/styles.scss
@@ -118,34 +118,43 @@
     }
   }
 
+  &.pt-disabled {
+    opacity: 0.6;
+  }
+
   &:not(.pt-minimal) {
     &.pt-intent-primary {
-      background-color: var(--xh-blue);
-      &:disabled {
-        background-color: var(--xh-blue-muted);
-      }
+      background-color: var(--xh-intent-primary);
     }
-
     &.pt-intent-danger {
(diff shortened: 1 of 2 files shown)


TomTirapani:
Also softened the color choices to be less 'eye-popping', particularly in dark theme. Note that disabled buttons now use opacity rather than a different color to convey the disabled state, in order to avoid specifying 2 vars for each intent.

amcclain:
Great thanks much - looks good. 👍

Issue #289, opened 2018-05-15
Grid - extended options for Excel exports


amcclain:
The default exports to Excel from ag-Grids are very basic, lacking formatting, data types, a table structure with headers, and fine-grained control over what fields are / are not exported.

We should decide if we want to invest more in learning / wrapping up ag-Grid's own [extensive Excel export functionality](https://www.ag-grid.com/javascript-grid-excel/) or if we instead want to build on what we have in hoist-sencha / hoist-core and use our server-side Excel generation via Apache POI.

amcclain:
@TomTirapani wondering if you would like to look at this one - a starting point might be to do a review of what ag-Grid offers and see if we think that could be a good / better alternative to our server-side stuff.

You know the server-side API and capabilities better than anyone, so if you think it would be quicker / more efficient to try extracting our store data first and POSTing it there, I would be fine with either.

Finally, we made some recent changes to GridModel and the column defs to make it easier to add and keep around extended attributes on those definition objects w/o them being fed to ag-Grid and triggering warnings.  So that should make it easier to standardize on an export formatter pattern or similar.

PR #435, opened 2018-07-10
Use hoist-core server-side export with agGrid. #289

TomTirapani:
+ Column prop `exportFormat` to control display in excel files.
+ Column prop `exportValue` to transform value prior to export, or to export a different field than the column.
+ Column prop `exportName` to use a different column header than the grid.
+ Add 'exportExcel' and 'exportCsv' context menu tokens.


cmp/contextmenu/StoreContextMenu.js
```diff
@@ -45,6 +47,26 @@ export class StoreContextMenu {
                         colChooserModel.open();
                     }
                 });
```
lbwexler: Could add a gridModel variable to this method

cmp/grid/GridModel.js
```diff
@@ -57,6 +63,9 @@ export class GridModel {
      * @param {string} [sortBy[].colId] - Column ID by which to sort.
      * @param {string} [sortBy[].sort] - sort direction [asc|desc].
      * @param {string} [groupBy] - Column ID by which to group.
+     * @param {string} [exportFilename] - Filename for exported file.
+     * @param {string} [exportFiletype] - Filetypes for export file. One of ['excel', 'excelTable', 'csv']
```
lbwexler: 'Filetype', not 'Filetypes'.  Also, why no default?

cmp/grid/GridModel.js
```diff
@@ -72,14 +81,20 @@ export class GridModel {
...
         stateModel = null,
         contextMenuFn = () => this.defaultContextMenu()
     }) {
         this.store = store;
         this.columns = columns;
-        this.contextMenuFn = contextMenuFn;
         this.emptyText = emptyText;
+        this.exportFilename = exportFilename;
+        this.exportFiletype = ['excel', 'excelTable', 'csv'].includes(exportFiletype) ? exportFiletype : 'excelTable';
+        this.enableExport = enableExport;
```
lbwexler: Why not simply throw if the filetype is not acceptable, rather than quietly ignoring and setting to excelTable?

utils/ExportUtils.js
```diff
@@ -0,0 +1,154 @@
...
+import {fmtDate} from '@xh/hoist/format';
+import {orderBy, uniq, isString, isFunction} from 'lodash';
+import download from 'downloadjs';
+
+
+/**
+ * Cell formats used in excel exports.
+ * Can be applied to columns via the 'exportFormat' prop.
+ */
+export const exportFormats = {
```
lbwexler: Seems like our naming convention would have this capitalized and singular -- e.g. ExportFormat

utils/ExportUtils.js
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Exports a grid to either Excel or CSV.
+ *
+ * Columns can install the following properties to manage how they are exported:
+ *
+ *      {string} exportName - specified title to appear at top of grid
+ *      {string|function} exportValue - modifies the value used in export
+ *              If string, can be used to point to a different field on the record
+ *              If function, can be used to transform the value
```
lbwexler: Need to add docs for parameters to this function.  Was momentarily confused by the docs on the columns (which are also useful.).  

Perhaps this whole thing could be made in to a class 'ExportManager' with a single public static method  exportToFile.  Might give you a bit more room to document everything around export in a class comment.

admin/tabs/activity/clienterrors/ClientErrorModel.js
```diff
@@ -49,6 +49,12 @@ export class ClientErrorModel {
         ]
     });
 
+    constructor() {
```
lbwexler: Would it be more straightforward potentially to just have exportFilename potentially be a closure as well as a string?

Would be nice to avoid the complexity of an autorun for this pretty common use case

TomTirapani:
I've added the option to export using agGrid - see https://github.com/exhi/hoist-react/pull/435/commits/b8b945c2f2eae89eac050b322ca02656ecb896dc.

In the end, I went for a flag you can set on GridModel *and* types. This allows you to set a grid to export either to server / local without having to use custom context menu items or export buttons, while still allowing to specify one-off exports should you want to.

admin/tabs/activity/clienterrors/ClientErrorModel.js
```diff
@@ -29,6 +29,7 @@ export class ClientErrorModel {
     gridModel = new GridModel({
         stateModel: 'xhClientErrorGrid',
         enableColChooser: true,
+        exportFilename: () => `Client Errors ${fmtDate(this.startDate)} to ${fmtDate(this.endDate)}`,
```
lbwexler: interesting that export is enabled by default.  Given the number of things that have to be done to make an export "look good" should we make it opt-in?

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -45,6 +48,26 @@ export class StoreContextMenu {
                         colChooserModel.open();
                     }
                 });
+            case 'exportExcel':
+                return new StoreContextMenuItem({
+                    text: 'Export to Excel',
+                    icon: Icon.download(),
+                    hidden: !gridModel || !gridModel.enableExport,
+                    disabled: !gridModel.store.count,
```
lbwexler: If gridModel can truly be null here (and the doc says that it can) line 56 and 66 are going to throw DNP

desktop/cmp/grid/GridModel.js
```diff
@@ -59,6 +63,8 @@ export class GridModel {
      * @param {string} [groupBy] - Column ID by which to group.
      * @param {boolean} [enableColChooser] - true to setup support for column chooser UI and
      *      install a default context menu item to launch the chooser.
+     * @param {boolean} [enableExport] - false to disable export context menu items.
+     * @param {function|string} [exportFilename] - Filename for exported file, or a closure to generate one.
```
lbwexler: This closure should probably be passed the gridModel as its first argument for good measure.

desktop/cmp/grid/GridModel.js
```diff
@@ -92,12 +102,35 @@ export class GridModel {
...
+        const {type, filename = this.exportFilename} = options;
+        XH.exportManager.exportAsync(this, filename, type);
     }
 
+    /**
+     * Exports the grid using agGrid's client-side export
+     *
+     * @param {string} fileName - Filename for exported file.
+     * @param {string} type - Type of export. One of ['excel', 'csv'].
+     * @param {Object} params - passed to agGrids export functions.
```
lbwexler: spelling of 'agGrid' a bit strange here

lbwexler:
Looks great and very tight!  Think its the minimal API for everything we want to expose right now.

PR #442, opened 2018-07-11
Update ClipboardButton to use elementRef button prop

### LINKED ISSUE #441: Clipboard button broken due to invalid ref handling
Clipboard button no longer functional due to invalid `ref` handling - possibly with move to Hoist button, maybe earlier.

amcclain:
+ Fixes #441


amcclain:
I can see how this might lead into a broad discussion of forwarding refs, what we want to bake into our HoistButton, etc. - but this also seemed like the correct immediate fix.

PR #443, opened 2018-07-11
Ensure client and session user in sync for relevant HoistImpl endpoints

amcclain:
+ Requires exhi/hoist-core#48 to be merged
+ Apologies as did not mean to loop in clipboard change as well - that also has a dedicated PR #442


lbwexler:
👍

PR #452, opened 2018-07-17
Ensure TabContainer activeTabId is in sync with the router state (Fixes #451)

### LINKED ISSUE #451: TabContainer active tab can become out of sync with the router state
When changing the active tab via `setActiveTabId` (such as with TabSwitcher), the active tab is set before attempting to navigate via the router. It is possible that the route change would be rejected (via middleware or a canDeactivate handler) and in that case the currently rendered tab is out of sync with the router state. When using routing, the TabContainer active tab should probably only be set via router state changes to ensure they remain in sync.

haynesjm42:
#451 

Just a slight re-organization of the code for setting the `activeTabId` to allow the router state to drive updating the `activeTabId` value when using routing. This is to fix a sync issue where a route change could be rejected by middleware of a canDeactivate handler, and the TabContainer would still change the active tab.

### DIFF
--- a/desktop/cmp/tab/container/TabContainerModel.js
+++ b/desktop/cmp/tab/container/TabContainerModel.js
@@ -23,7 +23,7 @@ export class TabContainerModel {
     tabs = [];
 
     /** Base route for this container. */
-    route = null
+    route = null;
 
     /** ID of the Tab to be active by default. */
     defaultTabId = null;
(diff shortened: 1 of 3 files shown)


desktop/cmp/tab/container/TabContainerModel.js
```diff
@@ -85,14 +85,14 @@ export class TabContainerModel {
...
         throwIf(!tab, `Unknown Tab ${id} in TabContainer.`);
 
-        this.activeTabId = id;
-        if (tab.reloadOnShow) tab.requestRefresh();
-
         if (route && !suppressRouting) {
             const tabRoute = route + '.' + id;
             if (!XH.router.isActive(tabRoute)) {
                 XH.navigate(tabRoute);
             }
```
lbwexler: John -- I can see wanting to return early and not set the activeTabId if we are going to navigate to it.  

But with the code as is, aren't we risking missing setting the id (and doing a needed refresh) if the route is already correct (e.g. conditional on line 90 returns false)

Does your code correctly do the reloadOnShow in all cases?

PR #453, opened 2018-07-17
Make Framework less opinionated about ContextMenus

lbwexler:
+ Apps now need to opt-in for App ContextMenus.  Browser menu available by default
+ Simplify Mixin Story in HoistComponent
+ Fix ContextMenu to restore ClipboardMenuItem support
+ Minor documentation enhancements.

### DIFF
--- a/core/HoistComponent.js
+++ b/core/HoistComponent.js
@@ -7,10 +7,8 @@
 import ReactDom from 'react-dom';
 import {XH} from '@xh/hoist/core';
 import {observer} from '@xh/hoist/mobx';
-import {ContextMenuTarget, HotkeysTarget} from '@xh/hoist/kit/blueprint';
 import {defaultMethods, chainMethods, overrideMethods} from '@xh/hoist/utils/ClassUtils';
 
-import {EventTarget} from './mixins/EventTarget';
 import {Reactive} from './mixins/Reactive';
 import {elemFactory} from './elem';
 
@@ -22,10 +20,16 @@ import {elemFactory} from './elem';
  * negatively impacted by the overhead associated with this decorator.
  *
  * Adds support for managed events, mobx reactivity, model awareness, and other convenience getters.
+ *
+ * @param {Object} [config] - configuration for the decorator.
+ * @param {boolean} [config.isReactive] - apply Reactive and associated mobx Observer mixin to the component Class.
+ * @param {boolean} [config.layoutSupport] - Does component support layout styles for flexbox as
+ *      first class properties?  If true, these properties will be parsed and placed into a
+ *      managed 'layoutConfig' on this component. See HoistComponent.layoutConfig, elem, and Box for more
+ *      information.
  */
 export function HoistComponent({
     isReactive = true,
-    isEventTarget = false,
     layoutSupport = false
 } = {}) {
 
@@ -33,23 +37,8 @@ export function HoistComponent({
         C.isHoistComponent = true;
         C.layoutSupport = layoutSupport;
 
-        //-----------
-        // Mixins
-        //------------
         if (isReactive) {
             C = Reactive(C);
-        }
-
-        if (isEventTarget) {
-            C = EventTarget(C);
-        }
-
-        if (C.prototype.renderContextMenu) {
-            C = ContextMenuTarget(C);
-        }
-
-        if (C.prototype.renderHotkeys) {
-            C = HotkeysTarget(C);
         }
 
         defaultMethods(C, {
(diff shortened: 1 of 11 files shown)


core/HoistComponent.js
```diff
@@ -22,36 +20,27 @@ import {elemFactory} from './elem';
  * negatively impacted by the overhead associated with this decorator.
  *
  * Adds support for managed events, mobx reactivity, model awareness, and other convenience getters.
+ *
+ * @param {Object} [config] - configuration for the decorator.
+ * @param {boolean} [config.isReactive] - apply Reactive and associated mobx Observer mixin to the component Class.
+ * @param {boolean} [config.layoutSupport] - Does component support layout styles for flexbox as
+ *      first class properties?  If true, these properties will be parsed and placed into a
+ *      managed 'layoutConfig' on this component. See HoistComponent.layoutConfig, elem, and Box for more
+ *      information.
```
amcclain: Thanks for filling these in!

desktop/cmp/contextmenu/ContextMenu.js
```diff
@@ -17,8 +17,9 @@ import {ContextMenuItem} from './ContextMenuItem';
 /**
  * ContextMenu
  *
- * To add a Context Menu to a component, specify a renderContextMenu() method on the
- * component that returns this object.
+ * Not typically used directly by applications.  To add a Context Menu to a
+ * component, decorate the component with ContextMenuSupport and implement
+ * getContextMenuItems().
```
amcclain: As we have one `@see` annotation on this doc comment already, might make sense to convert this note into a pointer to the decorator class.

desktop/cmp/contextmenu/ContextMenu.js
```diff
@@ -63,14 +65,15 @@ export class ContextMenu extends Component {
...
+                return menuItem({text: item});
+            }
+            
             const items = item.items ? this.parseMenuItems(item.items) : null;
-
             return menuItem({
                 text: item.text,
                 icon: item.icon,
-                onClick: () => start(item.action),    // do async to allow menu to close
+                onClick: item.action ? () => start(item.action) : null,    // do async to allow menu to close
```
amcclain: Nice approach here - good to know!  Always dislike it when the menu keeps floating around for a tick while the action commences behind it.

desktop/cmp/contextmenu/ContextMenuSupport.js
```diff
@@ -0,0 +1,53 @@
...
+import {ContextMenuTarget} from '@xh/hoist/kit/blueprint';
+
+/**
+ * Mixin to allow a Component to show a ContextMenu.
+ *
+ * Components that use this mixin should implement a getContextMenuItems()
+ * method.
+ *
+ * See Blueprint documentation for more information about the implementation of this
+ * behavior.
```
amcclain: I would tuck the orphaned words up with their respective lines above. Only mention as you were inquiring about my CLBOCD (comment-line-break-OCD) the other day 😀

desktop/cmp/contextmenu/ContextMenuSupport.js
```diff
@@ -0,0 +1,53 @@
...
+ */
+export function ContextMenuSupport(C) {
+    C.isContextMenuSupport = true;
```
amcclain: `hasContextMenuSupport` more natural (vs "is")?

desktop/cmp/contextmenu/ContextMenuSupport.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+import {defaultMethods} from '@xh/hoist/utils/ClassUtils';
+import {contextMenu} from './ContextMenu';
+import {ContextMenuTarget} from '@xh/hoist/kit/blueprint';
+
+/**
+ * Mixin to allow a Component to show a ContextMenu.
```
amcclain: Would it be helpful to nest within a mixins subdirectory, to separate it a bit from the "component Components"?  We could still export directly out of the index file, as per our standard.

desktop/cmp/contextmenu/ContextMenuSupport.js
lbwexler: yea, agree, 'has' would be more natural, but trying to balance that against the convention that
we put an 'isMixinName' on all the mixins.

I guess it points to the convention of 'xxxSupport' not being great.  We could call it 'ContextMenuHost' if you prefer? Want to stay from 'ContextMenuTarget' in this case to avoid confusion with Blueprint.

lbwexler: 'ContextMenuSource' also possible

mpeckerman:
This fixes https://github.com/exhi/hoist-react/issues/454

Issue #446, opened 2018-07-12
Fill in AppContainer services for Mobile Apps (e.g. Exception Handling, message rendering, login, impersonation, etc.)


lbwexler:
Currently the App Container for /mobile includes a bunch of dummy objects for rendering exceptions, impersonation, messages, app version, etc. etc.  We would like to implement these with Onsen and bring the mobile container up to parity with the desktop container.  

This may be a somewhat long-running ticket that spans a couple of submissions.

lbwexler:
Thanks for this Tom!

PR #456, opened 2018-07-19
Mobile app container

TomTirapani:
See issue #446 .

Implements the following for mobile:
+ AboutDialog
+ ExceptionDialog (with details)
+ FeedbackDialog
+ ImpersonationBar
+ LockoutPanel
+ LoginPanel
+ UpdateBar
+ VersionBar

Also adds a number of enhancements to the mobile toolkit:
+ Configurable AppMenu with standard items
+ Onsen button wrapper which adds support for icons
+ XH.app.isMobile flag
+ Mobile Messages, used via XH.alert(), XH.confirm() etc
+ Mobile Toasts, used via XH.toast()
+ Improved dropdown menus on mobile
+ New mobile fields TextAreaField && SelectField

These new mobile components will require some minor changes to client mobile apps, but otherwise are pretty self-contained. **There are, however, a couple of things that will affect existing desktop apps and require review:**

This changes removes ToastManager in favour of the cross-platform XH.toast(). Apps should update to use XH.toast(). Desktop can show several toasts at once (nicely stacked in the region), whereas mobile will queue toasts and show them one at a time.

Also note that impersonation bar visibility has been lifted up IdentityService, to help power the mobile implementation and give us a central place to determine visibility & check HoistAdmin permissions. This shouldn't require any changes to existing apps, but it does change the existing desktop ImpersonationBar.


TomTirapani:
If it makes it easier to review, I can split this into 2 pull requests - the new mobile components, and the framework changes required to support them (XH.toast(), IdentityService, etc). The framework changes may not make as much sense outside of the context of the mobile components, but may be easier to review.

lbwexler:
Big change -- looking at it now.  One quick concern I have is with moving the visibility state of the impersonation bar to the service.  

Have been increasingly looking at any references to GUI state within services as an anti-pattern.  
I think services are easier to understand as above the level of the actual GUI/platform/etc.

lbwexler:
Closing this in favor of amended one I just opened on a child branch of this one.

PR #460, opened 2018-07-20
Add xh-red-muted css class

febbraiod:
Think we should have this, and I have use of it in a client app. We already have the var, and symmetrical a green-muted class.


amcclain:
Thanks - agree completely.

Issue #459, opened 2018-07-19
Tab Activation Hooks


haynesjm42:
To bring TabContainer in line with framework routing libraries and to generally reduce the amount of wiring it takes to use routing it would be nice to have some hooks in place to help streamline the process of setting up routable components.

`onActivate(params)`
* Would eliminate the need for each Tab's content component to react to router state changes, check if they are the active route, etc.
* Would make sense to allow passing params to a Tab even if not using routing

`onDeactivate`
* Would not be as commonly used as `onActivate` but would allow a hook for a Tab to suspend some process if needed

`canActivate(params)`
* Most commonly used to reject a route if the user does not have necessary permissions, etc.
* Can return a boolean or Promise

`canDeactivate`
* Most commonly used if the current tab has unsaved changes which would potentially be lost if the user navigates away
* Can return a boolean or Promise

See https://router5.js.org/advanced/preventing-navigation for how to implement `canActivate` and `canDeactivate` with router5.

Also for reference see https://aurelia.io/docs/fundamentals/cheat-sheet#routing for details on how these same hooks work in the Aurelia router.

lbwexler:
This looks like a fantastic ticket.  In theory, you could imagine building it deeper into the framework, but I am hoping that building it into TabContainer will be a sweet spot for adding the most useful functionality with minimal complexity.

Very curious to know how you might be able to support parts of this lifecycle even in the case of a non-route enabled tab container.

Also --  assuming these hooks would be looked for on the Contents of the Tab, rather than the TabModel?

amcclain:
@haynesjm42 here is another candidate for Hoist work next week.

PR #461, opened 2018-07-22
AppContainer services for Mobile + AppContainer refactor

lbwexler:
+ Core ApplicationContainer Services in Mobile
+ Reuse Models across mobile/desktop for all core hoist function 
+ Provide Model to AppContainer
+ Simplify XH to delegate most work to other objects
+ Make all XH aliases explicit
+ Multiple Improvements to Message and Toast APIs.  These APIs are now symmetrical and immutable.
+ Remove all blueprint references from mobile app.
+ Simplify ExceptionHandling code
+ Remove "Implementation" services --  ErrorTrackingService,  FeedbackService -- from API.
+ Mobile and Desktop AppContainer are now decorated as HoistComponent


lbwexler:
Here are some (amended) notes from Tom's PR.
Implements the following for mobile:

Also adds a number of enhancements to the mobile toolkit:

Configurable AppMenu with standard items
Onsen button wrapper which adds support for icons
XH.app.isMobile flag
Mobile Messages, used via XH.alert(), XH.confirm() etc
Mobile Toasts, used via XH.toast()
Improved dropdown menus on mobile
New mobile fields TextAreaField && SelectField
These new mobile components will require some minor changes to client mobile apps, but otherwise are pretty self-contained. There are, however, a couple of things that will affect existing desktop apps and require review:

This changes removes ToastManager in favour of the cross-platform XH.toast(). Apps should update to use XH.toast(). Desktop can show several toasts at once (nicely stacked in the region), whereas mobile will queue toasts and show them one at a time.

core/XH.js
```diff
@@ -245,7 +272,13 @@ class XHClass {
     //---------------------------
     /**
      * Show a Toast.
-     * @param {Object} config - see ToastModel for available options.
+     *
+     * @param {Object} config - options for toast instance.
+     * @param {string} message - the message to show in the toast.
```
haynesjm42: Shouldn't these be `config.message`, `config.icon`, etc.?

core/appcontainer/MessageModel.js
```diff
@@ -61,15 +51,20 @@ export class MessageModel {
...
     doConfirm() {
         if (this.onConfirm) this.onConfirm();
+        this._resolver(true);
         this.close();
     }
 
     @action
     doCancel() {
         if (this.onCancel) this.onCancel();
+        this._resolver(false);
```
haynesjm42: For my specific use-case where I want to return an `XH.confirm()`  in a router5 `canDeactivate` callback, I would need the Promise to be rejected in order to prevent navigation. Just tried it in Apollo and resolving the promise with `false` will not prevent the navigation. I think I've seen both styles of using Promises with different dialog libraries, rejecting on cancel or passing the result in resolve, not sure which one I think makes the most sense from a API design perspective. If we do keep this approach, it will require a bit more code in my use-case to call the `done()` callback that router5 passes to the `canDeactivate` callback based on the result, so wouldn't actually clean up my code where I am currently using `onConfirm`. Thoughts?

lbwexler: Can we look at your code together again?  Definitely want to understand it -- it seems like it may be more complicated than the typical use case.

Logically, it seems more appropriate to resolve to a boolean than to throw.  Its really not an "exception" or an "error" -- it seems to me that cancel is  a totally expected and common code path that ought to be symmetrical with confirm.

core/XH.js
lbwexler: Yes, great catch

core/appcontainer/MessageModel.js
haynesjm42: Yeah it is definitely different from the typical use-case, and very specific to the router5 API. 100% agree with your point on resolving to a boolean. Happy to look at it with you again, I think #459 is the more appropriate place to cleanup my use-case now that I've thought about it more (perhaps returning a string from the `canDeactivate` hook would result in a confirm prompt to minimize the amount of wiring apps would need to do)

lbwexler:
Reviewed this with ATM.  Merging.  @tom, @atm, please feel free to look further at this diff in detail if you have a moment.  There is a lot here!

PR #466, opened 2018-07-23
Don't enforce focus on overlay unless covering entire viewport. Fixes #462

### LINKED ISSUE #462: Masks and inline LoadMasks sometimes block input to elements outside of their parent
Noticed this a while back and never looked into it deeply but this has become a major issue for some recent changes I've made to a client app to add a mask over a grid when the parameters used to fetch the data have changed, it is making the text boxes in my toolbar completely un-interact-able. Interestingly enough this issue only seems to occur for InputGroup components, doesn't happen on DatePicker (dayField) or Suggest (ComboBox) components, even though they use an InputGroup under the hood.

A little debugging and experimentation has shown that the issue lies with the `enforceFocus` property on the Blueprint Overlay component. The `handleDocumentFocus` event handler is being called on that component when I attempt to focus on a text box not covered by the overlay, and the handler is killing the focus change because the element attempting to receive focus is outside of the overlay's container.

Setting `enforceFocus` to false for Mask and conditional based on `isInline` for LoadMask seems to resolve the issue, and I didn't notice any undesired side-effects in some quick testing. Seems like it makes sense that `enforceFocus` should match the `usePortal` property in the majority of cases, though this does seem like a Blueprint bug.

haynesjm42:
This was easily reproduce-able in Toolbox on the components/loadMask page. Click the Load button and try to focus into the text boxes on the bottom toolbar.

### DIFF
--- a/desktop/cmp/mask/LoadMask.js
+++ b/desktop/cmp/mask/LoadMask.js
@@ -7,8 +7,8 @@
 
 import {PropTypes as PT} from 'prop-types';
 import {Component} from 'react';
-import {HoistComponent, elemFactory} from '@xh/hoist/core';
-import {vbox, vspacer, box} from '@xh/hoist/cmp/layout';
+import {elemFactory, HoistComponent} from '@xh/hoist/core';
+import {box, vbox, vspacer} from '@xh/hoist/cmp/layout';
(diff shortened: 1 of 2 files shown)


haynesjm42:
#466

Issue #463, opened 2018-07-23
Delegate leftIcon and rightElement properties for applicable HoistField controls


haynesjm42:
Would be nice to be able to set the `leftIcon` and `rightElement` properties on `HoistField` controls, have a user request for a clear button for individual fields which I would like to do with a simple right-aligned X button in the InputGroup. I also am mostly using HoistFields in toolbars where labels can make things a bit clunky and horizontal space is at a premium, so some icons or tags embedded in the controls to help identify them when appropriate would be cleaner.

lbwexler:
John -- My only concern here was that we were trying not to tie our control wrappers *too* tightly to the underlying Blueprint API. (realize this is getting to be a bit of an impossible task)

Do we feel that both 'leftIcon' and 'rightElement' are the right way to present this in our API?  To be honest, it seemed a little weird and asymetrical to me.  But perhaps most pragmatic? 

 @amcclain  -- curious to get your thoughts on this one.

haynesjm42:
It is definitely asymmetrical but I think the reason is that their css and dom structure only really supports icons in the `leftIcon`, even though it would be relatively trivial to handle any element there and call it `leftElement`, not sure why they decided to go that way.

Would probably make more sense to call it just `icon` to be consistent with our other components. `rightElement` is as good as any other name I can think of for the other property if we want to keep it completely generic and user defined. Could see us adding explicit support for the common use-cases there, by providing an `actionBtn` property for my specific use-case which let's you set the icon and click handler to use.

amcclain:
Only question I have is how / if the Bluerprint `ControlGroup` concept plays any role.

http://blueprintjs.com/docs/#core/components/control-group

Leaving that aside, I don't feel strongly, other than it would be good for us not to be blocking on existing functionality in the kit that we could put to good use.

If I had to vote - and I am just typing here on github and not using them myself in an app - I would vote for the original ticket to match the BP API and be as clear as possible about what the props take and where they go in the control.

PR #468, opened 2018-07-24
Expose leftIcon and rightElement properties on applicable HoistFields (#463)

haynesjm42:
#463 

Delegating `leftIcon` and `rightElement` where applicable. Also fixed up the css for displaying our icons inside of Blueprint input groups to match as closely as possible the Blueprint styling.

See [this commit](https://github.com/exhi/toolbox/commit/761feb545da9a73aab4f10d066eb212842c2ae69) for example usage (accidentally pushed to the master branch, but doesn't cause issues as the new props are just ignored)

### DIFF
--- a/desktop/cmp/leftrightchooser/LeftRightChooserFilter.js
+++ b/desktop/cmp/leftrightchooser/LeftRightChooserFilter.js
@@ -7,11 +7,11 @@
 import {Component} from 'react';
 import {PropTypes as PT} from 'prop-types';
 import {escapeRegExp, isEqual} from 'lodash';
-import {inputGroup} from '@xh/hoist/kit/blueprint';
 import {observable, setter} from '@xh/hoist/mobx';
 import {elemFactory, HoistComponent} from '@xh/hoist/core';
 import {Icon} from '@xh/hoist/icon';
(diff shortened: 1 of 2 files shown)


amcclain:
As noted via Slack, please see https://github.com/exhi/hoist-react/issues/475 as it relates to this ticket re.

* do we want to continue the delegateProps idea
* what is the standard for propTypes docs

lbwexler:
HoistField is just a really thorny issue -- feel like we need to sit together (maybe tomorow?) and come up with a plan for it.  Need to consider relationship between Mobile and Desktop.  

One thing that makes the documentation of it tricky is that we are -- for better or for worse -- using inheritance here, and it doesn't seem like  React provides a good way to document inherited props.

Want to also throw out the idea that we could use a collection of the form blueprintProps, to gather blueprint specific props.  This pattern has been useful with agGrid and others.  Perhaps helpful to clarify things?

Again -- happy to lose delegate props...

amcclain:
I am going to go ahead and merge this now to take support for the additional props.  I can see now that it's useful and allows us to replace some direct usages of inputGroup within the framework.

This change also includes a fix for #490 which I would like to fold in.

Would absolutely still like to revisit and focus on #475 !!

Issue #465, opened 2018-07-23
StoreFilterField should debounce filtering of the store


haynesjm42:
It currently calls `setFilter` on the bound store on every change of the value, which can result in a poor user experience with large data sets. The standard 200ms debounce time should suffice, though it might make sense to expose a prop for it so it could be modified to suit whatever store it is bound to.

PR #470, opened 2018-07-24
Debounce filtering of the store in StoreFilterField (#465)

haynesjm42:
#465

### DIFF
--- a/desktop/cmp/store/StoreFilterField.js
+++ b/desktop/cmp/store/StoreFilterField.js
@@ -7,10 +7,10 @@
 
 import {Component} from 'react';
 import {PropTypes as PT} from 'prop-types';
-import {escapeRegExp} from 'lodash';
+import {debounce, escapeRegExp} from 'lodash';
 import {inputGroup} from '@xh/hoist/kit/blueprint';
-import {HoistComponent, elemFactory} from '@xh/hoist/core';
-import {setter, observable} from '@xh/hoist/mobx';
+import {elemFactory, HoistComponent} from '@xh/hoist/core';
+import {observable, setter} from '@xh/hoist/mobx';
 import {button} from '@xh/hoist/desktop/cmp/button';
 import {BaseStore} from '@xh/hoist/data';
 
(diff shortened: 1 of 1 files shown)


desktop/cmp/store/StoreFilterField.js
```diff
@@ -25,11 +25,20 @@ export class StoreFilterField extends Component {
         /** Store to filter */
         store: PT.instanceOf(BaseStore).isRequired,
         /** Names of fields in store's records to filter by */
-        fields: PT.arrayOf(PT.string).isRequired
+        fields: PT.arrayOf(PT.string).isRequired,
+        /** Delay (in ms) used to buffer filtering of the store after the value changes from user input (default is 200ms). Set to 0 to filter immediately after user input. */
```
lbwexler: This long line should be broken up

desktop/cmp/store/StoreFilterField.js
```diff
@@ -25,11 +25,20 @@ export class StoreFilterField extends Component {
...
+        filterBuffer: PT.number
     };
 
     @setter @observable value = '';
 
+    constructor(props) {
+        super(props);
+
+        const {filterBuffer = 200} = this.props;
```
lbwexler: awesome syntax here.  Seems like this should be used everywhere we currently do 
>>filterBuffer = this.props.filterBuffer || 200;
It cleanly avoids any issues of 'falsy' values not getting set....

desktop/cmp/store/StoreFilterField.js
```diff
@@ -25,11 +25,20 @@ export class StoreFilterField extends Component {
...
+        this._debouncedFilter = debounce(this.runFilter, filterBuffer);
+    }
```
lbwexler: This kind of brings up the question I have long had about how to distiguish beween "fixed" props and "dynamic" props.  In this case,  this becomes a "fixed" prop.  Changing it after the first render of this component will have no effect.

lbwexler: (Not implying that there is a problem with that)

desktop/cmp/store/StoreFilterField.js
haynesjm42: Yes definitely, been trying to make sure I use it (and lodash `isNil()`) whenever appropriate after I introduced a couple bugs in Apollo from implicit boolean conversions to falsy values..

PR #482, opened 2018-07-30
Upgrade FontAwesome, use explicit imports

### LINKED ISSUE #480: Upgrade to FontAwesome 5.2.x
This brings with it an expanded icon pack, but also several differences to the package structures and imports.

amcclain:
The added imports and call to `library.add()` are clearly painful within this file, but it looks like this makes a huge difference in bundle size, with a large (several 100k) reduction even when gzipped. We are now only shipping the SVG paths for the icons we've used, rather than every icon in the library (in three different weights).

Adding a new icon now requires importing the three variants + adding the triplet to the library call. Annoying, but not terrible and not something that happens every day...

Note if an app wants to use an icon we don't import here on its own, it can do so - would just need to do the specific import itself.

Also fixes #480 taking the overall FA upgrade.


amcclain:
@lbwexler and @TomTirapani wanted to run this by you anytime you have a chance.

TomTirapani:
If it reduces several 100k of unused icons in the bundle, I think we can handle the extra amount of code!

I wonder how much we are using the 3 different variants? Perhaps we could simplify by having one variant as the 'hoist standard', while still allowing apps to import icons from the other variants if desired?

amcclain:
It's a good question... Right now we have this kind of funny API on our icon elemFactories where you can pass a "prefix" option of `far|fas|fal` - so all three options declared as supported for the icons we've defined. (Any suggestions for improving this? Seems a bit clunky as is.)

I am not sure how often we're using that - we could search...  I know I have used the "light" variant for some of our app header icons as I thought they looked a bit better, but nothing critical.

I would be interested to understand what the workflow would be re. imports if we didn't import the variants and add them to the library, but a developer wanted to use a variant. I am guessing they could import the icon and add to the library (in another dedicated call) someplace reasonable in their code...

If we were going to remove one variant, I would vote for solid, but (other than the boilerplate) I am not sure how much weight that would take off of the bundles.

PR #487, opened 2018-07-31
Wait for agGrid to finish updating its node model before attempting to sync selection.

### LINKED ISSUE #484: Setting grid model selection can throw if grid rendering not yet complete
We have a client app that's setting a selected record based on a "last viewed" preference / route and ensuring that gets set on the relevant Grid / GridModel via an autorun call to `gridModel.selModel.select()`.

The app code has long had a `delay: 1` opt on the auto-run "to account for grid rendering." Updated to latest hoist-react 11.0.0-SNAPSHOT and now that code is throwing within `Grid.selectionReaction()` as the node is not found. As a result, the row is not visible selected within the grid, and an error ends up on the console.

Trial and error showed that upping the delay to e.g. 300 works OK, but 100 or 200 does not - i.e. this is actually a time-based thing and not simply a matter of pushing to the end of the event loop.

Ideally we would not need any delays at all in client app code, and definitely want to avoid any fuzzy delays based on variable rendering performance.  Curious though as to what changed - and then happy to help brainstorm ideas to make more reliable all around at the Hoist level.

TomTirapani:
There is an indeterminate short delay between `agApi.setRowData()` and agGrid's node model getting updated. In between these two times, the store's record collection and agGrid's internal node model are out of sync, and as such, any logic using agGrid's node model is unreliable. The delays in the client app were added to bandaid this, but are far from a robust solution.

This branch aims to make the 'readiness' of agGrid's node model observable via the new `agNodesRendered` prop (open to suggestions for better naming). This is set to false before setting rowData in the `dataReaction`, and set to true upon agGrid's `onRowDataUpdated` event.

Any code that works with agGrid's node model (i.e. selection syncing) should use this observable to determine if the node model is 'ready' and is a true reflection of the store model's records. Waiting for `agApi` is not enough for these types of functions.

Tested in client app and fixes #484

### DIFF
--- a/desktop/cmp/grid/Grid.js
+++ b/desktop/cmp/grid/Grid.js
@@ -58,7 +58,6 @@ class Grid extends Component {
         this.addReaction(this.dataReaction());
     }
 
-
     render() {
         const {colChooserModel} = this.model,
             {layoutConfig, agOptions} = this.props;
@@ -122,7 +121,8 @@ class Grid extends Component {
             onSelectionChanged: this.onSelectionChanged,
             onSortChanged: this.onSortChanged,
             onGridSizeChanged: this.onGridSizeChanged,
-            onDragStopped: this.onDragStopped
+            onDragStopped: this.onDragStopped,
+            onRowDataUpdated: this.onRowDataUpdated
         };
     }
 
@@ -228,10 +228,10 @@ class Grid extends Component {
         const {model} = this;
 
         return {
-            track: () => [model.selection, model.agApi],
+            track: () => [model.selection, model.agApi, model.agNodesRendered],
             run: () => {
-                const {agApi, selModel} = model;
-                if (!agApi) return;
+                const {agApi, agNodesRendered, selModel} = model;
+                if (!agApi || !agNodesRendered) return;
 
                 const modelSelection = selModel.ids,
                     gridSelection = agApi.getSelectedRows().map(it => it.id),
@@ -286,6 +286,7 @@ class Grid extends Component {
             run: () => {
                 const {agApi} = model;
                 if (agApi) {
+                    model.setAgNodesRendered(false);
                     agApi.setRowData(model.store.records);
                 }
             }
@@ -304,6 +305,7 @@ class Grid extends Component {
     }
 
     onSelectionChanged = (ev) => {
+        if (!this.model.agNodesRendered) return;
         this.model.selModel.select(ev.api.getSelectedRows());
     }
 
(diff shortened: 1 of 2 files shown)


desktop/cmp/grid/Grid.js
```diff
@@ -304,6 +305,7 @@ class Grid extends Component {
     }
 
     onSelectionChanged = (ev) => {
+        if (!this.model.agNodesRendered) return;
```
amcclain: Curious about the return here. Would this fire when the nodes were not rendered? What would be the general state of the grid in such a case? Would it throw, or would the API return something strange / misleading for `getSelectedRows()`?

I ask because otherwise it seems like it would be better to still call `selModel.select()` and let the selModel update its record collection. The reaction above would be short-circuited, but presumably also be a no-op in that the grid itself triggered the selection event - and in any case it would ensure all in sync on the next go round...

desktop/cmp/grid/GridModel.js
```diff
@@ -34,7 +34,7 @@ export class GridModel {
     @observable.ref columns = [];
     @observable.ref sortBy = [];
     @observable groupBy = null;
-
+    @observable agNodesRendered = false;
```
amcclain: Need some kind of doc someplace to indicate what this is all about. It could be a comment on this property or an addition to the top-level class comment - I would favor the property.

Do we anticipate that applications will need to hook into this as part of the model's public API? If so, trying to think of a more general name so we don't necessarily bake "ag" into the model API...  Would `nodesRendered` / `rowsRendered` / `dataRendered` be an improvement?  Or `xxxRenderComplete` for any of those options?

Was also thinking something with `ready` in it to abstract a bit from the notion of "rendering" being required for certain operations, but ready seems more like a state change that happens once...

lbwexler: Yes, number of critical questions here.  I was actually thinking we might hold such implementation specific state in the component itself (rather than the model).  And hoping that this whole thing could be invisible to applications.  

I don't think its the case that our models have to hold *all* state.  I think the key is that they hold the state that we expect applications, and external users of the component to read/write.

In this case the GridModel is the central API object for applications.  Would be happy to leave this particular wart off of it, if applications don't need it.

Note that 'agAPI' made more sense to put on the model, as I assumed it would be a useful 'escape hatch/entry point' for applications to the broad world of the ag grid API.

amcclain:
See #495 - the underlying issue was that our grid's selectionReaction can run prior to its dataReaction on a store update, so the store/selModel know about records that have not yet been loaded into the grid by the dataReaction.

Note the ag-Grid loadData method is indeed synchronous, so this didn't turn out to be some kind of "rendering timing" issue, but instead exposed a lack of coordination / sequencing between those two reactions.

Issue #486, opened 2018-07-31
DateInput Field:  Cannot select a day after navigating several months


cnrudd:
This can be seen in toolbox:
http://localhost:3000/app/forms

After navigating several months back in the Start Date DateInput Datepicker popover, clicking to select a date should close the popover and put the selected date in the input field.
But this does not happen.  The popover does not close, and the date is not selected.  

Oddly, clicking in the input field causes the date to change and the date to be selected in the popover, and then clicking on the date in the popover causes the popover to close.

cnrudd:
This was fixed by adding prop 
commitOnChange: true

to the dayField declaration in toolbox.

cnrudd:
Hmm.  Looking into this more, I see that adding commitOnChange: true causes changing the month to cause the DayField's value to change, even without selecting a new date in a new month panel.  This is problematic in use cases where the date change triggers other actions.

Blueprint was already made aware of this issue: https://github.com/palantir/blueprint/issues/2418

and fixed it in v3 by adding another argument to their onChange prop function:
v3: http://blueprintjs.com/docs/#datetime/dateinput
v2: http://blueprintjs.com/docs/versions/2/#datetime/dateinput

We might need to migrate to blueprint v3.

amcclain:
Which we've now done on develop 😄 

@cnrudd would you be able to make whatever - hopefully incremental - changes are required to resolve the issue?

cnrudd:
I've made changes to DayField.js so that it uses the new isUserChange argument that BP3 added to the dateInput onChange function.
Changes are in: https://github.com/exhi/hoist-react/commit/b17c52f9d0bb0db357dcee5ae1244193e270dca9

This solves the problem of the selected date changing when a user navigates months. 

However, the popover stays open after date selection in some cases, due to this decision:
https://github.com/palantir/blueprint/blame/develop/packages/datetime/src/dateInput.tsx#L281
"for UX purposes, we want to close the popover only if the user explicitly clicked a day within the current month."

So, the popover will close on date selection only if no date was previously selected, or if the new date selected is in the same month as the previous date selected.

If you had selected 07/23/2018 and now select 01/22/2018, the new date is indeed selected and the date input value changes to the new date, but the popup stays open until you click a second time somwhere outside the popup.

amcclain:
Thanks for tracking that down and noting the issues / oddities here. Can't say I fully understand the "close only if in the same month" decision, but it sounds like that's what BP does and I assume they have their reasons.

[...]

PR #491, opened 2018-08-01
DayField: Prevent date selection change on month navigation. #486

cnrudd:


lbwexler:
Colin and I looking at this.  WIP

PR #494, opened 2018-08-02
HoistComponent support for baseCls / getClassNames()

### LINKED ISSUE #252: Standardize handling of baseCls on Hoist components
I would suppose a standard `baseCls` property that we define for all our core Hoist-provided components and wrappers.  This should be applied alongside any optional instance-specific `cls` props in a consistent way.

Even if we don't have any particular styling to apply, it is often useful to have a CSS handle on a given component or component-type in the tree.

amcclain:
Bundled up pattern for specifying a baseCls on the component class and producing a combined list of baseCls, instance-specific classNames provided via props, and any additional classes required at render-time. 

+ Fixes #252 
+ Adds support for Panel.maskText property.
+ Adds layoutSupport to relativeTimestamp (useful for e.g. margins).
+ Adds CSS vars for Panel background color and borders.
+ Updates to toolbar borders to restore bottom borders on panel top toolbars, add consistent borders for bottom/left/right tbars.
+ Some more icons (for Toolbox work).


amcclain:
@TomTirapani wanted to loop you in here re. the mobile components - I did not apply any of this there, yet. I would like to ensure we're happy with this approach (or decide on anything different) and then of course let's get things consistent.

lbwexler:
Reviewing this now.

icon/Icon.js
```diff
@@ -238,6 +246,7 @@ import {
     faSearch as faSearchSolid,
     faSignIn as faSignInSolid,
     faSignOut as faSignOutSolid,
+    faSkull as faSkullSolid,
     faSun as faSunSolid,
```
lbwexler: Where are we using this?   :-)

amcclain: In Toolbox > Toolbars.  Gotta have at least one skull someplace. 💀Maybe we can add it to some "fatal error" dialog as well to get our money's worth.

Issue #507, opened 2018-08-08
JsonField not rendering correctly in Dark Theme


lbwexler:

febbraiod:
+Logo on welcome page in toolbox not accounting for themes either

PR #511, opened 2018-08-08
Add reaction to jsonField to respond to theme changes, Fixes: #507

febbraiod:


desktop/cmp/form/JsonField.js
```diff
@@ -47,6 +47,14 @@ export class JsonField extends HoistField {
         editorProps: PT.object
     };
 
+    constructor() {
+        super();
+        this.addReaction({
+            track: () => XH.darkTheme,
+            run: () => this.editor.setOption('theme', XH.darkTheme ? 'dracula' : 'default')
```
lbwexler: Let's just make this null safe -- editor may be null, not setup immediately

lbwexler: Can we also cleanup the editor in destroy() instead of willUnmount()? (see below)

Issue #518, opened 2018-08-09
Context menu shows divider when items are hidden


TomTirapani:
<img width="248" alt="context menu divider" src="https://user-images.githubusercontent.com/3017757/43916050-6a358226-9c04-11e8-83e7-92fddac14a81.png">

TomTirapani:
Pull request for fix here: https://github.com/exhi/hoist-react/pull/519

We already had code in place to filter out consecutive dividers and filter out starting / ending dividers. However, both filtering steps were happening in the same pass over the array. This meant if there were _consecutive ending_ dividers, one would remain.

See the `defaultContextMenuItems` here: https://github.com/exhi/hoist-react/blob/develop/desktop/cmp/grid/GridModel.js#L40. You can see how a grid with neither export nor column chooser would end up with 2 consecutive ending dividers.

In the pull request, I split these filters into 2 different passes. Now starting / ending dividers are filtered out after any other mutations.

PR #519, opened 2018-08-09
Fix filtering out starting / ending context menu dividers. #518

TomTirapani:


TomTirapani:
See https://github.com/exhi/hoist-react/issues/518 for example of issue and explanation.

Issue #508, opened 2018-08-08
Provide method/affordances for "Restore All Defaults"


lbwexler:
We need to provide a method (and maybe a default button) on XH to clear all of a users local grid state and preferences.

PR #520, opened 2018-08-10
Added 'Restore Defaults' method and affordance. #508

TomTirapani:


lbwexler:
Thanks!

+ Can't really think of a case when you would ever call this method with false?   Could we hold off on that, until we actually need it?  YAGNI....

+ Can we improve the documentation a bit to describe what it actually does to preference service and local storage service?

+ Not sure we need the button, but it definitely matches the other similar buttons that we have for this kind of thing.  So great (plus, I asked for it in the ticket!)

desktop/cmp/button/RestoreDefaultsButton.js
```diff
@@ -0,0 +1,59 @@
...
+import {Component} from 'react';
+import {PropTypes as PT} from 'prop-types';
+import {elemFactory, HoistComponent, XH} from '@xh/hoist/core';
+import {button} from '@xh/hoist/kit/blueprint';
+import {Icon} from '@xh/hoist/icon';
+
+/**
+ * Convenience Button preconfigured for use as a trigger for resetting user
+ * customizations. Clears all grid state and preferences, then reloads the app.
+ * Accepts props documented below as well as any supported by Blueprint's Button.
```
amcclain: This is a Hoist button so we can skip the Blueprint reference here. Ideally that provides a nice abstraction over BP implementation, so if we swapped it in the future no need to go back and change these references then.

Could drop the line entirely - I think it's relatively clear from the implementation, and the idea is to typically take the defaults and not have to think about it. (Applies just as much to any of these other convenience buttons.)

TomTirapani:
Thanks for the feedback guys - I've updated the P/R. Only GridState local storage keys are removed, and the method always reloads the app.

amcclain:
I dunno - I think it's a good scope / amount of work for this function to just clear prefs/comp state and leave the existing local storage service method to zap any other data.  Forces us to think more precisely about what we're storing where in the framework and to maintain a better contract with the apps around infrastructure services like local storage.  Of course can discuss as we go forwards if this ends up being problematic...

(But FYI we're adjusting this right now to avoid importing the desktop model class into XH in case you are following along from the car 😄)

Issue #390, opened 2018-06-15
AppBar refresh button should have some default handler


amcclain:
We could discuss a standard template method on `HoistApp` decorator - e.g. `XH.app.fireGlobalRefresh()` which we use in Sencha world.

Immediate issue in my mind right now is that if you don't do anything - just take the defaults - you do get a refresh button but it's not wired to _anything_ which is a confusing default. 

We could replace `refreshButtonProps` with `refreshFn` or similar (I think you had suggested that) and then omit the button if that is not provided.  But I like the idea of a default handler and the `hide` prop consistent with the others if you don't want to show.

@lbwexler looping you in here re. the HoistApp level template method idea.

lbwexler:
like the idea of a global template method.  Should allow for indicating
autoRefresh vs. user initiated....

-Lee

PR #526, opened 2018-08-13
Global App refresh method + default AppBar RefreshButton handler

haynesjm42:
#390 

Went with `requestRefresh` method on the App since that was the pattern we are following with the Tabs (and was already what the Admin app was using). Decided against using an event for now since we don't really seem to be using events much, and the App will probably need to drive the refresh from the top-down most of the time.

### DIFF
--- a/desktop/cmp/button/RefreshButton.js
+++ b/desktop/cmp/button/RefreshButton.js
@@ -29,28 +29,20 @@ export class RefreshButton extends Component {
         title: PT.string,
         /** Function to call when the button is clicked. */
         onClick: PT.func,
-        /**
-         * Model to call loadAsync() on when the button is clicked. Only used if onClick was not
-         * provided.
-         */
+        /** Model to refresh via loadAsync(), if onClick prop not provided. */
         model: PT.object
     };
 
     render() {
         warnIf(
-            this.props.model && this.props.onClick,
-            'Both model and onClick were provided to RefreshButton properties. The model property will be ignored.'
-        );
-
-        warnIf(
-            !this.props.model && !this.props.onClick,
-            'Neither onClick nor model were provided to RefreshButton. Clicking the RefreshButton will have no effect.'
+            (this.props.model && this.props.onClick) || (!this.props.model && !this.props.onClick),
+            'RefreshButton must be provided either a model or an onClick handler to call (but not both).'
         );
 
         const {
             icon = Icon.sync(),
             title = 'Refresh',
-            onClick = this.model ? this.onRefreshClick : undefined,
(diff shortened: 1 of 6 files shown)


desktop/cmp/button/RefreshButton.js
```diff
@@ -9,41 +9,66 @@ import {Component} from 'react';
...
-        const {icon, title, onClick, model, ...rest} = this.props;
+        warnIf(
+            this.props.model && this.props.onClick,
+            'Both model and onClick were provided to RefreshButton properties. The model property will be ignored.'
+        );
+
+        warnIf(
+            !this.props.model && !this.props.onClick,
+            'Neither onClick nor model were provided to RefreshButton. Clicking the RefreshButton will have no effect.'
+        );
```
amcclain: Deciding how/when we want to add these kinds of warnings will be more of an art than a science. It's a balance between helping to document the component and avoid common pitfalls, without adding too much defensive code. 

I'm glad you used the pattern here, but think a slightly better balance would be to collapse these into one check that could output something like "RefreshButton must be provided either a model or an onClick handler to call (but not both)."

desktop/cmp/button/RefreshButton.js
```diff
@@ -9,41 +9,66 @@ import {Component} from 'react';
...
+            onClick,
             ...rest
         });
     }
 
-    //-------------------------
+    //---------------------------
     // Implementation
     //---------------------------
     onRefreshClick = () => {
```
amcclain: Suggest we rename this `refreshModel()` for some extra clarity.

desktop/cmp/button/RefreshButton.js
```diff
@@ -9,41 +9,66 @@ import {Component} from 'react';
...
     static propTypes = {
+        /** Icon to display for the button. Defaults to Icon.sync(). */
         icon: PT.element,
+        /** Tooltip text to display when the mouse is over the button. Defaults to 'Refresh'. */
         title: PT.string,
+        /** Function to call when the button is clicked. */
         onClick: PT.func,
+        /**
+         * Model to call loadAsync() on when the button is clicked. Only used if onClick was not
+         * provided.
```
amcclain: Sorry to nitpick you here 😄because I really do appreciate you keeping a consistent line length throughout this change. I usually try to avoid the one-word-on-last-line wrap if I can. A good approach can be to shorten up the comment to avoid the wrap if you can still keep it sensible, e.g. "Model to refresh via loadAsync(), if onClick prop not provided."

haynesjm42:
Thanks for the review! Requested changes implemented

PR #527, opened 2018-08-13
Cast headerClass and cellClass to array before pushing to it

febbraiod:
These props can be specified as strings (in our agOptions) as documented by ag-grid. However, if passing a string here along with an align prop, our method was throwing when we tried to push our xh alignment class onto that string.

### DIFF
--- a/columns/Column.js
+++ b/columns/Column.js
@@ -6,7 +6,7 @@
  */
 
 import {Component} from 'react';
-import {startCase} from 'lodash';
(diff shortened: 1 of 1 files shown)


amcclain:
As discuss in Slack, this does leave one possible problematic case where the user wishes to pass a function for cellClass or headerClass. That is supported by ag-grid, although we don't guarantee we will support every option within this grid.  Merging to take this clear improvement and we can revisit if we need the function support in the future.

PR #530, opened 2018-08-14
Doc updates (focus on Grid, Column, and related)

### LINKED ISSUE #336: Add @type jsdoc tags to key public properties
Just came across this using hoist-react in a client app (i.e. npm package mode, not inline). Had some application code like: `this.gridModel.selModel.select(id)`. 

I had constructed the gridModel in the same class via `new GridModel()`, so IntelliJ was able to provide ctrl+click navigation and some intellisense regarding that class - namely, it knew that it had a selModel property.  However it wasn't able to navigate to the implementation of `select()` or tell me anything about the selModel as it didn't know what class it was.

I added a docstring above the declaration of the GridModel.selModel property like so:
```
/** @type {StoreSelectionModel} */
selModel
```
and suddenly IntelliJ was happy to give me much more (see attached).  I think this is a really nice pickup when using the API.  Think this will work the same for @returns on a method as well (need to verify).

screen shot 2018-05-31 at 1 59 46 pm
screen shot 2018-05-31 at 1 59 23 pm

The docstring format is unfortunately a bit noisy, so I am not proposing we do this for every property of every class. But for a key one like this I would like to add - we have two complex classes existing within Hoist with public methods with interesting signatures we will use frequently.

### LINKED ISSUE #515: GridModel constructor doc comment params format incorrect
Need to specify that it takes a config object and namespace all configs under that to conform to spec / get nice doc preview to work correctly in IDE.

```
@param {Object} config
@param {string} config.foo - the foo string
...etc
```

Would be good to have a look at constructors for other key objects to ensure they are also in correct format.

### LINKED ISSUE #525: Document Column API, interaction with GridModel and related classes
WIP on https://github.com/exhi/hoist-react/tree/gridAndMiscDocs

amcclain:
+ Additional standardization / completion of public-facing API docs throughout.
+ Add constructor documentation for new Column API. 
+ Mirror service API signatures for XH aliases - they are not overly complex/dynamic, and XH is a primary entry point. (Ideally we would find a way to syndicate the actual docs to these methods...).
+ Remove `contextMenuItem.prepareFn` - this was documented as if it were a storeContextMenuItem but unused / not clearly applicable. Left / improved docs on the store variant.
+ Updated `GridStateModel.xhStateId` -> `gridId` - no need for "xh" prefix on a property of a class created and wholly managed by the toolkit.
+ Use of `@typedef` annotations to clarify some doc comments / APIs.
+ Use of `@member` annotations w/typing on certain key public properties.
+ Fixes #336
+ Fixes #515
+ Fixes #525

### DIFF
--- a/desktop/cmp/contextmenu/StoreContextMenuItem.js
+++ b/desktop/cmp/contextmenu/StoreContextMenuItem.js
@@ -5,9 +5,6 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
-/**
- *  Basic Model Object for ContextMenu for Stores.
- */
 export class StoreContextMenuItem {
 
     text;
@@ -20,17 +17,16 @@ export class StoreContextMenuItem {
     recordsRequired;
 
     /**
-     * @param {string} text - label to be displayed.
-     * @param {Object} icon - optional icon to be displayed.
-     * @param {Object[]} items - child menu items.
-     * @param {function} action - function of the form (item, record, selection) => {}
-     *      Executed when the user clicks the menuitem.
-     * @param {boolean} disabled - true to disable this item.
-     * @param {boolean} hidden - true to hide this item.
-     * @param {function} prepareFn - function of the form (item, record, selection) => {}
-     *      The prepareFn is a callback that is triggered before each time the menuitem is shown.
-     *      It can be used to modify the menuitem based on the record / selection.
-     * @param {(number|boolean)} recordsRequired - how many records must be 'active'
+     * @param {Object} c - StoreContextMenuItem configuration.
+     * @param {string} c.text - label to be displayed.
+     * @param {Object} [c.icon] - icon to be displayed.
+     * @param {Object[]} [c.items] - child menu items.
+     * @param {storeCtxMenuItemActionFn} [c.action] - called on store context menu item click.
+     * @param {boolean} [c.disabled] - true to disable this item.
+     * @param {boolean} [c.hidden] - true to hide this item.
+     * @param {storeCtxMenuItemPrepFn} [c.prepareFn] - called prior to context menu item show,
+     *      available to modify the item based on current record / selection at time of show.
+     * @param {(number|boolean)} [c.recordsRequired] - how many records must be 'active'
      *      (selected and / or clicked upon) for the menuitem to be enabled.
      *      int: specifies exactly n number of records. Defaults to 1 for single record actions.
      *          Can specify 0 to only enable menuitem if no records are active.
@@ -57,3 +53,18 @@ export class StoreContextMenuItem {
         this.recordsRequired = recordsRequired;
     }
 }
+
+/**
+ * @callback storeCtxMenuItemActionFn - called on store context menu item click.
+ * @param {ContextMenuItem} item - the menu item itself.
+ * @param {Object} [record] - row data object (entire row, if any).
+ * @param {Object[]} [selection] - all currently selected records (if any).
+ */
+
+/**
+ * @callback storeCtxMenuItemPrepFn - called prior to store context menu item show, available to
+ *      modify the item based on current record / selection at time of show.
+ * @param {ContextMenuItem} item - the menu item itself.
+ * @param {Object} [record] - row data object (entire row, if any).
+ * @param {Object[]} [selection] - all currently selected records (if any).
+ */
--- a/desktop/cmp/contextmenu/StoreContextMenuItem.js
+++ b/desktop/cmp/contextmenu/StoreContextMenuItem.js
@@ -21,10 +21,10 @@ export class StoreContextMenuItem {
      * @param {string} c.text - label to be displayed.
      * @param {Object} [c.icon] - icon to be displayed.
      * @param {Object[]} [c.items] - child menu items.
-     * @param {storeCtxMenuItemActionFn} [c.action] - called on store context menu item click.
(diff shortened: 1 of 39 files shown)


desktop/cmp/contextmenu/StoreContextMenuItem.js
```diff
@@ -56,4 +52,19 @@ export class StoreContextMenuItem {
         this.prepareFn = prepareFn;
         this.recordsRequired = recordsRequired;
     }
-}
\ No newline at end of file
+}
+
+/**
+ * @callback storeCtxMenuItemActionFn - called on store context menu item click.
```
lbwexler: interesting.

lbwexler: Cool -- Wonder if we could discuss the convention for this callback type  name?  Wondering if capital casing would be more appropriate.  Also, is this complete unscoped, name wise?

desktop/cmp/grid/Grid.js
```diff
@@ -16,40 +16,42 @@ import {agGridReact, navigateSelection} from './ag-grid';
...
+         * This constitutes an 'escape hatch' for applications that need to get to the underlying
+         * ag-Grid API.  It should be used with care. Settings made here might be overwritten and/or
+         * interfere with the implementation of this component and its use of the ag-Grid API.
          */
         agOptions: PT.object,
 
         /**
          * Callback to call when a row is double clicked.  Function will receive an event
          * with a data node containing the row's data.
+         * @see {@link https://www.ag-grid.com/javascript-grid-events/#properties-and-hierarchy|ag-Grid Event Docs}
```
lbwexler: Little nervous about adding too many direct urls to other systems -- I guess as long as we are committed to keeping them up to date its ok

desktop/cmp/grid/GridModel.js
```diff
@@ -207,24 +229,23 @@ export class GridModel {
     }
 
     /**
-     * Set sort by column
-     *
-     * This method is no-op if provided any sorters without a corresponding column
+     * This method is no-op if provided any sorters without a corresponding column.
+     * @param {(GridSorter|GridSorter[])} sorters - column(s) and direction for sorting.
```
lbwexler: Should we have a convention for these typedefs (and also the callbacks) to be clear that they are only documentation time artifacts/names?  e.g. 'DGridSorter' /'D_GridSorter'

desktop/cmp/grid/Grid.js
amcclain: Paradoxically, if the ag-Grid docs were easier to work with I would be less inclined to add these. As it is, it was consistently annoying enough to locate some key bits and pieces, I thought it might be helpful.  I'll keep an eye on them!

desktop/cmp/grid/GridModel.js
amcclain: Would you use `i` for interface, or does that seem too much like a reader should be able to find typescript lurking around somewhere?

The defs themselves are very useful in terms of structuring the docs w/o making the primary signature totally unreadable, so would like to keep using them - although in some cases I bet they will end up being placeholders for actual class implementations.

desktop/cmp/grid/GridStateModel.js
```diff
@@ -8,26 +8,42 @@ import {XH, HoistModel} from '@xh/hoist/core';
...
+     * Version of grid state definitions currently supported by this model.
+     * Increment *only* when we need to abandon all existing grid state that might be saved on
+     * user workstations to ensure compatibility with a new serialization or approach.
+     */
+    static GRID_STATE_VERSION = 1;
+    static STATE_SAVE_DEBOUNCE_MS = 500;
 
     gridModel = null;
-    xhStateId = null;
+    gridId = null;
```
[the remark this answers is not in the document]
amcclain: For me the key was that this is the identifying pointer back to the grid from the state model, so prioritized grid over state since the state part is implied by it being a member prop on the state model. But more than happy to have both - will rename.

desktop/cmp/contextmenu/StoreContextMenuItem.js
amcclain: How about `Cb` vs. `Fn` at the end? 

Scoping of the names follows the general rules for JsDoc, of course, which is to say your guess is as good as mine as to how these things get resolved. They should come up with some kind of scheme for package names - could use something super-clever like your company's internet domain name, but maybe switch the order around so the last part comes first and then...oh wait...

You will see I am experimenting in one or two places with the `@alias` tag to see how we could attach more specific / reliable names to some classes with very general name - `@alias HoistColumn`.  What matters to me in the end is what the tooling we care about does with such things. My next step as soon as I can breathe again is to work on jsDoc > md conversion to generate docs for publishing, and in my mind the format and readability/linkability of those docs will be a big input into what tags/names we use and if they are effective or not.

desktop/cmp/grid/GridStateModel.js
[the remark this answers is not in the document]
lbwexler: This is fine as is.  Realize now that the main entry point for applications is really the 'stateModel' param for the GridModel constructor.  That has the word state in it, which is really my main concern.

desktop/cmp/contextmenu/StoreContextMenuItem.js
amcclain: Swapping to `Cb` at end now, but realizing we might want to think about the constructor args themselves a bit more - we have `action` and `prepareFn` - not exactly consistent there.....

PR #531, opened 2018-08-14
Day field improvements: Fixes #502, Fixes #497

### LINKED ISSUE #497: DayField - with commitOnChange=false, selected date does not appear in input field
With commitOnChange=false (which is the default), when clicking on a date in the popover, the selected date does not appear in input field after the popover closes.

### LINKED ISSUE #502: DayField - typing in date beyond max/min bounds throws on commit
If you manually type in a date that's beyond the min/max limits of a DayField and press <enter>, the component will throw an exception on render.

febbraiod:


desktop/cmp/form/DayField.js
```diff
@@ -95,6 +95,11 @@ export class DayField extends HoistField {
 
     onChange = (date, isUserChange) => {
         if (!isUserChange) return;
+        const {minDate, maxDate} = this.props;
+
+        if (date < minDate) date = minDate;
+        if (date > maxDate) date = maxDate;
+
```
febbraiod: Debatable if this is the correct incremental step. I could also see just keeping the previous value. My thought here is that one might be entering some arbitrary 'really long time ago' or 'really long time from now' date that is meant to mean 'show me everything'.

Ext actually displays the invalid date in this case but shows an inline validation failure. Which Lee mentioned we are headed toward.

lbwexler: I agree its debateable what the right things to do is, but we definitely need to do something, and this seems like a good first step.  

We can revisit shortly during our validation work

PR #532, opened 2018-08-14
Add message to Promise models, display in loadMask, Fixes #426

### LINKED ISSUE #426: Make loadMask text dynamic
In a client app, I used a single loadMask for successive api calls and updated an observable  loadMaskText property on a model based on the call being made or based on a status being returned by the server. I fed that into the loadMask via the render method on the mask's parent component. When that prop changed the loadmask rerendered with the new text.

Lee thought it was a good move and should be built into loadMasks. One way of doing this might be to put an observable loadMaskText prop on our LastPromiseModel. A LoadMask could check its bound model for that prop and use that, before defaulting to it's own text prop. Applications could then dynamically set this prop on the LastPromiseModel. Lee also thought we might do this via a ref on the mask itself. Open ticket for discussion.

febbraiod:


promise/model/LastPromiseModel.js
```diff
@@ -9,11 +9,13 @@ import {observable, action, computed} from '@xh/hoist/mobx';
 /**
  * Tracks the resolution state of the last of a series of promise invocations.
  * Currently it only tracks the *last* invocation, but could be enhanced to maintain historical statistics.
+ * Has an observable message property that if set will be displayed in a linked loadMask
```
amcclain: I'd move this to a doc comment right on the property itself.

Also - do we want a constructor now to set this initially?

lbwexler:
I am reviewing this now...My first thought -- which I shared with Don -- is that this is starting to look like 'MaskModel' or a "WaitingModel".  Could imagine adding a "percentComplete" property that could be set as well.  Wondering if we should rename this model?    

Or would it be more appropriate for these to just be props on the load mask (that could be set to observable properties on its parent).  Would like to see the use case again...

amcclain:
Not to get to speculative, but I can imagine how we might have other displays (e.g. a progress bar or spinner in a toolbar or bottom status bar) for async operations, arguing for not necessarily linking this too closely to the mask UI. 

But yes something like WaitingModel (if not _quite_ that name) might help communicate the use case better....

febbraiod:
Re: props on load mask that can be set to observable parent prop.

That's how we did this in the client app. Using just the text prop, although I could imagine adding a percentage prop as well.  

The use case was that we make regular repeated calls to a report 'status' record. That status was expected to be updating via a background job on a separate service as it builds the report records. Once that status reads 'done' we know we can safely query for the entire report. There were a handful of intermediate statuses that we display in the mask along the way. Glad to show the code whenever is convenient.

lbwexler:
Anselm -- totally agree with your use cases -- "mask" not good.

What we have is accurate in that the thing really is tied closely to
promises,  but a bit too wonky.
The main way its listed to Promises, really, is that we want applications
to "wire" it up, by just calling Promise.link()

Ideas, in no particular order:
"ProgressModel",  "PromiseTrackingModel", "AsyncTrackModel",
"AsyncTrackingModel", "PromiseModel", "LoadModel"  (allthough realize UI
may not always be waiting on a load).

We could easily build in the "Last vs. Multi" as an option on a single
class, if that would simplify things.

lbwexler:
Closing this in favor of new PR, based on some of the comments.  I merged in Don's changes for Hoist React and Toolbox

PR #533, opened 2018-08-14
Accept Column objects or configs in GridModel.setColumns

haynesjm42:
Encountered this issue while updating a client app to v12. This client app makes heavy use of `setColumns` as the columns are dynamic based on a user-configured object.

### DIFF
--- a/desktop/cmp/grid/GridModel.js
+++ b/desktop/cmp/grid/GridModel.js
@@ -8,18 +8,7 @@ import {HoistModel, XH} from '@xh/hoist/core';
 import {action, observable} from '@xh/hoist/mobx';
 import {StoreSelectionModel} from '@xh/hoist/data';
 import {StoreContextMenu} from '@xh/hoist/desktop/cmp/contextmenu';
-import {
-    castArray,
-    defaults,
-    find,
-    findLast,
-    isPlainObject,
-    isString,
-    last,
-    orderBy,
-    pull,
-    uniqBy
-} from 'lodash';
+import {castArray, defaults, find, findLast, isPlainObject, isString, last, orderBy, pull, uniqBy} from 'lodash';
 import {Column} from '@xh/hoist/columns';
 import {throwIf, warnIf} from '@xh/hoist/utils/JsUtils';
 import {ColChooserModel} from './ColChooserModel';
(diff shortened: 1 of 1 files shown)


desktop/cmp/grid/GridModel.js
```diff
@@ -264,10 +252,11 @@ export class GridModel {
         return [...this.columns];
     }
 
-    /** @param {Column[]} cols */
+    /** {(HoistColumn[]|Object[])} cols - Columns, or configs to create them. */
```
amcclain: Missing `@param`

desktop/cmp/grid/GridModel.js
```diff
@@ -291,7 +281,7 @@ export class GridModel {
             pull(newCols, emptyFlex).push(emptyFlex);
         }
 
-        this.setColumns(newCols);
+        this.columns = newCols;
```
amcclain: Wondering if there's any value in consistently using the setter internally as well as for the public API, to consolidate the `@action` and any other logic / checks that might prove useful. We also set the cols at end of `setGroupBy()`.

Cost would be looping / rebuilding the array again, but don't see that as material given grids will only have so many cols.

@lbwexler do you have an opinion here? I do not feel strongly just wanted to ask the question.

desktop/cmp/grid/GridModel.js
```diff
@@ -264,10 +264,11 @@ export class GridModel {
         return [...this.columns];
     }
 
-    /** @param {Column[]} cols */
+    /** {(HoistColumn[]|Object[])} cols - Columns, or configs to create them. */
     @action
     setColumns(cols) {
-        this.columns = [...cols];
+        this.columns = cols.map(c => c instanceof Column ? c : new Column(c));
+        this.validateColumns();
```
lbwexler: Can we refactor this to validate *before* mutating the columns.  Think it makes more sense, and it allows a caller of this model to recover from a mistake without wrecking the state of a healthy GridModel.

Just refactor the validate method to take the proposed new columns.

desktop/cmp/grid/GridModel.js
lbwexler: Good thought, but I would be inclined to leave it as is, especially since the setter is now doing warning/validation.  Would be very annoying if my grid kept re-warning me every time I resized a column!

haynesjm42: That was my thought as well, want to avoid duplicate validation warnings if possible. One thing to note is that any caller of `setColumns` (such as the col chooser) could cause duplicate warnings. Actually, if the user removed all flex columns from the grid in the chooser and isn't using the emptyFlexCol, it would always cause a warning with this change. Not sure what (if anything) we should do for that case..

lbwexler:
Yes, I discussed this case with Anselm -- I think we may have some
additional tweaks in front of us.

The warnings don't worry me so much (we could have a second flag to
suppress that for internal callers, if it really bothered us).  I do think
we could consider having the grid itself just insert an empty flex at the
end if there are no flex columns?  Might save some boilerplate?

Also, wondering if it ever makes sense to have flex columns with a grid
column chooser?  Maybe that's the fundamental problem, and something we
should warn against.

PR #546, opened 2018-08-17
Chooser should check for groupable data before applying class, Fixes #544

### LINKED ISSUE #544: LeftRightChooser options should not be indented if not grouped
Appears as if `xh-lr-chooser__group-row` always applied.

febbraiod:


febbraiod:
Closing for improvement

PR #547, opened 2018-08-18
Pending task

lbwexler:


lbwexler:
Just wanted to add to this, that I decided to make the default for spinner to be false (even though it seems like we most frequently want it set).  It seemed somehow more appropriate, and the extra boilerplate not horrible, if it forces developers to think about it it a little. Happy to be convinced otherwise -- we could also bring back LoadMask, just to default that flag, but that seemed like extra complexity that would not be worth it.

Issue #317, opened 2018-05-24
Validation Support for HoistFields


lbwexler:
We would like to  sketch out a design for simple validation support in HoistFields.  

Should include a list of the template methods, events, etc. and how an app would use them to modify validation, and how a control writer would use them to display validation.

Note: I do not want to start coding this without a plan. This is a tricky thing to make useful to applications -- very easy to build something that is unusable.

Also, want to keep this as simple as possible, and still have it useful for applications (what else is new :-))

amcclain:
Re-routing to John for now given that he has put together some solutions around this for a particular client application. 

I liked the use of a toast in particular to provide additional details about validation errors in a lightweight way that doesn't interfere with the overall layout of a form.

amcclain:
@haynesjm42 would like to touch base on this next week to see if we can move forwards.  Also want to loop in @TomTirapani as he is planning to build out forms for a client app that will need clear validation status.

haynesjm42:
This may be somewhat larger in scope than what we were thinking for this ticket, but here is what I would propose ..

I took a look at [validate.js](https://validatejs.org/) and it looks like it would be a great library to bring in to handle the actual validation logic. I would see us building a ValidationModel class to wrap up the library similar to what we are doing with RouterModel. The form model/component would create a ValidationModel with the constraints configuration object under the assumption that the attribute names in the constraints match the field names that the HoistFields are associated with (in bound mode, if not using bound mode then it would be entirely up to the app to handle linking validation status to the HoistField components). The validation model would be passed to the HoistField components so they could trigger validation of their field (and only their field) when the value changes. The form model/component could trigger a full or partial validation programmatically based on what makes sense for the form (clicking the save button, etc.). Any ValidationModel methods that perform validation should return a Promise to make sure we allow for asynchronous validations.

The current validation results would be stored in the ValidationModel in an observable field (`@observable.ref results = {…}`) so the owning model/component and the individual field components are able to react to validation status changes as needed. 

For HoistFields I think we should just add a css class based on the current validation results for the associated field. We could provide default styling to add a red border around the input element and allow apps to override it as desired. I think that should be the extent of the styling we apply by default, if anything. We could even just let the app provide the "invalid" css class name to apply in the component props, and do nothing by default.

[...]

PR #552, opened 2018-08-21
Validation

haynesjm42:
#317 

See https://github.com/exhi/toolbox/compare/validation for example usage. It needs some TLC before it can be merged, not very pretty right now and the validations don't all make sense, but useful for testing and demonstrating the usage.

One thing I didn't realize at first was that the last commit to the [validate.js repo ](https://github.com/ansman/validate.js) was 9 months ago... not sure if it is completely dead or not, I couldn't find anything saying that it wasn't being maintained any longer. That being said, I think the approach the library takes is very solid and the source looks easily extendable, so a fork is always an option if we need bugs fixed or features added.

A couple notes on the changes:
* For the invalid styling I added scss files for various HoistField components as needed, though I did consider just adding everything to HoistField.scss to keep things in one place. Happy to change that if we think it's cleaner.
* I added `ValidationService` which handles some necessary init for the date validator so it knows how to parse and format dates. This is also where I would see us adding something like a `registerValidator()` method for apps (or us) to add custom validators, and potentially adding additional wrappers around the validate.js apis.


cmp/validation/ValidationErrors.js
```diff
@@ -0,0 +1,48 @@
...
+            {isValid} = model;
+
+        // Nothing to show if the model is valid
+        if (isValid) {
+            return null;
+        }
+
+        let errors;
+        if (!fields) {
+            errors = model.allErrorMessages;
```
lbwexler: Can these two names be more symmetric?  'allErrorMessages' and 'getFieldErrors' seems like different things, when one is just a filtered version of the other.

cmp/validation/ValidationErrors.js
```diff
@@ -0,0 +1,48 @@
...
+        fields: PT.oneOfType([PT.arrayOf(PT.string), PT.string])
+    };
+
+    baseClassName = 'xh-validation-errors';
+
+    render() {
+        const {model, fields} = this.props,
+            {isValid} = model;
+
+        // Nothing to show if the model is valid
```
lbwexler: code speaks for itself here.

desktop/cmp/form/TextField.js
```diff
@@ -41,10 +41,13 @@ export class TextField extends HoistField {
 
     delegateProps = ['className', 'disabled', 'type', 'placeholder', 'autoFocus', 'leftIcon', 'rightElement'];
 
+    baseClassName = 'xh-text-field';
```
lbwexler: great cleanup with the application of baseClassName?

svc/ValidationService.js
```diff
@@ -0,0 +1,31 @@
+import {HoistService} from '@xh/hoist/core';
```
lbwexler: Don't think this needs to be a service, if it has no entry points.  Why not do this setup within ValidationModel?  Can run code when the module loads (or lazily when first one is created, but I think simplest is better)

core/ValidationModel.js
```diff
@@ -0,0 +1,143 @@
...
+
+            this.addReaction({
+                track: () => model[field],
+                run: () => this.validateFieldAsync(field)
+            });
+        });
+    }
+
+    /**
+     * Checks whether a field is valid in the current validation state.
```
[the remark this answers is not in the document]
amcclain: Agree re. textual descriptions that don't add much beyond the name of the method / property itself. In my mind, the primary benefit of the doc strings rests with the typing info. 

To that end, would also propose we augment/replace the class member comments above with `@member {type}` comments. Desktop GridModel has some of these.

core/ValidationModel.js
```diff
@@ -0,0 +1,143 @@
...
+     * @param field The field to check
+     * @returns {boolean} True if the field is valid, false otherwise
+     */
+    isFieldValid(field) {
+        return !has(this.errors, field);
+    }
+
+    /**
+     * Returns all validation error messages in a flat list
+     * @returns {Array} array of error message strings
```
amcclain: The return type here can be `{string[]}`

core/ValidationModel.js
```diff
@@ -0,0 +1,143 @@
...
+                track: () => model[field],
+                run: () => this.validateFieldAsync(field)
+            });
+        });
+    }
+
+    /**
+     * Checks whether a field is valid in the current validation state.
+     *
+     * @param field The field to check
```
amcclain: Would add a type here - it's a `{string}` right?

core/ValidationModel.js
```diff
@@ -0,0 +1,143 @@
...
+                return {isValid: this.isValid, result: this.errors};
+            })
+            .catch((errors) => {
+                this.setResult(assign({}, this.errors, errors));
+                return {isValid: this.isValid, result: this.errors};
+            });
+    }
+
+    //---------------------------------------------
+    // Implementation (internal)
```
amcclain: Sorry to nit-pick 😃 but the "internal" is pretty well implied by this standard little implementation comment block.

core/ValidationModel.js
```diff
@@ -0,0 +1,143 @@
...
+        fields = castArray(fields);
+        return flatten(fields.map(field => errors[field] || []));
+    }
+
+    /**
+     * Validates all fields on the model which have constraints configured.
+     *
+     * @returns {Promise} A Promise that resolves to an object containing the current validation state
+     */
+    async validateAsync() {
```
amcclain: @lbwexler John and I had discussed that the async version of this method seemed more powerful and consistent with our use of promise-based APIs elsewhere in the framework. It does introduce some possible complexities, however - e.g. seems like out-of-order returns could be a possible source of bugs here if we had truly async server-side validation.

The fact that the validation library rejects the promise on a validation failure also seemed like a bit of an odd choice to me (on their part).

Not sure if we need to consider or do anything with this now. I didn't / don't want to over-design.

desktop/cmp/form/HoistField.js
```diff
@@ -176,4 +182,23 @@ export class HoistField extends Component {
...
+    /**
+     * Override of the default implementation provided by HoistComponent so we can add the xh-field
+     * and xh-field-invalid classes (when appropriate)
+     */
+    getClassName(...extraClassNames) {
+        return classNames(this.baseClassName, this.props.className, ...extraClassNames, 'xh-field', this.invalid ? 'xh-field-invalid' : null);
+    }
+
+    /**
+     * @returns {boolean} whether the currently bound field's value is valid or not
```
amcclain: For clarity, suggest `true if the field's current value is invalid, false if valid or no validation model provided`.

Either that or we swap the sense around and make this `isValid()` which better matches the validationModel method.

desktop/cmp/form/HoistField.js
```diff
@@ -176,4 +182,23 @@ export class HoistField extends Component {
...
+     */
+    @computed
+    get invalid() {
+        const {validationModel, field} = this.props;
+        if (!validationModel) return false;
```
amcclain: I do wonder if it's worth returning `null` here when there is no ValidationModel. Would that ever be useful?

desktop/cmp/form/JsonField.scss
```diff
@@ -0,0 +1,12 @@
+.CodeMirror pre.CodeMirror-line {
+  box-shadow: inset 0 0 0 0px rgba(16, 22, 26, 0.4);
+  -webkit-box-shadow: inset 0 0 0 0px rgba(16, 22, 26, 0.4);
```
amcclain: I know you didn't add these - but what *are* they?

mobile/cmp/form/HoistField.js
```diff
@@ -10,6 +10,8 @@ import {PropTypes as PT} from 'prop-types';
 import {upperFirst, isFunction} from 'lodash';
 import {throwIf} from '@xh/hoist/utils/js';
 import {observable, computed, action, runInAction} from '@xh/hoist/mobx';
+import {ValidationModel} from '@xh/hoist/core';
```
amcclain: Thank you for keeping the entire mobile package in mind here. 😄

haynesjm42:
Lee and I spent a while looking through these changes and made some improvements, as well as a start towards managing async validations better, but he also brought up other issues with async validations that we aren't yet dealing with. He also brought up the use case (which is an immediate client need) where validation rules on a field need to be enabled or disabled based on the value of another field, which I don't think could be handled with the current light wrapper around validate.js. validate.js also doesn't seem to prove a means of providing a simple function to perform some custom validation for a single field.

The more I think about it the more I feel we should abandon validate.js constraints in favor of our own system. We could pretty easily build the basic validation functions that most forms require, and give complete flexibility in the form of accepting functions to perform validation checks. I was able to throw something together quickly that works well, and can continue down that path if we are okay with abandoning the validators that validate.js provides (though we can actually still use the util functions that they provide).

Here is an example of what the rules looks like with my new approach, where the `text2` validation reaction tracks `text1` (in addition to `text2`) and check if the individual rule should be checked via the `active` function. So `text2` is only required when `text` equals 'test', and the validation is re-run whenever `text1` or `text2` changes.
 
```
{
    text1: {
        rules: [
            {
                check: ({value}) => (!isNil(value) && !isEmpty(value)) || 'text1 is required!'
            }
        ]
    },
    text2: {
        track: () => [this.text1],
        rules: [
            {
                active: () => this.text1 === 'test',
                check: ({value}) => isNil(value) || isEmpty(value) ? 'text2 is required when text1 === \'test\'!' : null
            }
        ]
    }
}
```

lbwexler:
Am looking at this now.  Will check it out and play with it a bit.

haynesjm42:
Pushed update to the toolbox branch. If this approach seems better I will clean up the implementation and update the doc strings, and generally flesh out the usage a bit with different types of validation rules. Also need to more closely review and test the async rule support, wrote most of this quickly in the car this morning :)

lbwexler:
Thanks John -- Give me a bit of time to play with this tonight -- I have a
few ideas.  Let's plan to circle back tomorrow A.M.?

lbwexler:
This work has been incorporated into a new branch that is under review in a separate PR

PR #557, opened 2018-08-23
New Decorators in hoist/mobx

lbwexler:
Plan is to roll-out the use of these gradually.  No need for any bombings, especially as we just bombed some of this same code in the opposite direction

### DIFF
--- a/mobx/decorators.js
+++ b/mobx/decorators.js
@@ -1,0 +1,40 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+import {action, observable} from 'mobx';
+import {capitalize} from 'lodash';
+
+/**
+ * Decorator to add a simple mobx action of the form 'setPropName()' to a class.
+ *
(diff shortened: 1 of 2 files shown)


amcclain:
Wondering if we would ever use `@settable` in practice, other than within this file to power `@bindable`. Would there be a use case for that? It seems to me that any property where `@setttable` makes sense would also be `@observable` - is that correct?

If that's the case, would suggest we put bindable first in this file, and potentially add a line to the comment either on that or settable pointing out that bindable is the primary public decorator. Could also not export settable for now if we weren't sure it would be needed / desirable in application code.

Let me know of course if I am missing something!

lbwexler:
Well, thought about this a while, and originally was going to stick with
just bindable.  But actually felt like settable would still be useful if
you wanted to use it with one of the variants of  observable (e.g.
observable.shallow, observable.ref, etc.)

Not sure that this is such an important use case -- in most cases like
that, we will have an action that is manipulating those observables.  They
aren't primitives that we bind controls to.  If we wanted to minimize the
API surface, would be happy to remove it.

amcclain:
👍 thanks that makes sense - I had not thought of the observable variants. I suppose if we find ourselves wanting to have easy bindable support as well as ref/shallow observability, we can add our own variants here.

PR #561, opened 2018-08-28
Add support for default tipFn:true formatter

### LINKED ISSUE #434: Number formatters - tipFn shortcut for original value
Would like to make it easy to enable a tooltip on formatted numbers that displays the original, full precision value with some improvements.

Right now I believe you can specify `tipFn: v => v` to output the original value, but would be nice to support a shortcut like `tipFn: true` that showed the original value in a tooltip with:

* Commas inserted if >= 1000
* Wrapped in () if ledger formatting also specified
* Potentially still cap precision at `MAX_NUMERIC_PRECISION` in case we get crazy floating point numbers?

The first two in particular could significantly improve the legibility / consistency of the output.  Believe this would be the equivalent of `tipFn: numberRenderer({ledger, forceLedgerAlign: false, precision: 12, zeroPad: false})`

nadkarniarjun:
Fixes #434

### DIFF
--- a/format/FormatNumber.js
+++ b/format/FormatNumber.js
@@ -5,7 +5,7 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
-import {defaults, isFinite, isString} from 'lodash';
+import {defaults, isFinite, isString, isFunction} from 'lodash';
 import numeral from 'numeral';
 
 import {Exception} from '@xh/hoist/exception';
@@ -220,7 +220,7 @@ function fmtNumberElement(v, opts = {}) {
 
     return span({
         className: cls.join(' '),
-        title: tipFn ? tipFn(originalValue) : null,
+        title: processTipFn(tipFn, originalValue),
         items: items
     });
 }
@@ -254,7 +254,7 @@ function fmtNumberString(v, opts = {}) {
     }
 
     if (tipFn) {
-        str = fmtSpan(str, {className: 'xh-title-tip', title: tipFn(originalValue)});
+        str = fmtSpan(str, {className: 'xh-title-tip', title: processTipFn(tipFn, originalValue)});
     }
 
     return str;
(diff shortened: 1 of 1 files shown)


format/FormatNumber.js
```diff
@@ -316,6 +316,17 @@ function isInvalidInput(v) {
     return v == null || v === '';
 }
 
+function processTipFn(tipFn, originalValue) {
+    if (tipFn === true) {
+        const formatter = numberRenderer(originalValue, {ledger: true, forceLedgerAlign: false, precision: 12, zeroPad: false});
+        return formatter(originalValue);
```
amcclain: We don't want to create a renderer here just to call it - we can simply call `fmtNumber(v, {...theFmtOpts}`.

The renderer functions (e.g. `numberRenderer`) have a different signature: they _only_ take the options object and return a curried function that then _only_ takes the value to be formatted. 

Given that, this code isn't really working as you think it is - the renderer that's returned and assigned to `formatter` is a version of `fmtNumber` with all of its default options. The options you pass in as a second argument on line 321 aren't being respected - they are just silently dropped in a very javascript-y way.

Try configuring `tipFn:true` on a grid column that shows negative numbers - e.g. [the P&L col in Toolbox's SampleGrid](https://github.com/exhi/toolbox/blob/b27a1c976fa52af4bf0b8e0ed4073c628c265204/client-app/src/desktop/common/SampleGrid.js#L78) and you will see that the tip values are not formatted in ledger format.

Also:
* We have a constant in the FormatNumber class for `MAX_NUMERIC_PRECISION` we can use instead of the 12 (sorry I know that number was directly from the ticket).
* If we pass this function the original opts, we can use the ledger formatting only if it was requested for the overall formatter.  I think that makes sense - if a ledger format is not desired / expected for the actual output, it would be strange to appear in the tip.
  * I think however that ledger is the only opt that makes sense to consult at this time in the tip use-case...

nadkarniarjun: Done! Code updated, committed and pushed. 

Thanks for the feedback. Much appreciated.

PR #562, opened 2018-08-28
onKeyPress added to TextField

shravnk:
Pass onSubmit function to password TextField in LoginPanel

### DIFF
--- a/desktop/appcontainer/LoginPanel.js
+++ b/desktop/appcontainer/LoginPanel.js
@@ -53,7 +53,8 @@ export class LoginPanel extends Component {
                             field: 'password',
                             placeholder: 'Password...',
                             type: 'password',
-                            commitOnChange: true
+                            commitOnChange: false,
+                            onKeyPress: this.onSubmit
                         }),
                         text({
                             omit: !model.warning,
--- a/desktop/cmp/form/TextField.js
+++ b/desktop/cmp/form/TextField.js
@@ -36,7 +36,8 @@ export class TextField extends HoistField {
         /** Icon to display on the left side of the field */
         leftIcon: PT.element,
         /** Element to display on the right side of the field */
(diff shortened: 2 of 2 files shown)


amcclain:
👍 Thanks for getting started here. Happy to review / continue with this when we speak tomorrow (although of course don't let that delay you from any other planned updates).

shravnk:
Cool - Will come prepared with questions on some of the other updates.

shravnk:
I see three imperfect options here. All involve setting a handleKeyDown (hKD) class property on Login Panel:

1. Pass hKD to TextField as a new propType. Works, but the new propType is likely too specific. (current solution) 
2. Pass hKD as an onCommit propType to TextField. Does not require changing TextField or HoistField, but does not allow onCommit = true.
3. Add event listeners to <input> elements in LoginPanel. See code below - I've actually got this working but I know it's bad practice in React.

`    componentDidMount () {
        const dom = ReactDOM.findDOMNode((this));
        const inputs = dom.getElementsByTagName('input');
        Array.from(inputs).forEach((input) => {input.addEventListener('keydown', (ev) => this.handleKeyDown(ev))})
    }`

amcclain:
Hi - sorry we didn't get to touch base on this today.  Will be sure to connect tomorrow AM.

Personally, I thought we were quite close with the first cut - accepting an `onKeyPress` prop for our component and calling that (if provided) when the BP input fires the same event.  That had the advantages in my mind of not needing to invent / define a new prop.

All we needed to do (I think) was bind that to a dedicated handler on LoginPanel that checked the event key and submitted if <enter>. 

We definitely want to stay in React component world for these kinds of very common events and listeners vs. dipping into the raw DOM. 

Again though let's touch base in case I'm missing something or you've been reviewing with anyone else on the team and I'm banging that up. 😄

PR #566, opened 2018-08-29
Support posting JSON body with content-type application/json.

cnrudd:
This change is part of improvements made to the framework to support a proxy service (https://github.com/exhi/hoist-core/issues/49)

Many api calls will be made with posting a json body, which our current fetch wrapper does not support.

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -309,6 +309,8 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
+    postJson(opts)              {return this.fetchService.postJson(opts)}
+    putJson(opts)               {return this.fetchService.putJson(opts)}
     getUser()                   {return this.identityService.getUser()}
     getUsername()               {return this.identityService.getUsername()}
     getConf(key, defaultVal)    {return this.configService.get(key, defaultVal)}
--- a/core/XH.js
+++ b/core/XH.js
@@ -309,8 +309,6 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
-    postJson(opts)              {return this.fetchService.postJson(opts)}
-    putJson(opts)               {return this.fetchService.putJson(opts)}
     getUser()                   {return this.identityService.getUser()}
     getUsername()               {return this.identityService.getUsername()}
     getConf(key, defaultVal)    {return this.configService.get(key, defaultVal)}
--- a/desktop/cmp/rest/data/RestStore.js
+++ b/desktop/cmp/rest/data/RestStore.js
@@ -69,11 +69,9 @@ export class RestStore extends UrlStore {
     async saveRecordInternalAsync(rec, isAdd) {
         let {url} = this;
         if (!isAdd) url += '/' + rec.id;
-        return XH.fetchJson({
+        return XH[isAdd ? 'postJson' : 'putJson']({
             url,
-            method: isAdd ? 'POST' : 'PUT',
-            contentType: 'application/json',
-            body: JSON.stringify({data: rec})
+            body: {data: rec}
         }).then(response => {
             const recs = this.createRecordMap([response.data]);
             this.updateRecordInternal(recs.values().next().value);
--- a/desktop/cmp/rest/data/RestStore.js
+++ b/desktop/cmp/rest/data/RestStore.js
@@ -69,7 +69,7 @@ export class RestStore extends UrlStore {
     async saveRecordInternalAsync(rec, isAdd) {
         let {url} = this;
         if (!isAdd) url += '/' + rec.id;
-        return XH[isAdd ? 'postJson' : 'putJson']({
+        return XH.fetchService[isAdd ? 'postJson' : 'putJson']({
             url,
             body: {data: rec}
         }).then(response => {
--- a/svc/FetchService.js
+++ b/svc/FetchService.js
@@ -6,7 +6,7 @@
  */
 import {XH, HoistService} from '@xh/hoist/core';
 import {Exception} from '@xh/hoist/exception';
-import {castArray} from 'lodash';
+import {stringify} from 'qs';
 
 @HoistService()
 export class FetchService {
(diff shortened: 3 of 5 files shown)


svc/FetchService.js
```diff
@@ -135,4 +133,28 @@ export class FetchService {
...
+     *          user: 'jDoe'
+     *      }
+     *  }
+     *
+     *  But it does handle arrays:
+     *   {
+     *      reportColumns: ['name', 'age', 'weight']
+     *   }
+     */
+    objectToQueryString(obj) {
```
amcclain: I think this is a great place to reach for a well-known library -  https://www.npmjs.com/package/qs appears to be one example.

haynesjm42:
Just thinking about the usage of this, it would be nice if we could avoid having to set the `contentType` every time we want to POST as JSON. I wonder if we should accept `content` in the options as an alternative to `params`, which could automatically set the content-type header based on the data type of `content` (text/plain and application/json are probably enough)? Seems like this will be a common enough use-case that we should streamline it as much as possible to avoid each app writing their own wrappers around our `fetch` method.

An alternative could be adding some common use-case wrappers around fetch (like `fetchJson`) for these things, such as `postJson`

cnrudd:
@amcclain I agree with going for a library for query string parsing.
@haynesjm42 postJson sounds like a good way to go.  
Any objections to postJson, anyone?

amcclain:
Cool - I dropped the qs lib into a client app today for an unrelated change and it seemed to do the trick without any drama.

I like `postJson` and the symmetry with `fetchJson` quite a bit. Seems to me that's exactly the kind of conveniences we can offer here.

cnrudd:
@amcclain , @haynesjm42 
I've added the qs lib to hoist-react and am using it in FetchService.
I've also added the postJson method to FetchService.

The branch/PR is ready for further review.

core/XH.js
```diff
@@ -309,6 +309,8 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
+    postJson(opts)              {return this.fetchService.postJson(opts)}
```
lbwexler: Minor nit, but I would argue that we don't need these aliases (and we could probably do without the fetch()) alias as well).  The idea here, is just to provide shortcuts for the most frequently used service methods.   FetchService is a public API, and it is totally valid to access it via XH.fetchService.postJson(),

desktop/cmp/rest/data/RestStore.js
```diff
@@ -69,11 +69,9 @@ export class RestStore extends UrlStore {
     async saveRecordInternalAsync(rec, isAdd) {
         let {url} = this;
         if (!isAdd) url += '/' + rec.id;
-        return XH.fetchJson({
+        return XH[isAdd ? 'postJson' : 'putJson']({
             url,
-            method: isAdd ? 'POST' : 'PUT',
-            contentType: 'application/json',
-            body: JSON.stringify({data: rec})
+            body: {data: rec}
```
lbwexler: nice!

svc/FetchService.js
```diff
@@ -67,26 +72,25 @@ export class FetchService {
...
-            Object.entries(params).forEach(v => {
-                const key = v[0],
-                    vals = castArray(v[1]);
-                vals.forEach(val => paramsStrings.push(`${key}=${encodeURIComponent(val)}`));
-            });
-            const paramsString = paramsStrings.join('&');
-
-            if (method === 'POST') {
+            let qsOpts = {arrayFormat: 'repeat', allowDots: true};
+            qsOpts = Object.assign(qsOpts, opts.qsOpts);
```
lbwexler: could do this in one line by spreading opts.qsOpts into the new variable with ...}

svc/FetchService.js
```diff
@@ -117,13 +121,70 @@ export class FetchService {
      *     If not explicitly set in opts then the contentType will be set based on the method. POST
      *     requests will use 'application/x-www-form-urlencoded', otherwise 'text/plain' will be
      *     used.
+     * @param {Object} [opts.qsOpts] - Object of options to pass to the param converter, qs.
```
lbwexler: Why are we repeating all this parameter documentation for every fetchJSON/postJSON/putJSON method?  Very difficult to maintain, and we already note that the options bundle is a bundle to "pass through to fetch().  Can we only document the new parameters for each method, and have the docs provide a direct {link} back to the fetch method?

svc/FetchService.js
```diff
@@ -6,10 +6,67 @@
...
-import {castArray} from 'lodash';
+import {stringify} from 'qs';
 
 @HoistService()
 export class FetchService {
+
+    /**
+     * Send an HTTP request to a URL, and decode the response as JSON.
+     *
+     * @param {Object} opts - options to pass through to fetch, with some additions.
```
lbwexler: Could we avoid re-documenting the shared parameters in each of the 3 xxxJson() methods?  How about saying the opts object contains "options for FetchService.fetch()" and providing a {link} to it that method?   The current setup seems like a maintenance issue -- lets put as much of the shared  documentation in FetchService.fetch().....

svc/FetchService.js
```diff
@@ -101,32 +162,16 @@ export class FetchService {
...
-    async fetchJson(opts) {
-        const ret = await this.fetch({acceptJson: true, ...opts});
-        return ret.status === 204 ? null : ret.json();
-    }
-
     //-----------------------
     // Implementation
     //-----------------------
+    async sendJson(opts) {
+        opts = Object.assign(opts, {
```
lbwexler: Can use ... operator here.

svc/FetchService.js
cnrudd: True. Done.

svc/FetchService.js
cnrudd: How would you do that without overwriting the stringify  of opts.body?

cnrudd: true. done.

svc/FetchService.js
cnrudd: Agreed.  I have edited the comments to reduce repetition.

core/XH.js
cnrudd: I like these aliases.  I would argue these are some of the most frequently used service methods.

svc/FetchService.js
```diff
@@ -104,29 +113,50 @@ export class FetchService {
...
-     * @param {Object} [opts.params] - parameters to encode and send with the request body (for POSTs)
-     *      or append as a query string.
-     * @param {string} [opts.method] - The HTTP Request method to use for the request. If not
-     *     explicitly set in opts then the method will be set to POST if there are params,
-     *     otherwise it will be set to GET.
-     * @param {string} [opts.contentType] - value to use in the Content-Type header in the request.
-     *     If not explicitly set in opts then the contentType will be set based on the method. POST
-     *     requests will use 'application/x-www-form-urlencoded', otherwise 'text/plain' will be
-     *     used.
+     * This method delegates to @see {fetch} and accepts the same options.
```
amcclain: Note that the `@see {foo}` annotation here doesn't really do anything - jsDoc needs those to be on their own line to be parsed / effective - and most likely to use the namespaced version `FetchService#fetch`

We will need to see what we end up doing w.r.t. docs like this.  It's unfortunate not to have the doc on the method that is most commonly used, as that is where you would get assistance and inline help from an editor. I use that consistently in my own development and get a lot of benefits from it. (I believe I had copied the relevant docs myself to provide this.)

We can revisit as we go. Just please be aware. Yes, this might in the end involve some careful duplication of comments, but I think we should treat the doc comments like we do any other code, reviewing them carefully and keeping them up to date.

PR #569, opened 2018-08-31
Pass the store through to the data view item component

haynesjm42:
Minor enhancement to pass the DataView store in the item component props (along with the item record). The need for this arose in a client app where my data view items have buttons to perform crud operations so they need access to the store.

### DIFF
--- a/desktop/cmp/dataview/DataView.js
+++ b/desktop/cmp/dataview/DataView.js
@@ -7,8 +7,7 @@
 
 import {Component} from 'react';
 import {XH, HoistComponent, elemFactory, LayoutSupport} from '@xh/hoist/core';
-import {grid} from '@xh/hoist/desktop/cmp/grid';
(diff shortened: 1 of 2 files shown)


lbwexler:
My first comment is that this is a great opportunity for us to use the {@callback} notation in jsdoc.  We have used it a few times in the source already and are trying to roll it out.  Take a look at other examples like the Column API, or the StoreContextMenuItem API.

Regarding the actual signature, would it make sense to have the second argument be a 'metadata' object that *contained* the store and/or the DataView?  Might be a bit more flexible for adding additional secondary bits, and also seems similar to the approach that grid elementRenderer is taking.

lbwexler:
Sorry for the delay here.  I discussed in detail with John and totally misunderstood this.  Didn't realize it is actually an elemFactory in the API, not an arbitrary callback.

We may have additional work around documenting the elemFactory type, but this seems appropriate to merge.

PR #570, opened 2018-08-31
Add Multiselect HoistField, first round tag picker

febbraiod:


amcclain:
👁ing at this now - would love to get something merged in shortly as we have a few use-cases built up for it.

amcclain:
I merged the latest in from develop on this branch, as well as the Toolbox demo branch.  (On Toolbox this essentially meant just recreating the example as the file had been renamed / reworked.)

Seeing an immediate issue where any query text entered to find an option is not cleared when the option is selected. Were you seeing the same earlier or is this a new issue?  Didn't dig into it, figured I would be lazy and check in here first 😄 

screen shot 2018-09-05 at 9 00 47 am

febbraiod:
Hey Anselm. Didn't notice that before but I think there might be an affordance for that in blueprint. Might we want to leave the query there in some cases? If you type 'c' and there are multiple results you want to select do we really want to pull that result out from under you?

One other issue that isn't immediately obvious from the example is that it appears our intended affordance for normalizing dynamically changing internalOptions isn't working. Lee asked me to test that out after CR yesterday and I haven't gotten it working yet (part of the reason this wasn't merged earlier). Was going to try again today but so far have been working with clients all morning.

amcclain:
I think once you select a value, you clearly don't want to have the text you used to find that just-selected value sitting in the form control in an ambiguous state. 

I don't think it's too likely that having typed "c" and selected California I would naturally find it convenient to have my "c" query still there and ready to be added on to again (e.g. to find Connecticut). More likely I am either done (and want the control to only show the selected tag(s)) or I am going to search for something else and it might well start with a different string.

Let me know of course if I've missed something here!

febbraiod:
No I don't think you are missing anything, I was just thinking about the blueprint example of favorite movies. Maybe I like The Godfather and The Princess Bride, and I used 'The' to narrow my results maybe I want to pick both and now I have to type 'The' again. I can see it both ways. Will look to clear the query in any case.

amcclain:
🎉 great!

PR #574, opened 2018-08-31
Tooltip API updates for Column + number/date formatters

nadkarniarjun:
Referenced in https://github.com/exhi/hoist-react/issues/509


amcclain:
@lbwexler do you want to provide a transitional alias for existing uses of `tipFn` on fmtNumber so this is not a breaking change?

PR #583, opened 2018-09-04
New "selectOnFocus" property for Text, TextArea, Number

nadkarniarjun:
See https://github.com/exhi/hoist-react/issues/554 for original issue

### DIFF
--- a/desktop/cmp/form/NumberField.js
+++ b/desktop/cmp/form/NumberField.js
@@ -43,7 +43,9 @@ export class NumberField extends HoistField {
         /** Alignment of numbers in field, default to 'right' */
         textAlign: PT.oneOf(['left', 'right']),
         /** Icon to display on the left side of the field */
-        leftIcon: PT.element
+        leftIcon: PT.element,
+        /** Whether text in field is selected when field receives focus */
+        selectOnFocus: PT.bool
     };
 
     static shorthandValidator = /((\.\d+)|(\d+(\.\d+)?))(k|m|b)\b/gi;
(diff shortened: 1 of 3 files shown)


lbwexler:
Code looks good to me.  

I do like the name 'selectOnFocus' much better than 'selectedOnFocus' -- difficult to articulate why -- perhaps it seems more natural, and inline with similar names in our toolkit? Or because there is no guarantee it will actually remain 'selected'?

(See commitOnChange, for an example in this control with my preferred name)

nadkarniarjun:
I've used `selectOnFocus` in the code, but have interchangeably used `selectedOnFocus` when describing the change in GitHub. I've changed those to be consistent.

lbwexler:
Got it Arjun -- sounds good - sorry I missed that

desktop/cmp/form/NumberField.js
```diff
@@ -114,5 +116,11 @@ export class NumberField extends HoistField {
         return parseFloat(value);
     }
 
+    onNumberFieldFocus = (ev) => {
+        if (this.props.selectOnFocus === true) {
```
amcclain: I think it's worth simplifying these to `if (this.props.selectOnFocus) {}` given that it should be reliably confirmed as a boolean via propTypes (and some other truthy value would not be the worst thing here).

More interestingly, @lbwexler did you and Arjun discuss the rather sticky bit of the superclass declaring `onFocus` as a property assigned to an arrow function (vs. a method that could be called via `super.onFocus()` - but would need to be properly bound for the handlers). 

This was my best suggestion - declare a more specific handler and use that - but I was treating that as somewhat provisional as it seemed like this could generally be a source of mistakes....

nadkarniarjun: I'll make the code change for `(this.props.selectOnFocus)` now as a separate commit.

Also, I did not have the discussion with @lbwexler about an arrow function vs. a method for these kinds of properties ... but will today.

PR #584, opened 2018-09-04
Support for markClass

lbwexler:
This formalizes (and fixes) our evolving ad-hoc method for marking classes in mixins with an identifying flag. 

Also, actually using this now to introspect the presence of FieldSupport.

### DIFF
--- a/utils/js/ClassUtils.js
+++ b/utils/js/ClassUtils.js
@@ -8,6 +8,23 @@
 import {forOwn, isPlainObject} from 'lodash';
 import {throwIf} from '@xh/hoist/utils/js';
 
+
+/**
+ * Decorate a class and its instances with a boolean flag set to true.
+ *
+ * Useful for providing an identifying flag for marking objects.  Frequently used
+ * in mixins to add 'isXXX' or 'hasXXXX' identifiers.
+ *
+ * For an example, see HoistComponent and the 'isHoistComponent' property.
+ *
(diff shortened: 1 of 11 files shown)


utils/js/ClassUtils.js
```diff
@@ -8,6 +8,23 @@
...
+ * in mixins to add 'isXXX' or 'hasXXXX' identifiers.
+ *
+ * For an example, see HoistComponent and the 'isHoistComponent' property.
+ *
+ * @param {String} flag
+ */
+export function markClass(C, flag) {
+    C[flag] = true;
+    C.prototype[flag] = true;
+}
```
amcclain: Looks great! Nice to have this buttoned up - and actually installing the flag where we want to find it!

1) What about `Defines the given boolean property, set to true, on a Class and its instances.` for the first line of the comment, only to avoid the use of the loaded term "decorate"?
2) Any advantage to using `Object.defineProperty(C, flag, {value: true, writable: false})` to make it clear this is not to be overwritten? (Called for `C` and `C.prototype`.) . I don't really foresee this being an issue in practice, but seems like it might be appropriate given the use-case here within ClassUtils.

Those were my only thoughts - please take or leave and merge anytime (or chat me and I can do so).

lbwexler: Will make both changes  and merge.  Excellent.

PR #587, opened 2018-09-05
Remove hoistComponentFactory()

lbwexler:
This seems like it is not worth the API complexity, and better to always create our internal components according to the standard patterns

Applications free to do this if they so wish.  Its a one-liner....


amcclain:
I was going to ask about this as had come across the function during recent docs-related work and was wondering how often we used. Agree better to keep the story as simple as we can.

PR #592, opened 2018-09-06
Make DropDownField internalOptions observable

### LINKED ISSUE #591: Investigate an observable internalOptions prop on hoist dropdown fields
When dynamically changing the option underneath one these fields we expect to see the options rerender in the field's menu. This seems to work as expected when using the @observable.ref decorator on set of options we pass in as props, and then reassigning that array to change it's value. However, if we use just @observable and then push to that array we are seeing the outer component that the field is nested in rerender, but not the internal field component itself.

While using the ref for our props is perfectly acceptable and somewhat more clear, there is a concern that we are relying on some implementation detail in mobx, and cannot rely on this always being the case. We should therefore look to somehow rerender these fields with certainty.

amcclain:
+ Ensures component implementations are re-rendered when props.options change and are "re-normalized" back into this internal collection.
+ Fixes #591

### DIFF
--- a/desktop/cmp/form/dropdown/ComboField.js
+++ b/desktop/cmp/form/dropdown/ComboField.js
@@ -39,13 +39,9 @@ export class ComboField extends BaseComboField {
 
     constructor(props) {
         super(props);
-        this.setInternalOptions(this.normalizeOptions(props.options));
+        this.addAutorun(() => this.normalizeOptions(this.props.options));
     }
-
-    componentDidMount() {
-        this.addAutorun(() => this.setInternalOptions(this.normalizeOptions(this.props.options)));
-    }
-
+    
     render() {
         const {style, width, disabled} = this.props,
             {renderValue, internalOptions} = this;
--- a/desktop/cmp/form/dropdown/ComboField.js
+++ b/desktop/cmp/form/dropdown/ComboField.js
@@ -39,11 +39,11 @@ export class ComboField extends BaseComboField {
 
     constructor(props) {
         super(props);
-        this.internalOptions = this.normalizeOptions(props.options);
+        this.setInternalOptions(this.normalizeOptions(props.options));
     }
 
     componentDidMount() {
-        this.addAutorun(() => this.internalOptions = this.normalizeOptions(this.props.options));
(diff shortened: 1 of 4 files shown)


desktop/cmp/form/dropdown/ComboField.js
```diff
@@ -39,11 +39,11 @@ export class ComboField extends BaseComboField {
 
     constructor(props) {
         super(props);
-        this.internalOptions = this.normalizeOptions(props.options);
+        this.setInternalOptions(this.normalizeOptions(props.options));
```
lbwexler: Realize this does not relate to this change, but I am not sure why we are not just running/setting up the auto-run in the constructor.  It seems to me the current setup will cause it to be run twice and adds extra complexity.
+

lbwexler:
Please let me know if my changes make sense -- meant to put these in a separate PR, but hopefully these are not objectionable

Mainly just prefer to simplify things by using observable.ref if that's all we are actually are using.  Also, agree that we might want to break things apart, but until we do that, we might as well handle this identical operation with the same mechanism in QueryComboField.

amcclain:
Thanks, I almost went further with my changes but then didn't want it to snowball (in terms of my personal afternoon and TODO list 😉) but I appreciate your doing the further cleanup.

Would like Don or Tom to review/test to ensure we're not missing anything, then would be happy for either one of those guys to merge. 

In particular I had wondered about why we were normalizing options in the constructor but only setting up the autoRun in willMount, and why the queryCombo had its own options field vs. re-using the internal options concept...  The cleanup certainly looks good and sensible to me, but I had been worried I was missing some subtlety.

Had also thought about having that method set the options - and maybe renaming to indicate it did both. I think I had my Ext hat on too much and was worried that we would have subclasses out there somehow that were treating that method as a protected API and overriding (i.e. normalizing options in some custom way) but I don't know why I thought that. Presumably we aren't extending these component classes anywhere outside of the framework.

amcclain:
@febbraiod when this is merged, can you please update #570 to match?

febbraiod:
yes will test/review/merge/update as soon as finish up with a current client request

PR #593, opened 2018-09-06
Fix: Column tooltip fn gets record data as second argument.

TomTirapani:
The ag-grid docs are misleading - the column tooltip function actually has a single `params` object, with the documented `value, valueFormatted, data, node , colDef, rowIndex and api` as properties (similar to renderer function). We need to deconstruct `data` from this object in order to pass it to tooltip fn.

### DIFF
--- a/columns/Column.js
+++ b/columns/Column.js
@@ -150,7 +150,7 @@ export class Column {
 
         if (this.tooltip) {
             ret.tooltip = isFunction(this.tooltip) ?
-                ({value}, valueFormatted, data) => this.tooltip(value, data, {colId: this.colId}) :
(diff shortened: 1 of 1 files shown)


nadkarniarjun:
Thanks for catching that, Tom!

@amcclain: will do. I'll add in an example for the Toolbox sample grid today.

nadkarniarjun:
FYI - I added the following tooltip for the city column in sampleGrid to show how the lower level data object can be referenced. 
```
tooltip: (value, data, meta) => `${data.company} is located in ${data.city}`
```
Let me know if you feel we need something more detailed.

PR #594, opened 2018-09-06
Add support for "unhideable" columns within grid column chooser, Fixes #524

### LINKED ISSUE #524: Add support for "unhideable" columns within grid column chooser
Column API should support option to mark columns as locked within grid col chooser, such that they appear in the interface but cannot be moved from one side to another. 

Would expect this to mean they are always locked to displayed/visible side in any imagined usage. Could implement as a `hideable` config, where the chooser would respect `false` with a lock.

febbraiod:


columns/Column.js
```diff
@@ -140,7 +144,7 @@ export class Column {
             field: this.field,
             colId: this.colId,
             headerName: this.headerName,
-            hide: this.hide,
+            hide: this.hide && this.hideable,
```
febbraiod: This is questionable, but I thought we show not only enforce this prop via the col chooser but protect also from programatic changes (possibly from state?). This way ag-grid never gets a hide property on an unhideable col.

I could see this behavior confusing an app developer if they intentionally try to hide an 'unhideable' col but it still pops up in the grid. On the other hand why would you try to do such a thing?

lbwexler: This one is a toss-up and I could see pros and cons either way

columns/Column.js
```diff
@@ -121,6 +124,7 @@ export class Column {
         this.chooserGroup = chooserGroup;
         this.chooserDescription = chooserDescription;
         this.excludeFromChooser = withDefaultFalse(excludeFromChooser, false);
```
lbwexler: realize that you did not add this, but line 126 looks like it has an extra argument

columns/Column.js
lbwexler: Thinking about this some more, I think its easiest to reason about if hide solely determines the visibility, and 'hideable' just governs the ability to hide it in the column chooser.  Also seems analogous to 'resizable' and 'moveable' in that it just governs user initiated GUI actions.

PR #595, opened 2018-09-06
Column Chooser and export manager handle grouped columns. Fix #588, Fix #589

### LINKED ISSUE #588: Column Chooser does not work with Grouped Columns

### LINKED ISSUE #589: Grid Export does not work with Grouped columns

febbraiod:
Open for review. Think this is good, but would like to test once more in updated group col demo in toolbox before merge.

### DIFF
--- a/desktop/cmp/grid/GridModel.js
+++ b/desktop/cmp/grid/GridModel.js
@@ -335,7 +335,7 @@ export class GridModel {
      * Grouped columns are a tree structure, this method returns a flat array representing the leaves
      */
 
-    getColumnArray() {
+    getLeafColumns() {
         const columns = this.cloneColumns();
         return this.gatherLeaves(columns);
     }
--- a/desktop/cmp/grid/GridModel.js
+++ b/desktop/cmp/grid/GridModel.js
@@ -331,6 +331,15 @@ export class GridModel {
         this.columns = newCols;
     }
 
+    /**
+     * Grouped columns are a tree structure, this method returns a flat array representing the leaves
+     */
+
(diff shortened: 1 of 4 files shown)


desktop/cmp/grid/GridModel.js
```diff
@@ -331,6 +331,15 @@ export class GridModel {
...
     }
 
+    /**
+     * Grouped columns are a tree structure, this method returns a flat array representing the leaves
+     */
+
+    getColumnArray() {
+        const columns = this.cloneColumns();
+        return this.gatherLeaves(columns);
+    }
```
febbraiod: This business used to live in grid state plug in, but as the colChooser and export plug in need this as well it makes sense to make it a public method on gridModels.

lbwexler: Can we rename this getLeafColumns().  Seems much clearer and we would not need the comment.  Also, you don't *always* need to clone the result of this (in one case above you are just walking through it).  So better to just clone the result of this function when you need to.

desktop/cmp/grid/GridStateModel.js
```diff
@@ -123,27 +123,17 @@ export class GridStateModel {
     }
 
     getColumnState() {
-        const {columns} = this.gridModel,
-            cols = this.gatherLeaves(columns);
+        const cols = this.gridModel.getColumnArray();
 
+        // Grouped columns are a tree structure but we store their state as a flat array of configs representing the leaves
```
lbwexler: put comment above line 126, ore remove completely

PR #596, opened 2018-09-06
Add XH.genId(), update LocalStore ID generation + validation

### LINKED ISSUE #590: Reliably generate unique store record IDs as required
LocalStore currently will auto-assign an ID to raw data objects if they don't contain an `id` property:

https://github.com/exhi/hoist-react/blob/0416d9413dac37fc6c49256497ee5b2b583d18c9/data/LocalStore.js#L116

This is currently based on the index of the object in the raw data array, and is problematic in a few ways:
+ It mutates the objects passed in to the store (we should be clear if we are OK doing this).
+ Use of raw number does not make it clear that these are locally generated IDs.
+ No check that an ID about to be assigned by idx does not overlap with another ID.  We had a particular use case where we ran a collection w/o IDs through `loadData`. They were mutated and idx-based IDs added starting at 0.  We then placed a new item in the collection at the start of it (code looked like `data = [newItem, ...oldItems]`) and ran through `loadData` again. The new item received an idx-based ID of zero, which was the same as the record right after it (which kept that ID as it had it already).
  + Note this was especially unfortunate as the store was powering a grid, and ag-grid will silently drop rows with dupe IDs.

Discussion items:
1) Are we OK mutating the objects passed to loadData()? Is that simply part of the store contract, or should we either avoid ever mutating or clone then mutate the copy?
2) Should we be auto-assigning IDs automatically at all, or should we instead throw if the objects to be loaded don't have an ID already?
3) Should we support a configurable `idProperty` on the store, so you could tell it to expect e.g. `Id` vs. `id` on the objects? This was a real use-case and has an equiv on the ExtJS side. Might make (2) more agreeable.
4) If we want to continue generating IDs as needed, how best to ensure uniqueness. Each store could have a monotonically increasing `idSeed` value it uses for the life of the store. Or we could have a global one via XH that would be used across all stores or other things that need IDs (again [see Ext example(https://docs.sencha.com/extjs/6.5.2/classic/src/Ext.js.html#Ext-method-id)).
5) And if we auto-generate, do we want to use a prefix such as `xh-id-` to make it clear these are generated locally?

amcclain:
+ Add XH.genId() to return locally-generated unique ID.
+ LocalStore uses XH.genId() if raw records do not contain an ID property.
+ LocalStore replaces null record IDs with generated IDs (not just undefined).
+ LocalStore warns if records contain non-unique IDs.
+ Minor re-org of some XH method blocks and comments.
+ Fixes #590

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -71,7 +71,7 @@ class XHClass {
     isDevelopmentMode = xhIsDevelopmentMode;
 
     //---------------------------
-    // Services
+    // Services + aliased methods
     //---------------------------
     configService = new ConfigService();
     environmentService = new EnvironmentService();
@@ -81,6 +81,16 @@ class XHClass {
     localStorageService = new LocalStorageService();
     prefService = new PrefService();
     trackService = new TrackService();
+
+    track(opts)                 {return this.trackService.track(opts)}
+    fetch(opts)                 {return this.fetchService.fetch(opts)}
+    fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
+    getUser()                   {return this.identityService.getUser()}
+    getUsername()               {return this.identityService.getUsername()}
+    getConf(key, defaultVal)    {return this.configService.get(key, defaultVal)}
+    getPref(key, defaultVal)    {return this.prefService.get(key, defaultVal)}
+    setPref(key, val)           {return this.prefService.set(key, val)}
+    getEnv(key)                 {return this.environmentService.get(key)}
 
     //-------------------------------
     // Models
@@ -127,10 +137,9 @@ class XHClass {
 
     /**
      * Transition the application state.
+     * Used by framework. Not intended for application use.
      *
      * @param {AppState} appState - state to transition to.
-     *
-     * Used by framework.  Not intended for application use.
      */
     @action
     setAppState(appState) {
@@ -187,11 +196,7 @@ class XHClass {
         return this.routerModel.currentState;
     }
 
-    /**
-     * Route the app.
-     *
-     * Shortcut to this.router.navigate.
-     */
+    /** Route the app - shortcut to this.router.navigate. */
     navigate(...args) {
         return this.router.navigate(...args);
     }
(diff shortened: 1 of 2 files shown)


amcclain:
(Apologies in advance for the extra noise shuffling around a few XH methods and comments while looking for a place for this new function.)

Alternate approaches would be to use [lodash's uniqueId method](https://lodash.com/docs/4.17.10#uniqueId) (which has a very similar counter-based implementation or a library like [uniqid](https://www.npmjs.com/package/uniqid) - either to power XH.genId() or within LocalStore (and no new XH method).

I obviously settled on this as it seemed like a generally useful XH util and I liked having IDs that were clearly very local and _not_ unique, from the server, hard to say, etc. But open to any suggestions as always of course.

I also debated warning vs. throwing when validating LocalStore ID uniqueness (or not adding the check at all). Since the bugs you get can be subtle (rows dropped quietly from grids) I thought it was worth the warning, but also seemed more conservative to warn vs. throw for this iteration.

lbwexler:
By the way -- we ran into a set of data that had non-unique ids (0) in sphere -- and it was tricky to figure out what was going on.  So great to add this warning!

core/XH.js
```diff
@@ -332,14 +323,26 @@ class XHClass {
...
+
+    /**
+     * Generate an ID string, unique within this run of the client application and suitable
+     * for local-to-client uses such as auto-generated store record identifiers.
+     *
+     * Deliberately *not* intended to be globally unique, suitable for reuse, or to appear as such.
+     * @returns {string}
+     */
+    genId() {
+        return `xh-id-${this._idSeed++}`;
```
TomTirapani: I think we could use lodash's `uniqueId('xh-id-')` here and lose the local _idSeed

lbwexler:
I think we might as well do this.  Just that much simpler...

amcclain:
Thanks. I had called that nethod out in the pull request. It's the same implementation (monotonically increasing counter) so I wasn't sure it was worth it to call into lodash. Mainly thinking that a developer who wanted to ensure they understood the nature of the id generation would then need to look at that source.

But - by way of counterargument - we are working to use lodash as a standard library wherever possible, so why not use it here, where it has a method that does exactly what we wanted. So I'm glad to have one more example of that pattern here!

PR #607, opened 2018-09-12
Allow passing additional user-defined data to DataView items via props

haynesjm42:
Need the ability to pass additional data through the data view to the items. In my specific use-case it is a shared model which the component which owns the DataView has access to, but is needed to be able to pass to a dialog model when the user clicks a button on the data view item.

### DIFF
--- a/desktop/cmp/dataview/DataView.js
+++ b/desktop/cmp/dataview/DataView.js
@@ -23,7 +23,7 @@ export class DataView extends Component {
 
     constructor(props) {
         super(props);
-        const {store, selModel, contextMenuFn, itemFactory, emptyText} = props.model;
+        const {store, selModel, contextMenuFn, itemFactory, itemData = {}, emptyText} = props.model;
         this._gridModel = new GridModel({
             store,
             selModel,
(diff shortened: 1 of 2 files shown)


lbwexler:
@amcclain, @haynesjm42:

Two things I would like to discuss in connection with this: 

1) I think it is a mistake to refer to the itemFactory prop as something that has to be an "elemFactory".   An elemFactory (as we all refer to it) is a very specific function for use in render methods, which we produce using an elemFactory() method in core.  While its true that the result of elemFactory() would be appropriate for this prop, more generally, this prop is just any function that is passed an object with a record and a store, and produces a React Element.

2) Having said that-- John couldn't you achieve the result you are trying to achieve by passing in a function that simply applies those default args that you want to apply internally?

Think both of these comments apply both to this property, and the "elementRenderer" property of GridColumn itself.  Perhaps we should standardize on a single naming convention for these properties.  Maybe best to stay away from the word 'factory'?

lbwexler:
Just to add a bit more color -- really trying to make sure that elemFactory does not appear anywhere explicitly in our Component APIs because we have users that want to use JSX and should not even need to know about elemFactory()...

haynesjm42:
It didn't even occur to me to not just provide an elemFactory function to `itemFactory` ... yes that approach solves my problem and makes this change unnecessary.

lbwexler:
well the name and the docs were throwing you off!

lbwexler:
Closing, per our discussion.  This can be done by building the parameters in to a new itemFactory function

Issue #575, opened 2018-08-31
Column elementRenderer and cellRenderer should have consistent params


haynesjm42:
`Column.elementRenderer` is currently passed the ag-grid `ICellRendererParams` as props, which is inconsistent with `Column.cellRenderer`. It would be good for these to be consistent, and for `Column.elementRenderer` to be more ag-grid-agnostic. These should also be consistent with what is passed to the `DataView` item components as props.

I propose we pass an object like this to each: `{ value, record, column, agParams }`

That should give the renderers/components everything they need, and additional data can be passed as needed in the params (no longer need `metaData` for `cellRenderer`).

Issue #549, opened 2018-08-20
Grid - action column


amcclain:
A common UI pattern is to have one or more grid columns that support row-level actions via minimalist inline buttons. Would like to review options for making such a column easy to create in a standardized manner.

Ideally a developer could provide a column config with a declarative set of action configurations and have it render and manage the relevant UI affordances.

Items to consider / discuss:
* What defines an action - handler, icon, tooltip, text (?), options to disable/hide.
* How such a feature would interact with the top-level Column API.
* Support for showing actions only on row-level hover, so they are not all printed all the time.

🤔 If we wanted to push the idea further, we could consider an alternate approach where the row-level actions are _not_ implemented as a grid column, but instead appear on hover as an overlay - even more like what Outlook/Gmail do...  That would presumably require more work to implement, but would a) be more space-efficient and b) avoid potentially confusing the current data-driven Column API with a different concept.

amcclain:
@TomTirapani was looking with @haynesjm42 at a not-yet-completely-generalized example he has of some of this. We talked about having him draw up a proposal for a Hoist-level API / feature here, but wanted to loop you in as we've been discussing and would put this to pretty immediate use.

lbwexler:
Hi Guys -- I wonder if we could see some psuedo-code here?  That's the easiest way to evaluate something like this.

haynesjm42:
Of what it would look like when adding an action column to a grid? An example can be seen here: https://github.com/exhi/toolbox/commit/035f035a044385770d2e1b9467fbe7c3b7bc94d3

```
{
    ...actionsCol,
    rendererParams: {
        actions: [
            {
                icon: Icon.delete(),
                action: (params) => console.log('Delete Clicked!', params),
                data: 'Here is some user data!',
                intent: 'danger'
            }
        ]
    }
}
```

lbwexler:
Hi:  Can we discuss?  I was thinking of something more like the following from the user perspective:

Still not sure about whether we want to use React Components for our custom column rendering or not, and it seems like a bunch of things are up in the air with our column rendering API, but I was thinking something more like the following from the application's perspective.  

const myActionCol = {
    colId: 'myActions',
    elementRenderer: (params) => actionBar({
        items: [
            deleteAction,
            cancelAction,
            {
                icon: Icon.skull(),
                action: (params) => console.log('Anonymous Skull action', params),
                intent: 'danger'
            }
        ]
    })
}

For some reason, not a big fan of breaking up the renderer and its params.

[...]

PR #608, opened 2018-09-12
Grid actions column - WIP

haynesjm42:
Initial pass at support actions grid columns. #549 

Extended the Column API to accept `rendererParams` to be passed through to the `renderer` or `elementRenderer` (the mechanism for passing the action definitions through to the ActionBar component)

ActionBar component which accepts a collection of action definitions, and simply renders them in a row as minimal buttons. Action definition properties can be a value or a function which returns a value (where it makes sense).

Also made some changes for #575 as part of this

@lbwexler did a quick review with @amcclain of this approach, wanted your feedback on whether this seems like a sounds approach or if something more object-oriented and explicit (ActionColumn, pass actions to the GridModel, etc.) would be preferable.

See https://github.com/exhi/toolbox/commit/035f035a044385770d2e1b9467fbe7c3b7bc94d3 for an action column example.

### DIFF
--- a/columns/Column.js
+++ b/columns/Column.js
@@ -39,11 +39,14 @@ export class Column {
      * @param {boolean} [c.resizable] - false to prevent user from drag-and-drop resizing.
      * @param {boolean} [c.movable] - false to prevent user from drag-and-drop re-ordering.
      * @param {Column~rendererFn} [c.renderer] - function to produce a formatted string for each cell.
-     *      Supports HTML as output. Passed both field value and entire row data object.
+     *      Supports HTML as output. Passed the field value as the first parameter, and an object
+     *      containing the record, column, and renderer params (destructured) as the second parameter.
      * @param {function} [c.elementRenderer] - elementFactory function to return a React component
-     *      for rendering within each cell. For ag-Grid implementations, an ICellRendererParams
-     *      object is passed to the rendered component as props.
-     *      @see ICellRendererParams
+     *      for rendering within each cell. Is passed the value, record, column, and renderer params
+     *      (destructured) in it's props. For ag-grid implementations, the ICellRendererParams are
+     *      also passed as agParams. @see ICellRendererParams
+     * @param {object} [c.rendererParams] - additional parameters to pass to the renderer function
+     *      or to be included in elementRenderer props.
      * @param {string} [c.chooserName] - name to display within the column chooser component.
      *      Defaults to headerName, but useful when a longer / un-abbreviated string is available.
      * @param {string} [c.chooserGroup] - group name to display within the column chooser component.
@@ -83,6 +86,7 @@ export class Column {
         movable,
         renderer,
         elementRenderer,
+        rendererParams,
         chooserName,
         chooserGroup,
         chooserDescription,
@@ -119,6 +123,7 @@ export class Column {
 
         this.renderer = renderer;
         this.elementRenderer = elementRenderer;
+        this.rendererParams = rendererParams;
 
         this.chooserName = chooserName || this.headerName || this.colId;
         this.chooserGroup = chooserGroup;
@@ -175,15 +180,28 @@ export class Column {
             ret.width = this.width;
         }
 
-        const {renderer, elementRenderer} = this;
+        const {renderer, elementRenderer, rendererParams} = this;
         if (renderer) {
             ret.cellRenderer = (params) => {
-                const metaData = {colId: params.column.colId};
-                return renderer(params.value, params.data, metaData);
+                return renderer(params.value, {record: params.data, column: this, ...rendererParams});
             };
         } else if (elementRenderer) {
+            const props = {
+                column: this,
+                ...rendererParams
+            };
+
             ret.cellRendererFramework = class extends Component {
-                render() {return elementRenderer(this.props)}
+                render() {
+                    const rendererProps = {
+                        value: this.props.value,
+                        record: this.props.data,
+                        agParams: this.props,
+                        ...props
+                    };
+
+                    return elementRenderer(rendererProps);
+                }
                 refresh() {return false}
             };
         }
--- a/desktop/cmp/actionbar/Action.js
+++ b/desktop/cmp/actionbar/Action.js
@@ -1,0 +1,41 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+export class Action {
+    /**
+     *
+     * @param {Object|Function} icon - icon to be displayed, or a function returning an icon to be
+     *      displayed.
+     * @param {string|Function} text - text to be displayed next to the icon, or a function returning
+     *      the text to be displayed.
+     * @param {ActionFn} action - called when the action button is clicked
+     * @param {*} data - data to be passed through to action when the button is clicked
+     * @param {string|Function} tooltip - tooltip to be displayed, or a function returning the tooltip
+     * @param {boolean|Function} disabled - disabled state for the action button, or a function
+     *      returning the disabled state
+     */
+    constructor({
+        icon,
+        text,
+        action,
+        data,
+        tooltip,
+        disabled
+    }) {
+        this.icon = icon;
+        this.text = text;
+        this.action = action;
+        this.data = data;
+        this.tooltip = tooltip;
+        this.disabled = disabled;
+    }
+}
+
+/**
+ * @callback ActionFn - called when the action button is clicked
+ * @param {*} data -
+ */
--- a/desktop/cmp/actionbar/ActionBar.js
+++ b/desktop/cmp/actionbar/ActionBar.js
@@ -1,0 +1,77 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+import {Component} from 'react';
+import {PropTypes as PT} from 'prop-types';
+import {elemFactory, HoistComponent} from '@xh/hoist/core';
+import {hbox} from '@xh/hoist/cmp/layout';
+import {button} from '@xh/hoist/desktop/cmp/button';
+import {isFunction} from 'lodash';
+
+@HoistComponent
+export class ActionBar extends Component {
+    baseClassName = 'xh-action-bar';
+
+    static propTypes = {
+        /**
+         * Actions to render in the bar. Action definitions support the following structure:
+         * {
+         *      icon,
+         *      text,
+         *      intent,
+         *      className,
+         *      action,
+         *      data,
+         *      tooltip,
+         *      disabled
+         * }
+         */
+        actions: PT.arrayOf(PT.object).isRequired
+    };
+
+    render() {
+        const {actions} = this.props;
+        return hbox({
+            className: this.getClassName(),
+            items: actions.map(action => this.renderAction(action))
+        });
+    }
+
+    renderAction(actionDef) {
+        let {icon, text, intent, className, action, data, tooltip, disabled} = actionDef;
+        icon = this.getValue(icon);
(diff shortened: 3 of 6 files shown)


haynesjm42:
@amcclain / @lbwexler per discussion with Lee, took a stab at making ActionBar into StoreActionBar and functioning similar to StoreContextMenu, and sharing item configs. Also went ahead and made RestGridModel configurable to show the Edit and Delete actions either in the toolbar or in a column.

See https://github.com/exhi/toolbox/commit/c7acfae2daecf6191e485c7b863d6d0bfe1d1085 for updated toolbox example

desktop/cmp/contextmenu/ContextMenuItem.js
```diff
@@ -11,13 +11,13 @@ import {assign} from 'lodash';
...
- *  @see StoreContextMenuItem for a more specific (and common) implementation tied to data views.
+ *  @see StoreActionDefinition for a more specific (and common) implementation tied to data views.
  */
 export class ContextMenuItem {
 
     text;
     icon;
-    action;
+    actionFn;
     items;
```
lbwexler: John and I discusssed this name change, due to the fact that the overall object is likely to be called "action".  Will need to determine if this is worth the risk and makes sense.

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -6,7 +6,7 @@
  */
 
 import {isString} from 'lodash';
-import {StoreContextMenuItem} from './StoreContextMenuItem';
+import {StoreActionDefinition} from '@xh/hoist/desktop/cmp/store/StoreActionDefinition';
```
lbwexler: Word 'Definition' seems a bit clunky.  Why not 'StoreAction'?

lbwexler: Or better yet "RecordAction" and "RecordActionBar" etc...

cmp/record/RecordAction.js
```diff
@@ -5,38 +5,44 @@
...
     disabled;
     hidden;
     prepareFn;
     recordsRequired;
 
     /**
-     * @param {Object} c - StoreContextMenuItem configuration.
-     * @param {string} c.text - label to be displayed.
+     * @param {Object} c - RecordAction configuration.
+     * @param {string|null} [c.text] - label to be displayed.
```
lbwexler: The 'string|null' seems clunky to me.  If we really want to express nullability, we could use {?string} or {!string}.  Not sure what we would consider the "default" but we would probably want to decide about that as well (probably non-nullable is the implicit default).

see http://usejsdoc.org/tags-type.html

columns/Column.js
```diff
@@ -41,11 +41,14 @@ export class Column {
      * @param {boolean} [c.resizable] - false to prevent user from drag-and-drop resizing.
      * @param {boolean} [c.movable] - false to prevent user from drag-and-drop re-ordering.
      * @param {Column~rendererFn} [c.renderer] - function to produce a formatted string for each cell.
-     *      Supports HTML as output. Passed both field value and entire row data object.
+     *      Supports HTML as output. Passed the field value as the first parameter, and an object
```
lbwexler: I think you should remove the detailed type info from this text, and rely on the jsdoc definition of Column~rendererFn instead (which is now out of date).

columns/Column.js
```diff
@@ -70,6 +73,8 @@ export class Column {
      *      desktop implementations. Note these options may be used / overwritten by the framework
      *      itself, and are not all guaranteed to be compatible with its usages of Ag-Grid.
      *      @see {@link https://www.ag-grid.com/javascript-grid-column-properties/|AG-Grid docs}
+     * @param {...*} [rest] - additional properties to store on the column, made available to column
+     *      renderers and other callbacks
```
lbwexler: Last clause of this comment confusing, would leave off

columns/Column.js
```diff
@@ -41,11 +41,14 @@ export class Column {
      * @param {boolean} [c.resizable] - false to prevent user from drag-and-drop resizing.
      * @param {boolean} [c.movable] - false to prevent user from drag-and-drop re-ordering.
      * @param {Column~rendererFn} [c.renderer] - function to produce a formatted string for each cell.
-     *      Supports HTML as output. Passed both field value and entire row data object.
+     *      Supports HTML as output. Passed the field value as the first parameter, and an object
+     *      containing the record, column, and renderer params (destructured) as the second parameter.
      * @param {function} [c.elementRenderer] - elementFactory function to return a React component
```
lbwexler: can we remove the word 'elementFactory' here.  It seems like a loaded term that makes me think of `elemFactory`.
Can just be a "function returning a react component"

columns/Column.js
```diff
@@ -192,13 +200,19 @@ export class Column {
...
+            const column = this;
             ret.cellRendererFramework = class extends Component {
-                render() {return elementRenderer(this.props)}
+                render() {
+                    const agParams = this.props,
+                        {value, data: record} = agParams;
+
+                    return elementRenderer({value, record, agParams, column});
+                }
                 refresh() {return false}
```
lbwexler: If we are going to mess with these, elementRenderer and renderer callbacks really should end up with the same exact signature.  I think I would prefer renderer(val, Object) but would also be ok with renderer(Object).  In any case, we need to choose one and use the JSDoc callback notation to clearly document it.

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -40,8 +40,8 @@ export class StoreContextMenu {
         this.gridModel = gridModel;
         this.items = items.map(it => {
             if (isString(it)) return this.parseToken(it);
```
lbwexler: Need to update JSDoc above here for type of items

desktop/cmp/record/RecordActionBar.js
```diff
@@ -0,0 +1,65 @@
...
+        /** RecordAction or configs to create. */
+        actions: PT.arrayOf(PT.oneOfType([PT.object, PT.instanceOf(RecordAction)])).isRequired
+    };
+
+    actions = [];
+
+    constructor(props) {
+        super(props);
+
+        this.actions = props.actions.filter(it => it !== null).map(it => new RecordAction(it));
```
lbwexler: This always creates a RecordAction, but prop type claims that application can pass in one 'already built'.

desktop/cmp/record/RecordActionBar.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    render() {
+        const {actionsShowOnHover = true} = this.props,
```
lbwexler: I might just call this "showOnHover".  Also, add it to the prop types.

desktop/cmp/rest/RestGridModel.js
```diff
@@ -118,26 +165,20 @@ export class RestGridModel {
...
                     icon: Icon.add(),
-                    action: () => this.addRecord()
+                    actionFn: () => this.addRecord()
                 },
                 {
-                    text: 'Edit',
-                    icon: Icon.edit(),
-                    action: (item, record) => this.editRecord(record),
-                    recordsRequired: 1
+                    ...this.editAction
```
lbwexler: Do these need to be cloned here?   also think Add should be defined like the other two for symmetry.

desktop/cmp/rest/RestGridModel.js
```diff
@@ -57,18 +80,42 @@ export class RestGridModel {
...
         unit = 'record',
         filterFields,
         enhanceToolbar,
         editors = [],
+        columns,
         ...rest
     }) {
         this.actionEnabled = Object.assign(this.actionEnabled, actionEnabled);
         this.actionWarning = Object.assign(this.actionWarning, actionWarning);
+        this.actionColEnabled = actionColEnabled;
```
lbwexler: If we exported the actions separately, would it be trivial matter for the application to simply specify a column of the form `{...actionsCol, actions: [recordEditAction, recordDeleteAction]}`.

Advantage here is that the application has flexibility to place and size the column, but also mix in its own actions in to the same column.

RestGridModel could even, say,  make its actions toolbar a public property, which could be enhanced along the same lines.

This generally suggests a great pattern for any API -- provide Action configs as public exports which can then be fed back to Grids as applications see fit.  Very similar to our column API.
Could be much more extensible than the 'token' based approach we were using.

cmp/record/RecordAction.js
haynesjm42: Didn't know about the ? and ! qualifiers, that's handy. I think I added the |null because I made the text optional, and then added the default of null in the ctor to be consistent with the other props and intellij complained about it not being null-able.  Looking at it again I don't think it's necessary to default to null, just let it default to undefined and be optional

columns/Column.js
haynesjm42: The only problem with renderer(val, Object) is that you then can't just provide an elemFactory for the elementRenderer. Maybe that's fine and we want to make it clear that the elementRenderer callback and elemFactory are not the same thing? 

I don't think we want to change the cellRenderer signature to take an object, @amcclain made a good point in #575 about that

lbwexler: OK -- totally fine with going that way.  I also think its totally fine that you can't provide a raw elemFactory to the renderer -- as you saw from my other comment want to move away from any possible conflation of the two concepts....

lbwexler:
Looking

desktop/cmp/grid/Grid.js
```diff
@@ -227,12 +224,14 @@ export class Grid extends Component {
...
         }
         if (!rec) selModel.clear();
-        const {count} = selModel;
+
+        const count = selModel.count,
+            selectedRecs = selModel.records;
 
         // Prepare each item
         const items = menu.items;
         items.forEach(it => {
```
lbwexler: Thinks this change makes sense -- seems more flexible to not tie the action to a selectionModel specifically.
I still wonder whether we would want to make prepareFn and actionFn take a config object to make them more flexible for gaining additional info in the future (e.g. column, cell, gridModel).  I am ok with waiting on that change if we feel we want to move forward with this now.

desktop/cmp/rest/RestGridModel.js
lbwexler: If we don't want to do my suggested change now (which I totally understand).  I do think we should remove this property 'actionColEnabled' from RestGrid.  I don't think its the right API, and we have no immediate need for it.   We can take some more time to think about the best pattern for exposing actions in this and other APIs.

desktop/cmp/grid/ColChooserModel.js
```diff
@@ -52,9 +52,9 @@ export class ColChooserModel {
```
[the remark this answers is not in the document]
haynesjm42: Sorry should have called this out and given some context. When testing the action col, I found that it wasn't working with the column chooser and found it was because the action col does not specify a field (nor should it), which led to this change. I would consider this a bug, as the field is an optional property on the column and is not guaranteed to be unique, we shouldn't be using it as a unique identifier for columns.

desktop/cmp/grid/Grid.js
haynesjm42: I have the change for passing a config object to prepareFn and actionFn on another branch, along with other changes to refine the RestGrid "actions" api as we had discussed. Agree that those changes don't need to hold up the action col being merged.

lbwexler: Perfect! -- if we can just unwind the intermediary RestGrid API changes in this branch, I would be fine to merge

desktop/cmp/grid/ColChooserModel.js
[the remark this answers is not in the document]
lbwexler: great! Nice fix

haynesjm42:
Implemented the changes we discussed and made a few tweaks/fixes.

+ Added support for specifying the header and cell css classes on Columns
+ Added a helper function for calculating the action column width, given the number of actions. This references an exported css var. Seems to work well, but have never done anything like this before so not sure if this is something we want to do or if I am following the desired pattern here
+ Changed the show on hover css logic to use `visibility` instead of `display` because I noticed that the icon centering was off when show on hover was turned on, and it was because I was overriding the `display: inline-flex` from the bp button css with `display: unset`

Issue #602, opened 2018-09-11
HoistInput addition: Datetime


amcclain:

amcclain:
Looking again, it appears as though the blueprint DateInput control has (new?) support for time precision, so we should be able to use it here. Most likely path will be to rename our DayField > DateField and make the date format more configurable, although still defaulting to day-level format/parse.

amcclain:
Bumping this up as it's needed in the short-term for a client project.  

⚠️ But please note we have some renames of all the input components going on as part of #579 which @TomTirapani has started.

febbraiod:
As mentioned in the closed PR: Uncovered a bug where the first time (n times until you take some as of yet unknown action?) you interact with the popover the new value doesn't commit. Lee says he's seen this with other controls as well.

I can confirm this is an existing problem as the existing month/year selectors do not commit their changes in the same manner.

febbraiod:
It seems that the onChange handler on the bp's date input fires when the time changes, but doesn't always (but sometimes!!??!!) pass the new time to the handler. However you can put an onChange handler on the timePicker which does seem to register the new time. We could probably find a way to switch between the handlers

PR #609, opened 2018-09-12
FormField component

TomTirapani:


lbwexler:
Thanks for taking this on including the bombing!

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,79 @@
...
+
+        return formGroup({
+            label,
+            item,
+            helperText: hbox({
+                height: 15,
+                items: [
+                    span({
+                        style: {color: 'red'},
+                        item: notValid ? fieldModel.errors[0] : ''
```
lbwexler: Not sure if this was intended for review yet, but some random thoughts:

+ Should/Can we provide a target for a tooltip that would potentially show *all* the errors?

+ Do we need to elide the error?

+ Use little spinner instead of text?

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    //--------------------
+    // Implementation
+    //--------------------
+    prepareChild() {
+        const {model, field, disabled} = this.props,
+            item = this.props.children;
+
+        throwIf(isArray(item) || !(item.type.prototype instanceof HoistInput), 'FormField child must be a single component that extends HoistInput.');
```
lbwexler: Let's use  markClass('isHoistInput') on the class to make this a bit easier/standard.

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,79 @@
...
+import {elemFactory, HoistComponent} from '@xh/hoist/core';
+import {PropTypes as PT} from 'prop-types';
+import {HoistInput} from '@xh/hoist/cmp/form';
+import {formGroup} from '@xh/hoist/kit/blueprint';
+import {div, span, hbox, hspacer} from '@xh/hoist/cmp/layout';
+import {throwIf} from '@xh/hoist/utils/js';
+import {isArray} from 'lodash';
+
+/**
+ * Wrapper around Blueprint's FormGroup component, which integrates
```
lbwexler: I think comment should emphasize the relationship to @Field, and @FieldSupport, and the fact that it shows Label/Validation/Required state in a standard way.

I think the fact that it is a FormGroup is, at best, a secondary characteristic (if we publicize it at all)

TomTirapani:
Thanks for your review and great suggestions Lee.

I've made the changes you suggested - eliding long error messages, spinner while pending and tooltip showing all errors.

Regarding the renaming, we hadn't discussed but its a great idea - I've renamed those components to lose the 'Input' suffix. Unfortunately, 'SwitchInput' remains as `switch` is a reserved word in JS (and couldn't think of a better alternative. Maybe `SwitchControl`, but not sure that's any better. Probably best to stick with Input for consistency)

Regarding the use of `markClass()`... I may be missing something, but isn't that only possible when using decorators? Unless you are suggesting we turn `HoistInput` into a decorator... :)

cmp/form/HoistInput.js
```diff
@@ -12,37 +12,37 @@ import {throwIf} from '@xh/hoist/utils/js';
...
  * Note that operating in bound mode may allow for more efficient rendering in a MobX context,
  * in that the bound value is only read *within* this control, so that changes to its value do not
  * cause the parent of this control to re-render.
  *
- * HoistFields support built-in validation when bound to a model enhanced by `@FieldSupport`.
- * When a HoistField control is linked to a property on the underlying model decorated by `@field`,
+ * HoistInputs support built-in validation when bound to a model enhanced by `@FieldSupport`.
+ * When a HoistInput control is linked to a property on the underlying model decorated by `@field`,
  * the model Field will be used to provide validation info and styling to the input component.
  */
```
lbwexler: Should probably rework this doc bit here to refer to the FormField class which is now providing most (all) of the validation support.

lbwexler:
Great Tom -- yes, thanks for looking at markClass.  Please ignore that comment (and yes, maybe someday HoistInput will become a decorator :-))

Got it re: switch.  Think SwitchInput sounds good.

Regarding the other names -- why not just 'select' and  'textArea' and 'multiSelect'?   Just think these are pretty standard names.  The first two mirror the names of the similar HTML elements

amcclain:
Should we take `dayInput` to `dateInput` at this time, to prep for #602? 

Unless we think there will be value in supporting two top-level components, I think we will most likely update the internals of current `DayField` within that ticket to support optionally showing / accepting time values.  Might be convenient to do the full rename here while we're already tweaking it - internal updates can be left to that ticket which we are hoping @febbraiod can look at when this has settled down.

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,89 @@
...
+import {isArray} from 'lodash';
+
+import './FormField.scss';
+
+/**
+ * Standardised wrapper around HoistInputs.
+ *
+ * Should receive a single HoistInput as a child element. If the HoistInput is bound
+ * to a model enhanced by `@FieldSupport`, FormField will automatically display a label,
+ * a required asterisk and validation state.
```
amcclain: I think a key thing to express in the doc here is that, when wrapping with a FormField, the inputs themselves don't specify their usual model/field. That's delegated to / controlled by the wrapper.  We have the nice warnings in place, so you will find that out pretty fast, but still seems like a key thing to know from reading the docs.

This also refers to the HoistInput being bound, which it will be, but probably clearer to say "If this component is bound"...

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+@HoistComponent
+export class FormField extends Component {
+
+    static propTypes = {
+        /** model to bind to */
+        model: PT.object,
+        /** name of property in model to bind to */
+        field: PT.string,
+        /** optional label for form field */
```
amcclain: We can lose "optional" here - I know it's scattered in a few places in the docs. They are all effectively optional. 

Sorry to nitpick - I need an emoji to preface annoying comments like this. Suggestions....maybe welcome

desktop/cmp/form/FormField.scss
```diff
@@ -0,0 +1,38 @@
...
+      height: 15px;
+    }
+
+    .xh-form-field-error-msg {
+      display: none;
+    }
+  }
+
+  &.xh-form-field-invalid .bp3-form-content {
+    padding: 0 0 calc(var(--xh-font-size-small-px) + var(--xh-pad-half-px));
```
amcclain: Cool!  I hope this is working / providing some value. It's exactly the kind of thing I hoped we could do with these vars.

(Is is OK to add the px values in these calc statements? I'm not well versed on when you need to use the underlying unit-less values.)

desktop/cmp/form/FormField.scss
```diff
@@ -0,0 +1,38 @@
+.xh-form-field {
+  .bp3-form-content {
+    position: relative;
```
amcclain: Why do we need this, as well as the `position: absolute` below? 

Would presumably be obvious if I was running and removed it, but I am always on the lookout for places where we are customizing the BP classes in non-obvious ways. (Non-obvious if you are not super great at CSS. Colors, font-sizes I think we all grok pretty well and seem less likely to break as BP updates their implementations.)

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,89 @@
...
+        const {model, field, ...rest} = this.props,
+            item = this.prepareChild(),
+            hasFieldSupport = model && field && model.hasFieldSupport,
+            fieldModel = hasFieldSupport ? model.getField(field) : null,
+            isRequired = fieldModel && fieldModel.isRequired,
+            isPending = fieldModel && fieldModel.isValidationPending,
+            notValid = fieldModel && fieldModel.isNotValid,
+            errors = fieldModel ? fieldModel.errors : [],
+            labelStr = this.props.label || (fieldModel ? fieldModel.displayName : null),
+            label = isRequired ? div(labelStr, span(' *')) : div(labelStr);
```
amcclain: Any sense we might want to customize or suppress this as the "required" hint?

I think the asterisk is pretty common, although it sometimes looks a bit meh. 

Just raising the question in case we have any ideas - *totally fine* with starting here and seeing if we are ever unhappy with it in practice.

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,89 @@
...
+
+    render() {
+        const {model, field, ...rest} = this.props,
+            item = this.prepareChild(),
+            hasFieldSupport = model && field && model.hasFieldSupport,
+            fieldModel = hasFieldSupport ? model.getField(field) : null,
+            isRequired = fieldModel && fieldModel.isRequired,
+            isPending = fieldModel && fieldModel.isValidationPending,
+            notValid = fieldModel && fieldModel.isNotValid,
+            errors = fieldModel ? fieldModel.errors : [],
```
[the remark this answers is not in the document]
lbwexler: Its available to us now in BabelScript :-)  (not saying we take it)

desktop/cmp/form/FormField.js
lbwexler: I wonder if we need to be so strict about the child?  Seems to me, that as long as it is a react element that can be cloned, we should accept it.  You could imagine an application wanting to decorate the actual control with additional layout adornments?  Or perhaps even inserting a non-HoistInput widget that is nevertheless successfully binding to the field

TomTirapani:
Thanks for the further feedback guys.

I've renamed `DayInput > DateInput`. Regarding `SelectInput > Select` and `TextAreaInput > TextArea`... my thinking there was that it is potentially problematic _because_ they are the same as the html tags. If used with JSX, it could lead to confusion between `<textarea>` and `<TextArea>`, and when used with elemFactory we don't even have the casing to differentiate (e.g. we have a textarea elemFactory in Tags.js).

desktop/cmp/form/FormField.scss
TomTirapani: Yes, perfectly fine (and in many cases necessary) to add the unit with calc. What's really cool is when you mix different units (e.g. `calc(100% - 30px)`)

desktop/cmp/form/FormField.js
```diff
@@ -0,0 +1,92 @@
...
+                div({
+                    omit: !isPending,
+                    className: 'xh-form-field-pending',
+                    item: spinner({size: 15})
+                }),
+                div({
+                    omit: !notValid,
+                    className: 'xh-form-field-error-msg',
+                    item: notValid ? errors[0] : null,
+                    title: notValid ? errors.join('. ') : null
```
lbwexler: This puts a pretty strong stake in the ground that errors cannot end in periods.  Could we divide these with pipes, or commas?

lbwexler:
Hi - I think we can (and will) continue to tweak the display of FormField here as we start to see it play out in client applications (and before and after this Hoist release).    But given that this includes the HoistField renaming, I would argue that we should merge it sooner rather than later, to avoid merging pain...  I am ok to merge with it in the current state.

TomTirapani:
I'm happy for you to merge now, and continue to tweak with the context of client app usage.

PR #616, opened 2018-09-13
Active User Switch Field on Admin User Panel

shravnk:
- Added switchFilter to UserPanel
- Added displayInactive observable boolean to UserModel
- Set filter on gridModel's store prior to calling loadAsync


lbwexler:
Hi Brendan -- 

This looks like a good start.  We might alternatively consider a server-side filter here -- perhaps we can go over what that might look like?  We would have to write a dedicated controller for that, but for something like this, where you potentially have a lot of data that you may not be interested in at all on the client, its a better pattern.

In any case, I do think you probably have a problem here with the use of the filter to pull out your data -on the client- won't that collide with the use of the quick filter  :-)  ?

In this case, best to preserve the grid store's filter  for use by the GUI --- you could implement your programmatic active filter by filtering the loaded data "manually" by just running through the raw data from the server and pulling out the inactive ones before feeding it to the store.

We can discuss -- its a pretty common pattern.

amcclain:
Thanks guys!

PR #617, opened 2018-09-13
MultiSelect Placeholder Bug

shravnk:
Still not sure why the 'Search...' placeholder displays after selecting an option. This is a workaround which explicitly sets inputProps.placeholder to an empty string. Per the documentation below I don't think we should have to do this.

screen shot 2018-09-13 at 5 17 01 pm


lbwexler:
Nice fix -- I merged it manually into the new file location directly on develop.

PR #620, opened 2018-09-14
Add Panel.contentFrameProps - options to customize inner vFrame

amcclain:
+ Name here is provisional - I am almost positive there is a better one.
+ See matching branch in Toolbox for example usage.
+ Alternate as-is approach requires nesting panel contents in another dedicated element to customize inner layout, which is not terrible but seems non-ideal.


amcclain:
Could also reconsider using a vFrame for that inner-most container...  Not positive that's the least surprising choice....

lbwexler:
On that point, I was sure that it was a Frame!  I was reviewing code with
Anand today and explicitly asked him why he was giving multiple children to
a panel, and told him to add an interstitial VBox!!!

(to your point about least surprising)

PR #628, opened 2018-09-17
Add time support to date inputs, Fixes #602

### LINKED ISSUE #602: HoistInput addition: Datetime

febbraiod:


desktop/cmp/form/DateInput.js
```diff
@@ -94,7 +115,7 @@ export class DateInput extends HoistInput {
     }
 
     parseDate(dateString) {
-        return moment(dateString, 'YYYY-MM-DD', true).toDate();
+        return moment(dateString, 'YYYY-MM-DD HH:mm:ss').toDate();
```
febbraiod: Briefly discussed with Lee about making this format string a prop. Such that the app developer could enable timePicking and also specify how the user might be expected to manually enter values. The problem was that in the implementation we could set the formatter string to this time specific string but the datePicker popover would still pass a dateString of the form 'YYYY-MM-DD' and because we were passing the 'strict' param to moment it would throw.
Thinking maybe we still want to add a formatter string prop (to for example enable millis) but I also think removing the stict param makes sense. Not sure why we wouldn't want to except any user input that might be acceptable to moment.

febbraiod: Actually should definitely still be a prop.

febbraiod:
Uncovered a bug where the first time (n times until you take some as of yet unknown action?) you interact with the popover the new value doesn't commit. lee says he's seen this with other controls as well. Closing PR to get to the bottom of this.

febbraiod:
'Bug' was caused by a bad test. We have a maxDate on our dateInput in toolbox. I was trying to set the time to the future. The handler was defaulting the date back to the maxDate. If we set maxDates we should probably set the time to 23:59:59, and as lee suggested, we should also probably warn when resetting an invalid date. This control is still probably worth a review as is.

PR #641, opened 2018-09-19
Tree store

amcclain:


amcclain:
This is pending review and confirmation from @lbwexler where it is at, but opening up this P/R for some more visibility on it so we all know it's in flight.  

(@cnrudd was investigating tree support today and pinged me on it, so I alerted him to the existence of this branch)

data/Record.js
```diff
@@ -0,0 +1,97 @@
...
+     * Will apply basic validation and conversion (e.g. 'date' will convert from UTC time to
+     * a JS Date object). An exception will be thrown if the validation or conversion fails.
+     */
+    constructor({raw, parent, fields, children = []}) {
+        this.id = raw.id;
+        this.parent = parent;
+        this.children = children;
+        this.fields = fields;
+
+        this.xhTreePath = parent ? [...parent.xhTreePath, this.id] : [this.id];
```
cnrudd: Changing line 38 to:
```
this.xhTreePath = this.xhTreePath ? this.xhTreePath : 
parent ? [...parent.xhTreePath, this.id] : [this.id];
```
would allow the treePath value to be already set in the event that some data did not need to have it built by this record class.

This would allow a source that already had its path values flattened 'ag grid style' (ie: ['grandparent', 'parent', 'child'] to just work, without having to reconvert the datasource into a parent/child hierarchy, only to have it converted back, here.

columns/Column.js
```diff
@@ -152,6 +156,18 @@ export class Column {
             ...this.agOptions
         };
 
+        if (this.isTreeColumn) {
+            ret.showRowGroup = true;
+            ret.cellRenderer = 'agGroupCellRenderer';
+            ret.cellRendererParams = {
```
cnrudd: Exposing the cellRendererParams would be an opportunity to not be too opinionated.

lbwexler:
Loading a "flat path" version will cause other things to break (namely
store filtering).

Thought the plan was to actually create the hierarchical records before
loading?

-Lee

lbwexler:
This is ready to be reviewed for mergeability

columns/Column.js
lbwexler: As discussed this was changed -- Client can override *all*

data/Record.js
lbwexler: Colin and I discussed this -- Tree Grid always expecting hierarchical data now.

PR #642, opened 2018-09-20
Relative time stamp features

shravnk:
RelativeTimeStamp takes two additional options, prefix and prefixMargin. Component is backwards compatible, but now includes a 'prefix' <span> if the option is included. Default prefixMargin is 10px.

Simple tooltip added to the timestamp <span> i.e. not the prefix span. Copied over old CSS from xh-hoist.


cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -102,6 +114,8 @@ export const relativeTimestamp = elemFactory(RelativeTimestamp);
  * @param {(Date|int)} timestamp - Date object or milliseconds that will be used as reference for this component
  * @param {Object} [options]
  * @param {boolean} [options.allowFuture] - Allow dates greater than Date.now()
+ * @param {string} [options.prefix] - Label preceding timestamp
+ * @param {int} [options.prefixMargin] - Right margin in pixels for the prefix label
```
amcclain: Could we just have a single space between the prefix and formatted string? That would adjust with the font size and not be something we need to specify / default in pixels.  I was thinking it would naturally be equivalent to the suffix. Any reason to treat them differently?

I also don't want to change the output of the getRelativeTimestamp method itself - that should continue to return a string (as documented and generally expected).

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -57,25 +59,35 @@ export class RelativeTimestamp extends Component {
...
+            items: [
+                span({
+                    item: prefix,
+                    style: {marginRight: prefixMargin}
+                }),
+                span({
+                    className: 'xh-title-tip',
+                    item: textItem,
+                    title: fmtDateTime(this.props.timestamp),
+                })
```
amcclain: See comment below which should simplify / make this fall out anyway, but it would be natural / friendly to have the tooltip appear if the user happened to hover over the prefix as well.

shravnk: I originally embedded prefix within the string as you suggest, but then probably started to overthink it. If you did want to offset the prefix, you could just leave the prefix option null and create a separate <span> label. Will make these changes, thanks.

amcclain:
👍 thanks will put this to immediate use

Issue #644, opened 2018-09-20
GridModel setGroupBy doesn't work properly after initial call


haynesjm42:
After initially calling setGroupBy with a field, subsequent calls fail to properly "ungroup" the previously set group column. This results in being unable to change the single group column (calling setGroupBy with a different field will add another grouping level) , as well as being unable to ungroup.

PR #645, opened 2018-09-20
Fix issues with changing groupBy after initial call

haynesjm42:
#644

Didn't dig too deep, but probably broke with the column definition refactoring, just needed to look for the `rowGroup` property in the correct place.

Added example of changing the groupBy field to toolbox in https://github.com/exhi/toolbox/commit/a3fa4485babde8af642db083e398f414c6c43005


amcclain:
Thank you for fixing!

PR #646, opened 2018-09-20
Grid row enhancements

TomTirapani:


lbwexler:
just curious about the decision to put the row class on the model instead of the grid.  Do we see the row class depending on business logic somehow?    Makes sense

Yes, good with this.

PR #648, opened 2018-09-20
Grid export allows for wrapped text, fixed width

shravnk:
- Added ExportFormat LONG_TEXT. Used to set an exportable column's type to 'longText'
- Added prop exportWidth on Column. Defaults to 100 characters if ExportFormat.LONG_TEXT = true


lbwexler:
Tom can you take a look at this and merge?

PR #649, opened 2018-09-20
Add RadioInput HoistInput, Fixes #601

### LINKED ISSUE #601: HoistInput addition: RadioField
Requested for client app.

febbraiod:


desktop/cmp/form/RadioInput.scss
```diff
@@ -0,0 +1,3 @@
+.xh-radio-input.bp3-control.bp3-inline {
+  margin-right: 20px;
+}
```
febbraiod: Our default margin for inline form fields made it difficult to understand with button belonged to what label.

desktop/cmp/form/RadioInput.js
```diff
@@ -0,0 +1,109 @@
+/*
```
febbraiod: Note: as it stands if you tab into the radio group you can select a button using the arrow keys, but the enter key will not select the highlighted options. Should we change that?

desktop/cmp/form/RadioInput.js
```diff
@@ -0,0 +1,109 @@
...
+import {isObject} from 'lodash';
+import {HoistComponent, elemFactory} from '@xh/hoist/core';
+import {observable, action} from '@xh/hoist/mobx';
+import {radio, radioGroup} from '@xh/hoist/kit/blueprint';
+import {withDefault} from '@xh/hoist/utils/js';
+
+import {HoistInput} from '@xh/hoist/cmp/form';
+import './RadioInput.scss';
+
+
```
febbraiod: Note: as it stands if you tab into the radio group you can select a button using the arrow keys, but the enter key will not choose the highlighted options. Should we change that?

PR #656, opened 2018-09-21
NumberInput renders unformatted value when focused

### LINKED ISSUE #639: NumberInput: display raw (unformatted) value whenever focused/editing
NumberInput formats its value on commit by inserting commas and fixing precision / zero padding. By design, this happens onCommit, so a user who is in the process of inputting a value sees their "raw" entry until they blur/commit, at which point the formatted value is flushed in for display.

When re-entering the field, however, to make another change, the formatted value remains in place. This means the user now has to deal with (or choose to ignore) commas. We have a request that the formatting be removed whenever editing, which I think does make sense and would be an improvement.

febbraiod:
Fixes #639 

+Slightly modify value handling - onValueChange and formatValue in lieu of toExternal, toInternal


amcclain:
Sweet thank you!

I see that the shorthand handling has also changed, in that as soon as you type `m` or `k` the number converts. I think this is an improvement - can see arguments both ways, but seems like it makes it clearer earlier what you are going to get.  Mainly wanted to ensure this was intentional and expected. (I have not had a chance to look at the new code.)

febbraiod:
Yes, Lee and I looked at it together and both thought it was a powerful improvement.

lbwexler:
yea -- I think I was maybe a little less aware of the full ramification of this -- it sorted of resulted by accident from a code cleanup  I suggested.  :-) Agree there could be pluses and minuses -- let's live with it for a bit.

Issue #659, opened 2018-09-24
Support absolute sorting in grids


TomTirapani:
Certain types of values (e.g. pnl values) are more useful sorted by absolute value. This is a custom feature we added to the Ext toolkit which we should bring to the react toolkit.

Columns should support an `absSort` flag. When enabled, sorting a grid column by clicking on its header would loop through 3 sort states (ascending, descending and absolute descending). Will likely require some UI customisation to display an appropriate sort icon when abs sorted.

TomTirapani:
Implemented using a custom header component, and a custom column comparator.
https://github.com/exhi/hoist-react/pull/662
https://github.com/exhi/toolbox/pull/83

lbwexler:
Fixed in https://github.com/exhi/hoist-react/issues/659

PR #662, opened 2018-09-25
Support absolute sorting in Grid. #659

TomTirapani:


lbwexler:
Looks great to me.  Tom will rename a file or two and make some additional tweaks we discussed, but I think the approach looks great.

@amcclain do you want to take a look at this as well?

desktop/cmp/grid/ag-grid/ColumnHeader.js
```diff
@@ -0,0 +1,132 @@
...
+    onClick = (e) => {
+        if (!this.props.enableSorting) return;
+
+        const {gridModel, activeGridSorter, colId} = this,
+            nextSortBy = this.getNextSortBy();
+
+        // Add to existing sorters if holding shift, else replace
+        let sortBy = e.shiftKey ? clone(gridModel.sortBy) : [];
+        if (activeGridSorter) {
+            sortBy = sortBy.filter(it => it.colId !== colId);
```
amcclain: Good use case for `_.pull()`?

desktop/cmp/grid/ag-grid/ColumnHeader.js
```diff
@@ -0,0 +1,132 @@
...
+            {sort, abs = false} = activeGridSorter || {};
+
+        if (!activeGridSorter) {
+            return {colId, sort: 'asc', abs: false};
+        } else if (sort === 'asc') {
+            return {colId, sort: 'desc', abs: false};
+        } else if (sort === 'desc' && column.absSort && !abs) {
+            return {colId, sort: 'desc', abs: true};
+        } else {
+            return null;
```
amcclain: This removes the sort when you have a non-abs col that was desc.  Wouldn't it be more natural to toggle that back to asc?

amcclain:
Realized we will need to fix `selectFirstRow` to account for abs sorting:

https://github.com/exhi/hoist-react/blob/d82d4861156cec18b6fd7739acde79c58e79f69b/desktop/cmp/grid/GridModel.js#L198

desktop/cmp/grid/ag-grid/ColumnHeader.js
amcclain: Sorry - just realized that is the default ag-grid behavior. Now not sure how standard that is - I found it odd and don't think most grids/tables work that way....  

"Unsorted" is a bit of a funny term too - assuming this is the order the data was provided to the store. But from a use perspective imagine the data happened to be name alpha asc. Let's also assume we've sorted on some other value by default.  Clicking the name header once would sort asc, again would sort dec, again would appear to sort asc but in a subtle way be "no sort" and then the fourth click would appear to do nothing...  

I dunno - not eager to change ag-grid functionality but found this weird.  E.g. see MacOS finder columns...

amcclain:
Also now that we have a sorter class don't think this check will ever pass and be a no-op...

https://github.com/exhi/hoist-react/blob/d82d4861156cec18b6fd7739acde79c58e79f69b/desktop/cmp/grid/Grid.js#L341

If that's the case we should remove - should we be concerned?

desktop/cmp/grid/GridSorter.js
```diff
@@ -6,14 +6,28 @@
...
+    /**
+     * Create a new GridSorter. Accepts a GridSorter configuration or a pipe delimited string
+     * generated using GridSorter.toString().
+     *
+     * @param {Object|String} [cfg] - GridSorter configuration or string representation.
+     */
+    static parse(cfg) {
+        if (isString(cfg)) {
+            const [colId, sort, abs] = cfg.split('|');
+            cfg = {colId, sort, abs: !!abs};
```
lbwexler: Love this use of array destructuring

lbwexler:
👍

PR #663, opened 2018-09-26
Refactor Hoist Inputs + new Select controls based on React Select

lbwexler:
+ Replace Blueprint Select package and related controls with React Select
+ Flatten Hierarchy
+ Remove Delegate Props 
+ Standardize handling of prop defaulting 
+ CheckBox is Checkbox
+ Better synchronization of model to controls
+ Workarounds to Blueprint Date Popover not closing
+ New and improved commenting on all props!


amcclain:
Pushed 0e73a18fc4a52c5898727f6ae22a76bf536f6104 to resolve issue noted in #725 where StoreFilterField did not render updated (empty) value on clear. We had another variant of this issue in the controlsPanel Toolbox example with the clear trigger in the "email" text input, which was clearing the value on the model but not updating itself internally since it maintained focus (and then would flush its stale value back to the model on blur).

cmp/form/HoistInput.js
```diff
@@ -7,98 +7,113 @@
...
+        super(props);
+
+        // Ensure that updates to the external value - sourced from either model or props - are
+        // always flushed to the internal value of this control and reflected in renderValue.
+        this.addReaction({
+            track: () => this.externalValue,
+            run: (externalVal) => {
+                this.setInternalValue(this.toInternal(externalVal));
+            },
+            fireImmediately: true
```
lbwexler: Do we need the fireImmediately here?  Does it replace something else that was initializing the internalValue?

desktop/cmp/form/inputs/DateInput.js
```diff
@@ -0,0 +1,226 @@
...
+
+            onChange: this.onChange
+        });
+    }
+
+    // Custom blur handler to account for focus potentially living in either input or popover.
+    // We want to call noteBlurred when focus has left both. Extra long delay here working around
+    // some kind of transition that happens when you use popover buttons to navigate between months.
+    // Focus appears to flap to focus for a tick, then back to the popover.... For review....
+    onBlur = () => {
```
lbwexler: Looks good -- thanks for this!

desktop/cmp/form/inputs/Select.js
```diff
@@ -0,0 +1,321 @@
...
+        loadingMessageFn: PT.func,
+
+        /** Placement of the dropdown menu relative to the input control. */
+        menuPlacement: PT.oneOf(['auto', 'top', 'bottom']),
+
+        /** Function to return message indicating no options loaded. Passed current query input. */
+        noOptionsMessageFn: PT.func,
+
+        /**
+         * Preset list of options for selection. Objects must contain a `value` property; a `label`
```
lbwexler: No convenience method of just providing an array of values?

desktop/cmp/form/inputs/Select.js
```diff
@@ -0,0 +1,321 @@
...
+            classNamePrefix: 'xh-select',
+            styles: this.getStylesConfig(),
+            theme: this.getThemeConfig(),
+
+            onChange: this.onSelectChange
+        };
+
+        if (this.asyncMode) {
+            reactSelectProps = {
+                ...reactSelectProps,
```
lbwexler: Why use the spread in these instances, creating 3 more clones.  We already create a new object on line 122, just write into it?

kit/blueprint/Wrappers.js
```diff
@@ -51,6 +51,9 @@ import {
 } from '@blueprintjs/core';
 
 import {DateInput} from '@blueprintjs/datetime';
+
+// TODO - if we could replace Omnibar (used in at least one client app) we could drop BP select dep.
```
lbwexler: Should ClientApp in question bring in Blueprint select directly on it own?

desktop/cmp/form/inputs/SwitchInput.js
```diff
@@ -0,0 +1,70 @@
...
+            className: this.getClassName(),
+
+            onBlur: this.onBlur,
+            onChange: this.onChange,
+            onFocus: this.onFocus
+        });
+    }
+
+    // Cast "false" -> false (see Admin config editor for a sample use case).
+    toInternal(external) {
```
lbwexler: This control explicitly has a prop type of boolean.  Seems a bit confusing here that we are bending over backward to allow it to bind to a string when in model mode.  Could we put the burden of this cast on the application?

Its sort of the same concern I have with the 'emitObjects'.  Would like the 'value' of the input, and its type, (where possible) to be consistent and easy to reason about.

PR #664, opened 2018-09-26
(WIP) Tag input component

shravnk:
Basic implementation of a Blueprint [Tag Input](https://blueprintjs.com/docs/#core/components/tag-input). API allows for further nuance, mainly by decomposing onChange into onAdd and onRemove and 
adding key press event handlers.

screen shot 2018-09-26 at 6 03 07 pm

### DIFF
--- a/desktop/cmp/form/TagInput.js
+++ b/desktop/cmp/form/TagInput.js
@@ -6,11 +6,10 @@
  */
 
 import {PropTypes as PT} from 'prop-types';
-import {action} from '@xh/hoist/mobx';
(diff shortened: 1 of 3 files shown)


amcclain:
Sorry to zap this (and thank you again for putting it together!), but the new react-select component can handle this pretty nicely if given an empty options list, enableMulti, and enableCreate. I do think there are some things we'll continue to tune about the sizing and displays, and it would be good to add an example of this config to Toolbox, but we'll build it off that base.

PR #671, opened 2018-10-02
dateIs change to accept "now" as a parameter

nadkarniarjun:
- Field validator dateIs could take special token for "now"

Updated Toolbox to show above as an example


field/validation/constraints.js
```diff
@@ -45,6 +46,9 @@ export function numberIs({min, max, notZero}) {
  */
 export function dateIs({min, max, fmt = 'YYYY-MM-DD'}) {
```
lbwexler: This change looks good.  Sorry if I wasn't clear -- was hoping we could go ahead and enhance the docs at this time, now that there is something important to document

amcclain: Something I've seen in a few other places and have meant to ask - why create a moment() vs new Date()?

nadkarniarjun: Sorry, yes, I can update the docs as well. Should have done that initially. 

This particular class didn't have any jsDocs documentation. Do you want me to create the jsDocs for this one, or just add comments to the functions as we have it right now?

nadkarniarjun: @amcclain No particular reason. I've used moment in the past because I find it easier to work with, but I can switch this to Date(). I'll do that while I'm adding the documentation.

nadkarniarjun: Also, not sure what other places you're referring to? I think this is the only place I've used moment. 

Do you want me to do a code search and see if we have Moment elsewhere?

lbwexler: Arjun -- no magic formula here as to how far to go with the doc enhancements  Usually good to clean-up the entire classes docs (or in this case all the docs in the file).  Its just 4 methods here, and would like to see the parameters to them described....

lbwexler: don't worry about auditing all changes to moment in the toolkit -- think anselm was just asking in general.  Let's get this change in

nadkarniarjun: Done!

PR #680, opened 2018-10-03
Ensure column size changes due to autoSize are flushed to GridModel

### LINKED ISSUE #670: Column width changes triggered by ag-Grid "autosize" feature not saved
* Enable the "autosize to data" context menu option as provided by ag-grid
* Auto-size a grid with gridState enabled
* Refresh - note column widths that were programatically set by ag-grid routine not restored.

amcclain:
+ We should look at debouncing `GridModel.noteAgColumnStateChanged()`. In normal usage this change does not appear to generate extra calls to the model, but e.g. double-clicking a column header border to auto-size appears to fire two dragEnd and one autoSize events, resulting in three calls.
+ Fixes #670

### DIFF
--- a/desktop/cmp/grid/Grid.js
+++ b/desktop/cmp/grid/Grid.js
@@ -143,6 +143,7 @@ export class Grid extends Component {
             onSelectionChanged: this.onSelectionChanged,
             onGridSizeChanged: this.onGridSizeChanged,
             onDragStopped: this.onDragStopped,
+            onColumnResized: this.onColumnResized,
 
             groupDefaultExpanded: 1,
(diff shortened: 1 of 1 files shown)


amcclain:
This can be tested in Toolbox by adding a stateModel string ID + 'autoSizeAll' context menu to SampleGrid.

amcclain:
Sorry also to bail on the actual debouncing of that method in this PR - if that is indeed something we want to do.  I don't think this change makes that *more* necessary in any way, and I wanted to get this fix in with a minimal diff.  If on review it's important we can do here or file an issue.

PR #681, opened 2018-10-03
cleanup action creation

lbwexler:


lbwexler:
The code was weirdly relying on an undocumented feature that RecordAction constructor would also clone a RecordAction given to it.

haynesjm42:
I wouldn't say it is an undocumented feature as much as a language feature, the RecordAction ctor doesn't case what you pass it as long as it is an object with the correct properties, and with the object destructuring pattern we are using for

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -42,7 +42,9 @@ export class StoreContextMenu {
     constructor({items, gridModel}) {
         this.gridModel = gridModel;
         this.items = flatten(items.map(it => {
-            return isString(it) ? this.parseToken(it) : new RecordAction(it);
+            if (it instanceof RecordAction) return it;
```
haynesjm42: This is dangerous since the RecordActions are now used as the backing state for multiple UI components, which can/will be on the screen simultaneously. If we do this then we will run into issues if users are passing the same RecordAction instance to the context menu and action bar.

I think we do want to always make a clone of the RecordActions which are passed to the StoreContextMenu and RecordActionBar. Agree that depending on the RecordAction ctor destructuring for cloning the way I was is also dangerous and likely to break in the future, should we instead do a lodash `cloneDeep` here?

desktop/cmp/record/RecordActionBar.js
```diff
@@ -45,7 +45,7 @@ export class RecordActionBar extends Component {
         super(props);
         this.actions = props.actions
             .filter(Boolean)
-            .map(it => new RecordAction(it));
+            .map(it => it instanceof RecordAction ? it : new RecordAction(it));
```
haynesjm42: Same danger here as in StoreContextMenu of the possibility of multiple UI components sharing the same backing RecordAction.

lbwexler:
Hi John -- Had a bit of time to think about this.  This pattern is actually what we are doing with columns.  (See GridModel, line 395).    Think this is a *great* example of why it would have been better if a class like RecordAction or Column were immutable

The reason its not a problem with Column is that we usually pass the *config* for a column in to GridModel -- the column class is really just an internal thing (plus, we don't share them).    We could certainly adopt the same pattern for RecordActions -- i.e. we document them as just not something applications typically create directly..

Not sure what the right thing to do is here.  I think automatically cloning them is weird, and inconsistent.  But agree there is a potential for misuse.  

The other alternative is that for both columns and RecordActions, we *only* tale a config.

haynesjm42:
Agree cloning is weird. If we never want an app to create instances of these classes then I think we should only accept configs for them in all app-facing interfaces, unless there is really a use-case for accepting a pre-constructed instance? I can't think of any..

lbwexler:
OK -- wanted to just get this very minor PR off the table, and know that we have larger changes to RecordAction coming.

Per John's suggestions, ended up essentially restoring the previous behavior, but clarifying with an explicit clone method.

amcclain:
Great - that's exactly what I was going to suggest - that pattern has served us well in the past and paved the way for better immutability support.  I think we should definitely look at making some of these core objects immutable. 

Another advantage in my mind is that then our validation and defaulting logic can stay in the constructor and not need to be repeated or spun out across setters - thinking of e.g. Column here were we want to have controlled relationships between e.g. width and flex.

(On the subject of immutability in general was just reading up on https://github.com/mweststrate/immer the other day and wondering how/if it could be useful. It automatically freezes objects as well, enforcing immutability when you have decided you want it.)

PR #684, opened 2018-10-03
Record action refactor

haynesjm42:
Refactoring of RecordAction and the RestGrid APIs for configuring which actions are enabled and where they should be available.

Overview of changes:
* RecordAction `prepareFn` has been replaced with `displayFn` which should return a display spec for rendering the action. This replaces the old behavior in the `prepareFn` callback where the action state would be modified.
* RecordAction `actionFn` and `displayFn` now take a single object as their param, instead of separate params. The callbacks are passed an object with the `record`, `selectedRecords`, `action`, `gridModel` and `column` (depending on what is available contextually)
* RestGridModel `actionEnabled` flags have been replaced with `toolbarActions`, `menuActions`, and `formActions` for configuring which actions are available and where
* GridModel now maintains any additional properties in the config object passed to the constructor as members
* RestForm now supports a read-only "view" mode. 
* RecordAction `items` property is now supported in StoreContextMenu/Grid component for creating sub-menus

See https://github.com/exhi/toolbox/tree/recordActionRefactor for associated toolbox changes


data/RecordAction.js
```diff
@@ -48,6 +52,15 @@ export class RecordAction {
      *          Can specify 0 to only enable action if no records are active.
      *      true: specifies that number of records > 0. Allows for arbitrary number of records.
      *      false: specifies any number of records (0 - infinity, inclusive). Always active.
+     * @param {(string|boolean)} [c.confirm] - whether the action needs to be confirmed before being
+     *      executed. Truthy values will determine the message displayed to the user. If provided the
+     *      action text will be used as the confirm button text and dialog title, otherwise "Confirm"
+     *      will be displayed.
+     *
+     *      string: specifies the exact message to display to the user
```
lbwexler: My first thought is that adding the confirm here --- and trying to standardize the language it contains, is going one step too far.  What if a user needs multiple confirms?  data driven confirms? etc.

data/RecordAction.js
```diff
@@ -71,14 +85,56 @@ export class RecordAction {
...
+
+        const confirmed = await promise;
+        if (!confirmed) return;
+
+        this.actionFn({action: this, record, selModel, metadata});
     }
 }
 
 /**
  * @callback ActionCb - called when the action's UI element is clicked or otherwise triggered.
```
lbwexler: would now be the time to consider simplifying this by just having a single bundle of records rather than 'record' and 'selection'.  Do we ever need to know what the primary record clicked on is?  If we did, could that be buried in the 'p.metadata'.?   Would any action operate on both the primary record and the selection?  Mostly thinking outloud here

desktop/cmp/record/RecordActionBar.js
```diff
@@ -32,51 +33,59 @@ export class RecordActionBar extends Component {
 
     static propTypes = {
         /** RecordActions to clone or configs to create. */
-        actions: PT.arrayOf(PT.oneOfType([PT.object, PT.instanceOf(RecordAction)])).isRequired,
+        actions: PT.arrayOf(PT.oneOfType([PT.object, RecordAction])).isRequired,
```
lbwexler: good catch

desktop/cmp/record/RecordActionBar.js
```diff
@@ -32,51 +33,59 @@ export class RecordActionBar extends Component {
...
+        selModel: PT.instanceOf(StoreSelectionModel),
 
-    constructor(props) {
-        super(props);
-        this.actions = props.actions
-            .filter(Boolean)
-            .map(it => new RecordAction(it));
-    }
+        /** Data to pass through to action callbacks */
+        actionMetadata: PT.object,
```
lbwexler: Just thinking about it -- not sure whether it really makes sense for the public component for use in a toolbar the same object as the private object used by the column (i.e.. leaves us with the wonky "showOnHoverOnly" property)

desktop/cmp/grid/GridModel.js
```diff
@@ -139,7 +139,8 @@ export class GridModel {
         enableExport = false,
         exportFilename = 'export',
         rowClassFn = null,
-        contextMenuFn = () => this.defaultContextMenu()
+        contextMenuFn = () => this.defaultContextMenu(),
+        actionMetadata = null
```
lbwexler: Let's discuss the metadata issue in person -- know you were concerned about it and think it best if you explain

desktop/cmp/record/RecordActionButton.js
```diff
@@ -0,0 +1,65 @@
+import {Component} from 'react';
+import {PropTypes as PT} from 'prop-types';
```
lbwexler: Should this be private?  Trying to minimize the API surface here....

desktop/cmp/rest/RestGridModel.js
```diff
@@ -12,6 +12,53 @@ import {Icon} from '@xh/hoist/icon';
 import {pluralize} from '@xh/hoist/utils/js';
 
 import {RestFormModel} from './RestFormModel';
+import {withDefault} from '../../../utils/js/LangUtils';
+
+export const restGridAddAction = {
```
lbwexler: Very cool -- wonder if we need the 'restGrid' prefix here.  Its implied by package we are importing from

desktop/cmp/record/RecordActionBar.js
haynesjm42: Didn't address this in my most recent changes, but I see the point re: the `showOnHoverOnly` prop, don't know that you would ever use that outside of the grid action col actionBar.

We could actually remove the prop entirely and have the actionCol partial conditionally add a css class to the element it returns in the render fn based on the `actionShowOnHoverOnly` column prop... thoughts?

data/RecordAction.js
```diff
@@ -41,7 +70,7 @@ export class RecordAction {
      * @param {ActionCb} [c.actionFn] - called on store action activation.
      * @param {boolean} [c.disabled] - true to disable this item.
      * @param {boolean} [c.hidden] - true to hide this item.
-     * @param {PrepareFnCb} [c.prepareFn] - called prior to showing the action in the UI.
+     * @param {DisplayConfigFn} [c.displayConfigFn] - called prior to showing the action in the UI.
```
lbwexler: Thinking we should use the name 'renderFn' or 'displayFn'?   This name is a bit of a mouthful, and I think of a 'config' as something you do once, when you construct an object...

columns/Column.js
```diff
@@ -160,31 +162,33 @@ export class Column {
         this.excludeFromExport = withDefault(excludeFromExport, !field);
 
         this.tooltip = tooltip;
+        this.gridModel = gridModel;
```
lbwexler: Are we no longer deep cloning these?

desktop/cmp/grid/GridModel.js
```diff
@@ -124,6 +124,7 @@ export class GridModel {
      *      Should return a string or array of strings. Receives record data as param.
      * @param {function} [c.contextMenuFn] - closure returning a StoreContextMenu.
      *      @see StoreContextMenu
+     * @param {Object} [c.actionMetadata] - additional data to be passed through to actions
```
lbwexler: This comment looks out of date

desktop/cmp/record/RecordActionButton.js
lbwexler: great -- can we put it in an impl sub package?

desktop/cmp/rest/RestForm.js
```diff
@@ -47,29 +49,26 @@ export class RestForm extends Component {
...
         return [
-            button({
-                text: 'Delete',
-                icon: Icon.delete(),
-                intent: 'danger',
-                onClick: this.onDeleteClick,
-                omit: !actionEnabled.del || isAdd
+            recordActionBar({
+                actions: toolbarActions,
+                actionMetadata: {restModel: this.model},
```
lbwexler: thought the plan was to have the actions all passed their grid, and to have the model hanging off of the grid?  Still find the passing around of the actionMetadata confusing.

data/RecordAction.js
```diff
@@ -5,6 +5,35 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
+import {Icon} from '@xh/hoist/icon';
+import {isBoolean, isNumber, isNil} from 'lodash';
+
+export const addAction = {
```
lbwexler: There are interesting.  More graphical, than anything, as they don't actually have actionFns.

Do we think this is really generalizable as a hoist API? Perhaps they should just be defined in the Rest package, where we actually have implementations?

admin/tabs/activity/feedback/FeedbackPanel.js
```diff
@@ -49,10 +49,7 @@ export class FeedbackPanel extends Component {
                 }
             ]
         }),
-        actionEnabled: {
-            add: false,
-            edit: false
-        },
+        toolbarActions: [deleteAction],
```
lbwexler: awesome!

desktop/cmp/rest/RestGridModel.js
```diff
@@ -22,11 +54,9 @@ export class RestGridModel {
...
     // Properties
     //----------------
-    actionEnabled = {
-        add: true,
-        edit: true,
-        del: true
-    };
+    toolbarActions;
+    contextMenuActions;
+    formToolbarActions;
```
lbwexler: cool -- would just call the last one 'formActions' for simplicity.

desktop/cmp/rest/RestFormModel.js
```diff
@@ -40,25 +46,21 @@ export class RestFormModel {
...
-        return (isAdd && actionEnabled.add) || (!isAdd && actionEnabled.edit);
-    }
-
     @computed
     get isDirty() {
         return !isEqual(this.record, this.originalRecord);
     }
 
-    constructor({parent, editors}) {
+    constructor({parent, editors, toolbarActions}) {
```
lbwexler: I think we should consider making all of these RestXXX classes private for the time being, with the only public entry point being RestGridModel and RestGrid. 

If we did this, then the toolbarActions property would not have a default here, but just get its default from RestGridModel

desktop/cmp/rest/RestGridModel.js
```diff
@@ -112,26 +165,15 @@ export class RestGridModel {
         this.formModel.openEdit(record);
     }
 
+    @action
+    viewRecord(record) {
+        this.formModel.openView(record);
```
lbwexler: great!

desktop/cmp/rest/impl/RestControl.js
```diff
@@ -74,7 +83,8 @@ export class RestControl extends Component {
         if (type === 'date') {
             value = value ? fmtDateTime(value) : '';
         }
-        return label(value);
+
+        return label(value ? value.toString() : null);
```
lbwexler: what if the value is falsey here?   Won't that be a problem?

PR #689, opened 2018-10-04
(Bug Fix) JSONInput reacts to value changes

shravnk:
Also passing props into the constructor.
This is required for Toolbox Other/Modals tab to work properly

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -105,7 +105,7 @@ class XHClass {
 
     /** State of app -- see AppState for valid values. */
     @observable appState = AppState.PRE_AUTH;
-f
+
     /** Currently authenticated user. */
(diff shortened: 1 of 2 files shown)


amcclain:
I found that this caused the editor to always flush a "new" value with each keystroke and set your cursor back to the start of the string, making it effectively impossible to edit the field.  You can see this now in currently deployed toolbox.xh.io (I came across it tweaking a config in a client app).

I pushed 4cb6ac18336793376aa112010348038502da5b40 which seems to resolve - @shravnk can you please verify that it still works in the use case that triggered this change? Thanks much.

shravnk:
Use case (currently mothballed) still works with this change. Thank you for the correction.

PR #702, opened 2018-10-12
Updated format number to use numbro instead of numeral.

cnrudd:
This PR has a corresponding ['numbro' PR](https://github.com/exhi/toolbox/pull/92) in toolbox, where a 'Formats' tab has been added to test number formatting.

This PR just replaces numeral.js with numbro.js, and rewrites buildFormatPattern to buildFormatConfig, and replaces fmtNumber option 'formatPattern' with 'formatConfig'.

This is a breaking change if any apps were using 'formatPattern', but I searched client apps and did not find any (though I do not have access to all).

Note:  this is a minimal drop in replacement with the minimal code changes needed to make it work.
numbro has other features we could take advantage of in later refactors, such as 'average' for smart number compacting, 'percentage', and maybe 'currency'.


lbwexler:
+ We could  take this change, just to get the new library in.  

+ We could go further and simplify our implementation by using new numbro features to implement our 'ledger' and 'withPlusSign' features

+ We could go even further by removing  'ledger'/'precision'/ 'zeroPad'/'withPlusSign' arguments in favor of raw numbro

+ We could go in the other direction by removing the new "formatPattern/formatConfig" property from the public API -- where it doesn't seem to be greatly used -- and just relying on it for our internal implementation.

Thoughts?

lbwexler:
Does anybody understand the difference between 'trimMantissa'? and 'optionalMantissa'? They seem to do the same thing -- feel like I need to understand this better to be confident this implementation replicates what we have/want.

cnrudd:
> Does anybody understand the difference between 'trimMantissa'? and 'optionalMantissa'? They seem to do the same thing -- feel like I need to understand this better to be confident this implementation replicates what we have/want.

explanation for optionalMantissa:  https://github.com/BenjaminVanRyseghem/numbro/blob/eac3859d2feec557eafed8bc2b25b2d3ca096c28/src/formatting.js#L476
explanation for trimMantissa:
https://github.com/BenjaminVanRyseghem/numbro/blob/eac3859d2feec557eafed8bc2b25b2d3ca096c28/src/formatting.js#L478

cnrudd:
We need to put this merge on hold until this numbro bug is fixed:
https://github.com/BenjaminVanRyseghem/numbro/issues/372

lbwexler:
I made this change.  If you test and feel comfortable with this, am happy
to merge it.

amcclain:
Excellent - glad this made it in!

PR #703, opened 2018-10-15
Mobile agGrid

TomTirapani:
Use agGrid to power mobile grid.

Note: Before merging, we should move cross-platform classes like GridModel, GridStateModel etc to /cmp/grid. They have been left in desktop/cmp/grid to mitigate merge conflicts during review.


columns/Column.js
```diff
@@ -221,22 +225,31 @@ export class Column {
             ret.width = this.width;
         }
 
-        const {renderer, elementRenderer} = this;
-        if (renderer) {
+        const {subFields, renderer, elementRenderer} = this,
+            column = this; // eslint-disable-line consistent-this
```
lbwexler: can we use me here, without this comment?

desktop/cmp/grid/Grid.js
```diff
@@ -104,15 +113,15 @@ export class Grid extends Component {
...
-                    showHover ? 'xh-grid-show-hover' : ''
+                    rowHeightCls,
+                    !mobile && showHover ? 'xh-grid-show-hover' : ''
                 ),
-                onKeyDown
+                onKeyDown: !mobile ? onKeyDown : null
             }),
-            colChooser({
+            !mobile ? colChooser({
                 omit: !colChooserModel,
```
lbwexler: can we either omit here *or* the ternery operator? using both seems weird

desktop/cmp/grid/Grid.js
```diff
@@ -85,9 +92,11 @@ export class Grid extends Component {
     }
 
     render() {
-        const {colChooserModel, compact} = this.model,
+        const {colChooserModel, hasSubFields, compact} = this.model,
             {agOptions, showHover, onKeyDown} = this.props,
-            layoutProps = this.getLayoutProps();
+            mobile = XH.app.isMobile,
```
lbwexler: consider using all caps 'MOBILE' to help distinguish it?  Sort of orthogonal to all other logic

desktop/cmp/grid/GridModel.js
```diff
@@ -163,6 +166,7 @@ export class GridModel {
         this.setSortBy(sortBy);
         this.setCompact(compact);
 
+        if (!selModel) selModel = XH.app.isMobile ? 'disabled' : 'single';
```
lbwexler: use `withDefault()`, here and remove the defaulting on line 134.  This will allow applications to set to null explicitly, if they wish.

desktop/cmp/grid/GridModel.js
```diff
@@ -228,6 +232,11 @@ export class GridModel {
         return this.selModel.singleRecord;
     }
 
+    @computed
+    get hasSubFields() {
+        return !!this.subFields.length;
```
lbwexler: 'subFields' are not observable (and in fact not even declared at top of class).  This @computed does nothing without that.  Make subFields observable.ref if so desired, but perhaps they really could be part of the "immutable" public properties.  Do we expect them to change?

PR #706, opened 2018-10-15
Implementation of fetchservice cancellable by requestKey

cnrudd:
This resolves issue: https://github.com/exhi/hoist-react/issues/698

There is a corresponding toolbox branch (fetchservice_cancel) which does not need to be merged in, but can be used to test this implementation by cancelling News refreshes.

### DIFF
--- a/core/ExceptionHandler.js
+++ b/core/ExceptionHandler.js
@@ -129,6 +129,10 @@ export class ExceptionHandler {
         const ret = Object.assign({}, options),
             isAutoRefresh = exception.fetchOptions && exception.fetchOptions.isAutoRefresh;
 
+        if (exception.name == 'AbortError') {
+            ret.showAlert = ret.logOnServer = false;
+        }
+
         ret.showAsError = ret.showAsError != null ? ret.showAsError : true;
         ret.logOnServer = ret.logOnServer != null ? ret.logOnServer : ret.showAsError;
         ret.showAlert = ret.showAlert != null ? ret.showAlert : !isAutoRefresh;
--- a/core/ExceptionHandler.js
+++ b/core/ExceptionHandler.js
@@ -7,7 +7,7 @@
 import {Exception} from '@xh/hoist/exception';
 import {XH} from './XH';
 import {stringifyErrorSafely} from '@xh/hoist/exception';
-import {stripTags} from '@xh/hoist/utils/js';
+import {stripTags, withDefault} from '@xh/hoist/utils/js';
 
 /**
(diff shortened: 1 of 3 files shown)


core/ExceptionHandler.js
```diff
@@ -129,6 +129,10 @@ export class ExceptionHandler {
         const ret = Object.assign({}, options),
             isAutoRefresh = exception.fetchOptions && exception.fetchOptions.isAutoRefresh;
 
+        if (exception.name == 'AbortError') {
```
lbwexler: These defaults are important, but should still not override what the user specified in handleException().

Also, can we change all of this code to use the much cleaner "withDefault" method?

svc/FetchService.js
```diff
@@ -45,11 +45,12 @@ export class FetchService {
      * @param {Object} [opts.qsOpts] - Object of options to pass to the param converter library, qs.
      *      The default qsOpts are: {arrayFormat: 'repeat', allowDots: true}.
      *      @see {@link https://www.npmjs.com/package/qs}
+     * @param {string} [opts.requestKey] - Key to the abort object to call when cancelling an in flight request.
```
lbwexler: This doc and name arr now a bit confusing.  It seems to me that you have implemented something that ought to be named 'autoCancelKey' and be described as.
     "If set, any pending requests associated with the same autoCancelKey will
     be immediately cancelled when a new request is made".

svc/FetchService.js
```diff
@@ -158,9 +172,19 @@ export class FetchService {
...
+        if (!ctrl) return;
+
+        delete this.abortControllers[key];
+        ctrl.abort();
+    }
+
     //-----------------------
     // Implementation
     //-----------------------
+    abortControllers = {};
```
lbwexler: let's put this state at top of the class, by convention?

lbwexler:
To me, the question here is do we only want to support the 'autoCancelKey'?  Or do we want to provide additional support for the more general ability to get a handle on the group of requests and cancel them (the way we did before). 

I think that someone that really wants to manually, can now just use the AbortController directly and provide a shared signal for the entire group.  That's just about as easy as anything we would build, so need to layer anything more complicated on.

exception/Exception.js
```diff
@@ -71,6 +71,14 @@ export class Exception {
         return this.createInternal(defaults, {});
     }
 
+    static fetchCancelled(fetchOptions, response) {
```
lbwexler: needs to be documented like its siblings

PR #708, opened 2018-10-16
ReadOnlyField

TomTirapani:
Add ReadOnlyField, a HoistInput wrapper which renders its readOnlyValue.

Test with toolbox branch https://github.com/exhi/toolbox/tree/readOnlyForms


amcclain:
I've reviewed this and discussed with Tom - will continue to do so and aim to incorporate the RO functionality into refactored HoistInput branch before/after merge of that project.

lbwexler:
A bit late to this party, but I wonder if would be simpler to build this into FormField?   

Its an existing layer surrounding our HoistInputs, and we only have one per platform, so perhaps a more manageable change.

PR #709, opened 2018-10-16
RecordAction sub-menu fix

haynesjm42:
Encountered a couple issues while using the new features in a client-app. The main change here is adding support for dynamic sub menus returned from the `displayFn` callback. Previously we were just taking them and assuming that they are RecordActions (which they would have been if they were statically defined in the RecordAction config).


desktop/cmp/grid/Grid.js
```diff
@@ -217,9 +218,10 @@ export class Grid extends Component {
             const displaySpec = action.getDisplaySpec(params);
             if (displaySpec.hidden) return;
 
-            let subMenu;
+            let childItems;
             if (!isEmpty(displaySpec.items)) {
-                subMenu = this.buildMenuItems(displaySpec.items, record, selectedRecords);
+                const menu = new StoreContextMenu({items: displaySpec.items, gridModel: this.gridModel});
```
haynesjm42: Doing this so we fully support all of the special string tokens for the sub menu items

Issue #704, opened 2018-10-15
Default 'Copy' on ContextMenu on DataView is bust


lbwexler:
The Copy Item appears to just copy out an '[Object] object'  string.

See Toolbox 'News' example and 'DataView' example in grids

lbwexler:
Yes -- absolutely.  Brendan -- can you whack the default context menu here on DataViewModel?  Think it probably was just a copy error from the very original version.

shravnk:
Sure thing

PR #711, opened 2018-10-16
Modify valueGetter to produce a string of each cell's data

shravnk:
For #704


lbwexler:
This seems reasonable enough, but I think in addition, we should just whack the default context menu on data view.  Neither 'copy' nor 'export' really make sense as default actions in this case.

PR #712, opened 2018-10-16
Hide actions for group rows

haynesjm42:


desktop/columns/Actions.js
```diff
@@ -41,7 +43,9 @@ export const actionCol = {
     chooserName: 'Actions',
     chooserDescription: 'Record Actions',
     excludeFromExport: true,
-    elementRenderer: (value, {record, column}) => {
+    elementRenderer: (value, {record, column, agParams}) => {
```
lbwexler: john -- I have a sinking feeling this is going to kill the action column for rows in our TreeGrid, which would *not* be acceptable....those rows definitely are bound to records

lbwexler: can you please confirm this is not the case?

haynesjm42: Rest assured that it is not the case 😄

lbwexler:
OK.  Am going to merge it and sleep quietly then.  😴

Issue #690, opened 2018-10-05
Enhance Copy Support in grid


lbwexler:
Would like to have a build-in ability to copy a *cell* value only  as well as the entire selection.
This is a facility we had in Hoist-Sencha.

Need to make sure our RecordActions are getting metadata for column/cell to do this.

amcclain:
We could consider generally re-implementing the built-in copy options of the ag-Grid context menu when we do this. Add the menu items ourselves, use our copy icon, have a hook for whatever other control or customization we might want to apply here.

lbwexler:
See the following ticket as well -- as part of this enhancement, need to be sure to handle hidden columns correctly.

https://github.com/exhi/hoist-react/issues/916

PR #714, opened 2018-10-16
Copy grid cells

shravnk:
For #690 
A couple questions:

1. Do we want to default to allow individual cell copy, as I have done in this PR? Might depend on how well we're able to control cell selecting behavior (see next point).
2. If we continue the behavior of clicking to select an entire row, then we probably do not want to allow multiple individual cells (i.e. range) to be selected. In theory ag-grid makes this possible via `supressMultiRangeSelection` but I couldnt get it to work in toolbox or even in their [demo](https://www.ag-grid.com/javascript-grid-range-selection). I wrote my own version of that option in this PR but it would likely need work.


shravnk:
I've gone back and forth with this one and think I have it in a place where it will solve the two relevant tickets (#690 and #916) while not forcing us to change the `Grid` API, especially wrt Range Selection. This PR:

1. Allows copy of individual cells without 'selecting' them. Used ag's `onCellContextMenu` handler and `GridModel` state. Another option would be to use ag's `copyFocusedCellToClipboard`. In both cases we're sort of hacking the ag api - I have tested pretty thoroughly though.

2. Replace default ag 'copy', 'copy w/ headers' context menu items with Hoist `RecordAction`s. This gives us control over which columns are included, standardizes the copy icons, and pluralizes menu text based on number of rows selected. This is an intermediate step - for full control over copy behavior would need to implement `[sendToClipboard](https://www.ag-grid.com/javascript-grid-clipboard/)`

lbwexler:
John -- Can you take a look at this one as well?  

Would be nice if there was a way to get the clicked on cell without having to save it in the cell context menu.

lbwexler:
@haynesjm42  -- please let me know your thoughts on this one?  Would like to close it out, or move it toward merging

haynesjm42:
Reviewing this now

haynesjm42:
So still a WIP, but played around with some changes to how this was originally implemented, and I think that at least for cell copying this is the correct approach to take.

- Cell copy now uses params passed to RecordAction `actionFn` (I added the Hoist column and ag-Grid params to the action params) instead of keeping track of which cell last got the cellContextMenu event
- Row copy now uses a new method in AgGridModel which handles serialization of row node data, with support for customizing the row and value delimiters, including column headers, hiding columns which are not visible, and also filtering columns via a callback fn. 
- Use [clipboard-copy](https://www.npmjs.com/package/clipboard-copy) library for copying data to clipboard, instead of using non-public ag-Grid APIs (or even using ag-Grid apis at all for actually putting data into the clipboard). Added this new dependency because clipboard-js doesn't seem to allow you do put anything in the clipboard without attaching to an element, though I may have missed something. clipboard-copy is a very simple and lightweight way of just putting something in the clipboard. Tested in Chrome and Safari and worked well.

I think this is moving in the right direction, but I was also thinking that it would be nice if the grid serialization was more closely tied with the export functionality. Seems like you would want a similar interface whenever you are serializing grid data. My hope with this approach was that the row copying would also work with a basic AgGrid component in pivot mode, but it appears that the ag-Grid api for getting a cell value does not work with non-primary column, though I didn't dig into too much to see if I was missing something.

haynesjm42:
Re-opening with cleaner non-wip cell-copy-only pr

PR #717, opened 2018-10-18
Add ag-Grid valueGetters to treeCol

amcclain:
+ For review - not sure what effect of valueGetter is - filterValueGetter does extract the intended values for the ag-grid filter menu if enabled.
+ Sorting still not right - did not look yet
+ Noticed that for treeCol we assume renderer is just extracting field from val - I think that's fine but is it worth throwing an exception or warning if someone tries to use a renderer on a treecol?


amcclain:
Closing this - will open another with this + tree column sorting fix proposal.

PR #719, opened 2018-10-18
Don't overwrite custom setter in @settable

amcclain:
+ To support this decorator getting called by @field() but you need a custom model setter


amcclain:
I think this is a good idea - wanted it while testing, do not need for a current application.

lbwexler:
👍

PR #726, opened 2018-10-21
Dimension Chooser

shravnk:
- Popover to set dimensions on tree grid
- Adjacent popover to clear all, add all, reset defaults. Includes nested history popover


shravnk:
See new commit for changes to bring in line with the Sencha functionality. A few remaining questions I had:

1. Should we warn users before switching a non-leaf column to a leaf column? i.e. warn that other currently selected column(s) will be lost
2. Should we throw if the component is initialized with multiple leaf columns selected?
3. Does it make sense to replace the final 'Add dimension' select with a Popover ButtonGroup which displays on hover?

amcclain:
Excellent! Glad to have this merged in - it's looking great.

I pushed a small change to Toolbox Bootstrap.groovy over the weekend to ensure we ended up with `[]` as our default value for the history preference there (this is on the new example app branch).  Without that, the default value was null and the dimChooser threw with an NPE on a call to `map()`.  The relevant code is really here:

https://github.com/exhi/hoist-react/blob/7f36df010b37c1104c01d7dfac849872b0eb981a/desktop/cmp/dimensionchooser/DimensionChooserModel.js#L143

Would it be safer to return an empty array in this case?  I think so but wanted to check.

shravnk:
Thanks, I just pushed an improved 'null handler'
Another question regarding the first time edge case: Should pending dims, i.e. the edit menu, be populated with the default value? Or should we keep as is, but not allow user to commit an empty value?
My preference would be the former, which is what I just pushed.

PR #727, opened 2018-10-21
Restructure Hoist Initialization and Top-Level App Structure

lbwexler:
+ New compact API for specifying App Properties and classes  -- `AppSpec`.
+ Top level App/AppModel follow our standard naming conventions.  No need for special instantiation of root model class.
+ XH.app is now XH.appModel
+ All Application code now constructed and initialized *only* when Hoist and its services are fully ready.
+ All HoistServices are now installed dynamically on XH.  This should allow better ability for overriding core hoist services, discovering "plug-in" services, and instrumenting all services.
+ Use Naming conventions more strongly for HoistServices.
+ Hoist no longer makes method calls on services before they have been initialized.
+ New support for mobx when() behavior.  Use for more decoupled service implementations.


lbwexler:
OK -- @amcclain -- please review.  This should be ready to merge.

Confirmed that when() fires at immediately if predicate is true when first hit!

PR #742, opened 2018-10-25
Fixes to value handling for treeColumn

### LINKED ISSUE #679: Tree grid column should implement valueGetter
It looks like we get an index or other ID instead of the value. This affects both sorting as well as the "copy" functionality exposed via context menu.

amcclain:
+ Our implementation of getDataPath() -> Record.xhTreePath results in treeColumn cells having the record ID as their underlying value via the node's groupData property. This is prioritized by default over the field value for filtering, sorting, and copying.
+ Fixes #679

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -169,10 +169,10 @@ export class Column {
      * Produce a Column definition appropriate for AG Grid.
      */
     getAgSpec() {
-        const {gridModel} = this,
+        const {gridModel, field} = this,
             me = this,
             ret = {
-                field: this.field,
+                field,
                 colId: this.colId,
                 headerName: this.headerName,
                 headerClass: this.headerClass,
@@ -190,14 +190,19 @@ export class Column {
                 headerComponentParams: {gridModel, column: this}
             };
 
+        // Our implementation of Grid.getDataPath() > Record.xhTreePath returns data path []s of
+        // Record IDs. TreeColumns use those IDs as their cell values, regardless of field.
+        // Add valueGetters below to correct + additional fixes for sorting below.
         if (this.isTreeColumn) {
             ret.showRowGroup = true;
             ret.cellRenderer = 'agGroupCellRenderer';
             ret.cellRendererParams = {
                 suppressCount: true,
                 suppressDoubleClickExpand: true,
-                innerRenderer: (v) => v.data[this.field]
+                innerRenderer: (v) => v.data[field]
             };
+            ret.valueGetter = (v) => v.data[field];
+            ret.filterValueGetter = (v) => v.data[field];
         }
 
         if (this.tooltip) {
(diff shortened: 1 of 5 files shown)


amcclain:
+ See Toolbox treeGrid for example of the current issue - when sorting by name, instrument names will remain out-of-order - it's the IDs that are being compared. Right-clicking and copying would also result in the ID being copied to clipboard.

cmp/grid/columns/Column.js
```diff
@@ -238,11 +243,22 @@ export class Column {
...
         const sortCfg = find(gridModel.sortBy, {colId: ret.colId});
         if (sortCfg) {
             ret.sort = sortCfg.sort;
             ret.sortedAt = gridModel.sortBy.indexOf(sortCfg);
-            ret.comparator = (v1, v2) => sortCfg.comparator(v1, v2);
+
+            // ag-Grid sort impl. sources its primary values from the node's `groupData` property,
+            // which is not what we want when sorting treeColumns.
+            if (this.isTreeColumn) {
+                ret.comparator = (v1, v2, node1, node2) => {
```
lbwexler: Could/Should our sort just always use the valueGetter for the column?  Isn't that more appropriate by default?

amcclain: For ag-grid produced aggregate values you need/want it to look at the groupData to get the computed agg.  That is - I think - why it always looks there - `data[field]` won't have that value.

cmp/grid/columns/Column.js
```diff
@@ -190,14 +190,19 @@ export class Column {
...
+        // Record IDs. TreeColumns use those IDs as their cell values, regardless of field.
+        // Add valueGetters below to correct + additional fixes for sorting below.
         if (this.isTreeColumn) {
             ret.showRowGroup = true;
             ret.cellRenderer = 'agGroupCellRenderer';
             ret.cellRendererParams = {
                 suppressCount: true,
                 suppressDoubleClickExpand: true,
-                innerRenderer: (v) => v.data[this.field]
+                innerRenderer: (v) => v.data[field]
```
lbwexler: could definite a valueGetter variable to a single instance of thsi function on line 197, and just set that three times.  Obviously not critical  :-)

cmp/grid/columns/Column.js
amcclain: Here is the ValueService within ag-grid where the value is sourced:
https://github.com/ag-grid/ag-grid/blob/1f82c57f6c0035a0f801a0ef58f05c85d0b3ade3/packages/ag-grid/src/ts/valueService/valueService.ts#L51

Note that is the link to the v18 version we are using now.  Here is the latest version, where they seem to have updated things for exactly this issue!!!

https://github.com/ag-grid/ag-grid/blob/master/packages/ag-grid-community/src/ts/valueService/valueService.ts#L51

Can I get three hours of my life back please?

lbwexler:
Merging this.  We need the fix now, and my comments were trivial.  We'll be unwinding this partially with the next release of agGrid soon anyway.

PR #743, opened 2018-10-25
Panel collapse button

haynesjm42:
Added 3 new properties to PanelSizingModel:
* showSplitter
* showSplitterCollapseButton
* showHeaderCollapseButton

For left and right collapse panels, I am rendering the header collapse button at the top of the panel header when collapsed. This seemed to make the most sense. At all other times the header collapse button is rendered at the end of the panel header, after all panel header items.

See https://github.com/exhi/toolbox/tree/panelCollapseButton branch in toolbox fo a demo of the new props.


desktop/cmp/panel/PanelSizingModel.js
```diff
@@ -28,6 +27,9 @@ export class PanelSizingModel {
     side;
     collapsedRenderMode;
     prefName;
+    showSplitter;
```
lbwexler: I think splitter is probably the right thing to call this.  Could alternatively call it 'resizeBar'.  

One simplification we could do, is just always show the splitter collapse button if the splitter was being shown, and the panel was collapsible.  Or do you think we might want to *not* show it, if we are already showing it in the header?  I think this is all probably right.

lbwexler: Also, one thing to consider is moving these to panel itself?  I think you *could* argue that they might go there?  Not 100% sure about this, and I do realize that 'side' is here on the model

(also realize that implementation-wise, a lot easier to have these on the model)

lbwexler:
I think the code here is pretty fine -- and I like the behavior and the choice of where to place the icons, and the choice of icons (although I will let others weigh in -- another option for icons would be the flat line, open box sort of thing that windows uses)

I do wonder how this dovetails with the title bar actions,  and what these title bar actions really are actually.  It seems like another approach would be to provide a title bar action that does this, and just leave users to use that API (and enhance that API if needed). On that note, I also wonder if we should have just had a single PanelModel, that could have reasonably housed the title bar actions, which are now weirdly *not* in a model.

Sorry to be waffling here, and bringing up some issues that go beyond the focus of this feature.  It may be that what we have here is the best way forward.  

Finally note, that if we go with these icons, we will definitely want to alter that toolbox example to use different core icons for the panels themselves!

desktop/cmp/panel/PanelSizingModel.js
haynesjm42: I had it as `resizeBar` or something similar originally, but realized it actually has nothing to do with resizing so I went with splitter. 

Yes was thinking you might not want the splitter button shown if you have the header button, would probably do that for my client app use-case. Don't feel too strongly about it though so happy to make a change if we want to eliminate a prop to simplify the api.

Also went back and forth on where these props should live, and ultimately decided that it seemed weird for them to be on the panel when the other props like `side` and `collapsible` which control if/how these things are rendered are on the model. Could see us moving the non-state props (`resizable`, `collapsible`, `side`, `collapsedRenderMode`) and these new props to the Panel component and limiting the PanelSizingModel to just maintaining the collapse state and size of the panel, though as you pointed out having everything on the model makes the implementation nice and clean for the child components.

PR #751, opened 2018-10-26
Button - default to minimal, add layoutSupport

### LINKED ISSUE #747: Button: basic sizing props, or layoutSupport
Currently need to use a style block to set a fixed width/height on a button, including specifying dims as px strings. Would like to support at least width/height as top level int props.

Could relate to the work on HoistInputs to make the sizing controls better defined and more regular. 

I wonder if we need a subset of @LayoutSupport that explicitly only handles e.g. sizing and margins.  Or maybe not....perhaps LayoutSupport is fine to use as is, with the caveat that specifying container configs on a button might well be problematic.  Would have the big advantage of also supporting margins out of the box....

amcclain:
+ Fixes #747


amcclain:
As discussed, we could consider extending our new `AppSpec` to take e.g. a `theme` object for style-related props that need to be set in JS.  It could support one `minimalButtons` config to start.

lbwexler:
I love the global reduction of the padding, and love the way the minimal buttons look generally in most of the contexts.

Am not 100% sure about the change of the default.  If we did take the new default, I think we would want to concurrently change our dialogs (alert/confirm/message) to *not* use the minimal style.

I do think the biggest problem with the non-minimal style is on the buttons that have intents -- its the coloring of the entire background there that becomes cartoonish.  Think the intent coloring for the minimal style is just less overwhelming.  I wonder if some more subdued colors for the intents would help in this regard?  Especially in the light theme? Is that a possibility?  Of course that could also be an orthogonal change....

lbwexler:
On a related note, would it be possible to reduce or remove the built-in left margin that's causing spacing between the buttons?  I find the adjacent buttons look gappy, especially in our toolbars.  Can't applications that want that manage that kind of spacing themselves?  For example, look at the text formatting buttons in this very github editor -- they are hard against each other.

Let me know if I am way off-base here -- realize I am out of my normal area of expertise :-)

TomTirapani:
Just adding my 2p:

+ I feel the minimal styling looks great for icon buttons, but doesn't work quite so well for text buttons (particularly ones without an icon). An icon implies interactivity in a way that text does not. I think the border separation provided by the standard button really helps with this for text buttons. For an example, see the 'Show Mask' button in the Toolbox mask example.

+ I think ButtonGroupInput has been negatively impacted by this change. If you consider a case with 2 buttons, darkening the button doesn't make it clear to me which one is selected, compared to the 3D 'pressed' effect provided by non-minimal buttons.

+ I must admit I agree with Lee that a lot of what we are trying to achieve could just be done by using less saturated intents and using them more sparingly.

+ That said, I think there's definitely a place for the minimal buttons, with the App header buttons and Toolbox Standard Grid examples looking significantly better. I must admit I agree with Lee that a lot of what we are trying to achieve could just be done by using less saturated intents and using them more sparingly. I think a considered case-by-case mix of the 2 styles would yield the best results.

shravnk:
I encountered some of the issues that Tom describes with the dimension chooser. Default styling was definitely too heavy for my buttongroup. The ‘pressed’ effect wasn’t needed in my case, but I think the minimal buttons looked much better with a light box shadow. Maybe we could apply custom styling with the ButtonGroup wrapper and CSS?

amcclain:
Cool, thanks for the thoughts.  How do we best proceed here?  A few options:
* The intents are always optional - I often apply one of our color classNames (xh-red, etc) to a button icon instead of using the intent.  But we do already customize the intent styling somewhat in terms of using our color vars, and we would get more into that if someone can take the time to look carefully at the Blueprint implementations and ensure any customizations work across the various button options / states (disabled/pressed/hover x standard/minimal x light/dark theme).
* If we did want to push for minimal as default, we could certainly swap back to non-minimal mode for ButtonGroup (good catch) and our standardized alert/confirm/exception dialogs. I don't think we want to get into cloning button elements in terms of auto-applying to toolbars, but perhaps there are options.
* The padding on the toolbar items could certainly be reduced - I think the padding you're referring to is applied by toolbar to its children in general (not just buttons).

I'd of course be totally up for looking at any alternate ideas. My main hope though is that whatever we come up with is as maintainable and consistent as possible, so we're not constantly patching styles in different contexts, but I liked the revised look enough that I think it's worth working towards,

amcclain:
I'm going to be a bit reckless and merge this so we can give it a go and see how we feel. I spent some more time with it and have continued to prefer the aggressive application of the minimal style, even on dialogs and text buttons.

* I updated to ensure buttonGroupInput applies the non-minimal style by default.
* I generally tightened up the toolbar padding and exposed a dedicated CSS var to customize if needed.
* We can of course add a way to customize this in JS or CSS if an app really wants to opt-out of the default.

PR #754, opened 2018-10-28
Parameterized decorator check

lbwexler:
I went back and forth on whether to throw, or just console.error, but ultimately decided to throw.

If you were to just spell the decorator wrong, (e.g. 'feild') it throws at runtime in the browser, so I thought it made sense to do the same here.


amcclain:
Very cool!  Merging now.

PR #755, opened 2018-10-29
Add MultiField support to Grids

TomTirapani:


cmp/grid/GridModel.js
```diff
@@ -315,6 +317,13 @@ export class GridModel {
         this.compact = compact;
     }
 
+    /** Return the minimum row height specified by the columns (if any) */
+    @computed
```
lbwexler: What's the value of exposing this on GridModel?  Can this very same getter be an implementation bit on the grid?   Can then easily combine it with the other logic around row height that is there.

 We don't give the creator of a GridModel any way to specify this at a gridmodel level in any case

cmp/grid/columns/Column.js
```diff
@@ -230,14 +235,14 @@ export class Column {
         const {renderer, elementRenderer} = this;
         if (renderer) {
             ret.cellRenderer = (agParams) => {
-                return renderer(agParams.value, {record: agParams.data, column: this, agParams});
+                return renderer(agParams.value, {record: agParams.data, column: this, agParams, gridModel});
```
lbwexler: great!  Could also enhance the "metadata" doc strings below to indicate 'column, record, and gridModel'

cmp/grid/impl/SubField.js
```diff
@@ -0,0 +1,27 @@
...
+ * @private
+ */
+export class SubField {
+    colId;
+    label;
+
+    /**
+     * @param {Object} s - SubField configuration.
+     * @param {string} s.colId - Column ID to render.
+     * @param {boolean} [c.label] - true to include label.
```
lbwexler: Could we allow this to be a string, or a boolean?  String would allow you to actually specify a label?  Maybe you want/need something shorter than the column header text, or just different based on the context?

cmp/grid/impl/MultiFieldRenderer.js
```diff
@@ -0,0 +1,105 @@
...
+
+import {SubField} from './SubField';
+
+/**
+ * Renders a collection of additional sub fields in a row beneath the main column field.
+ * Typically not used directly - instead created by using the MultiFieldRenderer elementRendererFn.
+ * @private
+ */
+@HoistComponent
+class MultiFieldRenderer extends Component {
```
lbwexler: Could we rename this to MultiFieldCellComponent or MultiFieldDisplay?  Renderer in the name of the component is confusing.

cmp/grid/impl/MultiFieldRenderer.js
```diff
@@ -0,0 +1,105 @@
...
+
+    static propTypes = {
+        /** Primary value to render */
+        value: PT.oneOfType([PT.string, PT.number, PT.bool]).isRequired,
+        /** The data Record to render. */
+        record: PT.instanceOf(Record).isRequired,
+        /** Reference to the Column */
```
lbwexler: I might just pass the column renderer metadata in as a single bundle, and read the  multiFieldRendererCfg right off the column from inside this component.  Makes the whole interface simpler -- the renderer truly just becomes a pass through.  Then you also have all the metadata available for passing on to the actual renderers on 91, without having to reconstruct -- less fragile

cmp/grid/impl/MultiFieldRenderer.js
```diff
@@ -0,0 +1,105 @@
...
+                label ? span(`${headerName}:`) : null,
+                span(this.renderValue(value, renderer || elementRenderer))
+            ]
+        });
+    }
+
+    renderValue(value, renderer) {
+        const {record, column, agParams, gridModel} = this.props;
+        let ret = value;
+        if (renderer) ret = renderer(value, {record, column, agParams, gridModel});
```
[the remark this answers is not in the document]
TomTirapani: Great catch! Agree its best to pass in the column it is derived from.

TomTirapani:
This branch has been updated to address feedback provided in last review.

cmp/grid/Grid.js
```diff
@@ -154,7 +162,7 @@ export class Grid extends Component {
             frameworkComponents: {agColumnHeader: ColumnHeader},
             rowSelection: model.selModel.mode,
             rowDeselection: true,
-            getRowHeight: () => model.compact ? Grid.COMPACT_ROW_HEIGHT : Grid.ROW_HEIGHT,
+            getRowHeight: () => Math.max(this.columnRowHeight, model.compact ? Grid.COMPACT_ROW_HEIGHT : Grid.ROW_HEIGHT),
```
lbwexler: would work this extra bit of max logic in to the computed header -- just cleaner to do all the rowheight logic in one place

cmp/grid/renderers/MultiFieldRenderer.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+
+import {multiFieldCell} from '../impl/MultiFieldCell';
+
+/**
+ * A grid elementRendererFn that renders a collection of additional sub fields in a row beneath the main column field.
+ *
+ * Requires the column to also specify a multiFieldRendererCfg, with the following params:
```
lbwexler: Describe what the SubFields are in *this* public file.  Don't want to refer user to that private impl class...  (which we may want to remove anyway, as mentioned)

Issue #741, opened 2018-10-24
Grid in TreeMode should not indent when no nodes have children


lbwexler:
We have similar special handling in hoist-sencha.

PR #756, opened 2018-10-29
Grid in TreeMode should not indent when no nodes have children. #741

TomTirapani:


data/BaseStore.js
```diff
@@ -56,6 +56,12 @@ export class BaseStore {
     /** Is this store empty before filters have been applied? */
     get allEmpty() {return this.allCount === 0}
 
+    /** Do any post-filtered records have children? */
```
lbwexler: I would lose these getters for now.  The names don't seem intuitive to me (might have used the word 'is') and they don't add a lot of value.  Just do the 'some' check in the grid.  

Also, think we should just always look at pre-filtered records.  Seems weird that the indentation might change just by tweaking a local filter (e.g. StoreFilterField)

cmp/grid/Grid.js
```diff
@@ -112,6 +112,7 @@ export class Grid extends Component {
                     'ag-grid-holder',
                     XH.darkTheme ? 'ag-theme-balham-dark' : 'ag-theme-balham',
                     compact ? 'xh-grid-compact' : 'xh-grid-standard',
```
lbwexler: See the comment on line 104.  Consulting the stores 'records' property here in render (via this tree getter, or not, see comment below) seems to go directly against this comment, and might yield that 'extra rendering and jumpiness' mentioned.  Perhaps it makes more sense to simply have a private 'isHierarchical' boolean flag on the grid that can be toggled in the dataReaction, and then used here? It can be observable if necessary, but it should be more far more stable than the records property itself.   Alternatively, the getter on the store could be made 'computed' but I think better not to have that getter (see my comment below).

cmp/grid/ag-grid/styles.scss
```diff
@@ -74,17 +74,23 @@
     }
   }
 
-  // Generate indentations for tree grids.
-  // Based upon https://github.com/ag-grid/ag-grid/blob/master/packages/ag-grid-community/src/styles/ag-grid.scss#L1018
-  $xh-tree-indent: 10px !default;
-  @for $i from 1 to 50 {
-    .ag-ltr .ag-row-group-indent-#{$i} {
-      padding-left: $i * $xh-tree-indent;
+  &.xh-grid-tree-mode {
```
lbwexler: *totally* minor tweak, but can this style be renamed to 'xh-grid-hierarchical' or 'xh-grid-indent-mode' or something like that?

I would expect from the current name, that these are styles that are always present when the model is treeMode= true  and that is not the case.

PR #763, opened 2018-10-30
Support larger excel exports

TomTirapani:


cmp/grid/impl/ExportManager.js
```diff
@@ -64,14 +64,15 @@ export class ExportManager {
             });
         }
 
+        const formData = new FormData(),
+            params = JSON.stringify({filename, type, meta, rows: rows.slice(0, 10)});
```
lbwexler: is the slice here a testing thing?

Issue #759, opened 2018-10-29
Expose Grid ExportManager as a service


amcclain:
Most if not all of the code within the current `grid/impl/ExportManager` class could be exposed as a `DataExportService`. 

This would have the advantage of using the service pattern for this kind of transform/post/fetch data task, and could also allow us to work on a more general API to POST data from sources other than a grid model to the server-side endpoints for Excel spreadsheet generation.

We currently have one client app that's assembling this data package manually at the moment, duplicating some of the logic within the impl class. It's a powerful feature to be able to further leverage our server-side export code to generate Excel sheets in a standard way, and we want to expose it consistently and with a maintainable API.

lbwexler:
Hi:  Just want to say that I definitely agree with this.  I wonder whether it might be more pragmatic  to create a 'GridExportService' that is GridModel aware, rather than trying to abstract into the more general DataExportService.  Its all pretty grid specific, and could be made to work for DataViews as well, but don't know that it needs to be more general than that.

amcclain:
We have this one use case (which could continue to be handled as a one-off) where we have a set of data that doesn't need to be / is not shown in a grid at all but is fetched on demand by the client for the express purpose of re-exporting it out to Excel through the Grails layer.  (I.e. the available input is a JSON API and the desired output is an Excel table).

Just occurred to me though....I don't think there's any reason why we couldn't construct a "headless" GridModel - i.e. just the model no grid component - and use the standard code paths from there.  Looking quickly at the existing GridModel and ExportManager classes, I don't *think* they depend on reading metadata or making API calls from an instantiated ag-grid.  We could give it a try.

Again, not to let this one (rather unusual) use-case drive the discussion here. Main goal is to have a service vs. the somewhat-ambiguous kind-of-class that is ExportManager.

lbwexler:
Think this is a great idea.

TomTirapani:
This issue has been folding into the related issue #701 in the following PR:
https://github.com/exhi/hoist-react/pull/764

PR #764, opened 2018-10-31
Enhance configurability of grid exports

TomTirapani:
+ Export supports 'includeHiddenCols' option (false if not set).
+ Export functions take map of export options.
+ Export buttons support map of export options.
+ GridModels can define map of export options.


cmp/grid/GridModel.js
```diff
@@ -171,13 +171,10 @@ export class GridModel {
     /**
      * Export grid data using Hoist's server-side export.
      *
-     * @param {Object} options
-     * @param {string} options.type - type of export - one of ['excel', 'excelTable', 'csv'].
-     * @param {(string|function)} [options.filename] - name for exported file or closure to generate.
+     * @param {Object} options - Export options. See ExportManager.exportAsync() for options.
      */
     export(options = {}) {
```
lbwexler: Any reason not to have this method be exportAsync as well and return the underlying promise?

cmp/grid/GridModel.js
```diff
@@ -171,13 +171,10 @@ export class GridModel {
     /**
      * Export grid data using Hoist's server-side export.
      *
-     * @param {Object} options
-     * @param {string} options.type - type of export - one of ['excel', 'excelTable', 'csv'].
-     * @param {(string|function)} [options.filename] - name for exported file or closure to generate.
+     * @param {Object} options - Export options. See ExportManager.exportAsync() for options.
```
lbwexler: A bit problematic that ExportManager is marked as a *private* class.  If that's how we want to handle it for now, we should move the documentation of these options here.  

Other option is to go ahead and promote ExportManager to something like GridExportService.  It would be a little unusual having a service that has a dependency on the GridModel, but perhaps appropriate in this case?  This of course relates to #759

TomTirapani:
I've updated this PR to include #759 as well.

TomTirapani:
Also, please not corresponding Toolbox changes: https://github.com/exhi/toolbox/pull/103

svc/GridExportService.js
```diff
@@ -16,28 +16,32 @@ import {isFunction, isNil, isString, orderBy, uniq} from 'lodash';
...
+     * @param {(string|function)} [options.filename] - name for exported file or closure to generate.
+     * @param {string} [options.type] - type of export - one of ['excel', 'excelTable', 'csv'].
+     * @param {boolean} [options.includeHiddenCols] - include hidden grid columns in the export.
      */
-    async exportAsync(gridModel, filename, type) {
+    async exportAsync(gridModel, {
+        filename = 'export',
+        type = 'excelTable',
+        includeHiddenCols = false
+    }) {
```
lbwexler: need to allow caller not to specify second argument at all with = {}

PR #766, opened 2018-10-31
Clone the pref value on set for the internal cache and event params (Fixes #765)

### LINKED ISSUE #765: PrefService.set() issues with Object/Array prefs
PrefService.set() is updating the cached data with the value passed to it, which could be an Object or Array, and could be the same reference which the App is holding onto. If the App modifies the Object properties or Array contents without cloning, then it is also modifying PrefService's cached copy of the value, and calls to PrefService.set() will never pass the `isEqual` check and thus never be updated on the server. This could also cause some interesting bugs with the deferred push to the server after a set, where the value persisted is different from what was passed to PrefService.set() since the App could have modified the value which the PrefService updates map is holding a reference to after calling set().

To be consistent with PrefService.get(), PrefService.set() should clone the value before updating the internal cache, and also pass a clone of the value to the `prefChange` event so Apps can be guaranteed to always be given a copy of the pref value.

haynesjm42:
#765 

Make sure that we don't hold a reference to the value passed to PrefService.set()

See https://github.com/exhi/toolbox/commit/4b8f805e5a86ff0c5f9489bbe5ab06656fcf9d3a for a demo of the issue if curious (not intended to be merged, will delete after this is pr merged)

### DIFF
--- a/svc/PrefService.js
+++ b/svc/PrefService.js
@@ -94,11 +94,11 @@ export class PrefService {
         if (isEqual(oldValue, value)) return;
 
         // Change local value and fire.
-        this._data[key].value = value;
-        this.fireEvent('prefChange', {key, value, oldValue});
+        this._data[key].value = cloneDeep(value);
+        this.fireEvent('prefChange', {key, value: cloneDeep(value), oldValue});
 
         // Schedule serialization to storage
-        this._updates[key] = value;
+        this._updates[key] = this._data[key].value;
         this.pushPendingBuffered();
     }
 
@@ -230,7 +230,6 @@ export class PrefService {
 
     validateBeforeSet(key, value) {
         const pref = this._data[key];
-
 
         throwIf(!pref, `Cannot set preference ${key}: not found`);
 


svc/PrefService.js
```diff
@@ -94,11 +94,11 @@ export class PrefService {
         if (isEqual(oldValue, value)) return;
 
         // Change local value and fire.
-        this._data[key].value = value;
-        this.fireEvent('prefChange', {key, value, oldValue});
+        this._data[key].value = cloneDeep(value);
```
lbwexler: + Would it make sense to "freeze" the right hand side here (after cloning) and also where we load the data from server and local storage below?    Could add a utility method to set a key value to a frozen object?

+ Seems like the second clone on 98 should not be needed (and we'll feel especially confident about that if we use freeze).

+ If we freeze, we could consider *not* cloning what comes out of get() and leaving that to applications.  That would be my preference, but that would be an API change....

haynesjm42: * Yes I think this is a great place to use freeze
* I do another clone on what is passed in the event params because I would consider this to be the same as a call to PrefService.get(), and whoever is subscribing would potentially want to grab that reference and start modifying the same way they would with the value returned from get()
* I think it makes sense from an API perspective that PrefService has an immutable store of preference values. My use case where I ran into this is probably not too common (especially considering this issue wasn't caught before now) where my pref is a json object where the keys are user-entered values. Usually when I have a json pref the structure is known, and I end up destructuring when getting the values out of the pref. Arrays may be a bit different, would expect a lot of the time to want code like `this.values = XH.getPref('myArrayPref');`, though often the array prop would be `@observable.ref` on a model and treated as immutable anyways. Perhaps we just provide a getClone() method on PrefService for convenience for those special cases?

lbwexler: Got it!  Re: the second clone().  Yes, makes sense, given the current expectation that apps get a clone made for them in get.

But having said that, what would people think of having PrefService just return frozen things?  Seems like a poster child for when you want to enforce immutability and freeze, and if we want our applications to start working in this way, what better way to start?

lbwexler:
Please see https://github.com/exhi/hoist-react/pull/768  for an alternative approach to this problem

lbwexler:
We merged 768 instead.  Thanks for uncovering this issue John.

PR #767, opened 2018-10-31
Special exception handling for EnvironmentService init

### LINKED ISSUE #758: Improve initial exception to user when grails server completely down
Due to recent refactorings, we now get the following obscure message to :  "Failure to initialize EnvironmentService;

amcclain:
+ Provides a clearer exception when Grails server is down or unreachable vs. generic service failure.
+ Fixes #758

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -390,7 +390,15 @@ class XHClass {
 
         try {
             await this.installServicesAsync(FetchService, LocalStorageService);
-            await this.installServicesAsync(EnvironmentService, TrackService, IdleService);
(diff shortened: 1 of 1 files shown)


amcclain:
Not sure if this is the approach you were thinking we could take....  

At first I kept envService in the second init with track/idle and wrapped those first two awaits in a try/catch, but then it seemed better to be more explicit in the code about exactly what we were aiming to catch here.

lbwexler:
Tried to think about other ways to handle this more generally but think this is really the most pragmatic.  Thank You!

PR #771, opened 2018-11-01
CheckBox and Switch Bind to Boolean Only

shravnk:
- handle boolean as string in restcontrolmodel => get value()
- remove boolean handler from switch, checkbox


desktop/cmp/rest/impl/RestControlModel.js
```diff
@@ -39,8 +39,8 @@ export class RestControlModel  {
 
     @computed
     get value()  {
```
lbwexler: If we are going to handle this here, would like to generalize this from handleBoolean to parseStringValue.  why are we handling just "false" => false? Why not "true" to true?  or "3" to 3.?

lbwexler:
Hi Brendan -- Thanks for this.  I reviewed carefully and think it was on the right track.  But thinking about it some more, I think its cleaner to handle on the server.  Can you review this (also gave to atm)

https://github.com/exhi/hoist-react/pull/772
https://github.com/exhi/hoist-core/pull/59

PR #776, opened 2018-11-05
Avoid mutating lookup results when adding 'null' in RestControl Select inputs. 

### LINKED ISSUE #775: Duplicate -null- entries in Log Level RestForm Select
Each time the LogLevel RestForm is opened, an extra `-null-` entry is added to the Select field dropdown options. It's not immediately clear if this is an issue with RestForm, Select or RestController lookups.

First:
screenshot 2018-11-05 at 12 54 36

Second:
screenshot 2018-11-05 at 12 54 48
etc.

TomTirapani:
Fixes #775

### DIFF
--- a/desktop/cmp/rest/impl/RestControl.js
+++ b/desktop/cmp/rest/impl/RestControl.js
@@ -93,7 +93,7 @@ export class RestControl extends Component {
 
         let options;
         if (field.lookup) {
-            options = field.lookup;
+            options = [...field.lookup];
         } else if (type == 'bool') {
             options = [true, false];
         } else {


lbwexler:
Wondering if the problem is not in our new select?  Seems like *it* should not be mutating the options presented to it.  Imaging if you wanted to pass the same options to a couple of select controls -- you want null added to those options twice...

TomTirapani:
I may be misunderstanding you, but I don't think the select is mutating its options. The problem is we're mutating `fields.lookup` on line 104, where we're unshifting a null onto the the front the array before passing to select. Each time renderSelect() is run, `fields.lookup` gets a new null until the lookups are refreshed.

lbwexler:
Sorry to malign your change, and select() in one fell swoop!  (Checked select, and it is indeed fine as well)

Think I need to get more context in my diffs when doing these reviews -- thanks for this one.

Issue #492, opened 2018-08-02
Toolbars should specify their size


amcclain:
Pretty sure that toolbars should specify a size - or some constraints - possible via CSS.

amcclain:
Relates to #541

PR #779, opened 2018-11-05
Toolbar css

shravnk:
For #492


amcclain:
Hard to argue with this, thanks.  If we need to make it more configurable somehow, etc. we can cross that bridge when we come to it, but as a first pass I was looking for some general sizing constraint, so - thanks!

PR #790, opened 2018-11-09
Dynamic imports

lbwexler:
Remove bleed of desktop classes in to mobile App  …
+ Support for dynamic platform imports
+ Avoid multiple imports of XH.
+ Freeze App Spec
+ Move ColChooser Button to desktop buttons package


lbwexler:
Thought about a couple of ways to handle this -- this seemed like the way that was the most straightforward, and did the least violence to the code.  

The one API change here is the move of ColChooserButton to the buttons package.

TomTirapani:
Wow, I had no idea this was even possible! Very nice and elegant :)

lbwexler:
Adding comments now -- wanted to make sure we were ok with the approach.
After much thought, I decided *not* to make it a formal, documented part of
the Hoist API for client use (e.g. XH.addDynamicImport()).  Just seemed
like a better way to do it may come along (dynamic imports is one proposed
js feature), and as we have no application need, better to avoid committing
to this in the API.

lbwexler:
Also, thanks for making the toolbox change -- had that in my sandbox -- had
to runout, and never pushed/PR it.....

PR #792, opened 2018-11-09
Upgrade ag-Grid to 19.1.2

amcclain:


amcclain:
Now that they've released this service-pack update, my hope is that we can take the latest ag-Grid. @haynesjm42 I wanted to check with you in particular before merging, as your active client work has some of the most specific implementations against AG APIs. @TomTirapani we would need to adjust imports and (??) any other relevant code in our active apps with ag-grid based filters, etc.

amcclain:
FYI @nadkarniarjun that I've set this up here to see if we can press go on the upgrade.

haynesjm42:
Unfortunately they made a change in 19.1.1 which is a breaking change for us and how we are managing ag-grid column state. They introduced delta column definition updates which means calls to `gridApi.setColumnDefs()` will no longer update the state of the columns (width, order, etc.). You can see this issue just by trying to use our column chooser on 19.1.1.

I updated to 19.1.1 on my branch where I am adding support for ag-grid pivot mode and have changes to use `columnApi.setColumnState()` (among other ag api methods) to handle the state updates. Part of the pivot mode support was also supporting the built-in ag-grid tool panel components, and updating the column defs can cause weird behavior there anyways, so it made sense to explore alternate approaches. Hoping to have a PR ready for those changes by the end of the day today..

haynesjm42:
Made changes to use the ag-grid Column API to set the column state (via `setColumnState`) instead of re-setting the column definitions (which no longer works in 19.x).

* Separate column state and column defs in GridModel, react to column state changes in Grid component
** Also required changing how the sort comparator works for columns so that the sort config is retrieved when the comparator is called instead of when building the ag column spec
* Support groupBy in GridStateModel
* React to ag-grid column grouping and visibility events and update our GridModel state, in order to support the ag-grid tool panel and grouping panel

cmp/grid/Grid.js
```diff
@@ -407,6 +454,20 @@ export class Grid extends Component {
         }
     };
 
+    // Catches row group changes triggered from ag-grid ui components
+    onColumnRowGroupChanged = (ev) => {
+        if (ev.source !== 'api' && ev.source !== 'uiColumnDragged') {
```
haynesjm42: This event (as well as columnVisible) will get fired when the user is dragging the column around in the ag-grid panels, so we want to ignore if source is uiColumnDragged and let the onDragEnd handler update our model once the change has been finished

cmp/grid/columns/Column.js
```diff
@@ -187,7 +186,15 @@ export class Column {
                 lockPinned: true, // Block user-driven pinning/unpinning - https://github.com/exhi/hoist-react/issues/687
                 pinned: this.pinned,
                 lockVisible: !gridModel.colChooserModel,
-                headerComponentParams: {gridModel, column: this}
+                headerComponentParams: {gridModel, hoistColumn: this},
```
haynesjm42: These `headerComponentParams` will actually get merged onto the params which ag-grid puts together for the header component, which already includes a `column` prop for the ag-grid Column object. Us adding our column here was causing it to be merged into the ag-grid Column object. Caused a lot of issues with the secondary columns created during pivoting, which isn't supported yet in this PR, but wanted to get this fix in asap.

Note the related changes in the ColumnHeader component.

cmp/grid/ag-grid/ColumnHeader.js
```diff
@@ -37,6 +38,7 @@ export class ColumnHeader extends Component {
     // Get any active sortBy for this column, or null
     @computed
     get activeGridSorter() {
+        if (!this.gridModel) return;
```
haynesjm42: Need this check for the auto group col, which will get this header component but wont have a gridModel. Will explore better support in general for the auto group col as part of the pivot mode support.

cmp/grid/GridModel.js
```diff
@@ -352,14 +346,15 @@ export class GridModel {
...
+            if (!col) return null;
             return {
                 colId,
                 hidden: hide,
                 width: col.flex ? undefined : width
             };
         });
 
-        this.applyColumnChanges(colChanges);
+        this.applyColumnChanges(colChanges.filter(Boolean));
```
lbwexler: I find this filter statement an awkward way to remove nulls.  Doesn't lodash have a better way to remove nulls?

cmp/grid/GridModel.js
```diff
@@ -375,15 +370,14 @@ export class GridModel {
      */
     @action
     applyColumnChanges(colChanges) {
```
lbwexler: We discussed renaming this to applyColumnStateChanges

cmp/grid/GridStateModel.js
```diff
@@ -114,9 +122,9 @@ export class GridStateModel {
     columnReaction() {
         const {gridModel} = this;
         return {
-            track: () => gridModel.columns,
+            track: () => gridModel.columnState,
```
lbwexler: Nice!

cmp/grid/ag-grid/ColumnHeader.js
lbwexler: super abbreviated form of this comment in code?

cmp/grid/columns/Column.js
lbwexler: really nice fix.  just wonder if we want to use xhColumn -- has a nice symmetry with agColumn.

cmp/grid/columns/Column.js
```diff
@@ -239,26 +247,34 @@ export class Column {
...
-            if (this.isTreeColumn) {
-                ret.comparator = (v1, v2, node1, node2) => {
-                    return sortCfg.comparator(node1.data[field], node2.data[field]);
-                };
-            } else {
-                ret.comparator = (v1, v2) => sortCfg.comparator(v1, v2);
-            }
+            ret.comparator = (v1, v2, node1, node2) => {
+                const sortCfg = find(gridModel.sortBy, {colId: ret.colId});
+                if (sortCfg) return sortCfg.comparator(node1.data[field], node2.data[field]);
```
lbwexler: Think this code might be easier to reason about if it were a single comparator, that just reset its v1 and v2 at the top, if it was a TreeColumn

cmp/grid/GridModel.js
haynesjm42: `without(colChanges, null)`?

lbwexler:
Absolutely -- I assumed we were just going to close that one.  Let's take
those changes and close out that PR

desktop/cmp/grid/impl/ColChooserModel.js
```diff
@@ -60,7 +60,7 @@ export class ColChooserModel {
             }
         });
 
-        gridModel.applyColumnChanges(colChanges);
+        gridModel.applyColumnStateChanges(colChanges);
```
amcclain: This doesn't relate directly to this change, but would there be any value in adjusting this method to only add columns to `colChanges` if the visibility changed? It seems like that would be more appropriate, and potentially avoid triggering the additional logic within gridModel related to sort order.

cmp/grid/columns/Column.js
```diff
@@ -239,31 +245,37 @@ export class Column {
...
+
+    //--------------------
+    // Implementation
+    //--------------------
+    comparator = (v1, v2, node1, node2) => {
+        const sortCfg = find(this.gridModel.sortBy, {colId: this.colId}),
+            // ag-Grid sort impl. sources its primary values from the node's `groupData` property,
+            // which is not what we want when sorting treeColumns.
+            val1 = this.isTreeColumn ? node1.data[this.field] : v1,
+            val2 = this.isTreeColumn ? node2.data[this.field] : v2;
```
amcclain: Could we test if the tree-related switching here is still required? I think that ag-Grid made some updates in 19.1 to respect the valueGetter for sorting, which we install on line 209 above.

cmp/grid/GridModel.js
```diff
@@ -350,66 +344,67 @@ export class GridModel {
...
-     * @param {ColumnState[]} colChanges - changes to apply to the columns. If all leaf columns are
+     * @param {ColumnState[]} colStateChanges - changes to apply to the columns. If all leaf columns are
      *      represented in these changes then the sort order will be applied as well.
      */
     @action
-    applyColumnChanges(colChanges) {
-        let {columns} = this,
-            newCols = [...columns];
+    applyColumnStateChanges(colStateChanges) {
+        let columnState = [...this.columnState];
```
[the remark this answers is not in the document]
haynesjm42: Yes we are only cloning here in order to trigger the reaction.

Agree that a deep clone is probably more appropriate, especially now that we are dealing with the simple column state objects instead of Columns

cmp/grid/GridModel.js
```diff
@@ -350,66 +344,67 @@ export class GridModel {
...
 
-        throwIf(colChanges.some(({colId}) => !this.findColumn(columns, colId)),
+        throwIf(colStateChanges.some(({colId}) => !this.findColumn(columnState, colId)),
```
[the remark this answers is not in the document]
haynesjm42: We should never hit this from any of our code, we filter out unknown/stale columns before calling this method from places where we know we can/will run into invalid columns.

This is the only public api for Apps to modify column state programmatically, so as you pointed out this is just to make sure the expectations are clear and programming errors in the api usage are immediately obvious to the developer.

cmp/grid/GridStateModel.js
```diff
@@ -170,6 +178,26 @@ export class GridStateModel {
...
+            run: (groupBy) => {
+                this.state.groupBy = groupBy;
+                this.saveStateChange();
+            }
+        };
+    }
+
+    updateGridGroupBy() {
+        const {gridModel, state} = this;
+        if (!state.groupBy) return;
```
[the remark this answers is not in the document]
haynesjm42: Yes makes sense, and the sorting should probably work the same way, good catch

PR #793, opened 2018-11-09
Improve sort performance

shravnk:
- remove sortBy from columnsReaction()
- column gets fixed comparator() which reacts to gridModel.sortBy


cmp/grid/columns/Column.js
```diff
@@ -243,27 +243,26 @@ export class Column {
...
-                    return sortCfg.comparator(node1.data[field], node2.data[field]);
-                };
-            } else {
-                ret.comparator = (v1, v2) => sortCfg.comparator(v1, v2);
-            }
+            val1 = this.isTreeColumn ? node1.data[this.field] : v1,
+            val2 = this.isTreeColumn ? node2.data[this.field] : v2;
 
-        }
+        return sortCfg.comparator(val1, val2);
```
amcclain: How are we avoiding possible NPEs here if there is no sortCfg returned on line 257?  Is it that the comparator won't be called generally if the column isn't in the sortBy? (Assume that's the case.)

I don't want to add needlessly defensive code here, but want to ensure I understand how it's working and see if there's a way we can be a bit more explicit about it.

TomTirapani: Good catch. Not expecting this code to be called without a sortCfg, but probably pragmatic to fall back to a default sort (and warn?) if not found.

TomTirapani:
FYI, noticed this change is included in @haynesjm42 's agGrid upgrade changes: https://github.com/exhi/hoist-react/pull/792

lbwexler:
closing this, as John has incorporated in to his broader changes.  Thanks for the great work on this guys.

PR #802, opened 2018-11-16
Column comparator falls back to default comparator.

TomTirapani:
Column.comparator is used by ag-grid's column filters to sort the available options in a set filter - we should fall back to default comparator if we don't have a sorter in the GridModel.


amcclain:
👍 Thanks again for the fix here.

PR #806, opened 2018-11-18
Improve IE 11 browser lockout, allow MS Edge

cnrudd:


lbwexler:
Merging this, as it seems like this is a much more accurate representation of our limitations, than the promise based test, which did not account for the fact that Babel is now providing Promise Polyfills.

Also seems like the critical step toward allowing us to actually try Edge and shake it out.  

We still need to allow some kind of app controllable lockout at some point, but this is a great improvement.

lbwexler:
I made a small tweak to update the comment.

PR #810, opened 2018-11-20
border-box styling for mobile

shravnk:
We have relied on blueprint css to default all DOM elements to `box-sizing: border-box`. See [here](https://github.com/palantir/blueprint/blob/4e93d2f5d8965aa2864759b663c70640b589ecd0/packages/core/src/_reset.scss#L9).

Since we're no longer applying blueprint styling to the mobile app, we need to set this property ourselves.


lbwexler:
Looks right to me, based on my research, but I have asked tom to look as well...

TomTirapani:
Looks good to me - will merge

PR #817, opened 2018-11-27
Mobile dimension chooser dialog

shravnk:


shravnk:
Had to tweak new mobile select component... mobile dialogs have a z-index of 20001, so I simply set the z-index of the select menu portal to 20002. Could be worth allowing zIndex as a prop on mobile select.

mobile/cmp/dimensionChooser/DimensionChooser.js
```diff
@@ -0,0 +1,197 @@
...
+import {size, isEmpty} from 'lodash';
+
+import './DimensionChooser.scss';
+import {DimensionChooserModel} from '@xh/hoist/desktop/cmp/dimensionchooser';
+import {withDefault} from '@xh/hoist/utils/js';
+
+
+/**
+ * Menu Component
+ */
```
lbwexler: Proper Comment here?

mobile/cmp/dimensionChooser/DimensionChooser.js
```diff
@@ -0,0 +1,197 @@
...
+import {div} from '@xh/hoist/cmp/layout';
+import {button} from '@xh/hoist/mobile/cmp/button';
+import {dialog} from '@xh/hoist/mobile/cmp/dialog';
+import {Icon} from '@xh/hoist/icon';
+import {select} from '@xh/hoist/mobile/cmp/form';
+
+import {size, isEmpty} from 'lodash';
+
+import './DimensionChooser.scss';
+import {DimensionChooserModel} from '@xh/hoist/desktop/cmp/dimensionchooser';
```
lbwexler: This would have to be moved to 'all'.  We'll never want the mobile package importing from the desktop package.

interesting pattern here, where we are able to share a model cross-platform, but not the view.  I think we have seen this elsewhere, and can expect to see it again.  It ends up being a bit strange for the application, which needs to import the two different bits of the component from different locations.  I wonder if a re-export of the model in the mobile  is in order.  

Alternatively, we could look to make the Component cross-platform.  How different are they?

mobile/cmp/form/inputs/Select.js
```diff
@@ -207,10 +207,10 @@ export class Select extends HoistInput {
 
     getOrCreatePortalDiv() {
         let portal = document.getElementById('xh-select-input-portal');
-
         if (!portal) {
             portal = document.createElement('div');
             portal.id = 'xh-select-input-portal';
+            portal.style.zIndex = '20002';
```
lbwexler: @TomTirapani  - any thoughts on how to assign these zIndices?

mobile/cmp/dimensionChooser/DimensionChooser.js
shravnk: Yes, I think we're partly just lucky that the desktop patterns fit neatly into the mobile dialog api. The main render methods are nearly identical. However I think there are enough context-specific helper methods/components that making it cross-platform would be confusing.

lbwexler:
Hi Tom -- I'll merge this now.  Think we need to play with a few patterns here on how we are going to share cross-platform code.  Shared models with separate components might be a recurring model, and eager to see how it plays out.

It seems like this is a good incremental change, and does not really further commit us to any one way of doing things....

lbwexler:
Hi Tom -- I'll merge this now.  Think we need to play with a few patterns here on how we are going to share cross-platform code.  Shared models with separate components might be a recurring model, and eager to see how it plays out.

It seems like this is a good incremental change, and does not really further commit us to any one way of doing things....

PR #818, opened 2018-11-27
Options panel and app bar menu

TomTirapani:


TomTirapani:
This branch has been updated to work with (and simplified by) the latest Forms model

cmp/form/validation/Rule.js
```diff
@@ -59,6 +59,7 @@ export class Rule {
     // Implementation
     //------------------------------
     isActive(field) {
+        if (!field.formModel) return;
```
lbwexler: This should explicitly return false

core/XH.js
```diff
@@ -327,6 +327,11 @@ class XHClass {
     //---------------------------
     // Miscellaneous
     //---------------------------
+    /** Show a dialog to for users to set app options. */
```
lbwexler: comment typo

core/appcontainer/OptionsDialogModel.js
```diff
@@ -0,0 +1,152 @@
...
+    @computed
+    get hasOptions() {
+        return !!this.options.length;
+    }
+
+    //-------------------
+    // Change management
+    //-------------------
+    @action
+    refreshChangedFields() {
```
lbwexler: I don't get the need for the extra state in the changes array.  Couldn't we just walk through the fields, both for the computed "requiresRefresh" and saveAsync()?

desktop/cmp/appbar/AppBar.js
```diff
@@ -47,6 +47,9 @@ export class AppBar extends Component {
          */
         hideAdminButton: PT.bool,
```
lbwexler: Can we take advantage of this change to combine the admin button/feedback button/theme button/options button within a single framework menu of a single button?  Think its too many framework buttons that apps will rarely want to dedicate their prime real estate to.

lbwexler:
Think this is going to give us some great consistency, and be a huge improvement for apps.
Some comments inline: 

+ Do getValue() and setValue() need to be getValueAsync() and setValueAsync()?

lbwexler:
I am ok with this change and would like to get it in.  It actually has a couple of form fixes that are important to take. @amcclain -- thoughts?

My one concern is that I would really like to combine all of the app bar buttons in this version:
        hideAdminButton: PT.bool,
        hideOptionsButton: PT.bool,
        hideFeedbackButton: PT.bool,
       hideThemeButton: PT.bool
       hideThemeButton: PT.bool

into a single "Hoist Menu", equivalent to Chromes three button menu.  

It can include menus items for all of these, and we can generally be more consistent across desktop and mobile, where we definitely don't have room for all these buttons. 

Note:   *Apps that wish to add individual buttons can still do so -- they are all part of the public API*

TomTirapani:
@lbwexler I just updated this branch to move the various app bar buttons into a 'App Menu'.

lbwexler:
Made some additional changes to harmonize with how we are specifying forms in RestGrid

Removing WIP as we are getting ready for merge

lbwexler:
I think the options panel itself is great and ready to be merged now

Want to get @atm thoughts on moving all the toolbar items into a single menu.  

Also wonder if there is an easier way to mix and match the items and also combine them with application logic?  Maybe appModel should have a method 'getAppMenuItems(){}, which can be specified to make the hamburger menu?  And a set of constants somewhere (on the appContainer?) which define the available built-in partials?

amcclain:
I will take a look.  Mainly would really like to favor consistency so key actions (especially launch admin) are readily accessible across apps in the same place.  We tried the hamburger menu recently in a desktop app and I was surprised at how much it threw off my muscle memory and expectations vs. all other apps.  

In general, I think it's a good differentiation between desktop and mobile that desktop gives you quicker access to more functions since you have so much more space available.

But I will look of course.  I was thinking that one advantage of the options panel is that the theme switching had a more natural place to live, so I was hoping that would effectively be replaced by the options button.

lbwexler:
Got it -- We can certainly roll back the AppBar changes for future consideration, and simply add a new button for this to AppBar.  Just want to get it merged....

amcclain:
> I will take a look. Mainly would really like to favor consistency so key actions (especially launch admin) are readily accessible across apps in the same place. We tried the hamburger menu recently in a desktop app and I was surprised at how much it threw off my muscle memory and expectations vs. all other apps.

and for the record I decided to change my mind about this 😛

PR #822, opened 2018-11-28
Store filter field no store fix

### LINKED ISSUE #821: StoreFilterField includeFields ignored if no store is associated with the component
If you don't provide a `store` or `gridModel` prop then the `includeFields` value will always end up being ignored when building the list of active fields if there is no store associated with the component.

This effectively makes it impossible to use the StoreFilterField if you are also providing your own filter function to the store.

haynesjm42:
Fixes #821 

Make sure that includeFields is used if there is no store associated with the component.

Note that in Toolbox, the News grid had a bug where the store filter field would overwrite the source filter, because a store was being provided. Removing that (https://github.com/exhi/toolbox/commit/4140f328f96b565f51b88a1e15a6629d6941845b) reproduces the bug this pr fixes.

### DIFF
--- a/desktop/cmp/store/StoreFilterField.js
+++ b/desktop/cmp/store/StoreFilterField.js
@@ -15,7 +15,7 @@ import {textInput} from '@xh/hoist/desktop/cmp/form';
 import {GridModel} from '@xh/hoist/cmp/grid';
 import {Icon} from '@xh/hoist/icon';
 import {BaseStore} from '@xh/hoist/data';
-import {withDefault, throwIf} from '@xh/hoist/utils/js';
+import {withDefault, throwIf, warnIf} from '@xh/hoist/utils/js';
 
 /**
  * A text input Component that generates a filter function based on simple word-boundary matching of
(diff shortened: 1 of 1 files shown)


desktop/cmp/store/StoreFilterField.js
```diff
@@ -88,6 +88,7 @@ export class StoreFilterField extends Component {
 
         throwIf(props.gridModel && props.store, "Cannot specify both 'gridModel' and 'store' props.");
         throwIf(props.includeFields && props.excludeFields, "Cannot specify both 'includeFields' and 'excludeFields' props.");
+        warnIf(!props.gridModel && !props.store && isEmpty(props.includeFields), "Must specify one of 'gridModel', 'store', or 'includeFields' or the filter will be a no-op");
```
haynesjm42: Had this as a throw initially, but that would mean you can't just add an empty StoreFilterField, which I often find useful for setting up initial layouts/mocks, so changed to a warning

lbwexler:
👍

PR #827, opened 2018-11-29
Number input parses in toExternal method

febbraiod:


desktop/cmp/form/inputs/NumberInput.js
```diff
@@ -157,6 +160,8 @@ export class NumberInput extends HoistInput {
     }
 
     parseValue(value) {
+        if (isNumber(value)) return value;
```
febbraiod: If value is committed via the enter key the external and internal vals will both be a number. This method gets called again onBlur via hoist input's logic. We don't need to parse and the next line would throw.

febbraiod:
A small wrinkle I see is that if one were to commit a value like 1m and we rightly set the external val to 1000000, if you reenter the field and type '1m' again and hit 'enter' the input does not get rerendered with 1000000 as you might expect. This is because the doCommit method will not set the value on the model if the newValue == the externalValue therefore we don't get the render that would format that value.

Probably only a very small chance this would ever happen, and the external value is still correct. Also, in any case, onBlur we get the correct format (I think because hasFocus is observable and triggers the render)

lbwexler:
Don -- Thanks or this.  Looks great.  Please look at my last commit, which fixes the subtle issue you mentioned (and generally seemed like the right thing to do on commit)

https://github.com/exhi/hoist-react/pull/827/commits/f54148626

PR #828, opened 2018-11-29
Mobile toolbars

TomTirapani:
Onsen's toolbars are absolutely positioned, which raises a number of issues (you can only have 1 top and / or bottom toolbar per page, they can overlap other components etc).

This change brings symmetry between the desktop and mobile toolbar implementations. Prompted by an immediate need in a client app.

Please also see Toolbox example:
https://github.com/exhi/toolbox/pull/116

mobile toolbar example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### 🎁 New Features
+
+* `Toolbar` and `ToolbarSeparator` have been added to the mobile component library.
+
 ## v16.0.0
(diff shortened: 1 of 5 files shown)


TomTirapani:
No, no particular reason, except wanting to minimise disruption to existing desktop apps - happy to make cross platform

TomTirapani:
I've updated the branches to make toolbar cross-platform and moved to /cmp

lbwexler:
Should we put these classes into layout, so we just have one less import line?

Issue #836, opened 2018-12-04
Select - render selected option with a checkmark in dropdown


lbwexler:
Right now the *focus* style of our Select component is much more accented than the *selected* style.
This to me seems to be exactly the opposite of what we want. 

The net result, is that it is very easy to get confused, and think that the focused item is currently selected.  

This is especially a problem if you use the Select box for just two items (which we do sometimes.)  The focused item simply looks like the selected item.  You end up thinking that there is nothing for you to change, and wondering why you went in to the select box in the first place!

shravnk:
I had noticed this as well... played around with CSS vars a little and came up with the following:

https://github.com/exhi/hoist-react/tree/selectMenuStyling

lbwexler:
Let's just ensure that what we are doing is standard here.  Worth looking at how other toolkits and os controls handle this.

amcclain:
I hadn't really paid attention to this before, but seeing now that MacOS uses a checkmark in a left-hand gutter to indicate the currently selected option.  That now seems like an obvious idea and one that makes for maximum clarity.  
<img width="670" alt="screen shot 2018-12-14 at 12 48 24 am" src="https://user-images.githubusercontent.com/225806/49985909-b6cf6a80-ff3b-11e8-9350-deb6d9c3e949.png">
<img width="747" alt="screen shot 2018-12-14 at 12 52 16 am" src="https://user-images.githubusercontent.com/225806/49985825-5e986880-ff3b-11e8-82df-49263c60a366.png">

I think the native browser select inputs must be getting this same behavior, as I see it consistently on some web pages loading what look like basic / native controls. Interestingly Gmail seems to provide an example of a custom implementation of the same:
<img width="381" alt="screen shot 2018-12-14 at 12 52 47 am" src="https://user-images.githubusercontent.com/225806/49985899-acad6c00-ff3b-11e8-8325-776f76d3b44d.png">

I think we should look at how difficult this would be to implement with react-select. (I would not be surprised if it's been tried or discussed in the wild.)

PR #846, opened 2018-12-10
Select menu options css tweak

shravnk:
- add to select component css
- define xh-select css vars

See #836

### DIFF
--- a/desktop/cmp/form/inputs/Select.scss
+++ b/desktop/cmp/form/inputs/Select.scss
@@ -86,14 +86,15 @@
     &__option {
       &--is-selected {
         color: var(--xh-text-color);
-        background: var(--xh-bg-alt);
+        background: var(--xh-select-selected);
       }
       &--is-focused {
         color: var(--xh-text-color-highlight);
-        background: var(--xh-bg-highlight);
+        background: var(--xh-select-focused);
       }
     }
   }
 
 }
+//--xh-bg-highlight: var(--bg-highlight, #{mc-trans('blue', '200')});
 
--- a/styles/vars.scss
+++ b/styles/vars.scss
@@ -192,6 +192,10 @@ body {
   --xh-scrollbar-bg: var(--scrollbar-bg, #{mc('blue-grey', '100')});
   --xh-scrollbar-thumb: var(--scrollbar-thumb, #{mc('blue-grey', '300')});
 
+  // Select menus
+  --xh-select-selected: var(--select-selected, #{mc('blue', '100')});
+  --xh-select-focused: var(--select-selected, #{mc('blue', '50')});
+
   // Tabs
   --xh-tab-border-width: var(--tab-border-width, var(--xh-border-width));
   --xh-tab-border-width-px: calc(var(--xh-tab-border-width) * 1px);


desktop/cmp/form/inputs/Select.scss
```diff
@@ -86,14 +86,15 @@
     &__option {
       &--is-selected {
         color: var(--xh-text-color);
-        background: var(--xh-bg-alt);
+        background: var(--xh-select-selected);
```
TomTirapani: Do we *need* to declare new vars for this? Wouldn't switching the existing selected / focused styles be sufficient?
I'm wary of adding further vars if not necessary.

desktop/cmp/form/inputs/Select.scss
```diff
@@ -86,14 +86,15 @@
...
       &--is-focused {
         color: var(--xh-text-color-highlight);
-        background: var(--xh-bg-highlight);
+        background: var(--xh-select-focused);
       }
     }
   }
 
 }
+//--xh-bg-highlight: var(--bg-highlight, #{mc-trans('blue', '200')});
```
TomTirapani: Remove this

styles/vars.scss
```diff
@@ -192,6 +192,10 @@ body {
   --xh-scrollbar-bg: var(--scrollbar-bg, #{mc('blue-grey', '100')});
   --xh-scrollbar-thumb: var(--scrollbar-thumb, #{mc('blue-grey', '300')});
 
+  // Select menus
+  --xh-select-selected: var(--select-selected, #{mc('blue', '100')});
+  --xh-select-focused: var(--select-selected, #{mc('blue', '50')});
```
TomTirapani: Here we should use `--select-focused` to allow the user to customise it.

TomTirapani:
Some general thoughts in addition to the inline comments:
- If we use new vars, please add dark theme variants - currently hard to read in dark theme.
- Please apply the same fix to the mobile Select

shravnk:
Thanks Tom, few thoughts:

- To avoid creating new variables, we could just flip the ones we had been using, i.e. 
selected -> `xh-bg-highlight` and focused -> `xh-bg-alt`. Also removes the need for dark mode styling . 

- I would suggest setting `!important` on the `is-selected` background (i.e. so that selected option does not change color when focused).

- Our code currently focuses the first option when the menu opens. That's normally fine, but can be confusing when there are two options and both are immediately shaded. One way to avoid this is to set a custom `renderValue` on `Select`:
`   @computed
    get renderValue() {
        return this.toInternal(this.externalValue)
    }`

Maybe overthinking that last point, but at least the first two CSS tweaks should offer an improvement.

TomTirapani:
> To avoid creating new variables, we could just flip the ones we had been using

Yes, that's what I suggested. We can also then flip the order of `--is-selected` and `--is-focused`, so that the selected style overrides the focused style. Then we wouldn't need the `!important`.

Don't think we need to worry about your last point just yet - let's keep this a quick styling-only change.

shravnk:
Right, sorry missed that top comment. Will amend the PR to only include CSS tweaks. Thank you.

PR #851, opened 2018-12-13
Action Column Performance

### LINKED ISSUE #784: Investigate Performance Issue with RecordAction column
We have seen in a client app that RecordAction column can really slow down scrolling.  There may be ag-Grid <=> react optimizations we are not taking advantage of that we could be.

haynesjm42:
Fixes #784 

Just did a direct port of the logic and resulting html from our react components, everything looks and behaves the same as it did previously, and performance while scrolling grids with actions columns is much improved.

### DIFF
--- a/desktop/columns/Actions.js
+++ b/desktop/columns/Actions.js
@@ -5,9 +5,8 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
-import {buttonGroup} from '@xh/hoist/desktop/cmp/button';
-import {recordActionButton} from '@xh/hoist/desktop/cmp/record/impl/RecordActionButton';
 import {RecordAction} from '@xh/hoist/data';
+import {convertIconToSvg} from '@xh/hoist/icon';
 import {isEmpty} from 'lodash';
 
 import {actionColPad} from './Actions.scss';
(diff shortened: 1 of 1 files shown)


lbwexler:
Thanks for getting on this one!

What are the performance implication of actually creating these dom nodes and returning it here vs. returning html as a string?  Does Ag-Grid say anything about this?   Could we have done this with strings?

haynesjm42:
I didn't see anything about it in the docs, and creating dom nodes is the way their examples work for cell renderer components. The recommendation they make for fast cell renderers is to use plain Javascript: https://www.ag-grid.com/javascript-grid-performance/

What would the performance difference be at the browser level? Wouldn't it end up constructing dom nodes anyway from the html string, thus being slightly slower than appending the nodes already provided? I'll be the first to admin that I know very little about what happens under the hood here...

Could pretty easily build an html string here instead of creating nodes, but what would be the means of adding the click handler for each button? I've done so little web dev without a framework that I don't really even know what the options are :)

lbwexler:
Actually, in that second article they talk about the fact that setting
innerHtml in bulk versus mucking with the dom directly was much faster for
them.

(I certainly didn't know anything about that when I made my comment).

I think I would be inclined to have us generate the raw html, and let the
framework/browser deal with efficiently updating the Dom.  That's what we
typically did with Sencha renderers, and what we have the most experience
with.

But the most important thing is to get the React Components out of there,
especially for our built-in renderers!

haynesjm42:
I tried building html string instead of creating dom elements, and performance was the same.

I believe that since the browser isn't really doing anything expensive until the elements are actually attached to the dom, which is likely being handled the same way by ag-grid whether we return a dom element or an html string (though admittedly my performance testing was quick and not very scientific, watching the console log for warnings about requestAnimationFrame taking a long time while scrolling the grid).

lbwexler:
ok-- agreed the dom elements are fine for now

Issue #801, opened 2018-11-14
Support disabling tabs


amcclain:
To still show them in the tab bar but not allow user to click into them - e.g. based on roles.

lbwexler:
+1 on this feature.  Sounds cool

Issue #796, opened 2018-11-12
Convenience configs to conditionally omit a tab/route


amcclain:
A tab / route could commonly be accessible by users with a particular role, or switched based on some other condition.  The arrays fed into TabContainerModel and the router can be built up / manipulated directly, but it would be useful to support an easy way to specify inline - perhaps via an `omit` config?

PR #856, opened 2018-12-13
Tab enhancements

haynesjm42:
#801, #800, #796

Added support for the following:

- `omit` can be passed in the tab config to `TabContainerModel`
- `TabModel.title` is now `bindable` to allow changing at runtime (can not get a reference to a `TabModel` via `TabContainerModel.getTabById`)
- Added `TabModel.disabled`, will (hopefully) make the tab unnavigable by any method
- Added `TabModel.excludeFromSwitcher`, will hide the tab from the switcher but will allow it to be navigated to programmatically or via routing (something I need for a client app)

Tried to be careful with the disabling of tabs to make sure the tab could not be navigated to via the switcher or routing if it was disabled, and when disabling a tab making sure that we switch to a different tab.

Toolbox example: https://github.com/exhi/toolbox/pull/129


desktop/cmp/tab/pane/TabModel.js
```diff
@@ -18,7 +19,9 @@ import {startCase} from 'lodash';
 @HoistModel
 export class TabModel {
     id;
-    title;
+    @bindable title;
+    @observable disabled;
+    @bindable excludeFromSwitcher;
```
lbwexler: Does excludeFromSwitcher need to be bindable?  Very curious to see the app that needs this.

desktop/cmp/tab/container/TabContainerModel.js
```diff
@@ -136,6 +156,11 @@ export class TabContainerModel {
                     });
 
                     if (activateTab) {
+                        if (activateTab.disabled) {
+                            console.warn(`Attempted to route to disabled tab ${activateTab.id}`);
```
lbwexler: Think you could skip this warning.

desktop/cmp/tab/pane/TabModel.js
```diff
@@ -64,6 +73,18 @@ export class TabModel {
...
     }
 
+    @action
+    setDisabled(disabled) {
+        this.disabled = disabled;
+        if (disabled && this.isActive) {
+            const {containerModel} = this,
+                tab = containerModel.firstEnabledTab;
+
+            throwIf(!tab, 'Could not find an enabled tab to activate!');
```
lbwexler: Ideally, you might throw *before* disabling tab if you see that all tabs will be disabled.    Its a real corner case, but in general we want to throw *before* we actually change the state, if the change is going to be untenable.

desktop/cmp/tab/pane/TabModel.js
haynesjm42: It doesn't, I made it bindable for symmetry with the other props and because there is no real reason that it _can't_ be bindable. I would have no objections to removing the bindable here if we think it is a simpler/clearer api without it and don't think there would be any real-world use-cases for it.

desktop/cmp/tab/container/TabContainerModel.js
haynesjm42: Should I also remove the warning in `activateTab`?

lbwexler:
I think so -- we were already quietly doing nothing if the tab did not
exist (e.g. was omitted for this user).

These are real corner cases, that I think are unlikely to cause major
confusion --- in this case, probably worth avoiding having the warnings
dominate what is otherwise pretty straightforward code.

lbwexler:
👍

PR #865, opened 2018-12-14
Form Support v2

lbwexler:
Still a work in progress, especially on the visible side with our FormField implementations.  Hoping Tom and I can collaborate there to bring this in for a landing.

To start, Tom and I discussed porting the readOnly display in FormField over to this branch 

Also, would like to propose moving inputs to a new "input" package as part of this change, but would suggest we do that just before merging into develop to avoid merge difficulties.


cmp/form/FormModel.js
```diff
@@ -0,0 +1,204 @@
...
+        return [
+            ...this.fields.map(fn),
+            ...this.children.map(fn)
+        ];
+    }
+
+    someMember(fn) {
+        return this.fields.some(fn) || this.children.some(fn);
+    }
+
```
TomTirapani: Rather than these `xxxMember` methods, wouldn't it be more straightforward to have a getter than returns the combined array of members and then directly operate on that wherever these are used?

e.g.
```
get members() {
  return [...this.fields, ...this.children];
}

this.members.forEach()
this.members.map()
this.members.some()
```

desktop/cmp/form/FormField.js
```diff
@@ -6,141 +6,246 @@
...
         });
     }
 
-    leftIcon(item) {
-        const leftIcon = item.props.leftIcon || (this.props.leftErrorIcon ? Icon.warningCircle() : null);
-        return item.type.propTypes.leftIcon ? {leftIcon} : {};
+    renderReadonly() {
+        const {fieldModel} = this,
+            value = fieldModel ? fieldModel['value'] : null,
+            renderer = this.props.readonlyRenderer || this.defaultReadonlyRenderer;
```
lbwexler: use withDefault

mobile/cmp/form/FormField.js
```diff
@@ -5,94 +5,195 @@
...
 
-        return React.cloneElement(item, {model, field});
+        return readonly ? this.renderReadonly() : React.cloneElement(item, overrides);
     }
 
-}
+    renderReadonly() {
+        const {fieldModel} = this,
+            value = fieldModel ? fieldModel['value'] : null,
+            renderer = this.props.readonlyRenderer || this.defaultReadonlyRenderer;
```
lbwexler: use withDefault

lbwexler:
I think this really looks great!  A few things.

1) The main item I want to raise is whether we want some built-in border around the read-only controls...  Or a seperator?  Something to distinguish from the labels....

2) We need to update changelog.  Am happy to take a crack at that.

3) Think we are ready to move the input package out, if others agree.

4) We need good toolbox examples of this.  Generally need to revamp our validation form, including adding support for subforms.  Also, I would leave the read-only toggle off of the first "controls" tab to avoid conflating forms and controls.  The form on that tab is really just an implementation thing....

TomTirapani:
Thanks for taking a look!
1. I went back and forth on this... I know what you mean, but also think we shouldn't style it too much, allowing it to take the "natural" styles of the application. That said, I think perhaps reducing our _label_ font size could go a long way. Would be interested to hear Anselm's view on this.
2. That would be great - I think you have a better understanding of how to capture the whole model in English :)
3. Yep - happy to make this change tomorrow, as long as no objections.
4. Agreed. Will spend some time tomorrow updating and incorporating subforms.

amcclain:
Just ran into a frustrating bug in v17 where applying the minimal validation mode somehow knocks out a custom className I am defining on a Select component.  I think it must have to do with how the target input item is cloned or something (I never really understood how that all worked).  I can show you guys - need to determine if this will still be a problem in this version and resolve if so.

PR #866, opened 2018-12-14
Select input:  stopPropagation of key event only when menuIsOpen = true

cnrudd:
Hi,

implementing this fix requires using a ref on the react select props, which will be overwritten if a user assigns their own ref via rsOptions.  

Just pointing that out for consideration.  One solution might be to only create this ref and wrap the reactSelect in a box if 'popoverMode' = true.  Happy to discuss.

-Colin

### DIFF
--- a/desktop/cmp/form/inputs/Select.js
+++ b/desktop/cmp/form/inputs/Select.js
@@ -191,16 +191,16 @@ export class Select extends HoistInput {
             className: this.getClassName(),
             width: props.width,
             onKeyDown: (e) => {
-                // Esc. can be used within the select to clear value / dismiss dropdown menu.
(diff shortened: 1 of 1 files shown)


amcclain:
Thanks - looks good to me!  Anything within `rsOptions` is buyer-beware, so if a user specifying such a thing breaks the esc handling here, that's in line with the component's guarantees.

(In that case, though, would we end up throwing an NPE on https://github.com/exhi/hoist-react/pull/866/files#diff-7c57d69a6dcea1ae69f31992eab0ca6bR196?  If so, maybe that's OK as per the above - although we could also ensure our expected ref has been assigned and log vs. throw if not.   Either way, I am glad we have this change in)

Issue #869, opened 2018-12-17
RestGrid should support read only mode


shravnk:
- RestGrid opens an editor onRowDoubleClicked - can't override
- To suppress CRUD controls, it's required to override three RestGridModel properties. Would be nice to bundle into a single read only/editable property.

lbwexler:
👍

PR #870, opened 2018-12-17
RestGrid Editable Mode

shravnk:
For #869 

- RestGrid takes an editable prop with default = true
- RestGrid.editable controls default menu buttons
- allow onRowDoubleClicked prop
- rowDoubleClicked defaults to edit if editable, view if not

### DIFF
--- a/desktop/cmp/rest/RestGrid.js
+++ b/desktop/cmp/rest/RestGrid.js
@@ -14,12 +14,32 @@ import {fragment} from '@xh/hoist/cmp/layout';
 import {restGridToolbar} from './impl/RestGridToolbar';
 import {restForm} from './impl/RestForm';
 import {RestGridModel} from './RestGridModel';
+import PT from 'prop-types';
 
 @HoistComponent
 @LayoutSupport
 export class RestGrid extends Component {
 
     static modelClass = RestGridModel;
+
+    static propTypes = {
+        /**
+         *
+         * This constitutes an 'escape hatch' for applications that need to get to the underlying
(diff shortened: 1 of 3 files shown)


desktop/cmp/rest/RestGrid.js
```diff
@@ -25,7 +25,7 @@ export class RestGrid extends Component {
 
     render() {
         const {model} = this,
-            {extraToolbarItems, agOptions} = this.props;
+            {extraToolbarItems, onRowDoubleClicked = this.onRowDoubleClicked, agOptions} = this.props;
```
lbwexler: need to document this with prop type

desktop/cmp/rest/RestGridModel.js
```diff
@@ -80,6 +80,7 @@ export class RestGridModel {
     get selectedRecord() {return this.gridModel.selectedRecord}
 
     /**
+     * @param {boolean} [editable] - make CRUD controls visible. Defaults to true.
```
lbwexler: Make this description more general -- e.g. allow 'user to edit/update/delete data.'
 
The default toolbars are probably only part of the story, and any app can always throw in whatever toolsbars they want -- for consistency, we should probably throw in RecordForm.saveRecord and RecordGridModel.deleteRecord() if somebody tries to edit a non-editable form.

lbwexler:
One more general comment -- can we rename the property to 'readonly' and default it to false?

As ATM points out, that is more consistent for objects that primarily exist to facilitate editing, and is what we are doing for our new forms package.

PR #873, opened 2018-12-17
Add file extension and contentType args to downloadjs download call.

cnrudd:
(Fixes [hoist-core #65](https://github.com/exhi/hoist-core/issues/65))

### DIFF
--- a/svc/GridExportService.js
+++ b/svc/GridExportService.js
@@ -85,8 +85,11 @@ export class GridExportService {
             headers: new Headers()
         });
 
-        const blob = response.status === 204 ? null : await response.blob();
-        download(blob, filename);
+        const blob = response.status === 204 ? null : await response.blob(),
+            fileExt = this.getFileExtension(type),
+            contentType = this.getContentType(type);
+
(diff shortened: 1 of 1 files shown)


cnrudd:
We were seeing csv exports appear without a file extension on windows computers, ie:  'export', not 'export.csv'.

And in my test Windows 10 machine, Excel exports were appearing the same way.

In GridExportService.js our call of function download(..) from the [downloadjs lib](https://www.npmjs.com/package/downloadjs)  was missing the file extension on the filename argument and missing the content type argument.   Any reason these pieces were left off? Or just an oversight?

The reason I saw the .xlsx extension dropped in my Windows 10 is because I don't have Excel installed on my Windows 10 machine.

It looks like if a computer has a mimetype associated to an extension, it will be smart enough to add the file extension to the download.

amcclain:
@TomTirapani please have a look when you can tomorrow to ensure we're not missing anything. I'm going to go ahead and merge now as it seems pretty clear, but we can certainly take more time to discuss if needed!

TomTirapani:
@cnrudd great stuff, thanks for catching!

PR #889, opened 2018-12-21
RestGrid default mask

shravnk:


lbwexler:
Thanks for taking the time to give RestGrid a bit of needed love all around.  Two quick things:

1) Can we call this property ' mask' and give it the semantics of `Panel.mask`, as much as possible? Specifically, its doc could be like the following, 
       /**
         * Mask to render on this Component. Set to:
         *   + a ReactElement specifying a Mask instance - or -
         *   + true for a simple default mask.
         */

  In the case of the first option, you would just clone the element passed, and install the appropriate PendingTaskModel.  The default (unlike for Panel) would be true.    

2) I would prefer that we use multiple withDefault() statements to get prop values with defaults in our render methods, rather than the destructuring with defaults.  It seems like we have been standardizing on that.

Thanks!

lbwexler:
Merging this.  Brendan will do some additional cleanups on the extraToolbarItems in a seperate request.

👍

PR #891, opened 2018-12-24
Chart resize

cnrudd:
This PR is just a placeholder, for review.  Merge not yet requested.

This is a chart resize solution that uses resizeObserver to observe the hbox size and redraws the chart when the hbox dimensions change.

This works in all uses cases I have tested, and adds support for maintaining an aspect ratio (which is a client requested feature, which seems like a good feature addition).

This resize solution may not be necessary (leave out consideration of aspectRatio feature)  if chart resizing can be proven to work consistently in all situations.
I have one use case where this solution appears to be necessary, but I think exploring how to make it not needed is worth pursuing further.


cnrudd:
I looked more into why using a resizeObserver solution may not be necessary:
I found that Highcharts 7, to which we just upgraded, fixed a bug that it describes in its change notes as:
"Corrected chart width for sub-pixel container width. Prevented overflow in flex-grids and table cells."
(first line of "Bug fixes" - unfortunately no Issue # attached to this line)

I think this fix creates this difference in charts created with v7 vs v6:
In charts with v7 (can be seen in toolbox):
<div style="flex: 1 1 auto;overflow: hidden;" data-highcharts-chart="2">
In charts with v6 (can be seen in my client app):
<div style="flex: 1 1 auto;" data-highcharts-chart="2">

When I edit the v6 chart element in the client app in the inspector to add overflow:hidden, the chart resizes correctly.


So, it looks likes a more complex 'resizeObserver' solution may not be necessary, but, here are some arguments for using a resizeObserver solution:

1. There may be some panel>>vframe>hframe>hbox> layout combos that break resizing.   It might be a lot to ask developers to discover the right combos or force them into using known working combos.
2. Resizing with resizeObserver is much smoother than the "native" highcharts resizing (even over VDI)
3. The resizeObserver solution in the PR adds a 'maintain aspect ratio' feature which requires calculated width/height based on readings of the chart hbox dimensions.   I think there is a strong argument for this feature for charts showing yield curves or options profit potentials in which comparing curve shapes is the purpose of the chart.

amcclain:
Hi - just looked over this - a lot to digest, but thank you much for all the details.

A possible immediate blocker, however - ResizeObserver appears to be Chrome only, and somewhat experimental at that - https://caniuse.com/#feat=resizeobserver.  I think we're pretty well committed to supporting Edge...  Let's discuss as soon as we have a chance.

cnrudd:
> Hi - just looked over this - a lot to digest, but thank you much for all the details.
> 
> A possible immediate blocker, however - ResizeObserver appears to be Chrome only, and somewhat experimental at that - https://caniuse.com/#feat=resizeobserver. I think we're pretty well committed to supporting Edge... Let's discuss as soon as we have a chance.

More details in the issue: https://github.com/exhi/hoist-react/issues/852
The NPM lib I chose:
https://www.npmjs.com/package/react-with-dimensions
uses another lib:
https://github.com/que-etc/resize-observer-polyfill
that has polyfills for other browsers.

amcclain:
Hi - me again 😄. 

Coincidentally, I just came across this Blueprint component tonight - https://blueprintjs.com/docs/#core/components/resize-sensor - which looks awfully similar in its general concern to the with-dimensions library.  Running `yarn why resize-observer-polyfill` on the current develop branch, I see we do indeed already have resizer-observer-polyfill in the project (who knew?).

Do you think we could accomplish the same goals with this one and avoid the new dependency? I'm not against adding new library whenever we need or would benefit, but figure it's worth asking the question since this does look like a close cousin.

cnrudd:
@amcclain 
Definitely worth trying https://blueprintjs.com/docs/#core/components/resize-sensor.  I will give it a try.  Thanks for finding that!
Colin

cnrudd:
Hi,

I have replaced react-with-dimensions with blueprintjs' ResizeSensor.  To test the functionality, you should switch toolbox to the new 'chartResize' branch.

Toolbox also has a [PR](https://github.com/exhi/toolbox/pull/147), in case you want to keep the changes in the chart to demo resizability and aspectRatio.

PR #893, opened 2018-12-26
Fix to Panel Resizing

lbwexler:
Fixes https://github.com/exhi/hoist-react/issues/685

I don't think this is the last word in panel resizing by any means -- in particular, would like to have a better way for components that are layout siblings with a resizer to declare a minimum-size (based on their own header or splitter) so they don't get completely overwhelmed.

But in any case, this prevents the horrible situation of any particular resize container going beyond the bounds of its own parent.


lbwexler:
Merging this to get it in the RC for testing, usage.  Easy to amend if needed.

lbwexler:
one of the little goodies we picked up in es6

-Lee

Issue #782, opened 2018-11-06
Match Select placeholder color to standard input placeholder


amcclain:
Select component draws its own custom placeholder element, which appears to have slightly different styling (more opaque/darker) than the built-in textinput placeholders.  

Not a crisis, but is noticeable on forms where you have a few of them side by side.

I wonder if there is a way to get exactly what style is used for placeholders by the browser (or at least, by Chrome).

lbwexler:
Giving this to Tom and Brendan, as they are working on the related select styling ticket: 
 
https://github.com/exhi/hoist-react/issues/836

PR #894, opened 2018-12-28
Enforce standard placeholder color across all xh inputs, mobile/desktop

shravnk:
For #782 - using color picker tool, determined that the default input placeholder (Mac, Chrome) is rgb(175, 183, 191). Rather than try to set `select`'s placeholder color for each browser/device, thought it would be better to enforce placeholder color across all XH inputs. Used `::-webkit-input-placeholder` selector for that purpose.

### DIFF
--- a/desktop/cmp/input/Select.scss
+++ b/desktop/cmp/input/Select.scss
@@ -37,7 +37,7 @@
     }
 
     &__placeholder {
-      color: rgb(175, 183, 191);
+      color: var(--xh-input-placeholder-text-color);
     }
 
     &__single-value,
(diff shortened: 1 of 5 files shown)


amcclain:
Looks like a good change to use the pseudo-selector, thanks.  

Can you please put this into an `--xh-input-placeholder-color` CSS var as per our standard? Even if it's not something we expect apps/clients to customize, it helps to DRY up and define the color vs. having the rgb value in a few places.

Also just confirming - does the same color look good in light and dark mode?

shravnk:
Updated with new XH css variable for placeholder text color. I think it looks good in dark mode as well. Thanks

Issue #862, opened 2018-12-14
Dialog should clip content overflow


amcclain:
This might be more general than just the Exception dialog - we should review and ensure a fix is applied at the most efficient level. In general I think think we would ever want a dialog to overflow like this.

`XH.handleException('This is an error with a reallyLongStringLikeTheKindOfThingThatCanBeProducedByExceptionsAndStuffLikeThatWeHaveSeen and it overflows')`

<img width="721" alt="screen shot 2018-12-14 at 3 06 25 am" src="https://user-images.githubusercontent.com/225806/49990933-72e56100-ff4d-11e8-8643-1f4e792281a7.png">

TomTirapani:
And interesting question is do we want to elide the overflowing long word (using `text-overflow: ellipsis`), or break it (using `word-break: break-word`)?

amcclain:
Looks like Github must do break-word here? If that works nicely, sounds like the better option to me.

`XH.handleException('This is an error with a reallyLongStringLikeTheKindOfThingThatCanBeProducedByExceptionsAndStuffLikeThatWeHaveSeenAlsoreallyLongStringLikeTheKindOfThingThatCanBeProducedByExceptionsAndStuffLikeThatWeHaveSeen and it overflows')`

lbwexler:
this would be a nice one to fix -- overflow errors to me are the most unsettling to me.  Seems just so deeply broken to have text escaping from its borders...

shravnk:
My mistake here. I set word-wrap to break-all to avoid unsightly white spaces preceding these long strings. 

Easiest change would be to set the rule to break-word. Could also look into handling special cases with more complex CSS logic.

amcclain:
Thanks for the explanation - swapped to `break-word` and will keep an eye on it.

cnrudd:
well, it worked for 5 years...
turns out 
break-word breaks numbers in a date-picker opened in a dialog:
<img width="1191" alt="image" src="https://github.com/xh/hoist-react/assets/2573928/892e579b-fd83-47cf-b576-10bc48c49e76">

amcclain:
Resolved: dialog body content overflow is handled via overflow-wrap: break-word in kit/blueprint/styles.scss, and the date-picker wrapping regression noted above was fixed in #3708 (v66.0.0) with an explicit datepicker carve-out. Both fixes reference this issue in code comments. Closing - please reopen if any new overflow case surfaces.

PR #895, opened 2018-12-28
Dialog text overflow

shravnk:
Two css tweaks to handle #862. Note that I used `break-all` instead of `break-word`. Think it looks better at least in the case of error messages, i.e. not forcing very long strings to occupy a new line.

### DIFF
--- a/kit/blueprint/styles.scss
+++ b/kit/blueprint/styles.scss
@@ -101,6 +101,10 @@
       margin-right: var(--xh-pad-px);
     }
   }
+
+  .bp3-dialog-body {
+    word-break: break-all;
+  }
 }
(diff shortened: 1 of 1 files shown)


amcclain:
Thanks! 

I am slightly concerned that the first rule could inadvertently target text within other nested components that happen to be inside of a dialog body - e.g. a popup with a grid or other controls inside that get unexpected styling here - but I don't have any real examples or strong reason to believe this will be the case.  I'll merge and we can get the fix in, look out for any issues. 

If we _do_ see this being applied in an over-eager way, we could restrict to a selector that targeted our message box for alerts/confirms/etc only.

PR #898, opened 2018-12-29
plugin to zoomout on RtL drag added to Chart component (fixes #545)

### LINKED ISSUE #545: Charts - zoom-out on click-and-drag from RtL
Assuming w/Highcharts plugin implementation we have used in Sencha toolkit.

cnrudd:


desktop/cmp/chart/plugin/zoomout.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable consistent-this */
```
lbwexler: Couple of questions:

+ Any comment on where this comes from?  Why do we need the chart variable instead of this?

cnrudd: I've removed the chart variable the eslint-disable.  The code works fine, but I think the feeling was that it was confusing to track when reading the several event handler methods that 'this' always refers to the chart.

PR #902, opened 2019-01-02
Add enableClear prop on select.

lbwexler:


amcclain:
I was seeing the X still on multi-value fields by default 🤔...  Weird?

Note additional commit to explicitly default to false which provides the behavior documented in the changelog.  I do think it's better to have a consistent default and don't think the multi-value fields clearly "need" the X more than single value.

PR #906, opened 2019-01-03
Make select menu's current value editable

shravnk:
Discovered an existing select menu bug - typing a string which exceeds the input area's width pushes the chevron icon to the right and out of view, and further input causes it to flap back and forth. Fixed this behavior by replacing `flex: 1` with `width: 100%` on the outer react-select div. Changing that css also solved a bug specific to this PR.

Minor but annoying problem was being unable to set `text-overflow: ellipsis` on the value-populated `<input>` element. In the interest of completing a first working version, I implemented a workaround (see 3c2d467) which I suspect we will deem not worth the trouble...


lbwexler:
Thanks everybody for the incredible work on this one.  I admit I don't fully understand what is going on here, but it looks just unequivocally better in my testing, and really fixes a very bad bug.

I made some small tweaks to clarify -- @amcclain  please feel free to merge this if it looks good to you?

shravnk:
@lbwexler see your point and agree there is no reason to add the `onFocus` prop. Question would be whether we should instead call `this.noteFocused()` within Select's `onFocus` handler. Tested with/without and don't observe any difference in behavior. Would at the least keep us consistent with other HoistInputs

lbwexler:
great -- thanks for confirming.  I can apply the cleanup and will merge

amcclain:
*Thank you* for getting this fix in - a big improvement to this control. 

Can we please add a changelog entry to highlight?  This has been annoying enough and noted by more than a few client developers, so good to call out the fix.  I would also like to ensure we cut a release with this change before merging any breaking changes into develop.

PR #908, opened 2019-01-04
export column definitions from grid packages

shravnk:


desktop/cmp/grid/columns/Actions.js
```diff
@@ -5,8 +5,8 @@
  * Copyright © 2018 Extremely Heavy Industries Inc.
  */
 
-import {RecordAction} from '@xh/hoist/data';
-import {convertIconToSvg} from '@xh/hoist/icon';
+import {RecordAction} from '@xh/hoist/data/index';
+import {convertIconToSvg} from '@xh/hoist/icon/index';
```
amcclain: Looks like IJ might have added this change

cmp/grid/index.js
```diff
@@ -8,3 +8,9 @@
 export * from './Grid';
 export * from './GridModel';
 export * from './GridStateModel';
+
+export * from './columns/Core';
+export * from './columns/DatesTimes';
+export * from './columns/ExportFormat';
+export * from './columns/Column';
+export * from './columns/ColumnGroup';
```
amcclain: I would vote for leaving `columns/index.js` in place and then re-exporting that from here. Would help with app-level updates - in that the `/cmp/grid/columns -> `/cmp/grid` import change would still be a good idea but be optional. More so though it seems like helpful to have sub-packages like this one with several nested files manage their own "directory listing".  Thoughts? Would defer to you and @lbwexler if you've discussed and don't want to do this.

lbwexler:
Brendan -- can we resolve these issues so we can go ahead and merge this.  Would like to prepare for a final 18.0.0 release

shravnk:
Sorry missed this - agree that we should try to avoid introducing breaking changes here. I've updated the PR accordingly.

There's still a breaking change because I moved /desktop/columns to /desktop/cmp/grid. Made a note of that in the changelog. Think this one is more reasonable because a columns package in the same directory as appcontainer and cmp seems out of place.

PR #911, opened 2019-01-08
Refactor handling of subforms

lbwexler:


lbwexler:
+ Refactor handling of subforms.
+ Fix to commitOnChange prop on form() and formField()
+ form should inherit from parent.
+ inputs should use autoComplete: 'nope' not 'no'! 

cr: atm

PR #914, opened 2019-01-11
PropTypes for LayoutSupport components

### LINKED ISSUE #900: PropType definitions for layoutSupport components
We should be able to define a set of PropType definitions that can be mixed into the propTypes for all Hoist components implementing our `@layoutSupport` decorator.

I think this would be worth the effort to more comprehensively document all props supported via the decorator, and in general make the layout props more discoverable for any given component that has them.

amcclain:
+ Checkpoint for review before applying to any additional comps.
+ Fixes #900


amcclain:
Pls. see comments in linked issue.  Obviously WIP - did not want to bomb all the components without checking that this seems worthwhile.

lbwexler:
I am all for this.  What would think about having the decorator rewrite the static prop to spread these in?  I think it has access to static properties like this!

amcclain:
Closing this - will revisit in Hooks world.

PR #928, opened 2019-01-24
Add @managed and @ManagedSupport

lbwexler:


lbwexler:
I definitely looked having HoistComponent use this system to tear down its own managed 'model', but ended up not doing so.

First of all, from a technical perspective turned out to be not straightforward given the special handling of model as a prop config, checkin on it, etc.  But also I think -- as discussed --  we are moving away from the class based definition of HoistComponent, and hopefully toward a future where we are specifying all HoistComponents as functions with hooks, rather than classes.  In that world the model will really be managed by a 'hook'.


Regarding the name -- agree @autodestroy is the best descrition right now, but I didn't like the word "destroy" being the introduction to every one of these properties.  It seemed to overemphasize that technical (albeit important)  detail to the reader who should be focused on what the thing is....was actually looking for something a bit more vague that would interfere less with your reading of the code...

PR #932, opened 2019-01-28
Rest grid form refactor

lbwexler:


lbwexler:
Still some work on the layout that I want to discuss here.  Namely -- would like to have a good story for getting the side-by-side labels, and also a better story about the default rendering of ReadOnly labels

lbwexler:
Think this is getting close for review -- Tom, Brendan, could you guys take a look?

Tom -- am especially interested if you have any ideas on how to make the FormFields look better "out-of-the-box".  Should the inputs be automatically stretching?  Should we make the label width configurable at the form level so that you can more easily set a width that will be long enough for all the fields?

desktop/cmp/form/FormField.js
```diff
@@ -80,7 +80,17 @@ export class FormField extends Component {
          * Focus or toggle input when label is clicked.
          * Defaulted from containing Form, or true.
          */
-        clickableLabel: PT.bool
+        clickableLabel: PT.bool,
+
+        /**
+         * Alignment of label text, default 'left'.
+         */
+        labelAlign: PT.oneOf(['left', 'right']),
```
amcclain: This seems like it could easily be confused with the overall position of the label relative to the input.  I would propose that this be customizable via CSS instead of a prop.  

In general, I have been trying to keep in mind that CSS exists as a very configurable, easily controlled system for style overrides - provided we easy-to-select classes in place consistently for our components.  Would apply to labelWidth as well - I almost added that last week, then realized that a CSS rule was easy.  

Having the inline styles could make adjustments trickier to target, given that they will get a high level of specificity out of the box.  Thoughts on unwinding?

amcclain: Also thinking that we can use CSS targeted at the rest-form to make the labels stand out a bit better in this context, without over-committing to formField changes.  E.g.
```
.xh-rest-form .xh-form-field-label {
  text-align: right;
  font-weight: bold;
  width: 130px;
  margin-right: var(--xh-pad-px);
}
```

desktop/cmp/rest/impl/RestForm.js
```diff
@@ -7,106 +7,90 @@
...
-                gridModel: parent.gridModel
+                actions: model.actions,
+                record: model.currentRecord,
+                gridModel: model.parent.gridModel
             }),
             filler(),
             button({
-                text: isWritable ? 'Cancel' : 'Close',
-                onClick: this.onCloseClick
+                text: formModel.readonly ? 'Cancel' : 'Close',
```
amcclain: Assume we want this nice bit of polish flipped - "Cancel" for the edit case.

desktop/cmp/rest/impl/RestForm.js
```diff
@@ -7,106 +7,90 @@
...
-    getForm() {
-        const {isWritable} = this.model;
-        return vframe(
-            this.model.controlModels.map((model,
-                idx) => restControl({model, disabled: !isWritable, autoFocus: idx === 0}))
+        return dialogBody(
+            form({
+                model: model.formModel,
+                fieldDefaults: {
+                    commitOnChange: true,
```
amcclain: Makes me again think we should consider flipping the default here.  This seems like a recurring pattern where we set to true to avoid bugs around edits not flushing as quickly as expected to update validation, etc.  

We could still keep the feature, but make it opt-in for controls where not waiting to commit will thrash a back-end search-as-you-type, etc. Those do seem like important use cases, but also perhaps the exception vs. the rule.

desktop/cmp/rest/impl/RestForm.js
```diff
@@ -7,106 +7,90 @@
...
+                    minimal: true,
```
amcclain: I wonder if this should be configurable.

desktop/cmp/rest/impl/RestForm.scss
```diff
@@ -1,5 +1,7 @@
 .xh-rest-form {
-  &__control {
-    margin-bottom: var(--xh-pad-px);
+  width: 470px;
```
amcclain: Here we're using styles for an even more critical width - the overall width of the form - which I think is fine but again could argue for having the label width and other styles be done primarily in CSS as well.

desktop/cmp/rest/impl/RestFormField.js
```diff
@@ -0,0 +1,93 @@
...
+            name = this.props.editor.field,
+            type = model.types[name],
+            storeField = model.store.getField(name),
+            fieldModel = model.formModel.fields[name];
+        
+        if (storeField.lookup) {
+            return this.renderSelect({
+                options: [...storeField.lookup],
+                enableClear: !fieldModel.isRequired,
+                enableCreate: storeField.enableCreate
```
amcclain: These are nice 👍

desktop/cmp/rest/impl/RestFormField.js
```diff
@@ -0,0 +1,93 @@
...
+            case 'bool':
+                return this.renderBoolean(fieldModel);
+            case 'number':
+                return numberInput({width: 300});
+            case 'json':
+                return jsonInput({width: 300, height: 150});
+            case 'date':
+                return dateInput({width: 300});
+            default:
+                return textInput({width: 300});
```
amcclain: Great that the need to default these widths here goes away in #943

desktop/cmp/rest/impl/RestFormField.js
```diff
@@ -0,0 +1,93 @@
...
+            switchInput() :
+            this.renderSelect({
+                options: [true, false],
+                enableClear: !fieldModel.isRequired,
+                enableCreate: false
+            });
+    }
+
+    renderSelect(args) {
+        return select({...args, width: 300});
```
amcclain: This default should hopefully go away with #943  as well

PR #933, opened 2019-01-29
RestStore returns the new Record after a save

haynesjm42:
Pretty common need when using a RestStore outside the context of a RestGrid (or grid in general) is to be able to get a reference to the new Record that was just saved/added. I specifically ran into this when doing a "Save As.." type of operation and I wanted to update the selection to the new Record after the save operation completed.

### DIFF
--- a/desktop/cmp/rest/data/RestStore.js
+++ b/desktop/cmp/rest/data/RestStore.js
@@ -7,7 +7,7 @@
 
 import {XH} from '@xh/hoist/core';
 import {UrlStore, Record} from '@xh/hoist/data';
-import {pickBy} from 'lodash';
+import {pickBy, filter} from 'lodash';
 
 import {RestField} from './RestField';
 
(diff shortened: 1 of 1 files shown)


amcclain:
Seems very reasonable to me.  Only thing that caught my eye is how the multiple promises make it relatively unclear what, exactly, is being returned. We have two explicit `return`s and one implicit one in the new arrow function.  

What do you think about this? (Also swapped to `_.filter()` which we could take or leave).
```
async saveRecordInternalAsync(rec, isAdd) {
    let {url} = this;
    if (!isAdd) url += '/' + rec.id;

    // Only include editable fields in the request data
    const editableFields = filter(this.fields, 'editable').map(it => it.name),
        data = pickBy(rec, (v, k) => k == 'id' || editableFields.includes(k));

    const fetchMethod = XH.fetchService[isAdd ? 'postJson' : 'putJson'],
        response = await fetchMethod({url, body: {data}}),
        newRec = new Record({fields: this.fields, raw: response.data});

    if (isAdd) {
        this.addRecordInternal(newRec);
    } else {
        this.updateRecordInternal(rec, newRec);
    }

    await this.ensureLookupsLoadedAsync();
    return newRec;
}
```
The linking to the loadModel could then be done in the two public methods so it still spans both the main update and the potential load for lookups.

We don't need to make this change, but I've been looking recently at how the async/await vs. Promise.then() styles compare and this is a case where having the standard return statement seems to help...

haynesjm42:
I think that looks great, much clearer what is going on with the return value with async/await. Code updated to use this method

PR #936, opened 2019-01-30
Add excludeHidden opt to GridModel.getLeafColumns()

### LINKED ISSUE #935: GridModel columnState not consistently used as source for current column visibility info
The grid export process reads the GridModel columns directly for information on hidden state - but the cols no longer update their internal hidden property when the state is changed.  GridModel is doing that bookkeeping separately in its columnState collection.
 
Same applies to intended behavior of StoreFilterField, where if provided a GridModel it should by default search over visible columns only.

amcclain:
+ Use in grid export + storeFilterField integrations, to reliably query for visible cols only when required.
+ Some smaller doc comment and syntax updates for clarity
+ Fixes #935

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -10,6 +10,7 @@ import {StoreSelectionModel} from '@xh/hoist/data';
 import {
     castArray,
     defaults,
+    find,
     findLast,
     isEmpty,
     isPlainObject,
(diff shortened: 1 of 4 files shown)


amcclain:
I think this has been broken since #694 - in October... Is that possible?  Just noticed by a client who didn't see his grid column changes reflected as expected in the export.

cmp/grid/GridModel.js
```diff
@@ -413,29 +414,59 @@ export class GridModel {
...
+     * @param {boolean} [opts.excludeHidden] - true to return currently-visible leaves only.
+     * @returns {Column[]}
+     */
+    getLeafColumns({excludeHidden = false} = {}) {
+        let ret = this.gatherLeaves(this.columns);
+
+        if (excludeHidden) {
+            ret = ret.filter(col => {
+                const state = this.getStateForColumn(col.colId);
+                return !state || !state.hidden;
```
amcclain: I wanted this to be safe to call at any point, even before state has been constructed. In practice we don't call it with excludeHidden:true that early, but still was wary of depending on the state object to be populated...

amcclain: ...although wondering if it would be advantageous to have `getStateForColumn()` always return a state object - either from the state collection or constructed on the fly from the column defaults...  Not sure if that would be an improvement.

haynesjm42: Something to note about `getLeafColumns` and the `columnState` is that `getLeafColumns` is used to initialize the `columnState` in `setColumns`, which makes things kind of weird here where this method now has a dependency on the `columnState` (albeit via an optional param).

I remember struggling with this when we were making the Grid changes which introduced this bug, at one point I remember having `getLeafColumns` merging in the state for each col so it would return a plain object which contained all of the `Column` props and the current state props, but that didn't feel right either, especially since it was eliminating the `Column` class and any helper methods we have there (or will add in the future).

Perhaps it would be best to clearly separate our primary method for getting all leaf Column definitions from getting a list of Column definitions for the cols which are currently visible. Could have a separate `getVisibleLeafColumns` method which does this. Alternatively could just leave it up to the user of the api to filter out the Column definitions they don't want, and we could provide helper methods like `isColumnVisible` to make it easy, and the code to do this would be:
```
gridModel.getLeafColumns()
    .filter(({colId}) => gridModel.isColumnVisible(colId));
```
Really what you have here is probably fine, but something has always bugged me about GridModel and clearly separating the column state from the definitions, but I'm not sure exactly what (if anything) to do about it.

amcclain: Yes, exactly my concerns as well.  I had the distinct `getVisibleLeafColumns()` in an earlier iteration, but it didn't seem especially clear that this method would have different characteristics in terms of when it would be ready to call.

Also had the callers doing the filtering, calling the getStateForCol method as needed, but that seemed like they needed to know too much about the (somewhat odd) internals of GridModel and its state-related bookkeeping.  

I like the idea of the `isColumnVisible()` helper method a lot - I think it strikes the right balance. What do you think it should do in the hypothetical case where gridState has yet to be initialized?  Return false? Return the column object's default?

haynesjm42: Is that even a case we need to handle? In the `GridModel` ctor we are calling `setColumns` which initializes the `columnState` with the defaults from the `columns`, so shouldn't the default state always be there?

Another case would be if someone is asking about a column which doesn't exist anymore, in which case I would think we want to just return false?

Issue #938, opened 2019-01-30
HoistInput should fill space of FormField


TomTirapani:
Now the FormField has LayoutSupport, HoistInputs within FormFields should fill the available space of the FormField.

lbwexler:
Tom -- sounds great -- when we fix this, we should rollback the explicit sizing in the restGridRefactor branch

amcclain:
Yes - agree very much in the case of inputs that are embedded within a formField.  In those cases, as a developer I would want to deal with sizing only one component - the outer formField wrapper.

We should still aim to provide some kind of sensible default for HoistInputs used on their own, outside of a formField. We'd like to stick to a general principle where "the simplest possible use" of a component should work well and look reasonable. A specific example would be putting a textInput or select in a toolbar - you would not expect it to stretch to fill the entire bar.

Note it would be great to review and close out #327 as part of this change.

TomTirapani:
See PR for an explanation of approach taken:
https://github.com/exhi/hoist-react/pull/943
https://github.com/exhi/toolbox/pull/149

Issue #327, opened 2018-05-29
Tighten up default 'width', 'height' props on JsonField and TextAreaField.


cnrudd:
It's not clear from the code (to me) how these fields behave when width is not specified.

**For the JsonField:**
I know from experience (if I remember correctly) that the codemirror library used for the JsonField will try to make the textarea it uses as wide as the container of the textarea if a width is not specified.  But, if a width is not specified, codemirror does not provide scrollbars correctly.
Maybe this info should be put into a comment in the JsonField class, maybe a default width should be provided so that scrollbars always appear, maybe a jsonfield demo or two demonstrating these behaviors should be put into toolbox.

**For the TextAreaField:**
 if the width prop is not provided, there is no default.
After reading the blueprint docs: http://blueprintjs.com/docs/v2/#core/components/forms/text-area
I think it might be worth considering: if no width is provided pass class 'pt-fill' to the blueprint textarea element so that the textarea takes up the fill width of its parent (if that isn't already somehow happening).

amcclain:
Hi @cnrudd  - when you have a chance can you clarify what you mean here by "button down"?  Do we need defaults defined, or need to improve the ones we have?  I wasn't sure from looking at the ticket what needed to happen, or if it was a bugfix / general enhancement / docs issue.  Thanks much.

lbwexler:
Assigning to Colin -- please see ?s by atm

amcclain:
See example in just-reworked toolbox form fields tab - some sizing issues there I don't understand that would be a good test-bed for any changes and fixes.

amcclain:
Also see this patch fix that was required to toolbox - better control over sizing would hopefully avoid need for this: https://github.com/exhi/toolbox/commit/bef7cf8fee9037a24c77b544c3103b99ae260a3e

amcclain:
This is still outstanding - moved the patch referenced above to the component itself after we broke the Toolbox JSX example again - https://github.com/exhi/hoist-react/commit/9906ee4ce9a0a9711166d7f596f5cfd763771dce.

We need someone to sit down and really carefully think through how these fields should size themselves - in terms of the default behavior and the input component API - and then determine how that should be implemented / made possible.

PR #943, opened 2019-02-01
Rationalise sizing of HoistInputs and FormField

TomTirapani:
This branch includes some changes that will hopefully make sizing HoistInputs and FormFields more intuitive:

- HoistInputs can be split into two categories: "sized" and "unsized".
- Sized HoistInputs are now all exclusively sized using LayoutSupport. We previously had some sized inputs that were sized using named `width` & `height` props, and others using css.
- All sized inputs use `defaultProps` for reasonable, out-of-the-box widths. `TextArea` and `JsonInput` also have a default height. This means that all sized inputs always have an explicit size for their pertinent dimensions, which should mitigate unexpected sizes.
- When a sized input is inside a FormField, FormField unsets the width and flexes HoistInputs within it. It will also unset the height if the FormField has a specified height (but *only* if specified, allowing TextArea and JsonInput to use their default heights).
- Unsized HoistInputs (Checkbox, SwitchInput, RadioInput) do not have LayoutSupport, and do not expand to fill the dimensions of the FormField. Instead, these take their natural size from their content.

See issues #938 & #327
See Toolbox PR: https://github.com/exhi/toolbox/pull/149


desktop/cmp/input/DateInput.js
```diff
@@ -79,10 +80,11 @@ export class DateInput extends HoistInput {
...
          * If undefined, control will not show time.
          */
-        timePrecision: PT.oneOf(['second', 'minute']),
+        timePrecision: PT.oneOf(['second', 'minute'])
+    };
 
-        /** Width of the control in pixels. */
-        width: PT.number
+    static defaultProps = {
+        width: 150
```
lbwexler: we were moving away from specifying defaults with this mechanism

lbwexler: instead, use withDefault in the render method.

desktop/cmp/form/FormField.js
```diff
@@ -196,6 +213,19 @@ export class FormField extends Component {
             disabled: fieldModel && fieldModel.disabled,
             id: idAttr
         };
+
+        // Item should fill the available size of the FormField, unless dimensions are specified.
+        // Note: We explicitly set width / height to null to override defaults.
+        if (!item.props.width && !item.props.flex) {
```
lbwexler: both the checks on 219 should be isUndefined() to allow zero and null settings to be respected

amcclain:
Already discussed with Tom, but adding here so we have a note on it and can pick it up again next week.  Saw some sizing irregularities on textarea and selects when switching between the various modes:

screen shot 2019-02-01 at 9 47 22 am
screen shot 2019-02-01 at 9 47 27 am
screen shot 2019-02-01 at 9 49 58 am

desktop/appcontainer/FeedbackDialog.js
```diff
@@ -38,7 +38,9 @@ export class FeedbackDialog extends Component {
             items: [
                 textArea({
                     placeholder: 'Please enter your comments...',
-                    style: {height: 250, marginBottom: 2},
+                    width: null,
```
amcclain: I am rushing a bit and reading this on github / not running the code, but I see a  few cases here and below where we are setting the width to null on these components.  Why is that?  It seems like, as a general goal, that should be a pretty unusual thing to need to do.

PR #945, opened 2019-02-01
delete minWidth/Height and width/height from layoutProps if collapsed…

### LINKED ISSUE #912: minWidth on Collapsed Resizable panel does not allow the panel to collapse correctly
If you put a minWidth prop on a Panel component that has a sizing model on it, the panel does not collapse correctly.
Here is what you get when you try to collapse the panel:
image

Dragging the panel to a small size still works fine.

cnrudd:
… (fixes #912)

should be viewed with corresponding toolbox branch: https://github.com/exhi/toolbox/tree/resize-panel-912

Do we want want to go further and:
1. prevent drag to smaller than a minWidth?
or 
2. warn if the 'defaultSize' is smaller than minWidth?

### DIFF
--- a/desktop/cmp/panel/Panel.js
+++ b/desktop/cmp/panel/Panel.js
@@ -93,8 +93,14 @@ export class Panel extends Component {
             resizable = false,
             collapsible = false,
             collapsed = false,
-            collapsedRenderMode = null
+            collapsedRenderMode = null,
+            vertical = false
         } = model || {};
+
+        if (collapsed) {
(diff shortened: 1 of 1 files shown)


lbwexler:
Don't think we need to take those extra steps -- very tricky to control the drag operation dynamically, and if sizing below minSizes is not causing crazy behavior, not as much of a problem.

Certainly the rendering shown on ticket is quite bad, so this is an important fix, however.

PR #947, opened 2019-02-05
Upgrade agGrid to v20.0.0

TomTirapani:


lbwexler:
I would like to go ahead and take this fix in Hoist 19.   AP changes are minimal and can be documented in changelong "breaking changes".

Most importantly, Tom reports promising results with mobile column scrolling!  Also we are doing some performance optimizations for Expo, so very helpful to be on the latest and greatest

lbwexler:
atm -- let us know if you are fine with that

PR #948, opened 2019-02-06
Force Users to opt-out of stable ids on Records

lbwexler:
This fixes https://github.com/exhi/hoist-react/issues/844

Motivation was that a client was recently hit very hard by these auto-generated ids.  auto-refreshing grid with tens of thousands of rows was in production for weeks without ids, causing performance problems and memory issues.

Also, client noted that agGrid itself warns about missing ids, so we were actually papering over that warning by silently creating these ids for users.

### DIFF
--- a/data/LocalStore.js
+++ b/data/LocalStore.js
@@ -9,7 +9,7 @@ import {XH} from '@xh/hoist/core';
 import {PendingTaskModel} from '@xh/hoist/utils/async';
 import {throwIf} from '@xh/hoist/utils/js';
 import {observable, action} from '@xh/hoist/mobx';
-import {isString, isUndefined} from 'lodash';
+import {isString, isNil} from 'lodash';
 
 import {RecordSet} from './impl/RecordSet';
 import {Record} from './Record';
@@ -151,7 +151,7 @@ export class LocalStore extends BaseStore {
 
             raw.id = idGen(raw);
             throwIf(
-                isUndefined(raw.id),
+                isNil(raw.id),
                 'Cannot load record without a unique id.  Provide a unique id on each raw record using the ' +
                 '`idSpec` property of this store.'
             );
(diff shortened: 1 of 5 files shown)


data/LocalStore.js
```diff
@@ -33,12 +35,25 @@ export class LocalStore extends BaseStore {
...
      * @param {...*} [c.baseStoreArgs] - Additional properties to pass to BaseStore.
      */
-    constructor({processRawData = null, filter = null, ...baseStoreArgs}) {
+    constructor(
+        {
+            processRawData = null,
+            filter = null,
+            idSpec = 'id',
+            ...baseStoreArgs
+
```
haynesjm42: Blank line here

data/LocalStore.js
```diff
@@ -128,11 +143,18 @@ export class LocalStore extends BaseStore {
...
         return rawData.map(raw => {
             if (this.processRawData) this.processRawData(raw);
 
-            // All records must have a unique, non-null ID - install a generated one if required.
-            if (isNil(raw.id)) raw.id = XH.genId();
+            raw.id = idGen(raw);
+            throwIf(
+                isUndefined(raw.id),
+                'Cannot load record without a unique id.  Provide a unique id on each raw record using the ' +
```
haynesjm42: This message may be a bit misleading, as we aren't really checking that the id is unique. Maybe something like "Cannot load record without an id. Please ensure the `idSpec` property of this store resolves to a unique identifier" or something?

haynesjm42: Also curious why we are just checking `isUndefined` here now instead of `isNil`? Are we ok with a record in the store having a `null` id?

lbwexler: great catch with isUndefined -- I will change that.    Regarding the message, yes was trying to reinforce to the user that when they do specify an id, they ought to make the id unique.  So I guess I can either change the message or do the unique check.

Thanks for the careful look

data/LocalStore.js
lbwexler: Actually the Ctor of Recordset should be checking the unique -- but on closer examination, its busted.  Fixing now.

PR #949, opened 2019-02-06
Add routing support to mobile navigation.

TomTirapani:
This PR adjusts the mobile navigation model to use routing, enabling deep-linking, usage with device back button, etc.

Navigation state is represented as a stack of pages. Previously, developers would navigate by using `pushPage()` and `popPage()` to adjust the page stack directly.

This PR takes a "route-first" approach, where the page stack is inferred from (and reacts to) the current route. Developers instead navigate by updating the route using `XH.navigate()`, or the provided `appendRoute()` & `back()` convenience methods.

Please also see Toolbox PR: https://github.com/exhi/toolbox/pull/152


amcclain:
Discussed just now with Tom.  He's going to look at handling for modal popups to get those to automatically dismiss on a routing change - so we don't e.g. leave a dimension chooser modal floating there if its entire page-level module gets routed away with a back-button press.

We also discussed:
* Support for specifying routes that aren't directly routable - i.e. you can't start there - and have the app respond in a sane way by unwinding the route to a valid starting point or defaulting to a base route. This avoids us setting up an expectation that all routes must be reachable from a cold-start, which in turn gives us more options for not having to encode *all* required state in the route to support things like detail panels.
* Support for routing tabs - similar to desktop. Can prioritize later as a follow-on feature, given that tabs are less of a top-level organizing concept on mobile.

mobile/cmp/navigator/NavigatorModel.js
```diff
@@ -4,78 +4,161 @@
...
-    /**
-     * @param {Object} page - configuration for NavigatorPageModel
-     * @param {function} [callback] - function to execute after the page transition
-     */
-    pushPage(page, callback) {
-        this._navigator.pushPage(new NavigatorPageModel(page));
-        this._callback = callback;
+        this.addReaction({
+            track: () => XH.routerState,
+            run: () => this.onRouteChange()
```
lbwexler: minor point, but these run functions can all just be direct references to the method, and don fat arrow functions.  `addReaction()` already calls the run function as a method on `this` by default...

core/XH.js
```diff
@@ -271,6 +271,20 @@ class XHClass {
         return this.router.navigate(...args);
     }
 
+    /** Add a routeName to the current route, preserving params */
```
lbwexler: Should we move these to routerModel?  Will application code need to call them?  Trying to make XH be more about common aliases than implementation, and to keep it from getting too bloated

TomTirapani: Mobile apps will use these as the primary methods for navigation (far more often that `XH.navigate()`). I've moved the implementation to RouterModel and left the XH aliases.

PR #957, opened 2019-02-07
Update Router5 to version ~7.0.0

febbraiod:
Migration guide can be found here: https://router5.js.org/migration/migrating-from-6.x-to-7.x

It appears that most of their changes do not apply to us, so minimal changes to Hoist needed.

### DIFF
--- a/core/RouterModel.js
+++ b/core/RouterModel.js
@@ -8,7 +8,7 @@
 import {HoistModel} from '@xh/hoist/core';
 import {observable, action} from '@xh/hoist/mobx';
 import createRouter from 'router5';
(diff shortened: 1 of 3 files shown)


lbwexler:
Great that these are minimal changes, to us, but based on the migration note, it appears its a gut renovation, and needs to be tested very carefully.

(interesting that it got rewritten in Typescript)

lbwexler:
To be clear, I vote we wait for Hoist 20 for this, unless Tom is finding stuff with his mobile routing work that would benefit from this.

lbwexler:
Thanks for this -- integrated into mobile routing PR

PR #959, opened 2019-02-08
Add dropUp property to new dateInput

febbraiod:


lbwexler:
Merging this, but my guess is we are about to whack this branch shortly :-(

PR #965, opened 2019-02-10
Misc cleanup

lbwexler:
+ This applies @managed in a few places that it was missing.

+ Also fixes an important issue with SubformFields not cleaning up after itself correctly:
*Decorators* should not be placed on super classes, but only concrete leaf classes.  Otherwise, the methods they "chain" in can get lost!

+ Also a minor refactor to DataForm to simplify and push more state into the model class.


core/appcontainer/AppContainerModel.js
```diff
@@ -27,25 +27,26 @@ export class AppContainerModel {
...
-    exceptionDialogModel = new ExceptionDialogModel();
-    optionsDialogModel = new OptionsDialogModel();
-    feedbackDialogModel = new FeedbackDialogModel();
-    impersonationBarModel = new ImpersonationBarModel();
-    loginPanelModel = new LoginPanelModel();
-    messageSourceModel = new MessageSourceModel();
-    toastSourceModel = new ToastSourceModel();
-    themeModel = new ThemeModel();
-    refreshContextModel = new RefreshContextModel();
+    @managed aboutDialogModel = new AboutDialogModel();
```
haynesjm42: Should these decorators be moved to the preceding line to be consistent with the rest of the code? Is that the standard we are trying to follow in general? I wonder if it would be worth adding to the eslint config if possible?

Admittedly this set of declarations is nice and symmetrical with the decorator on the same line ...

desktop/cmp/dataview/DataView.js
```diff
@@ -29,50 +28,20 @@ export class DataView extends Component {
...
-                    colId: 'data',
-                    flex: true,
-                    elementRenderer: itemRenderer,
-                    agOptions: {
-                        valueGetter: this.valueGetter
-                    }
-                }
-            ]
-        });
-    }
```
amcclain: This looks to me like a clear improvement - funny that we had this hanging on the component for so long, when we've always had a dedicated model class....

desktop/cmp/dataview/DataViewModel.js
```diff
@@ -40,11 +44,32 @@ export class DataViewModel {
         emptyText,
         contextMenuFn = null
     }) {
+
+        selModel = this.parseSelModel(selModel, store);
```
amcclain: This was pre-existing and applies to GridModel as well.  Given that these ctors can take a StoreSelectionModel instance or a config/string to create, we might/might not want to clean those models up when disposing of the parent model. 

I suspect most of the time we're creating them. Is this something we should be sure to handle?

I also realize I don't have a good grasp on the impact of _not_ destroying something like a StoreSelectionModel....

amcclain:
Oh dear sorry - I think I was a few minutes late and pushed my tweaks after merge.  They were in ccdfa6b8f0259829637d1b3cf96749a9fa0af495

lbwexler:
I will merge in.  Thanks for the improvement.

PR #979, opened 2019-02-15
Mobile core updates

TomTirapani:
+ Add Panel to mobile toolkit. This is a simplified version which supports standardised headers, tbar and bbar, but does not support resizing / collapsing via a PanelModel
+ Update mobile AppBar to be closer to the desktop version
+ Use hoist toolbar rather than onsen toolbars.
+ Use our mobile button wrapper for all preset mobile buttons

See Toolbox PR: https://github.com/exhi/toolbox/pull/155


amcclain:
Thanks for getting this in Tom! We have a long weekend here coming up in the US - should be able to review some time in the next three days.  Can you please add an relevant entries to the Changelog file to help outline these updates for the a release?

TomTirapani:
Great, thanks for taking a look. I've updated CHANGELOG.md

TomTirapani:
I've updated this branch:
+ Clarified how "Pages" work. "Pages" are now simply panels that can fit into Onsen's navigation and tab systems. 
+ `NavigationModel` is no longer responsible for managing and updating the app header title. Instead, the app header title is static (default app name), and the user is expected to set titles on the page itself.
+ Reorganised mobile styles, moving `xh-` styles out of the onsen kit package into associated component level style sheets.

mobile/cmp/page/Page.js
```diff
@@ -8,42 +8,69 @@
...
-            })
-        );
+        const {
+            className,
+            mask: maskProp,
+            children,
+            ...rest
+        } = this.props;
+
+        // The page itself takes the mask and className.
```
TomTirapani: Explain why page needs its own mask Tom!

lbwexler:
This looks great to me.  The use of Panel pushes us much closer to a unified model across mobile and desktop and seems like absolutely the right direction here.

Will merge now. We'll certainly continue to tune this before and after v20 release.

PR #980, opened 2019-02-18
Cross-platform Tab Container

TomTirapani:
Make TabContainerModel, TabModel and TabRefreshContextModel cross-platform.

The presentation components (TabContainer, Tab and TabSwitcher) remain platform specific.

See toolbox PR: https://github.com/exhi/toolbox/pull/156


lbwexler:
Here are my thoughts -- will try to go over with Tom in person as well:
* Can we have a cross-platform "TabContainer" as well?  If necessary,  it can delegate or re-export an implementation class to actually do the rendering.  This would be similar to our cross-platform grid package setup, and would mean the entire tab setup would just appear cross-platform to most application usages.  (I see why the tab switcher is not cross-platform, and that certainly makes sense, but most apps won't need to create one of those).

* Is icon now supported on desktop?  It did not seem like it was.  This might be confusing to users.

* Should we take advantage of this opportunity to just move the switcher position to the TabContainerModel, so that the component becomes prop-less? ( As a related matter -- its fine if mobile is unable to support various switcher positions and throws.)

TomTirapani:
Thanks for taking a look:

- I was following the DimensionChooser pattern of having a shared model with platform-specific views. Can certainly see the advantage from a user's perspective of having a TabContainer from a cross-platform package. Onsen's tabbar complicates things somewhat by not allowing us to easily separate a TabContainer from a TabSwitcher on the mobile side - the TabContainer _is_ the switcher. Delegating to a dynamic implementation class is one solution, or perhaps it's time to wind down Onsen's tabbar and roll our own TabSwitcher?

- icon is not supported on desktop, good catch. Will see if we can support on desktop. Icon is quite important on mobile.

- Sure thing, will do.

- Tab routing is not supported on mobile - trying to set up a route for TabContainer on mobile will be ignored with a warning. My thinking is that, on mobile, the routing is very closely tied to the navigation page stack, with changes to the route representing moving "forwards" and "backwards" through this stack. Using route changes to handle moving "sideways" between "siblings" like tabs seemed to add a lot of complexity, and wasn't sure it was worth it. Tabs are not a "primary navigation" method on mobile like they are on desktop. Happy to discuss further.

PR #981, opened 2019-02-19
Admin RestForm readonly formatting

shravnk:
- Display readonly admin grid lastUpdated with hh:mm:ss


lbwexler:
Hi -- Thanks for this -- was very helpful.  

Ended up going with the following -- thinks its more appropriate for a Date to give the full info by default.  We have an outstanding ticket for better support for a "ValueDate" -- a date without time -- hopefully we will be moving toward using that for most places where we just want to save a calendar day.

PR #990, opened 2019-02-25
Add clone action to rest grid

lbwexler:


lbwexler:
Seem like its a good start.  Can we build in to an admin grid (e.g. config grid?) to see what that looks like and have a built in example there right in hoist-react?

I do agree that this as a variant of 'add', and in particular closer to an 'add' that an 'edit'.  So I agree with your choice to return isAdd as `true` in this case. However very counter-intuitive to enter into it by reusing the openEdit() method.  Why not add an `openClone` method, or alternatively add an optional `clone` property to the the `openAdd` method?

haynesjm42:
Closing in favor of https://github.com/exhi/hoist-react/pull/992 with master branch as target

amcclain:
@haynesjm42 / @nadkarniarjun can we delete this branch?  I assume so but didn't want to jump the gun - please delete if not needed.

haynesjm42:
Yes we can, doing so now.

PR #991, opened 2019-02-25
Remove Label Component from input package

cnrudd:
Replaced with Tags/label in two files where it was used.  (EXHI-986)


amcclain:
Could you please add a one-line CHANGELOG entry?  Much appreciated.

cnrudd:
> Could you please add a one-line CHANGELOG entry? Much appreciated.

Done.

PR #992, opened 2019-02-25
RestGrid Clone action

haynesjm42:
Initial working version of Clone action.


haynesjm42:
Encountered an issue/caveat with FormModel with how this is currently implemented: Since we are initializing the values of the FormModel with the clone object data, the initial state of the form is not dirty, so the Save button is disabled. 

We have a couple options here:
1. Init the FormModel with an empty object, and set the field values with the clone data. This sort of makes sense since we are effectively just auto-filling the fields for an Add operation, but not sure if this is ideal. Would probably necessitate addition of a `setData` method on FormModel.
2. Allow initializing the FormModel as dirty. Seems like this could be useful for edge-cases like this, but may make the FormModel api more confusing?

haynesjm42:
Also, while looking at the RestFormModel code, looks like we are potentially leaking a FormModel every time we call `initForm`? Should we have a call to `XH.destroy(this.formModel)` before creating the new one?

data/Record.js
```diff
@@ -67,6 +67,14 @@ export class Record {
         });
     }
 
+    /**
+     * Return an object containing the current values of all fields in this record.
+     * @returns {Object}
+     */
+    getData() {
```
haynesjm42: This seemed like a generally useful method to have for Records, and something we needed for setting up the clone data, so added this here

lbwexler: Yes, although does really suggest we should have stored the data in its own bundle from the get go,

lbwexler:
Yes, regarding the leak.  Good Catch.  

Regarding the disabled add button, I think the right move is just to have it always enabled for an "isAdd || formModel.isDirty"  Note that there might be cases with raw add as well as clone where the default values are enough and user doesn't need to actual make any changes.  If any required values are missing, that will be picked up when the user tries to add!

lbwexler:
Will also need a description in changelog as well as a usage in toolbox and/or admin in order to merge.

nadkarniarjun:
Hi all, 

Just checked in changes I discussed with John. Specifically:
- Changed the `Save` button disabling logic to also include the `isAdd` flag
- Called `XH.safeDestroy(this.formModel);` before instantiating a new `FormModel`
- Added description to `CHANGELOG.MD`

I also created a restCloneAction branch in Toolbox, where I added cloneAction as an example in the Rest Grid tab, and passed in a sample prepareCloneFn to show how we can transform the record prior to form init.

amcclain:
Just tried this out in the Toolbox RestGrid example.  I see that the lastUpdated and lastUpdatedBy fields are both brought forward in the cloned record, and appear in the new "add record" dialog. This is weird and doesn't seem correct in this instance.

We could customize this particular example via the prepFn, but in general I think we should avoid populating non-editable fields. Presumably these are always null when adding a new record (is that accurate?) and the clone op is closer I believe to adding a new rec than it is to editing an existing one.

Does that sound correct?  Pls do think through carefully as I admit I am juggling a few things and don't want to miss any subtleties.

admin/tabs/general/configs/ConfigModel.js
```diff
@@ -68,6 +75,19 @@ export class ConfigModel {
...
+            editAction,
+            cloneAction,
+            deleteAction
+        ],
+        menuActions: [
+            addAction,
+            editAction,
+            cloneAction,
+            deleteAction
+        ],
```
amcclain: I don't have a better proposal at the moment, but it does seem unfortunate that we need to specifically list all these out (in a particular, hopefully standard order) to enable the clone action.  

I understand how having the API take a list of actions for these various sections of the grid makes sense - and how boolean flags would rapidly spin out of control.  

Mentioning in case anyone had any ideas.  Happy to proceed though as again I don't have a better plan 🤷‍♂️

data/Record.js
amcclain: Yes - filed #998 to track / review.  

We have discussed how it is disquieting that we mix in the record data fields as top-level properties - at the least it necessitated https://github.com/exhi/hoist-react/blob/395e285e9281373aad81574e1d6dca5c302fa24b/data/Record.js#L18

desktop/cmp/rest/RestGridModel.js
```diff
@@ -95,6 +106,7 @@ export class RestGridModel {
      * @param {Object[]|RecordAction[]} [menuActions] - actions to display in the grid context menu. Defaults to add, edit, delete.
      * @param {Object[]|RecordAction[]} [formActions] - actions to display in the form toolbar. Defaults to delete.
      * @param {Object} [actionWarning] - map of action (e.g. 'add'/'edit'/'delete') to string.  See default prop.
+     * @param {function} [prepareCloneFn] - a function used by the clone action to prepare a clone of a record.
```
amcclain: Please look at documenting the signature here via an `@callback` typedef - e.g. https://github.com/exhi/hoist-react/blob/395e285e9281373aad81574e1d6dca5c302fa24b/data/RecordAction.js#L146

desktop/cmp/rest/impl/RestFormModel.js
```diff
@@ -93,9 +100,11 @@ export class RestFormModel {
     //---------------------
     initForm(rec) {
         this.currentRecord = rec ? rec : {id: null};
-        this.isAdd = !rec;
+
+        this.isAdd = !rec || !rec.id;
```
amcclain: Would use `isNil` - not positive we have complete support for id:0 but no need to trip up here.

data/Record.js
haynesjm42: Would we prefer to remove this new method for now in favor of just doing this work in RestGridModel (where it is also going to need to omit non-editable properties anyways)? Probably best to avoid adding this new method if we are going to provide a different interface for retrieving the data in a future release

admin/tabs/general/configs/ConfigModel.js
haynesjm42: I had to use this pattern for adding an extra action for a few rest grids in a client-app and ended up creating a separate list to hold these actions for setting `toolbarActions` and `menuActions` on the RestGridModel, which at least avoids the dupe.

I wonder if we should just put the add, edit, delete actions in some `defaultRestActions` constant or something so apps could do `toolbarActions: [...defaultRestActions, myAction]`, though if you didn't want the default ordering you'd still need to unroll the list. There is something to be said about having Add, Edit, Delete always together in that order whenever you see a RestGrid, with any non-default actions coming after, though I always tend to put Clone/Copy actions next to Edit when I am building something like this.

Could see us having a few pre-build lists of actions: `defaultRestActions` (Add, Edit, Delete), `allRestActions` (Add, Edit, Clone, Delete)

haynesjm42:
Re: the inclusion of non-editable fields when cloning, I think we should definitely exclude those when cloning, using the same logic as here: https://github.com/exhi/hoist-react/blob/develop/desktop/cmp/rest/data/RestStore.js#L76

A non-editable fields would never be required when creating a new one so I think this is the correct approach.

admin/tabs/general/configs/ConfigModel.js
amcclain: Exactly - I was thinking that a `defaultRestActions` collection could be a small but meaningful aid to consistency - ala https://github.com/exhi/hoist-react/blob/395e285e9281373aad81574e1d6dca5c302fa24b/cmp/grid/GridModel.js#L90 in the grid context menu.  

But you are also right that it seems weird to have clone after delete....  

I vote let's leave it alone for now and go with the approach we have here (no changes to this code). We can feel it out a bit more - hesitant to add more utils or defaults if we're not really sure on it.  Thanks for discussing.

CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 
 ## v20.0.0-SNAPSHOT (unreleased / under development)
 
-* TBD
+### 🎁 New Features
```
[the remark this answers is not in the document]
nadkarniarjun: Done. Updated to v19.1.0 in changelog.

desktop/cmp/rest/RestGridModel.js
nadkarniarjun: Done. Added following documentation, and renamed callback function to `prepareCloneCb`.

`    /**
     * @callback prepareCloneCb
     * @param {record} original record
     * @param {clone} record to be cloned
     */
`

desktop/cmp/rest/impl/RestFormModel.js
nadkarniarjun: Done.

data/Record.js
amcclain: Yes I would support that change, for exactly those reasons.

amcclain:
I'm sorry @nadkarniarjun that I wasn't more clear - I did not mean to suggest renaming the actual `prepareCloneFn` config - I was just looking to add the typedef - staying as consistent as possible with existing examples like https://github.com/exhi/hoist-react/blob/be03128c8b8b3a396da99b202187fb93c0c05d26/data/RecordAction.js#L44.

I would also move the callback definition to the bottom of the class, consistent with the others so it's not right in the flow of the code.  The main help it provides is with the tooling integration and at some point hopefully generated docs.  Putting it in between the constructor and its doc comment likely breaks the ctor doc.

Finally....the callback def isn't quite correct, in that it accepts an object and that object has two properties.  The RecordAction callbacks provide an example there too.

nadkarniarjun:
No worries @amcclain. I made the documentation changes, and changed it back to `Fn` from `Cb`. Let me know if there's anything else you would like me to correct.

data/Record.js
lbwexler: great point john H, and I agree.  Looks like this still needs to be done

lbwexler:
Re-opening against develop

PR #1000, opened 2019-03-01
Customize RestGrid toolbar

shravnk:
This was for an old ticket - original goal was to simply allow RestGrid's extraToolbarItems prop to take an array or function.

I've also included the ability to customize the top right toolbar buttons with an additional prop. I recently worked on an app in which those buttons needed to be excluded. 

I don't really like two separate toolbar props on RestGrid, but it does have the benefit of backwards compatibility.


lbwexler:
Not crazy about the lack of symmetry here in the API with the additional change to add rightItems.
Would prefer we do nothing with regard to that.  In the future, we could provide a way to override the entire toolbar (including the actions) by creating a tBar property that would default to the current toolbar, but could be nulled out, or replaced entirely.

PR #1004, opened 2019-03-04
Allow client-side timer intervals to be changed dynamically

shravnk:


shravnk:
A couple of questions:

1. Assuming we want a single app-level timer to fetch configs (as I have implemented), should that interval itself be a config?

2. I started by mirroring Hoist Core's Timer, and am still using the `coreTimer` [pattern](https://github.com/exhi/hoist-core/blob/0c5e0cc0c685f46efc1ce7c827706e1139bb53cf/src/main/groovy/io/xh/hoist/util/Timer.groovy#L135). Am wondering if instead ConfigService's `_data` object should be an observable.ref, and the timer instance should update its interval via an `addReaction`.

svc/ConfigService.js
```diff
@@ -51,5 +60,20 @@ export class ConfigService {
...
+        Timer.create({
+            runFn: this.checkConfigUpdates,
+            interval: this.REFRESH_INTERVAL * SECONDS
+        });
+    }
+
+    checkConfigUpdates = async () => {
+        const data = await XH.fetchJson({url: 'xh/getConfig'});
+        if (!isEqual(data, this._data)) this._data = data;
+    }
```
lbwexler: OK -- Actually having auto-reloading configs on the client is something we are very much thinking of supporting in hoist central, but it was not my intention to have as part of this "Timer-Only" change.  Would like to shelve the change in this file for now.

lbwexler: And in fact, what's most likely for a client -side timer is going to be responding to a pref.  

This all relates to some related changes Paul just did in Hoist Central -- To make this properly flexible, we should take a "function/closure" as an option to get the interval dynamically!

lbwexler: This looks very promising, Brendan -- let's discuss in person -- there a couple of bits I think we need to get, especially handling of calls that take longer than the interval, and avoiding piling them up.

shravnk: I see - so we won't be able to fully replicate the server side functionality until configs are auto-reloaded. But for now, a timer should be able to take three arguments for its `interval`:

1. Integer (current)
2. String (checks for config, purely a convenience at this point)
3. Closure, which could be used to get either a pref or config + add custom logic

lbwexler:
Just 1 && 3.  2 becomes unneccessary.

lbwexler:
Guys -- I am really sorry for this -- we need to hold off on this. Thanks so much for your hard work on this Brendan, but I am concerned that putting this on the client was potentially a mistake.  

My immediate concern is that this would run the closure to check the interval and timeout once a second. (the server has a seperate timer for checking its config). Seems weird (and against user expectations) to poll this closure once a second, but would also be weird to have an arbitrary delay on the client.

lbwexler:
Checking this in because I had it -- would allow dynamically changing the interval.  Not sure if the complexity is worth it, but I think it is solid

lbwexler:
Renaming to represent where we landed on this one.

Issue #1014, opened 2019-03-08
Store should implement destroy(), be @managed as appropriate


amcclain:

lbwexler:
I put in the base implementation for Store.destroy().  I did not see it or any of its sub-objects creating any objects that created reactions or otherwise would warrant explicit destruction, so I think we can probably close out this bug at this point.

PR #1015, opened 2019-03-08
GridModel.store config accepts a plain object, creates LocalStore

### LINKED ISSUE #1013: GridModel - accept plain object for store config, construct default LocalStore
Currently the GridModel ctor store config requires a fully instantiated store. Often we are constructing a new LocalStore inline, requiring the class import and the extra boilerplate of the store constructor.

We could support a plain object instead, which if passed would construct a LocalStore (LocalStore being the most common / basic concrete Store class.)

amcclain:
+ Clean up instances where we were creating a new LocalStore inline
+ Mark created store as managed within GridModel. Note that stores do not yet actually implement destroy() but will when #1014 complete.
+ Fixes #1013

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v21.0.0-SNAPSHOT (under development)
+
+### ⚙️ Technical
+
+* The `GridModel.store` config now accepts a plain object and will internally create a `LocalStore`.
+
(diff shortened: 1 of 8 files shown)


amcclain:
Note there are a number of instances in Toolbox that could be cleaned up.  I did not jump on that yet, though, pending this review. Figured we had enough concrete examples within this project to test.

PR #1016, opened 2019-03-08
Refresh ideas

lbwexler:


amcclain:
Looking forward to discussing.  

On a related note, I was about to file a ticket to see if we can make `lastLoadCompleted` observable, to facilitate its use with a relativeTimestamp.  Thoughts there? Happy to ticket if doable, unless you thought it expedient to simply fold in here.

(I was also wondering if relativeTimestamp should take an optional model prop, from which it could source lastLoadCompleted as a default timestamp - avoiding potential outer re-renders.)

lbwexler:
Great ideas -- I'll take a look.

Regarding this would we need to distinguish between successful load and
failed load?

Or to put it another way:  What is the contract for an app that fails to
load its data due to some kind of business condition?  Should it still
resolve cleanly, and indicate the failure through other means?  Or should
it throw from doLoadAsync() and expect that to be recognized as a separate
state?




-Lee

PR #1017, opened 2019-03-08
DataViewModel - remove dupe references to GridModel properties

amcclain:
+ Did not see a need to save extra references to the ctor params, added getter for store and selModel.
+ Rely on GridModel to parse selModel and store configs, including recently added support for plain object store configs.


amcclain:
Note this would be a breaking change if any apps were relying on the references to itemRenderer or contextMenuFn on the model class, but I don't see why they would be.  Let me know if I am missing anything.

I did review the `parseSelModel()` function and didn't see any meaningful difference in defaults, etc.

lbwexler:
Great cleanup! 👍

PR #1022, opened 2019-03-12
Make GridModel.store config optional, source store fields from columns

amcclain:


lbwexler:
This removes a lot of boilerplate --so am sympathetic.

But it does seems to really discourage the use of type information or other information about store fields.  

For example, if I wanted to provide a bit of metadata for some fields -- Label?  nulllable? defaultValue?
I have to specify *all* of the fields.  I am much less likely to do that, if I can be tempted by this auto-generate capability.

Could I suggest an alternative -- that if a field specified in a column is missing from the fields array of an auto-generated store,  then and only then do we populate it?  This I think is more along the theme of passing a proper partial for the store, which include whatever bits (including, say 'idProp') that are neccesary.

amcclain:
Good point - please take a look at latest and see if that's what you were thinking.  👍

PR #1023, opened 2019-03-12
Remove 'delay' workaround to avoid focus blurring.

lbwexler:
This workaround was causing tricky problems for apps that were getting stale data during this interval.


amcclain:
Will merge and happy to avoid exactly this kind of "competing delay" issue. I confess I don't really understand why it was there to start ("prevents extraneous 'flapping' of focus state at this level" ?) but lets drop it.

PR #1025, opened 2019-03-12
Grid store fields from cols

lbwexler:


lbwexler:
Take 2.  Thanks for catching my error.

PR #1028, opened 2019-03-15
Don't re-render tabs that are not changing their active state

lbwexler:


amcclain:
Nice - this seems like a cool fix.  Am curious how you came to find this issue / solution - hoping we can discuss anytime.

PR #1029, opened 2019-03-18
Allow FormField labelWidth to override min width.

TomTirapani:
I moved the application of the min width out of the css and into JS, and only apply it if a specific labelWidth has not been provided. This maintains the current expected behaviour, while allowing labels narrower than 80px.

Another potential solution would be to remove the concept of a min width entirely - labels are either sized explicitly, or use their natural size. This may be a simpler API, but may potentially be a breaking change to any apps currently depending on the min width.

### DIFF
--- a/desktop/cmp/form/FormField.js
+++ b/desktop/cmp/form/FormField.js
@@ -6,7 +6,7 @@
  */
 import React, {Component} from 'react';
 import PT from 'prop-types';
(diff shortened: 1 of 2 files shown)


amcclain:
Thanks for fixing. I agree that the minWidth might is likely not adding much at this point, and I would support removing it if we found ourselves bumping up against this again and wanting to change it.  

For now, I am happy to merge this as a clear incremental improvement.  We can keep an eye on it.

PR #1030, opened 2019-03-18
Hooks

lbwexler:


admin/tabs/activity/feedback/FeedbackPanelModel.js
```diff
@@ -0,0 +1,84 @@
...
+            {field: 'browser'},
+            {field: 'device'},
+            {field: 'appVersion'},
+            {field: 'appEnvironment'},
+            {field: 'dateCreated'}
+        ]
+    });
+
+    async doLoadAsync(loadSpec) {
+        await this.gridModel.doLoadAsync(loadSpec);
```
amcclain: We want `loadAsync()` vs. the impl method, right?

admin/tabs/activity/clienterrors/ClientErrorDetail.js
```diff
@@ -4,79 +4,69 @@
...
-            }),
-            jsonInput({
-                value: rec.error,
-                disabled: true,
-                height: 450,
-                width: '100%',
-                editorProps: {lineWrapping: true}
+//----------------
+// Implementation
+//----------------
```
amcclain: Trivial point, but I am thinking we could lose all these Implementation block comments where we are building up a component from non-exported functions.

core/hooks/Model.js
```diff
@@ -0,0 +1,114 @@
...
+
+    // Recognize any disallowed dynamic change to the model
+    if (state.instance && !state.isOwned && propsModel !== state.instance) throwModelChangeException();
+
+    // If we don't yet have a model, potentially source from props, instantiating if appropriate.
+    if (!state.instance && propsModel) {
+        if (isPlainObject(propsModel)) {
+            if (modelClass) {
+                state.instance = new modelClass(propsModel);
+                state.isOwned = true;
```
amcclain: As discussed on IM today, wary of mutating state directly here. Should we revisit with `useRef()`?

admin/tabs/activity/clienterrors/ClientErrorDetail.js
lbwexler: Agree

admin/tabs/activity/feedback/FeedbackPanelModel.js
lbwexler: Great catch, whacked this whole class in favor of anonymous model

core/hooks/Mount.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import {useEffect} from 'react';
+
+/**
+ * Hook to run a function once after component has been mounted.
+ * @param {function} fn
+ */
+export function useOnMount(fn) {
+    useEffect(fn, []);
```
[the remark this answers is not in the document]
lbwexler: very subtle point and nice catch - fixed

core/HoistComponent.js
```diff
@@ -5,57 +5,78 @@
...
+    return [component, elemFactory(component)];
+}
+
+
+/**
+ * Create a Class Component in Hoist.
  *
  * Adds support for MobX reactivity, model awareness, and other convenience methods below.
+ *
+ * NOTE: This decorator provided the original method for specifying class-based components within
```
haynesjm42: I would like to understand exactly what you lose (or gain) by using a functional component over a class component. (using this comment for my own note taking and thoughts)

ManagedSupport is lost, and seems that the preferred way to handle any additional cleanup in components is to use the `useOnUnmount` hook (which makes perfect sense and is consistent with React best practices). Could see some benefit to a `useManagedState` or similar hook which similar functionality, though in general I think we want to encourage putting state in the model as much as possible

ReactiveSupport is lost, which makes sense since there doesn't seem to be any hook equivalent to a class component constructor (and also should not be needed in most components except in very special cases)

XhIdSupport is lost, and we could write a hook to provide the same functionality, though I've never used it on the component before, only the model. Have there been use-cases where a component id is needed instead of the model id? Don't see how that would be useful on the component since its own key should come from its parent.

core/hooks/ClassName.js
```diff
@@ -0,0 +1,25 @@
...
+ * outermost (or otherwise most appropriate) rendered component.
+ *
+ * @param {String} baseClassName - base class name for the component.
+ * @param {Array} props
+ * @param {...string} extraClassNames - additional classNames to append.
+ *
+ * @returns {String[]} - Array of class names appropriate for html className attribute
+ */
+export function useClassName(baseClassName, props, ...extraClassNames) {
+    return classNames(baseClassName, props.className, ...extraClassNames);
```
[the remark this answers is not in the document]
lbwexler: At this point, I guess not -- I was just being overly conservative with the existing structure.  

Should we continue to take props generally and have it be a wrapper around classNames?  If we keep the wrapper, should we take the opportunity to have it take an object for more flexibility going forward? e.g.

xhClassName({baseName, props, extraNames:[]})

core/hooks/LayoutProps.js
```diff
@@ -0,0 +1,34 @@
...
+ * Components will typically delegate this responsibility to a child component that also
+ * implements LayoutSupport. `Box` is typically the Component that is ultimately rendered
+ * and will handle this by outputting a div with appropriate styles.
+ *
+ * @param {Array} props
+ * @returns {Object[]} - Array containing a bundle of layout props, and a bundle of all non-layout props.
+ */
+export function useLayoutProps(props) {
+    // Consider caching this comp if props haven't changed.
+    return [getLayoutProps(props), getNonLayoutProps(props)];
```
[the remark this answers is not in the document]
lbwexler: I think this is a better candidate for a hook than the class names because it is such a core cross-cutting bit of Component functionality, and I could see it relying on the component lifecycle, or caching, or some specialized logging/correctness checking that happens  only on mount?

If I understand useMemo correctly, we would actually need to specify the full lot of the parameters in the dependencies array, so I am not sure that is actually a good use at the moment -- although I wonder if it might change if we ever did more complicated layout computations.

I could actually imagine an optimization that just compares the provided props *that are not functions* against each other (for example)..... Or an option on the hook that declares itself for
"static" layout only .

cmp/layout/Box.js
```diff
@@ -17,50 +16,33 @@ import {div} from './Tags';
...
-@LayoutSupport
-export class Box extends Component {
-    render() {
-        let {children, ...props} = this.getNonLayoutProps();
-        props = merge(
-            {style: {display: 'flex', overflow: 'hidden', position: 'relative'}},
-            {style: this.getLayoutProps()},
-            props
-        );
+export const [Box, box] = hoistComponent(function Box(props) {
```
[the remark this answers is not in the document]
lbwexler: John:  named function here is for the benefit of how the component will appear in the React developer tools.  Otherwise our component will appear anonymous.  At Anselm's suggestion, I did this for a number of core low-level components.

Its actually annoying needing to specify the name of the component *three* times -- and spoiling the fun of removing the elemFactory() and its requirement to repeat the Component's naem.

core/HoistComponent.js
```diff
@@ -5,57 +5,78 @@
...
- * Exceptions include highly specific low-level components provided to other APIs which may be
- * negatively impacted by the overhead associated with this decorator.
+ * @param {function} renderFn - function defining a React component.
+ * @returns {Object[]} - Array containing the Component and an elemFactory for the Component.
+ *
+ * @see HoistComponent decorator for a ES6 class-based approach to defining a Component in Hoist.
+ */
+export function hoistComponent(renderFn) {
+    const component = observer(renderFn);
+    return [component, elemFactory(component)];
```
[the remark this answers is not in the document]
lbwexler: Well, I am personally with you John, but the design of this was actually designed to *encourage* the export of the Component itself, and thereby keep us honest, in providing full support for JSX for our public APIs.

Agree the `[, myComponent]` pattern is a bit annoying , and your suggestion also crossed my mind.  I would certainly be open to it.  Especially useful for local components where you don't even intend to export anything!

-Lee

core/HoistComponent.js
lbwexler: As Joni Mitchell wrote in 'Both Sides Now'
>>Well something's lost, but something's gained ...  In living every day

Think you pretty well stated things.  I think just staying with the emerging standard React architecture going forward is the overriding value here which trumps a lot of the pros and cons that we could discuss.

Yes, I agree that the change toward functional components ideally obsoletes much of the lost functionality you mention.   The idea is to just aggressively push as much state as possible in to our mobx models, leaving the components as anonymous rendering functions....

So hopefully not too much lost, and wondrous things still to be gained.....

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@
   TreeMap visualization based on the Highcharts library. Supports optional binding to a GridModel,
   which syncs selection and expand / collapse state.
 
+### 💥 Breaking Changes
+
+* If using JSX, component ToolbarSeparator has been renamed ToobarSep
```
haynesjm42: ToobarSep? 😄

haynesjm42:
Just noticed the ToolbarSeparator rename in these changes (thanks to Colins changelog update), and looks like we still have AppBarSeparator, and mobile still has ToolbarSeparator... if we are having the breaking change for ToolbarSeparator can we just make all of our separator component names consistent?

PR #1032, opened 2019-03-19
Updates to Grid style / visual config and CSS

amcclain:
+ Support rowBorders and stripeRows observables on GridModel.
+ Borders off by default, row striping remains on.
+ Move Grid.showHover prop -> GridModel.highlightOnHover.
+ Update grid variant CSS class names to BEM conventions.
+ Ensure Hoist border-color CSS var actually applies, then set to a lighter value.

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -83,12 +83,7 @@ export class Grid extends Component {
          * Callback to call when a cell is double clicked. Function will receive an event
          * with a data node, cell value, and column.
          */
-        onCellDoubleClicked: PT.func,
-
-        /**
-         * Show a colored row background on hover. Defaults to false.
-         */
-        showHover: PT.bool
+        onCellDoubleClicked: PT.func
     };
 
     static ROW_HEIGHT = 28;
@@ -124,8 +119,9 @@ export class Grid extends Component {
     }
 
     render() {
-        const {compact, treeMode} = this.model,
-            {agOptions, showHover, onKeyDown} = this.props,
+        const {model, props} = this,
+            {treeMode, compact, highlightOnHover, rowBorders, stripeRows} = model,
+            {agOptions, onKeyDown} = props,
             {isMobile} = XH,
             layoutProps = this.getLayoutProps();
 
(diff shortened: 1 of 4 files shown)


amcclain:
Note this is technically a breaking change due to the rename of the (relatively minor) Grid.showHover prop and tweaks to grid CSS classes, but I think these are small / edge enough to take into a minor version release if we wanted.

cmp/grid/GridModel.js
```diff
@@ -479,6 +494,12 @@ export class GridModel {
         return c.children ? new ColumnGroup(c, this) : new Column(c, this);
     }
 
+    // Misc setters for visual/style observables.
+    @action setCompact(compact)                     {this.compact = compact}
+    @action setHighlightOnHover(highlightOnHover)   {this.highlightOnHover = highlightOnHover}
+    @action setRowBorders(rowBorders)               {this.rowBorders = rowBorders}
+    @action setStripeRows(stripeRows)               {this.stripeRows = stripeRows}
+
```
lbwexler: why not just make these @bindable?

cmp/grid/Grid.js
```diff
@@ -144,9 +140,11 @@ export class Grid extends Component {
                 className: this.getClassName(
                     'ag-grid-holder',
                     XH.darkTheme ? 'ag-theme-balham-dark' : 'ag-theme-balham',
-                    compact ? 'xh-grid-compact' : 'xh-grid-standard',
-                    treeMode && this._isHierarchical ? 'xh-grid-hierarchical' : '',
-                    !isMobile && showHover ? 'xh-grid-show-hover' : ''
+                    treeMode && this._isHierarchical ? 'xh-grid--hierarchical' : 'xh-grid--flat',
+                    compact ? 'xh-grid--compact' : 'xh-grid--standard',
+                    rowBorders ? '' : 'xh-grid--no-row-borders',
+                    stripeRows ? '' : 'xh-grid--no-stripes',
```
amcclain: We could go ahead here and flip the sense of these classes to consistently add formatting, especially since it looks like we'll land on a default of no borders.

cmp/grid/GridModel.js
amcclain: I would be happy to, and had started out that way, but we had a pre-existing `@action setCompact()` and I recalled some discussion a while ago about whether or not we should use bindable for core toolkit classes such as GridModel that expose the setters as public API methods.  The fact that we didn't have bindable imported on the class led me to think we might be avoiding it on purpose.

I can see some advantage to having the explicit methods for IDE support as well, although I didn't add jsDoc comments (yet).  Having gone through all that, I think I slightly favor having these methods written out in the class, but don't feel very strongly.

lbwexler:
Hi -- making some tweaks here and will merge shortly.  I think its ok and preferable to continue to rely on bindable now (and since you said you didn't feel too strongly... :-))  If we decide it's really inappropriate and should be replaced with setXXX(), I think we could look to do that in a consistent bombing.

PR #1033, opened 2019-03-20
Preserve tree grid filters at all levels

shravnk:


amcclain:
Thanks for putting this together - I understand from Lee it's a critical fix.  Is there a comment we could add here, or expand on, to clarify the logic here?  

I'm of the opinion that whenever we fix a subtle-but-critical bug it's worth a careful comment, and scanning this section of the code (admittedly without a lot of context) I'm still not sure exactly what's going on... 😃

amcclain:
Please also add a changelog entry in this branch to record the v19 patch release - that way when we merge this we'll catch the changelog as well.  

(Those special hotfix releases can be especially important to get into the changelog, as is is very easy to forget all the details after the initial rush to fix, and they can cause confusion later on.)

amcclain:
Hi @shravnk could you please update with the above tomorrow when you have a chance?  Would love to release a v20.2 and this important fix should definitely be in it.  Ping me anytime if you want to go over any wording or anything - can keep it very minimal, just want to stick to our good changelog process.

shravnk:
Hey, sorry this slipped through the cracks. Should be good to go now

PR #1035, opened 2019-03-22
Add optional queryBuffer to use with select's queryFn.

### LINKED ISSUE #1012: Select - optional debounce for async queryFn
Our wrapper around react-select accepts an async queryFn and wraps that with some post-processing to pass on to the library's `loadOptions` prop:

https://github.com/exhi/hoist-react/blob/c678dacfe3143c79b754f7c3089f23397e337215/desktop/cmp/input/Select.js#L276

React-select maintainers are [looking at adding debounce support](https://github.com/JedWatson/react-select/pull/3254), but if we can't wait or our implementation requires some custom support, we will need to provide on our own.

TomTirapani:
Fixes  #1012


amcclain:
Hi Tom - I tried this out in Toolbox, looking at the existing select example we have in our form controls panel with and without the new queryBuffer configured.  

Unfortunately the control did not appear to be working properly. Simple searches did not come back with any / expected results - it wasn't anything too subtle - give it a go and compare with the deployed toolbox-dev version and you should see what I mean. Let me know if you can replicate?

I am sorry I don't recall more of the details of my initial troubleshooting, but I recall going into this thinking it would be relatively straightforward, but after one or two attempts that might have been similar to this decided it was not... Happy to discuss more anytime.

PR #1037, opened 2019-03-26
GridModel option to show cell focus

### LINKED ISSUE #946: Allow Grid to show cell selection
Our current Grid css hides cell selection with:

 .ag-cell-focus {
    border-color: var(--xh-grid-cell-focus-border-color) !important;
    outline: 0 !important;
  }

It is sometimes nice to see cell selection.  We should use a boolean prop to the Grid component that when true will add a css class to the grid to use alternate css that shows cell selection:

 .ag-cell-focus {
    border-color: var(--xh-green) !important;
    outline: 1px !important;
  }

While on the subject of cell navigation....
When cell navigation is visible, a tree group should open only when the left most cell is selected and the user hits the right arrow.  Right now, the group opens when any cell in the group row is selected and the user hits the right arrow.

Same for closing a tree group:  When cell navigation is visible, a left arrow event should only close the tree group when the left most cell is selected.

amcclain:
+ Fixes #946
+ Tweak to also avoid (existing) shifting on right-aligned focused cells
+ Lighten up grid row borders in general if shown

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -86,13 +86,15 @@ export class GridModel {
     @observable groupBy = null;
 
     /** @member {boolean} */
-    @bindable compact = false;
+    @bindable compact;
     /** @member {boolean} */
-    @bindable highlightOnHover = false;
+    @bindable highlightOnHover;
     /** @member {boolean} */
-    @bindable rowBorders = false;
+    @bindable rowBorders;
     /** @member {boolean} */
-    @bindable stripeRows = true;
+    @bindable stripeRows;
+    /** @member {boolean} */
+    @bindable showCellFocus;
 
     /** @member {GridApi} */
     @observable.ref agApi = null;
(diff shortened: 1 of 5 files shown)


amcclain:
Note there is a corresponding branch in Toolbox to add a new switch to our sample grids and DRY that up a bit.  They should be merged together if looking good.

amcclain:
I admit that renaming `showHover` does not make much sense in the light of then choosing `showCellFocus` as this new prop.  I dunno - something about `showHover` seemed like it was not quite clear/right at the time, but I'm having a hard time justifying the rename now and would be happy to keep as it was - but still move to the gridModel vs. a grid prop.

cmp/grid/GridModel.js
```diff
@@ -86,13 +86,15 @@ export class GridModel {
...
+    @bindable compact;
     /** @member {boolean} */
-    @bindable highlightOnHover = false;
+    @bindable highlightOnHover;
     /** @member {boolean} */
-    @bindable rowBorders = false;
+    @bindable rowBorders;
     /** @member {boolean} */
-    @bindable stripeRows = true;
+    @bindable stripeRows;
```
lbwexler: glad you removed these meaningless defaults

Issue #1040, opened 2019-03-28
HoistInput: Use onKeyDown rather than onKeyPress


TomTirapani:
Should we consider replacing the numerous uses of `onKeyPress` in HoistInputs with `onKeyDown`? There are two motivating factors:

1) The `keypress` event is deprecated: https://developer.mozilla.org/en-US/docs/Web/Events/keypress. The recommendation is to replace with `keydown`. Not expecting browser support to disappear anytime soon, but I think would be better to be up to date.

2) The `keypress` event only fires for certain keys - specifically, keys that "produce a character value (alphabetic, numeric, and punctuation keys)". We recently wanted to use it to listen to arrow keys - it did not fire for these keys, yet `keydown` would have.

amcclain:
Sounds like a pretty clear move to me - could there be any downside / usages we can anticipate breaking?  I say let's go ahead.

TomTirapani:
https://github.com/exhi/hoist-react/pull/1043

amcclain:
Great thanks - will review and test it out soon.  These inputs are tricky enough that I do want to ensure we test thoroughly, but change looks good and I am sure we will take it shortly.

PR #1043, opened 2019-03-28
Use onKeyDown rather than onKeyPress consistently across HoistInputs.

TomTirapani:
Also added onKeyDown prop to desktop TextArea, for symmetry with other inputs & mobile.
#1040


lbwexler:
This looks great to me.  See my tiny tweak to help ensure a safe transition to the API.

Issue #874, opened 2018-12-18
Managed display of docked / modal / minimized sub-components


amcclain:
Add support for Gmail-style management of dialogs and related "pop-up" components, such as editor or detail forms, that can be registered with a parent component and be toggled by the user between the following states:

* Docked - snapped to bottom edge of parent. Non-modal to permit ongoing interaction with parent and its primary children.
* Modal - centered floating modal dialog, with mask over parent component to focus user and block interaction.
* Minimized - replaced in same docked position with minimal title placeholder.

Using the word "parent" here somewhat loosely. We could determine if it's better to support this only for the overall AppContainer, or if we can make a feature that could be enabled for other subsections of the interface. Discussions around this and other implementation matters TBD.

PR #1049, opened 2019-04-01
Docked views

TomTirapani:
Took another pass at this, and got it into a place I'm happy with. The intention is to closely mirror the TabContainer API.

See issue #874 and corresponding Toolbox PR https://github.com/exhi/toolbox/pull/165


lbwexler:
Reviewing this now Tom

desktop/cmp/dock/DockContainer.js
```diff
@@ -0,0 +1,50 @@
...
+export class DockContainer extends Component {
+
+    static modelClass = DockContainerModel;
+
+    baseClassName = 'xh-dock-container';
+
+    render() {
+        const {direction} = this.model;
+
+        return hbox({
```
lbwexler: Agree with the decision to focus on the horizontal layouts!  Vertical layouts only theoretically interesting and could be added some day.

desktop/cmp/dock/DockContainer.js
```diff
@@ -0,0 +1,50 @@
...
+            className: this.getClassName(`xh-dock-container-${direction}`),
+            items: this.model.views.map(model => {
+                return dockView({key: model.id, model});
```
lbwexler: Would xhId be safer here?  Don't we want to re-render if we get a new model with the same Id?

desktop/cmp/dock/DockContainerModel.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @param cfg - Configs for DockViewModel to be added.
+     */
+    addView(cfg = {}) {
+        const exists = this.findView(cfg.id);
+
+        if (exists) {
+            this.expandView(cfg.id);
```
lbwexler: curious as to why we don't replace? (and expose, say, getView() as a public method, for applications that want to reshow/expand a view)

desktop/cmp/dock/DockContainerModel.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    //-----------------------
+    // Implementation
+    //-----------------------
+    findView(id) {
+        return this.views.find(it => it.id === id);
+    }
+
+    @action
```
lbwexler: Since this setter is 'private' could just remove it and mark addView() and removeView() as actions.

desktop/cmp/dock/impl/DockView.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    renderContent() {
+        const {content} = this.model;
+        if (isReactElement(content)) return content;
+        if (content.prototype.render) return elem(content);
+        return content();
+    }
+
+    onToggleDocked = () => {
```
lbwexler: You get *no* value from these functions on class properties.  You still are using fat arrow functions in render anyway!  Just ditch these.

desktop/cmp/dock/DockContainerModel.js
```diff
@@ -0,0 +1,88 @@
...
+        if (exists) {
+            this.expandView(cfg.id);
+        } else {
+            const views = [new DockViewModel({...cfg, containerModel: this}), ...this.views];
+            this.setViews(views);
+        }
+    }
+
+    removeView(id) {
+        const views = this.views.filter(it => it.id !== id);
```
lbwexler: Would be good to destroy the views that we are removing.

desktop/cmp/dock/impl/DockViewModel.js
```diff
@@ -0,0 +1,123 @@
...
+        containerModel,
+        title,
+        icon,
+        content,
+        docked,
+        collapsed,
+        allowClose,
+        allowDialog
+    }) {
+        this.id = withDefault(id, XH.genId());
```
lbwexler: + Could use xhId getter for the second arg here -- we already have a unique id.

+ constructor declarations like this could be a good case for using the defaults object initialization syntax.  Just a thought.  Its exposes the defaults a bit more.

desktop/cmp/dock/impl/DockViewModel.js
```diff
@@ -0,0 +1,123 @@
...
+    // Docked state
+    //-----------------------
+    toggleDocked() {
+        if (this.collapsed) {
+            this.showInDock();
+            this.expand();
+        } else if (this.docked && this.allowDialog) {
+            this.showInDialog();
+        } else {
+            this.showInDock();
```
lbwexler: Isn't there a missing this.collapse() here?  Tough to reason about these state changes.  Also wonder if this should be a more generic name like 'toggle' if doesn't just deal with dock vs. dialog.

lbwexler:
This is an extremely cool component and really well implemented!  Can't wait to merge it.

One question -- could it be useful on tablet?  What is desktop specific here?

TomTirapani:
Thanks for taking a look Lee, and for the great suggestions - I've pushed an update implementing them.

**Toggling Behaviour:**

The toggling behaviour is inspired by Gmail's docking behaviour. Toggling between docked / dialog and collapsed / expanded are toggled independently of each other, **except for one specific case**: going from docked+collapsed to dialog inserts an intermediate docked+expanded state.

Happy to change to a different approach if we feel it makes more sense. I'll admit that I feel the Gmail system is quite hard to reason about, but doesn't feel surprising in actual use. Would also like to get @amcclain's thoughts on this, since we had a back-and-forth on this behaviour in the early days of this component.

**Tablet:**

I didn't consider tablet usage, but think it would make sense and be a worthy addition. One potential complication is the mobile and desktop toolkits currently use different implementations for the dialogs - but I expect an approach using dynamic imports of platform-specific view implementations could work. Do we want to go ahead and do that now?

lbwexler:
I trust the toggle behavior well thought out and intended -- just thought
the code was a bit unclear to reason about. Maybe rename the method to just
toggleState or something more generic to make it clear that it is not a
binary thing....not a big deal.

I don't think there is a huge need to do anything on the tablet now -- the
question is -- if we intend to, and the API is going to be generic, should
we put this object in the cross-platform package now, so clients won't need
to adjust imports later....

Not sure about this one, but it seems like if we are trying to model it
after tab container, that would be the way to go.

cmp/dock/DockViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+    toggleDocked() {
+        if (this.collapsed) {
+            this.showInDock();
+            this.expand();
+        } else if (this.docked && this.allowDialog) {
+            this.showInDialog();
+        } else {
+            this.showInDock();
+        }
+    }
```
amcclain: Lee mentioned this but I find myself also weird unable to quite follow the sequence here. A comment or two might help or a little x > y > z kinda thing in a class comment. I leave it to your discretion.

desktop/cmp/dock/impl/Dock.scss
```diff
@@ -0,0 +1,85 @@
...
+.xh-dock-container {
+  position: absolute !important;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  pointer-events: none;
+  align-items: flex-end;
+  justify-content: flex-end;
+
+  &.xh-dock-container-ltr {
```
amcclain: I dunno how you feel about the BEM style or if you want to keep at that - I know it's awkward, but maybe adds some value.  I am not going to advocate strongly for it at this point, but if we wanted to use it there are a few pretty clear cut cases here where we have the sub-elements (header/body) and modifiers (ltr, collapsed)

cmp/dock/DockContainerModel.js
```diff
@@ -0,0 +1,80 @@
...
+     * @param {Object} c - DockContainerModel configuration.
+     * @param {Object[]} c.views - DockViewModel configs to be displayed.
+     * @param {string} [c.direction] - direction in which docked views build up as they are added to
+     *      the container. Valid values are 'ltr', 'rtl' - the default of 'rtl' causes the first
+     *      view to be docked to the bottom right of the container with each subsequent view docked
+     *      to the left of view before it.
+     */
+    constructor({
+        views,
+        direction = 'rtl'
```
amcclain: I flipped this to match the gmail default of docking right-aligned. That somehow feels more natural in terms of the overall goal of keeping docked components available yet "out of the way."

cmp/dock/DockViewModel.js
```diff
@@ -0,0 +1,125 @@
...
+        containerModel,
+        title,
+        icon,
+        content,
+        docked = true,
+        collapsed = false,
+        allowClose = true,
+        allowDialog = true
+    }) {
+        this.id = withDefault(id, this.xhId);
```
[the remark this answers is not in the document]
TomTirapani: Thinking about it, I think it's probably best to always require an id. The id is the only method of interacting with the view after its creation, so seems weird to have it autogenerated and effectively hidden in a black-box. Also, requiring an id brings it closer to the tab container API.

If you want to create "throwaway" views, then the application can use XH.genId() to do so - but I think it should be conscious choice, and not baked into the component.

TomTirapani:
Thanks for taking a look. A couple of notes:

**DockViewModel.id**
In the end, I made it so that dock views always require an id. It seemed the most useful and stable to make apps responsible for providing the id, and I think the stability is worth the inconvenience of not having the id generated for you.

**BEM classes**
I've updated the styling to use BEM standards, but must admit that BEM is something that never really felt right for me, not entirely sure why.

I think a large part is because you have to use the fully qualified name in the markup, and so the SASS and component don't read the same. Do you think it might be overkill to add a JS util function to generate BEM class names using the `baseClassName` as the root? Then both the JS and SASS will be labelling a component as e.g. '__header'. Just a thought, and obviously a larger question than this branch.

**Dialog Toggle**
I agree that its difficult to reason about.

I tried something a bit different - I tried hiding the dialog toggle button when a view is collapsed. I think its a good solution that has effectively the same behaviour (it prevents going from fully collapsed to dialog) but is clearer for both users and code. It allows the behaviour of the dialog toggle button to be a simple binary toggle. It also has the nice side-effect of minimising the space on collapsed docked views.

Issue #1045, opened 2019-03-29
FetchService - accept headers as a plain object, merge in defaults


amcclain:
Currently FetchService will apply default headers to a request only if no headers at all have been specified by the caller. This can make it a bit more annoying to e.g. add a custom header but then also remember to add in content-type.

https://github.com/exhi/hoist-react/blob/76197105fe34791a9bf6c2f6351bf3958ceb04f2/svc/FetchService.js#L75

An approach more in line with our pattern elsewhere would be to:
1) Accept incoming headers config as both a plain object and an actual `Headers` object
2) If not provided or a plain object, merge in our defaults and then instantiate a `Headers` object to use.
3) If provided as a `Headers` object already, don't touch it - caller is specifying exactly what it wants.

amcclain:
It might be worth thinking about a way for an app to register one or more default headers that would be applied to each request.  Thinking about the particular client use case we've been discussing. We could add app-specific util methods to DRY that up - and maybe that's the best option - but I'm tempted to cleanly support consistent custom headers somehow without that step.

A complication might be - would we be OK with those headers also being included in the calls made back to the Hoist /xh endpoints?

Not looking to over-design or over-complicate this ticket - raising here for discussing next week while we're thinking on it....

PR #1050, opened 2019-04-01
Fetch service headers

TomTirapani:
FetchService accept headers as a plain object and merges in defaults. Also provides a mechanism to set app wide headers to be sent with each request.

See #1045

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -100,6 +100,7 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
+    setAppHeaders(headers)      {return this.fetchService.setAppHeaders(headers)}
     getConf(key, defaultVal)    {return this.configService.get(key, defaultVal)}
     getPref(key, defaultVal)    {return this.prefService.get(key, defaultVal)}
     setPref(key, val)           {return this.prefService.set(key, val)}
--- a/core/XH.js
+++ b/core/XH.js
@@ -100,7 +100,6 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
-    setAppHeaders(headers)      {return this.fetchService.setAppHeaders(headers)}
     getConf(key, defaultVal)    {return this.configService.get(key, defaultVal)}
     getPref(key, defaultVal)    {return this.prefService.get(key, defaultVal)}
     setPref(key, val)           {return this.prefService.set(key, val)}
--- a/svc/FetchService.js
+++ b/svc/FetchService.js
@@ -6,7 +6,7 @@
  */
 import {XH, HoistService} from '@xh/hoist/core';
 import {Exception} from '@xh/hoist/exception';
-import {throwIf, warnIf, withDefault} from '@xh/hoist/utils/js';
+import {throwIf, warnIf} from '@xh/hoist/utils/js';
 import {stringify} from 'qs';
 import {isFunction, isPlainObject, isNil, omitBy} from 'lodash';
 
@@ -156,7 +156,7 @@ export class FetchService {
             ...opts,
             headers: {
                 'Accept': 'application/json',
-                ...(withDefault(opts.headers, {}))
+                ...opts.headers
             }
         });
         return ret.status === 204 ? null : ret.json();
@@ -228,7 +228,7 @@ export class FetchService {
             body: JSON.stringify(opts.body),
             headers: {
                 'Content-Type': 'application/json',
-                ...(withDefault(opts.headers, {}))
+                ...opts.headers
             }
         });
     }
--- a/svc/FetchService.js
+++ b/svc/FetchService.js
@@ -17,7 +17,8 @@ import {isPlainObject} from 'lodash';
  * the most common use-cases. The Fetch API will be called with CORS enabled, credentials
  * included, and redirects followed.
  *
- * Todo: Explain setting headers app-wide / per request
+ * Custom headers can be provide to fetch as either a plain object or a Headers object. App-wide
+ * headers can be set using setAppHeaders (aliased to XH for convenience).
  *
  * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API|Fetch API Docs}
  *
@@ -31,8 +32,10 @@ export class FetchService {
     appHeaders = {};
 
     /**
-     * Todo: Doc
-     * @param headers
+     * Set App-wide headers for all subsequent requests. App-wide headers are not
+     * sent with core framework requests.
+     *
+     * @param {Object} headers - Headers to be sent with all requests.
      */
     setAppHeaders(headers) {
         throwIf(!isPlainObject(headers), 'Todo');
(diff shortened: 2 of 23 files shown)


core/XH.js
```diff
@@ -100,6 +100,7 @@ class XHClass {
     track(opts)                 {return this.trackService.track(opts)}
     fetch(opts)                 {return this.fetchService.fetch(opts)}
     fetchJson(opts)             {return this.fetchService.fetchJson(opts)}
+    setAppHeaders(headers)      {return this.fetchService.setAppHeaders(headers)}
```
amcclain: Concerned it might not be clear enough what we mean by "headers" here - e.g. what "app" means in this context + HTTP request headers vs. some kind of visual element.

What about making the method `FetchService.setDefaultHeaders()`, and then foregoing the XH alias? Presumably an app would need to call the method at most once, so I don't think there's a strong case for hanging an alias off of XH.  Does the word "default" make it clear enough that these will be added on?

svc/FetchService.js
```diff
@@ -25,6 +29,18 @@ import {stringify} from 'qs';
...
+    appHeaders = {};
+
+    /**
+     * Set App-wide headers for all subsequent requests. App-wide headers are not
+     * sent with core framework requests.
+     *
+     * @param {Object} headers - Headers to be sent with all requests.
+     */
+    setAppHeaders(headers) {
+        throwIf(!isPlainObject(headers), 'Todo');
```
amcclain: I think we can leave this off - again it would be called once, presumably by a dev who has read up on or been informed of this feature and its requirements as documented above.

svc/FetchService.js
```diff
@@ -40,6 +56,9 @@ export class FetchService {
      * @param {string} [opts.contentType] - value to use in the Content-Type header in the request.
      *     If not specified, contentType is set based on method: 'application/x-www-form-urlencoded'
      *     for POSTs, 'text/plain' otherwise.
+     * @param {Object|Headers} [opts.headers] - headers to send with this request. If an Object,
+     *      will be merged with existing default headers. Alternatively, provide an
+     *      instantiated Headers object to replace entirely.
```
amcclain: Tempted to remove contentType as a top-level input to this API - seems like the dev could/should just pass it as a header since we now support the easy-to-use object form and that is where it will land anyway.  Could note in the headers doc comment that "a Content-Type header will be set if not provided by the caller directly or via one of the  dedicated `xxxJson` methods on this service."

This would be another breaking change but I think it's worth it.  If we have reason to believe this is used in more than one or two places, we could add a warnIf to easy the API transition. (I am in favor of those when they allow us to take cleanups we want to make while not being too annoying during upgrades.)

svc/FetchService.js
```diff
@@ -66,13 +85,15 @@ export class FetchService {
...
+        headers = this.getHeaders(headers, contentType, url);
+
         // 2) Prepare merged options
         const defaults = {
                 method,
                 cors: true,
                 credentials: 'include',
                 redirect: 'follow',
-                headers: new Headers({'Content-Type': contentType})
+                headers: new Headers(headers)
```
amcclain: Won't these get overwritten all the time by the unmodified version off of opts? 🤔

svc/FetchService.js
```diff
@@ -25,6 +29,18 @@ import {stringify} from 'qs';
 export class FetchService {
 
     autoAbortControllers = {};
+    appHeaders = {};
+
+    /**
+     * Set App-wide headers for all subsequent requests. App-wide headers are not
+     * sent with core framework requests.
```
amcclain: I'm thinking that an app would almost always need to set these up right at the top of its `initAsync()` method to ensure they are used for all app fetches. If that sounds right, we might want to include that as a pointer here in the doc string.

We should ensure @lbwexler thinks this is a worthy addition to the service overall, and discuss if it needs to support a closure to be useful in practice across more scenarios (i.e. if you wanted to include a timestamp or token from some auth API that might expire or something else like that).

core/XH.js
lbwexler: +1 on these ideas.   One concern is would we have the situation where an app would want these headers set *before* the app starts it Hoist bootstrap sequence?  Imagine a certain type of header is needed at a client-site get past certain internal proxies to Hoist Central itself?

In that case, we would want to add a static defaultFetchHeaders property to the appSpec itself.

svc/FetchService.js
lbwexler: agree

svc/FetchService.js
lbwexler: well that closure bit is really interesting.  Seems like just the kind of thing that might come along.

Would a more general thing be a way to attach a closure to pre-process *all* requests?  FetchService.preprocessRequest(fn) and (potentially) FetchService.postProcessRequest?

TomTirapani:
Thanks for taking a look guys, and thanks for the great suggestions.

I've remove the XH.alias & renamed AppHeaders > DefaultHeaders to avoid confusion. You can set defaultHeaders to either a plain object or a closure.

You can also set them via AppSpec.defaultFetchHeaders, for the case Lee brought up where you need the headers before the app starts the Hoist bootstrap sequence. I think its important we still keep the setDefaultHeaders method though, for the opposite use case - where the headers depend upon some business logic not yet available this early in the lifecycle.

I also removed contentType as a top level parameter, requiring the user to pass it as a header instead.

svc/FetchService.js
```diff
@@ -50,44 +72,54 @@ export class FetchService {
...
 
-        // 2) Prepare merged options
+        // 2) Compute headers
+        if (!(headers instanceof Headers)) {
+            const baseHeaders = {
+                'Content-Type': (method === 'POST') ? 'application/x-www-form-urlencoded': 'text/plain'
+            };
+            if (opts.acceptJson) {
+                baseHeaders['Accept'] = 'application/json';
+                delete opts.acceptJson;
```
amcclain: I know this was pre-existing, but caught my eye. 

Any justification for `acceptJson` as a top-level option, vs. again passing in the header? Especially since we go ahead and mutate the opts provided by the caller, which in general we avoid doing.

TomTirapani: Good catch - will treat it the same as ContentType and just pass in the header

svc/FetchService.js
```diff
@@ -50,44 +72,54 @@ export class FetchService {
...
+            }
+
+            const {defaultHeaders} = this;
+            headers = new Headers(Object.assign(
+                baseHeaders,
+                isFunction(defaultHeaders) ? defaultHeaders() : defaultHeaders,
+                isPlainObject(headers) ? headers : {}
+            ));
+        }
+        delete opts.headers;
```
amcclain: Ah - I see before we weren't mutating opts (we cleaned up the merged `fetchOpts` instead), but now we are in two places with headers and acceptJson.  I can't think of any instances of an app re-using / holding a reference to the opts passed in here, but still as a general practice best to leave them alone and might be a bit of a code smell that the logic around building up the merged / defaulted opts could be clarified.

lbwexler: Can we just delete line 83?

svc/FetchService.js
```diff
@@ -37,10 +52,12 @@ export class FetchService {
      *      with the request body (for POSTs/PUTs sending form-url-encoded).
      * @param {string} [opts.method] - HTTP Request method to use for the request. If not specified,
      *      the method will be set to POST if there are params, otherwise GET.
-     * @param {string} [opts.contentType] - value to use in the Content-Type header in the request.
-     *     If not specified, contentType is set based on method: 'application/x-www-form-urlencoded'
-     *     for POSTs, 'text/plain' otherwise.
-     * @param {boolean} [opts.acceptJson] - true to set Accept header to 'application/json'.
+     * @param {(Object|Headers)} [opts.headers] - headers to send with this request. If an Object,
```
lbwexler: Out of curiosity, why do we take an actual Headers object?  Do we have a use case for that right now?  Seems like we could leave it out for now, leaving us more flexibility to add it with different semantics in the future when we need it.

TomTirapani: It's there to offer a way to take full control of the headers for that particular request - i.e. bypass any default headers the app is providing. We have a use-case with export right now where we need to make sure we *don't* set a Content-Type header:
https://github.com/exhi/hoist-react/blob/develop/svc/GridExportService.js#L81

lbwexler:
got it -- but would it be better in that case for the app to explicitly set
that particular header to undefined (or null) and have that be the API for
fetch service to 'remove' it?

Might there be other headers that ExportService should not be zapping?

Seems weird to have the behavior be *so* different depending on the type
being passed.

PR #1052, opened 2019-04-02
Tab improvements

lbwexler:
Improve handling of currentlyActive tab and route handling in TabContainer.

### DIFF
--- a/cmp/tab/TabContainerModel.js
+++ b/cmp/tab/TabContainerModel.js
@@ -86,7 +86,7 @@ export class TabContainerModel {
                 run: this.syncWithRouter
             });
 
-            this.markActiveTabWithRouter(this.activeTabId);
+            this.forwardRouterToTab(this.activeTabId);
         }
     }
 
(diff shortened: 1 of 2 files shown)


cmp/tab/TabContainerModel.js
```diff
@@ -135,16 +138,22 @@ export class TabContainerModel {
...
         if (router.isActive(route)) {
             const tab = tabs.find(t => router.isActive(route + '.' + t.id));
-
             if (tab && !tab.isActive && !tab.disabled) {
                 this.setActiveTabId(tab.id);
             }
         }
     }
 
+    markActiveTabWithRouter(id) {
```
TomTirapani: I think if this method takes the id we shouldn't have `ActiveTab` in the name - either `markActiveTabWithRouter()` (using this.activeTabId internally) or `markTabWithRouter(id)`.

In general, I find this name a bit confusing - maybe `forwardRouterToTab(id)` or something?

lbwexler: Thanks for the careful review tom -- agree your name is clearer.   Will change and merge!

PR #1053, opened 2019-04-02
Update ag-grid from 20.1 -> 20.2

shravnk:


amcclain:
Need yarn.lock in this PR as well - please run yarn install and push the changes there too!

Glad no other changes appear to be required.

shravnk:
My bad - updated PR with lock file

lbwexler:
I think this yarn.lock diff is bigger than it needs to be because your older version of yarn (<1.9?) is removing the integrity flag

shravnk:
Thanks, had to upgrade yarn

PR #1058, opened 2019-04-04
Select first fix

amcclain:


cmp/grid/GridModel.js
```diff
@@ -232,11 +232,15 @@ export class GridModel {
...
         if (agApi) {
-            const idx = (this.groupBy && !this.treeMode) ? 1 : 0,
-                first = agApi.getDisplayedRowAtIndex(idx);
+            let dataNode = null;
+            agApi.forEachNodeAfterFilterAndSort(node => {
+                if (!dataNode && node.data) dataNode = node;
+            });
 
-            if (first) selModel.select(first);
+            if (dataNode) selModel.select(dataNode);
```
haynesjm42: Should this be passing `dataNode.data` here to the selection model? Looks like we have a lot of dependencies on the ag-grid row node id matching the record id (meaning we depend on `getRowNodeId` not being overridden in `agOptions` for instances of the grid component) which is probably fine, but we might want to explicitly check that `getRowNodeId` doesn't get overridden in `agOptions` since that would break a lot of core Grid functionality.

It does seem odd to be passing the ag-grid row node here when we can easily access the `data` of the node which is the actual record, and this is only working because Record and ag-grid RowNode objects each have an `id` prop. This is also the case here: https://github.com/exhi/hoist-react/blob/develop/cmp/grid/Grid.js#L513

amcclain: Thanks agreed and updated to pass the record.

The link to the Grid implementation is calling `getSelectedRows()`, which returns an array of Records (not nodes) as provided to the grid, so we should be good there.

Issue #1061, opened 2019-04-08
ag-grid Wrapper Component


haynesjm42:
There is a growing need to use some of the more advanced features of ag-grid (such as pivoting) which are difficult or undesireable to support in our core Grid/GridModel components, so would be nice to have a wrapper around AgGridReact which provides a look and feel consistent with our Grid component while being flexible in terms of how you interact with the ag-grid component and APIs while also incorporating some of the commonly useful functions for using the grid that we have developed in our framework and in apps (saving/restoring expand state, serializing grid config, knowing when the grid is "ready", etc.)

PR #1062, opened 2019-04-09
AgGrid Component

haynesjm42:
#1061 

The goal here was to allow an app to use an AgGrid component which had a consistent look and feel to our Grid but would allow apps to use something that is very close to stock ag-Grid. Also attempted to push whatever made sense for utility methods into an `AgGridModel` class to simplify Grid/GridModel where possible and provide a place for us to continue adding logic that has been built in apps for carrying out certain actions with ag-Grid (such as pivot-compatible grid state serialization).

Decided to go with `AgGrid` and `AgGridModel` for class names, had `AgGridWrapper` at one point but found that clunky - open to suggestions.

Toolbox branch with example tab here: https://github.com/exhi/toolbox/tree/agGridWrapper


amcclain:
Great will review shortly - eager to get this in.

cmp/grid/GridModel.js
```diff
@@ -87,22 +94,6 @@ export class GridModel {
...
-    @bindable stripeRows;
-    /** @member {boolean} */
-    @bindable showHover;
-    /** @member {boolean} */
-    @bindable showCellFocus;
-
-    /** @member {GridApi} */
-    @observable.ref agApi = null;
-    /** @member {ColumnApi} */
-    @observable.ref agColumnApi = null;
```
amcclain: Should we add pass-through getters for agApi and agColumnApi? They were part of the public API, and saving references to these two is a useful feature of GridModel. Any reason not to do so, i.e. do we think it's important that the developer realize these are pulled from the agGridModel?

(We could use also those in the implementation code below FWIW.)

cmp/grid/GridModel.js
```diff
@@ -217,29 +210,26 @@ export class GridModel {
...
-            this.agApi.exportDataAsCsv(params);
+            agApi.exportDataAsCsv(params);
         }
     }
 
     /** Select the first row in the grid. */
     selectFirst() {
-        const {agApi, selModel} = this;
+        const {agGridModel, selModel} = this;
+        warnIf(!agGridModel.isReady, 'Called selectFirst before the grid was ready!');
```
lbwexler: Should we bail out here, if we have already warned?  Otherwise what's the sense of this warning?

cmp/grid/ag-grid/AgGridModel.js
```diff
@@ -0,0 +1,148 @@
...
+
+        this.addReaction({
+            track: () => this.compact,
+            run: () => {
+                if (this.isReady) this.agApi.resetRowHeights();
+            }
+        });
+    }
+
+    /** @returns {boolean} - true if the grid is ready to be interacted with (ag-grid onGridReady event has been fired) */
```
lbwexler: thought we were planning to use @computed here?

cmp/grid/GridModel.js
```diff
@@ -269,15 +259,20 @@ export class GridModel {
         return this.selModel.singleRecord;
     }
 
-    @action
-    setAgApi(agApi) {
-        this.agApi = agApi;
-    }
+    get compact() { return this.agGridModel.compact}
```
lbwexler: The problem with trampolining these is that they also need documented -- I still wonder if not easier/more powerful to just reveal to the applications the underlying model as a getter, and to take in the constructor these options as a bundle....

I guess the change as is is more backward compatible, but I doubt many apps are actually setting *any* of these dynamically other than compact

lbwexler:
👍

PR #1070, opened 2019-04-12
Add gridModel.groupSortFn for custom group sort

amcclain:
+ Relays to defaultGroupSortCompator, avoid need to use agOptions.
+ Keep existing default that sorts groups according to their natural order.
+ Tweak contextMenuFn config handling for symmetry (requires review).
+ Remove unused/dupe init values on class fields declarations.

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -66,9 +66,9 @@ export class GridModel {
     colChooserModel;
     /** @member {function} */
     rowClassFn;
-    /** @member {function} */
+    /** @member {GridStoreContextMenuFn} */
     contextMenuFn;
-    /** @member {function} */
+    /** @member {GridGroupSortFn} */
     groupSortFn;
     /** @member {boolean} */
     enableExport;
@@ -141,8 +141,8 @@ export class GridModel {
      * @param {object} [c.exportOptions] - default options used in exportAsync().
      * @param {function} [c.rowClassFn] - closure to generate css class names for a row.
      *      Called with record data, returns a string or array of strings.
-     * @param {function} [c.groupSortFn] - closure to sort full-row groups. Called with two nodes to
-     *      be compared, returns a number as per a standard JS comparator.
+     * @param {GridGroupSortFn} [c.groupSortFn] - closure to sort full-row groups. Called with two
+     *      group values to compare, returns a number as per a standard JS comparator.
      * @param {GridStoreContextMenuFn} [c.contextMenuFn] - function to optionally return a
      *      StoreContextMenu when the grid is right-clicked (desktop only).
      * @param {*} [c...rest] - additional data to attach to this model instance.
@@ -647,8 +647,7 @@ export class GridModel {
         });
     }
 
-    defaultGroupSortFn = (nodeA, nodeB) => {
-        const a = nodeA.key, b = nodeB.key;
+    defaultGroupSortFn = (a, b) => {
         return a < b ? -1 : (a > b ? 1 : 0);
     }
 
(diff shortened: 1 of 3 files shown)


amcclain:
Customizing group sort order is pretty common, and we had an example in an app where it was requiring the use of agOptions, which we want to avoid.

See corresponding branch in Toolbox for an example usage in sampleGrid.  The win/lose grouping should sort winners first.

While I was here I wanted to get more symmetry between this new config and the contextMenuFn, but I really was confused by why we had the ctx menu fn setup as we did.  Please review carefully and ensure I did not break anything w.r.t. scope!  Thank you.

cmp/grid/GridModel.js
```diff
@@ -171,13 +167,15 @@ export class GridModel {
         enableExport = false,
         exportOptions = {},
         rowClassFn = null,
-        contextMenuFn = () => this.defaultContextMenu(),
```
amcclain: This was the extra closure layer that tripped me up.  I also thought this was a bit much to destructure here.

cmp/grid/GridModel.js
```diff
@@ -640,6 +638,20 @@ export class GridModel {
         const Model = XH.isMobile ? MobileColChooserModel : DesktopColChooserModel;
         return this.markManaged(new Model(this));
     }
+
+    defaultContextMenuFn = (agParams, gridModel) => {
+        if (XH.isMobile) return null;
+        return new StoreContextMenu({
+            items: GridModel.defaultContextMenuTokens,
+            gridModel
```
amcclain: The impl in grid calls the ctx menu fn with the grid model as the second argument, so I am using that here vs. relying on `this` as it had been.....

cmp/grid/GridModel.js
```diff
@@ -55,23 +55,25 @@ export class GridModel {
...
-    contextMenuFn = null;
+    contextMenuFn;
+    /** @member {function} */
+    groupSortFn;
     /** @member {boolean} */
-    enableExport = false;
+    enableExport;
     /** @member {object} */
-    exportOptions = null;
+    exportOptions;
```
amcclain: Sorry for the extra churn here but we've been trying to knock these out where applicable.

amcclain:
@haynesjm42 if you have a chance to look at this I would appreciate it.  Looking to remove another occasional use of `agOptions` here.

amcclain:
As discussed briefly on the phone just now, I liked this idea and went to implement, but ran into a case right away where our little Toolbox example has multi-level grouping and wants to know what grouped field it is comparing on.   That information is available on the node.

Here is the revised signature I had come up with for the groupSortFn API:
```
/**
 * @callback GridGroupSortFn - comparator for custom grid group sorting, provided to GridModel.
 * @param {*} groupAVal - first group value to be compared.
 * @param {*} groupBVal - second group value to be compared.
 * @param {GridModel} gridModel - controlling GridModel instance.
 * @param {Object} metadata - additional metadata with raw ag-Grid group nodes.
 * @returns {number} - 0 if group values are equal, <0 if `a` sorts first, >0 if `b` sorts first.
 */
```

So the nodes are still available - making the Toolbox example code:

```
groupSortFn: (a, b, gm, {nodeA}) => {
            const groupField = nodeA.field;

            if (a == b) return 0;
            if (groupField == 'winLose') {
                return a == 'Winner' ? -1 : 1;
            } else {
                return a < b ? -1 : 1;
            }
        }
```

We could bake the field in as the third arg if we wanted - so `sort(aVal, bVal, fieldName, gridModel, {nodeA, nodeB})`....  Is that starting to push it? It does seem that, if you are supplying a custom group sort, you almost always want to change your sort based on the field, and reading from gridModel is problematic in the multi-level case.

amcclain:
Thanks for the review @haynesjm42 - updated to pass groupField as third arg as suggested, with the gridModel and rowNodes in the final, metadata arg.

PR #1072, opened 2019-04-12
Ensure that Forms and FormFields are usable within Toolbars.

### LINKED ISSUE #1059: Review styling for forms in toolbars
This is a bit of an edge case, but we have real use cases where a Form is desired within a Toolbar to manage a set of query input controls.

FormField and related classes install margins, etc. that should not be applied in the Toolbar use-case.

TomTirapani:
See Toolbox PR: https://github.com/exhi/toolbox/pull/172. Fixes #1059


amcclain:
Looked great thanks much!

PR #1074, opened 2019-04-15
Workaround for AG-2879

lbwexler:
Brendan and Lee co-dev

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -347,17 +347,14 @@ export class Grid extends Component {
             run: ([api, records]) => {
                 if (api) {
                     runInAction(() => {
+                        const now = Date.now();
 
                         // Workaround for AG-2879.
-                        // pre-emptively dump rows with internal ag API to avoid expensive removals
-                        if (this.pendingDeletionCount(records, api) > 50) {
-                            api.selectionController.reset();
-                            api.clientSideRowModel.setRowData([]);
-                        }
+                        this.clearDataIfExpensiveDeletionPending(records, api);
 
                         // Load updated data into the grid.
                         api.setRowData(records);
-                       
+
                         // Size columns to account for scrollbar show/hide due to row count change.
                         api.sizeColumnsToFit();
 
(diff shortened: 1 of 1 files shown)


lbwexler:
Variant on Brendan's initial suggestion that will standalone and also help our filtering.

cmp/grid/Grid.js
```diff
@@ -489,6 +498,14 @@ export class Grid extends Component {
         };
     }
 
+    // Get the number of records to be removed in upcoming update. For AG-2879 Workaround.
+    pendingDeletionCount(newRecords, api) {
+        const ids = new Set();
+        api.forEachNode((node, index) => ids.add(node.id));
+        newRecords.forEach(rec => ids.delete(rec.id));
+        return ids.size;
```
amcclain: Obviously depending on these loops and especially `Set` being fast, which I am sure they are. 

So ironic that we're doing here exactly what ag-grid does that is so slow - building up a flat list then winnowing it down one-by-one - but the use of Set makes all the difference.  I wonder how babel transpiles this code - I confess I would need to load it back in my brain, but is it relying on the browser supporting Set or replacing with a different backing implementation?

cmp/grid/Grid.js
```diff
@@ -347,9 +347,17 @@ export class Grid extends Component {
             run: ([api, records]) => {
                 if (api) {
                     runInAction(() => {
+
+                        // Workaround for AG-2879.
+                        // pre-emptively dump rows with internal ag API to avoid expensive removals
+                        if (this.pendingDeletionCount(records, api) > 50) {
```
amcclain: Woops - I had typed a big comment here but lost it somehow.  

Discussed this number with Lee - I would have started with something more like 500-1000 - the idea being that the really slow behavior was happening when we replaced 50k records with another 50k, so it's starting with a 100k array and cutting it down in 50k passes.  Total time depends on count being removed as well as interim size, but in the end we don't need to be too fancy - just find a good balance.

In general the incremental updates we make (which we want to ensure are stable and fast) don't tend to remove many rows, so this is  a low-pressure choice either way.

Finally I would be up for moving this to a constant just to doc it there...

lbwexler:
Relaxed the threshold where we dump state substantially at Anselm's suggestion -- testing in toolbox yields good results.

Also added runtime debug logging that could be generally useful.    Please merge if you are comfortable with this!

Issue #674, opened 2018-10-02
Namespace local storage keys by username


amcclain:
With non-SSO apps, we could have multiple users logging into a web application under one OS/machine-level account.  It's a relatively rare edge-case given our standard corporate use-cases, but it is not that far-fetched.

Given that, we should consider mixing the username into the key used to fetch/create localStorage instances - see https://github.com/exhi/hoist-react/blob/f4a205c40473aeab1b0f335e50695ef8652f7e7d/svc/LocalStorageService.js#L58.

It might be good at this same time to swap to appCode (vs name) and maybe do some simple hashing of both appCode and username to provide some light obscurity around who has keys stored under a given account.  Again should stress here that none of this is intended as a security measure - truly sensitive stuff should not be put into local storage at the app level unless we know that the app will only be used on single-user machine logins.

TomTirapani:
PR: https://github.com/exhi/hoist-react/pull/1075

PR #1075, opened 2019-04-16
Local storage namespace uses appCode and username.

TomTirapani:
See issue #674

### DIFF
--- a/svc/LocalStorageService.js
+++ b/svc/LocalStorageService.js
@@ -9,6 +9,12 @@ import {XH, HoistService} from '@xh/hoist/core';
 import {throwIf} from '@xh/hoist/utils/js';
 import store from 'store2';
 
+/**
+ * Service to provide simple key/value access to browser local storage, appropriately namespaced
+ * by application code and username.
+ *
+ * Relied upon by Hoist features such as local preference values and grid state.
+ */
 @HoistService
 export class LocalStorageService {
 
@@ -71,6 +77,9 @@ export class LocalStorageService {
         return `${XH.appCode}.${XH.getUsername()}`;
     }
 
+    // Added in April 2019 to support a switch to the new user-specific namespace without current
+    // users losing their local state. Remove when we are confident essential apps have been updated
+    // and accessed by end users to run this routine.
     migrateLegacyNamespace() {
         try {
             const oldSpace = store.namespace(XH.appName);
--- a/svc/LocalStorageService.js
+++ b/svc/LocalStorageService.js
@@ -12,13 +12,10 @@ import store from 'store2';
 @HoistService
 export class LocalStorageService {
 
-    _supported = !store.isFake();
-
     async initAsync() {
-        this.addReaction({
-            when: () => XH.appIsRunning,
-            run: this.migrateNamespace
-        });
+        if (this.supported) {
+            this.migrateLegacyNamespace();
+        }
     }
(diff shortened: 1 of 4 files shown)


amcclain:
Thanks for grabbing this.  We need to decide if/when to merge as this will be a breaking change and reset any current local state, which currently includes grid state...

Some things to consider:
* Most real users will be the only ones logging in to a site under their overall OS profile. However for apps like Toolbox where you can log out / log in interactively - and also importantly for impersonation - this closes a very real gap in how we were namespacing local storage.
* I can see the value of hashing the appCode and username to not have usernames embedded directly into local storage keys, although I don't know if we have any real-world concerns here given that everything is under an OS profile. The tradeoff is that this makes troubleshooting potentially more difficult.

Would it be overkill to consider a migration strategy, where LocalStorageService would check on init if there was a local namespace with the appName and, if so, migrate its contents to the new desired namespace then delete it?  I know that sounds fiddly and is something we would eventually want to unwind, but we probably need to do it if we want to get this done without complaints and bug reports about lost state from multiple clients/users.  @lbwexler your thoughts?

lbwexler:
Yes, as you know I think that kind of runtime data cleanup/conversion fiddly, and normally not in favor of it,  but here we really would not have another choice, so I think we should carefully do it!

lbwexler:
I could skip the hashing.  That's a false sense of security that just buys us troubleshooting difficulty.  The data will still be totally visible, and not appropriate for any real multi-user situation...

I think impersonation is the main reason to do this!

TomTirapani:
Thanks for taking a look guys. I've added some migration logic, and removed the hashing.

svc/LocalStorageService.js
```diff
@@ -55,6 +63,21 @@ export class LocalStorageService {
     //------------------
     getInstance() {
         throwIf(!this._supported, 'Local Storage is not supported');
-        return store.namespace(XH.appName);
+        return store.namespace(this.getNamespace());
+    }
+
+    getNamespace() {
+        return `${XH.appCode}.${XH.getUsername().toLowerCase()}`;
```
amcclain: Technically we are requiring that HoistUser normalize usernames to lowercase:

https://github.com/exhi/hoist-core/blob/ff105dce5a563ba9ac27fb876a736449324012c8/src/main/groovy/io/xh/hoist/user/HoistUser.groovy#L22

Any reason to believe that mixed-case usernames are making it to the client?  I can understand the instinct to be defensive here and don't feel strongly about it, but that lowercase requirement is probably something we should take as a requirement / given throughout our client-side Hoist code.

svc/LocalStorageService.js
```diff
@@ -55,6 +63,21 @@ export class LocalStorageService {
...
     }
+
+    migrateNamespace() {
+        const oldNamespace = store.namespace(XH.appName),
+            namespace = store.namespace(this.getNamespace());
+
+        if (oldNamespace.size() && !namespace.size()) {
+            namespace.setAll(oldNamespace.getAll());
+            oldNamespace.clear();
```
amcclain: That certainly is convenient!  😄

lbwexler:
Implemented ATM's suggestion and tweaked this a bit more -- just being a bit paranoid, and looking to minimize our exposure to it ever bringing down the app.

Feel free to merge!

PR #1077, opened 2019-04-16
Force app-level refresh on SessionMismatchException.

### LINKED ISSUE #675: Force app-level refresh on clientUsername mismatch exception
We validate that, for certain Hoist-level calls, the current user as per the client app matches the user as per the server session. This was added to both hoist-react and hoist-core - the former POSTs the username and the latter checks it and throws if there is a mismatch. See:

https://github.com/exhi/hoist-core/blob/cf5e9412e5dd8eefc55f780f63708748f41fa943/grails-app/controllers/io/xh/hoist/impl/HoistImplController.groovy#L236

You can trigger this behavior by opening multiple browser tabs and impersonating or logging out / in again with a different user account across the two tabs, then try doing something like updating a preference or submitting feedback.

We could improve the error reporting by throwing a dedicated exception and, at the client level, prompting for a reload when that exception is encountered. This would be similar to our handling of sessionExpired exceptions, see:

https://github.com/exhi/hoist-react/blob/f4a205c40473aeab1b0f335e50695ef8652f7e7d/core/ExceptionHandler.js#L135

TomTirapani:
Fixes #675
See hoist-core PR: https://github.com/exhi/hoist-core/pull/75

### DIFF
--- a/core/ExceptionHandler.js
+++ b/core/ExceptionHandler.js
@@ -145,6 +145,12 @@ export class ExceptionHandler {
         ret.title = ret.title || (ret.showAsError ? 'Error' : 'Message');
         ret.message = ret.message || exception.message || exception.name || 'An unknown error occurred.';
 
+        if (this.sessionMismatch(exception)) {
+            ret.title = 'Session Mismatch';
+            ret.message = 'Your session may no longer be active, or no longer matches with the server. Please refresh.';
+            ret.requireReload = true;
+        }
+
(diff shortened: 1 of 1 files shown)


amcclain:
Great thank you for closing out this old ticket.  Released hoist-core change as 5.5.3 and will ensure we update any client plugins to fold this in on both sides. 

(For the record, no blocking dependencies between the two.)

Issue #1056, opened 2019-04-03
Update Highcharts to 7.1.x


amcclain:
New minor version release with a pretty decent set of changes:

https://www.highcharts.com/blog/changelog/

Worth taking after a careful look / test.

PR #1080, opened 2019-04-18
Upgrade Highcharts to ~7.1.1, Fixes: #1056

febbraiod:
From the HC side this upgrade is largely additive, with few breaking changes:

1. Changed default X axis tickWidth option. Now tick marks are not rendered by default on category axes. Set the tickWidth to 1 to bring back the old behaviour. 
-See note in PR below

2. Renamed plotOptions.treemap.drillUpButton to traverseUpButton. 
- We do not use this in HR or Toolbox, but could theoretically effect client apps.

3. Changed default states.inactive.opacity for map and mapline series to prevent dimming map shapes.
-  We do not use this in HR or Toolbox, but could theoretically effect client apps. But not sure I would consider this a 'breaking change'

### DIFF
--- a/desktop/cmp/chart/theme/Dark.js
+++ b/desktop/cmp/chart/theme/Dark.js
@@ -252,4 +252,5 @@ export const DarkTheme = {
     textColor: '#C0C0C0',
     contrastTextColor: '#F0F0F3',
     maskColor: 'rgba(255,255,255,0.3)'
+
 };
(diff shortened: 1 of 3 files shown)


desktop/cmp/chart/theme/Dark.js
```diff
@@ -64,6 +64,7 @@ export const DarkTheme = {
         lineColor: '#707073',
         minorGridLineColor: '#505053',
         tickColor: '#707073',
+        // default tickWidth here ??
```
febbraiod: Could see doing this to ensure consistant behavior in client apps, but could also make the argument that we should embrace the library's decision regarding this default, and make it the app's responsibility to adjust if necessary

PR #1081, opened 2019-04-18
Grid Context Menu should suppress Browser Menu

shravnk:
Not the most elegant solution, but tough because Ag's context menu API is pretty minimal. 

Limitation is that this fix fails when `onCellContextMenu` passed via agOptions. I suppose that falls under our agOptions disclaimer, though.

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -238,7 +238,8 @@ export class Grid extends Component {
             ret = {
                 ...ret,
                 allowContextMenuWithControlKey: true,
-                getContextMenuItems: this.getContextMenuItems
+                getContextMenuItems: this.getContextMenuItems,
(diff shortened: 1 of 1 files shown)


shravnk:
Although it's hard to see how this change would ever conflict with a handler passed through agOptions. At some point could be worth moving cases like this to a separate object and chain the handler to whatever is passed in from agOptions

lbwexler:
Assigning john for your insight (also @amcclain  take a look too.)  
Is this worth doing, or should we just abandon it, as just the way ag-Grid works. 
This is starting to seem overly fiddly and invasive to me.

haynesjm42:
I think we should close this, and take no action ourselves. If anything I would consider reporting this as a bug to ag-grid, seems like a custom context menu should swallow the dom contextmenu event on itself to avoid this situation as it can be confusing to users.

PR #1082, opened 2019-04-19
Additional improvements to Store and Record API

lbwexler:
+ Restore 'raw' data property on record.
+ More maintainable handling of name collision
+ Don't require user to delete bad 'id' property on raw data.  idSpec always wins.
+ Cleanup to left/right chooser


data/Record.js
```diff
@@ -43,20 +42,31 @@ export class Record {
      * requiring children to also be recreated.)
      *
      * @param {Object} c - Record configuration
+     * @param {Object} c.data - data for constructing the record.
      * @param {Object} c.raw - raw data for record.
```
amcclain: I know that apps aren't expected to construct records, but this is still a bit confusing here. How about:
```
c.data - data used to construct this record, pre-processed if applicable by store.processRawData().
c.raw - the same data, prior to any store pre-processing.
```
Would that be incrementally more clear?

data/Record.js
```diff
@@ -43,20 +42,31 @@ export class Record {
...
      * @param {BaseStore} c.store - store containing this record.
```
amcclain: Realizing that LocalStore defines `idSpec`, which means that's what you need to be passing in here.  Yet `BaseStore` also declares interfaces that deal in Records/RecordSets.

We've discussed how we might want to unwind BaseStore, taking LocalStore -> Store and having it be the primary/parent store class.  I dunno if we want to deal with that now.

Alternatively we could move idSpec to BaseStore, or adjust the docs here to punt a bit and deal with LocalStore only.

data/Record.js
```diff
@@ -43,20 +42,31 @@ export class Record {
...
-        const id = raw.id;
+    constructor({data, raw, store, parent}) {
+        const {idSpec} = store,
+            id = isString(idSpec) ? data[idSpec] : idSpec(data);
+
+        throwIf(isNil(id), "Record has an undefined ID. Use 'LocalStore.idSpec' to resolve a unique ID for each record.");
 
         this.id = id;
         this.store = store;
+        this.raw = raw;
```
amcclain: We should add an `@member` declaration

data/Record.js
```diff
@@ -43,20 +42,31 @@ export class Record {
...
         this.parentId = parent ? parent.id : null;
         this.xhTreePath = parent ? [...parent.xhTreePath, id] : [id];
 
         store.fields.forEach(f => {
-            this[f.name] = f.parseVal(raw[f.name]);
+            const {name} = f;
+            if (name == 'id') return;
+            throwIf(
+                name in this,
+                `Field name "${name}" cannot be used for data. It is reserved as a top-level property of the Record class.`
```
amcclain: Great that this check moved into this class.

data/impl/RecordSet.js
```diff
@@ -164,28 +164,22 @@ export class RecordSet {
...
-        raw.id = isString(idSpec) ? raw[idSpec] : idSpec(raw);
-
-        throwIf(
-            isNil(raw.id),
-            "Record has a null/undefined ID. Use the 'LocalStore.idSpec' config to resolve a unique ID for each record."
-        );
+        let data = raw;
+        if (store.processRawData) {
+            data = store.processRawData(raw);
+            throwIf(!data, 'processRawData should return an object. If writing/editing, be sure to return a clone!');
```
amcclain: Should we enforce processRawData always returning a new object? Is there a likely/important use case where we have processRawData making decisions to *not* edit some objects (and it would be terribly inefficient to force it to clone anyway)?

desktop/cmp/leftrightchooser/LeftRightChooserModel.js
```diff
@@ -173,21 +172,21 @@ export class LeftRightChooserModel {
...
             });
     }
 
     moveRows(rows) {
         rows.forEach(rec => {
             if (rec.locked) return;
-
-            const rawRec = this._data.find(raw => raw === rec.raw);
-            rawRec.side = (rec.side === 'left' ? 'right' : 'left');
+            rec.raw.side = (rec.side === 'left' ? 'right' : 'left');
```
amcclain: I'll be honest - I am not gonna dive into LRChooser code right now, but this seems very weird to be modifying the raw object.

amcclain:
@haynesjm42 - Lee and I just talked this over a bunch more and he is going to press forward with some additional refactoring.  Looking to collapse BaseStore+LocalStore -> Store + provide an efficient children getter on record (not directly related to this raw data change but will roll it all up into a bigger data package redo).

lbwexler:
This should be good to review and merge.  Thanks to everyone for getting this in a good place.  I think we now have a very tight, efficient TreeStore, with a very usable API.

data/impl/RecordSet.js
```diff
@@ -17,52 +15,38 @@ import {Record} from '../Record';
...
+            } else {
+                children.push(record);
+            }
+        });
     }
 
-    /** Total number of records contained in this RecordSet. */
     get count() {
-        return this.records.size;
+        return this.records.size();
```
haynesjm42: Don't think this is correct, `records` is a Map and `size` is a property of Map.

lbwexler:
made a few tiny improvements -- john, can you take one more look ?

data/Store.js
```diff
@@ -97,11 +97,11 @@ export class Store {
 
     /**
      * Remove a record (and all its children, if any) from the store.
-     * @param {(string|number)} id - ID of the the record to be removed.
+     * @param {(string[]|number[])} ids - IDs of the records to be removed.
      */
     @action
-    removeRecord(id) {
-        this._all = this._all.removeRecord(id);
+    removeRecords(ids) {
```
haynesjm42: We could do a `ids = castArray(ids)'` here just to make it nice and clean for apps to remove a single record (something I tend to do a lot in inline editable grids)

amcclain:
Great - excellent to see this landed!  🚀

PR #1084, opened 2019-04-20
Layout components review

TomTirapani:
See issue https://github.com/exhi/hoist-react/issues/859


mobile/cmp/mask/Mask.js
```diff
@@ -60,7 +60,7 @@ export class Mask extends Component {
                 items: [
                     showSpinner ? progressCircular({indeterminate: true}) : null,
                     showSpinner ? vspacer(10) : null,
-                    message ? box({className: 'xh-mask-text', item: message}) : null
+                    message ? div({className: 'xh-mask-text', item: message}) : null
```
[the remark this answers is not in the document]
TomTirapani: No, no particular reason - except we're not using LayoutSupport. There's no problem with using box here, just a question of preference :)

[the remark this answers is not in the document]
TomTirapani: My 2 rials:
I was thinking about it slightly differently from you, where `div` is the "base unit" and we should only switch to `box` when actually using `layoutSupport` props. I think its important because `box` _isn't_ a drop-in replacement for `div` - it changes how a bunch of props are handled, and while div > box changes shouldn't cause issues, box > div changes can break layouts. We shouldn't think of them as interchangeable, since they serve different purposes.

CHANGELOG.md
```diff
@@ -33,6 +33,14 @@
   than mutating it.
 * The method `Store.removeRecord` has been replaced with the method `Store.removeRecords`. This will
   facilitate efficient bulk deletes.
+* The low-level `Box` component no longer applies default flex-related styles. Previously `Box` and
+  `HBox` were functionally identical, given that `Box` declared `display: 'flex'` and the default
+  for flex boxes is `flex-direction: 'column'`. Now `Box` makes no assumptions about flexing its
```
lbwexler: isn't the default flex-direction: 'row'?

cmp/layout/Box.js
```diff
@@ -23,7 +25,6 @@ export class Box extends Component {
     render() {
         let {children, ...props} = this.getNonLayoutProps();
         props = merge(
-            {style: {display: 'flex', overflow: 'hidden', position: 'relative'}},
             {style: this.getLayoutProps()},
```
lbwexler: wasn't this position: 'relative' important for something?

cmp/layout/Frame.js
```diff
@@ -7,21 +7,23 @@
...
 export class Frame extends Component {
+    baseClassName = 'xh-frame';
     render() {
         return box({
             ...this.props,
-            flex: 'auto'
+            className: this.getClassName()
         });
     }
 }
```
lbwexler: so according to the logic of this change isn't Frame just a quiet alias for HFrame?  Shouldn't we just drop it?

lbwexler:
I am generally ok with this, but would just love to talk over these concerns.

Sorry for my misunderstanding the obvious point that even when you have just one child, the row-direction is critical....

CHANGELOG.md
amcclain: Yes you are correct - fixed.  I always get confused because flex "row" == "lay out children in a row, causing them to appear as, er, columns"

cmp/layout/Frame.js
amcclain: Yes good catch.  I think by our logic `.xh-frame` should just declare `flex: 1 1 auto;` (as a child we want it to be space-filling) but not `display: flex` (as a parent we don't want to say it must flex).

cmp/layout/Box.js
amcclain: I also would like to know what that's about in general - we could add a comment to its new home in the CSS.

I don't see why it should be something that an app would depend on being set for box if we are OK with its new role as "div with layout support," which I think remains a reasonable and easy-enough to explain reason for its existence.

mobile/cmp/mask/Mask.js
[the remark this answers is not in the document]
amcclain: Yes I agree with you there - no reason to reach for box if you truly just need a div and don't plan on using any layout props (things like padding included).  

Also (to state the obvious) nothing wrong with a div + CSS class to style however required.

lbwexler:
Having a few second thoughts about maintaining 'position: relative' and 'overflow:hidden' as defaults on Box.  I vaguely remember some subtle issue that required the former on all our components (Gil may have helped on it.)  The latter just seems intrinsic to what we mean by 'Box'....

TomTirapani:
Thanks for catching the Frame == HFrame issue. I've updated it to be consistent with our logic, so that Frame fills up its parent flex container but doesn't flex its children.

This required some app changes where we relying on the flex behaviour, but also allowed for a cleanup for the agGrid wrapper where we were fighting _against_ the enforced `display: flex`.

TomTirapani:
I wish I understood more about the reasoning behind `position: relative` and `overflow: hidden` everywhere. Neither seem like they should be necessary by default, but I'm sure there is a good reason to keep applying it to our layout classes.

That said, I'm inclined to not apply it to Box, since it muddles our "div with layout support" story. Unless of course we decide it *is* necessary, in which case we can apply with an `.xh-box` class and encourage the use of `box` over `div` as the lowest-level building block.

lbwexler:
Well, very eager to merge this, and also like the idea that box is just "div with layout support".

It's just interesting that *all* of the other layout classes are quietly putting on 'position:relative' which effects how the component fits in its own parent, not how it lays out.  Is it important or not?

Should we just whack Frame?  Doesn't it just complicate our story unnecessarily by adding a multiple inheritance of sorts to how you think of these container behaviors?  For most apps, they can and should simply be replaced with HFrame or VFrame.

amcclain:
Another alternative to frames would be to simply use the box classes with the desired flex value.  In my experience, setting `flex: 1` seems to typically do what you want, although I realize this is not the same as `flex: 1 1 auto`.

I would support dropping frames if we thought this would be the common update/replacement needed.

TomTirapani:
Okay, will try whacking frames all together. I think explicitly setting `flex: 1` is easier to understand, and may even allow us to reduce some nesting in our components.

lbwexler:
OK -- to be clear, I was thinking we could possibly just whack Frame, and
leave VFrame and HFrame.

Would be nice to avoid the boilerplate of flex: 1 -- especially if that is
often the only attribute we would need to set.

Would love to discuss -- just wanted to make sure we consider that option.

-Lee

amcclain:
I can revisit this in the context of our new functional world.  Will report back here and either set it up again for review or close out with any further thoughts.  Thanks!

PR #1099, opened 2019-05-01
Allow more flexible export formatting with exportType col prop, Fixes #1097

### LINKED ISSUE #1097: GridExportService should be more flexible about date formats
This JS service prepares the columns for export to conform to what the server expects. This includes setting the column 'type' ('date' or 'datetime') and formatting the date to a string specific to the server's needs.

Currently it will only do this if the Column.exportFormat == ExportFormat.DATE_FMT for 'date' column type, or Column.exportFormat == ExportFormat.DATETIME_FMT for 'datetime'. Essentially we are providing a exportFormat property but then limiting that string to only one possible value.

We could start by allowing the column def to specify a exportColType property and using that to make these determinations before falling back to the Column.exportFormat.

febbraiod:


febbraiod:
Changes to admin/tabs/activity/tracking/ActivityGridModel.js are for demo only, should be removed prior to merge

lbwexler:
Thanks all.  I looked at this with Don.   It seems like we are leaving the user with a potentially more confusing story around this new 'exportType'.  

As an application developer I will be thinking do I need to set this or not?  If I don't set it, will I always get a string.  Let's continue to think on this?  Would rather keep the API to a single 'exportFormat'.

Also, note, we have already given users the option to type the field in the store.  Now they (maybe) need to do it *again* for export?  Can we just use the type of the field if we need additional info for the server?

lbwexler:
Closing this, per discussion above.  Will leave the ticket open and we can take another crack at it.

PR #1100, opened 2019-05-02
Cube cleanups

lbwexler:
+ Rename Record -> CubeRecord
+ Remove additional cruft related to streaming updates.
+ Cleanup documentation, use ES6 collections, iterations

### DIFF
--- a/data/cube/Query.js
+++ b/data/cube/Query.js
@@ -7,7 +7,7 @@
 
 import {XH} from '@xh/hoist/core';
 import {ValueFilter} from '@xh/hoist/data/cube';
-import {values} from 'lodash';
+import {pick, map} from 'lodash';
 
 /**
  *  Specification used to define the shape of data to be consumed from cube.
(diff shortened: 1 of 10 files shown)


data/cube/View.js
```diff
@@ -108,26 +107,23 @@ export class View {
...
-                    const lVal = rec.get(f.name);
-                    if (lVal != null && !vals.includes(lVal)) {
-                        vals.push(lVal);
-                    }
-                });
-
-                ret.push({
-                    name: f.name,
-                    displayName: f.displayName,
-                    values: vals
```
shravnk: Don't think we're using this right now, but making `values` a Set would change the API, no? Obviously benefit in terms of code cleanup

data/cube/Query.js
```diff
@@ -89,17 +89,8 @@ export class Query {
     // Implementation
     //---------------------------
     parseFields(names) {
-        const cubeFields = this.cube.fields;
```
shravnk: See the benefit of updating to ES6/lodash, just want to confirm that this safety check is unnecessary/unwarranted

data/cube/View.js
lbwexler: yes -- but we are essentially at a clean slate with the API -- the sencha upgrade is never going to be a dropin at this point.

data/cube/Query.js
lbwexler: This was probably me going overboard trying to clean things up.  Agree the safety check is probably worth it.  Will restore post-merge

PR #1104, opened 2019-05-06
Hoist Admin always see VersionBar in the bottom of the page

YanaNirvana:
Applies to both desktop and mobile!

💎😁


amcclain:
Thanks - this was my pet ticket, as I always like to see that footer in production mode.  👍

Issue #1065, opened 2019-04-11
Option to anchor toasts to a component


amcclain:
Not expecting this to be supported by current toast implementations, but would be really nice to trigger a toast that was positioned relative to a specified component.

For complex layouts / wide screens / etc the app-wide corner toast can be far away from the context of whatever operation triggered the toast.

amcclain:
This is great, thank you.  I tweaked the Toolbox branch you had a bit to use the existing toasts we were showing off the SampleGrid row-level action buttons - those are now nicely anchored to the grid itself.

PR #1106, opened 2019-05-07
Anchor toasts

shravnk:
Adds a `containerRef` param to XH.toast which accepts a ref to a HoistComponent. `ToastSource` uses the ref to pull out the component's underlying DOM element and generate a unique id for the toaster. Importantly, this implementation requires a Blueprint CSS override (see App.scss in linked Toolbox branch):
```
  .bp3-toast-container {
    position: absolute;
  } 
```
@lbwexler discussed this implementation, and considered whether it's safe to add the override as a Hoist default. 

For #1065. See https://github.com/exhi/toolbox/tree/anchorToasts for demo. 

Screen Shot 2019-05-07 at 1 18 54 PM


lbwexler:
@TomTirapani -- Can you review this with Brendan?

TomTirapani:
Looked at this and discussed with Brendan. I brought the CSS positioning override into hoist, with the caveat that its only applied when the Toaster is anchored.

Looks great 👍

core/XH.js
```diff
@@ -336,10 +336,11 @@ class XHClass {
      *
      * @param {Object} config - options for toast instance.
      * @param {string} config.message - the message to show in the toast.
-     * @param {element} [config.icon] - icon to be displayed
+     * @param {Element} [config.icon] - icon to be displayed
      * @param {number} [config.timeout] - time in milliseconds to display the toast.
      * @param {string} [config.intent] - The Blueprint intent (desktop only)
      * @param {Object} [config.position] - Position in viewport to display toast. See Blueprint Position enum (desktop only).
+     * @param {Element} [config.containerRef] - Parent component ref (defaults to document.body if undefined)
```
lbwexler: The icon "{Element}" is actually a react element.

The containerRef "{Element}" is actually an HtmlElement in the blueprint API, but a Component in our API.    We need to be clearer on this.  If we want it to be either,  let's say that and support that.  Otherwise, let just say its a Component defaulting to `null`...  
 

 const containerElement = document.createElement("div");
        container.appendChild(containerElement);

desktop/appcontainer/ToastSource.js
```diff
@@ -53,13 +56,18 @@ export class ToastSource {
      * If non-default values are needed for a toaster, a different method must be used.
      *
      * @param {string} [position] - see Blueprint Position enum for allowed values.
+     * @params {Element} [containerRef] - see Blueprint 'container' parameter
      */
```
lbwexler: See comment below on the type of this thing.

desktop/appcontainer/ToastSource.js
```diff
@@ -53,13 +56,18 @@ export class ToastSource {
...
-    getToaster(position = Position.BOTTOM_RIGHT) {
-        const toasters = this._toasters;
+    getToaster(position, containerRef) {
```
lbwexler: Ah, I see, this is a private class so we are just whacking this extraneous default

desktop/appcontainer/ToastSource.js
```diff
@@ -53,13 +56,18 @@ export class ToastSource {
...
+        const toasters = this._toasters,
+            container = containerRef ? containerRef.getDOMNode() : document.body,
+            containerId = containerRef ? containerRef.xhId : 'viewport',
```
lbwexler: The need for a unique id to recognize the toaster definitely argues for asking the user pass in a Component.  Got it! 🌮

shravnk: Yes, but your point about the comment on line 59 is well taken. E.g. 
`[containerRef] component which passes its underlying DOM node as Blueprint's containerId params, and is used to set a unique ID for this Toaster instance

Issue #1086, opened 2019-04-23
StoreFilterField loses state on unmount


shravnk:
Specifically referring to the `value` observable which populates the component's TextInput. `value` is reset on unmount because it's defined on the StoreFilterField component itself. @lbwexler and I discussed:

1. Binding to a `LocalStore` property
2. Requiring the value to be manually defined (e.g. on the enclosing `Panel`'s model)

lbwexler:
(comment from initial PR, which placed the state for the comment on the Store itself:)

Cool idea, but would like to really keep Store unaware of Store filter-field's state. Its an inversion of the layers, and not worth it to fix this bug.

If we really want to associate the filter state with that store in a stateful way, and we don't want to build a dedicated model in store filter field, or have it use its parent to hold the state, we could consider a static map in store filter field, of store.xhId to filter text.

PR #1107, opened 2019-05-07
Move StoreFilterField's state to Store

shravnk:
Simplest fix for the unmounting bug described in #1086. Could also make the StoreFilterField's parent model responsible for managing its value, which improves clarity and follows Hoist pattern. Concern is that would be a pervasive breaking change for a relatively minor bug fix

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -5,7 +5,7 @@
  * Copyright © 2019 Extremely Heavy Industries Inc.
  */
 
-import {observable, action} from '@xh/hoist/mobx';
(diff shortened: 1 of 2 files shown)


lbwexler:
Cool idea, but would like to really keep Store unaware of Store filter-field's state.  Its an inversion of the layers, and not worth it to fix this bug.  

If we really want to associate the filter state with that store in a stateful way, and we don't want to build a dedicated model in store filter field, or have it use its parent to hold the state,  we could consider a static map in store filter field, of store.xhId to filter text.

PR #1108, opened 2019-05-07
Add built in Hoist AutoRefresh Service

shravnk:

### DIFF
--- a/core/AppSpec.js
+++ b/core/AppSpec.js
@@ -47,6 +47,7 @@ export class AppSpec {
         isMobile,
         isSSO,
         checkAccess,
+        autoRefreshEnabled = true,
         trackAppLoad = true,
         idleDetectionEnabled = false,
         idleDialogClass = null,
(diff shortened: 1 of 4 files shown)


shravnk:
@lbwexler and I discussed the distinction between enabling auto-refresh across a data environment (i.e. with a boolean config) vs enabling for specific apps (e.g. exclude mobile, admin clients). This PR currently uses a global boolean config and a new autoRefreshEnabled property on `AppSpec`. We agreed that a single default JSON config might be cleaner - something like:
```
    xhAutoRefreshConfig = {
        enabledApps: ['app'],
        allowedIntervals: [-1, 15, 30, 60]
    }
```
There's also the question of creating a default UI control for auto-refresh enabled apps. We currently don't support default `appOptions`, so perhaps beyond the scope of this PR. For now could just tighten up our auto-refresh app option in Toolbox. @amcclain wanted to get your opinion before proceeding

amcclain:
Closing - will open up replacement on a new branch shortly.

PR #1114, opened 2019-05-09
Rename AppOption.requireRefresh -> requireReload

amcclain:
Trigger a _refresh_ of the app (data) if any options changed that don't require a _reload_.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,21 +1,29 @@
 # Changelog
 
-## Upcoming Release
+## Upcoming Release - under development
 
 ### 🎁 New Features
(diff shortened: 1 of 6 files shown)


amcclain:
Note trivial matching PR exhi/toolbox#181

amcclain:
We could easily unwind the auto-refresh part of this change - I thought that would be handy for a current client app project, and it still could, but realized I could also setup a listener on prefService and do a more targeted refresh of the component that matters.  That ends up being a better behavior, although the code to setup and tear down the listener is some extra complexity within the app component.

PR #1118, opened 2019-05-09
Adjust background color for ButtonGroup active input

amcclain:


amcclain:
Great thank you - a clear improvement.  I played around a bit more with the color on light theme and looking at the gradients, but nothing that seemed obviously better.  We can keep tuning but this fixes the immediate legibility issue for sure.

PR #1123, opened 2019-05-15
Add Store.removeAllRecords()

amcclain:
A low pressure pull request here. This seemed like a useful-enough convenience method, and I can imagine it being helpful to have an official API for efficient clearing vs. apps calling `loadData([])` as they do currently.

Initially called it `clear()` and would like the short name, but consistency with existing `removeRecords(ids)` seemed worth it.


lbwexler:
Agree, but I think we should call it clear😄

PR #1125, opened 2019-05-16
Date input focus

lbwexler:


desktop/cmp/input/DateInput.js
```diff
@@ -31,9 +38,6 @@ export class DateInput extends HoistInput {
         ...HoistInput.propTypes,
         value: PT.instanceOf(Date),
 
-        /** True to commit on every change/keystroke, default false. */
```
lbwexler: Is this no longer supported?

lbwexler:
Thanks for the tremendous work here Tom.  I am a huge fan of this one and would be inclined to merge it as soon as we get some clarity regarding commit on change....  There are still a number of issues with the blueprint calendar picker, but I suppose we will eventually get to them.

TomTirapani:
Thanks for taking a look Lee.

Yes, I took `commitOnChange` out with this change - it was tricky to get it working well with the new separate `TextInput`. The committed parsed date would stomp on whatever you were typing, meaning the input value could change under you as you type.

I have a few ideas of how we could work around this, but I'm not convinced `commitOnChange` is worth supporting here.

I think the ideas of "commit as you type" and "parse a date from the typed in string" are fundamentally at odds with each other.

As we enter / change date strings, the partial strings are parsed and, more often than not, interpreted as different date from intended. To see what I mean, try playing around with the second `DateInput` on https://toolbox.xh.io/app/forms with `commitOnChange` switched on - it's remarkably easy to get it into a state where the committed value doesn't match the partially entered string (e.g. start deleting the 2019 and the committed date will jump to 2020).

With other `commitOnChange` inputs, the typed-in value always matches the committed value, whereas here the committed value is the "best guess" parsed date of whatever partial string you have typed-in. I think committing a different value to what the user sees is only going to lead to unexpected validation behaviour and error reports.

If we decide to *do* want to support `commitOnChange`, we have a couple of options:

1) Tighten up the date parsing, so it only accepts a full date string, avoiding the partial date mismatch (although this begs the question of why use `commitOnChange` at all).

2) Use a different `inputValue` observable to control the text input, so we can avoid stomping on it when we commit. I believed we've used a similar pattern before in our `QueryComboField` days.

lbwexler:
Hi Tom — I am fine with the Calendar Control — think its great!  

The only other thing I could see doing is have an option to control whether the picker comes up.  It could be a boolean — enablePicker.

Can you make a changelog entry and we can prepare for merge?

Issue #925, opened 2019-01-24
Add client-side support for 'value dates'


lbwexler:
Its a very common problem for our business clients where the a field they want to display is a calendar date (e.g. settle date, trade date).  There is no time co-ordinate, and we don't want to express it as a javascript (or UTC) date that might be subject to timezone handling on different machines.

Support should be built around a YYYYMMDD representation of the date, and could include a dedicated HoistInput and built-in column type.

Probably useful for us to build this on top of our new DateInput control.

lbwexler:
Again, good for some thoughtful work during your trip!

lbwexler:
'CalendarDate' is ATM and my best take on it!  Feel free to suggest other names, but we like this one to start!

TomTirapani:
PRs here:

https://github.com/exhi/hoist-react/pull/1130
https://github.com/exhi/toolbox/pull/191

Added a HoistInput, column type and form validator. Anything else we can think of to include?

lbwexler:
Gents was thinking about this again -- was just looking at a client app that is really suffering due to the lack of support here -- they go back and forth all over the place between dates and strings,  in their management of a business day.  

Need to get support here for binding a dateinput to a YYMMDD date pronto.  Would like to re-up the priority here

TomTirapani:
Happy to get back on this next week 👍

PR #1130, opened 2019-05-16
Add client-side support for 'calendar dates' - issue #925.

TomTirapani:
+ Added CalendarDateInput
+ Added calendarDateIs validation constraint
+ Added calendarDateCol

See corresponding Toolbox PR: https://github.com/exhi/toolbox/pull/191


format/FormatDate.js
```diff
@@ -15,7 +15,8 @@ import {fmtSpan} from './FormatMisc';
 export const DATE_FMT = 'YYYY-MM-DD',
     DATETIME_FMT = 'YYYY-MM-DD h:mma',
     TIME_FMT = 'h:mma',
-    MONTH_DAY_FMT = 'MMM D';
+    MONTH_DAY_FMT = 'MMM D',
+    CALENDAR_DATE_FMT = 'YYYYMMDD';
```
lbwexler: Not sure about this format?  Aren't these formats primary for display -- wouldn't we stick with YYYY-MM-DD

TomTirapani: I just used 'YYYYMMDD' as that was the format requested in the ticket - happy to use either format

lbwexler: Tom -- think I meant here in the ticket that this would be the format for transferring it and storing it in the control, not displaying it!

(seemed like the tighter usage was our standard for API serialization)

lbwexler:
Other than the format, my only other question is whether `DayInput` would be less of a mouthful.

We have been using just "Day/day" in a Client app to signify the YYYYMMDD concept and it works pretty nicely!

TomTirapani:
Apologies for the confusion - didn't realise at first that we didn't want CalendarDates to be user-facing. I've updated the branch to avoid using our display formats, instead using our standard DATE_FMT to when displaying.

desktop/cmp/input/CalendarDateInput.js
```diff
@@ -0,0 +1,61 @@
...
+import {HoistInput} from '@xh/hoist/cmp/input';
+import {toCalendarDate, parseCalendarDate} from '@xh/hoist/utils/datetime';
+import {warnIf} from '@xh/hoist/utils/js';
+import {omit} from 'lodash';
+import moment from 'moment';
+
+import {DateInput, dateInput} from './DateInput';
+
+/**
+ * A wrapper around DateInput, that publishes a CalendarDate string as its value.
```
lbwexler: The fact that this is a 'wrapper' is an implementation detail

cmp/form/validation/constraints.js
```diff
@@ -99,4 +99,19 @@ export function dateIs({min, max, fmt = 'YYYY-MM-DD'}) {
         if (minMoment && minMoment.isAfter(value)) return `${displayName} must not be before ${minMoment.format(fmt)}.`;
         if (maxMoment && maxMoment.isBefore(value)) return `${displayName} must not be after ${maxMoment.format(fmt)}.`;
     };
+}
+
+/**
+ * Validate a CalendarDate against allowed min/max boundaries.
+ * @see dateIs
+ */
+export function calendarDateIs({min, max}) {
```
lbwexler: seems like min and max here should be other CalendayDates/Strings  (not Date)

cmp/grid/columns/DatesTimes.js
```diff
@@ -35,4 +36,12 @@ export const compactDateCol = {
     renderer: compactDateRenderer(),
     exportFormat: ExportFormat.DATE_FMT,
     width: 100
+};
+
+export const calendarDateCol = {
+    ...defaults,
+    renderer: (v) => fmtDate(parseCalendarDate(v)),
+    exportValue: (v) => parseCalendarDate(v),
```
lbwexler: Why wouldn't this be exported in its raw form?

lbwexler:
Hi -- Did a bit more research, and looked at Toolbox examples here.
A couple of random thoughts:  

1) The following link is interesting:  
https://www.npmjs.com/package/js-joda
Not saying we would go to this over moment.js, but the fact that they use the term 'LocalDate' is very interesting.

2) We would like our 'Field.type' enumeration take one of these things, and have stores/grids seamlessly validate and format them.  Should we add that now? (Think we had type= 'day' at some point, but we removed it beacuse it was unused)

3) Made some other comments inline in the PR

lbwexler:
Tom and I had a great discussion. In short:

+ The export is correct as is -- Excel needs to know this is a Date.

+ The validator for calendarDay should take the CalendarDay string representation for Max and Min.

+ The biggest question is whether we go a step further and either provide or import a simple CalendarDay object that will let users manipulate these things on the client in their string representation.  Otherwise, to just set contraints or initialize these things, you are required to go back and forth between Moment and JS Dates.   Options are 
https://github.com/js-joda/js-joda or https://www.npmjs.com/package/local-date or something homespun...

amcclain:
It's a tough one....mainly because joda-time does look very compelling given its link to Java and the fact that it advertises exactly the things we want, LocalDate included. If it is "2 to 10 times faster," that would be another nice plus when doing things like processing large datasets into date-ish objects.

Of course the big downside being it really doesn't seem right to have both this and moment as dependencies of Hoist, and it's clearly less common/well-known than moment (+ the migration effort)... 

I also looked only briefly, but it appeared as though the common task of formatting dates is a bit more verbose, requiring the use of a `DateTimeFormatter` (although those could presumably be imported as constants).

Sorry I know I haven't added much/anything to the conversation - mainly wanted to let you know I saw this was in progress and am thinking on it / happy to discuss more tomorrow...

PR #1133, opened 2019-05-17
Short formats for relative timestamps, Fixes #1115

### LINKED ISSUE #1115: Support short variants of RelativeTimestamp strings
In hoist-sencha we had a useful option on relativeTimestamp to use shorter variations on the output strings: https://github.com/exhi/hoist-sencha/blob/89aad521933850ca0af467c38fa7739bd284f5a0/grails-app/assets/src/xh/all/util/RelativeTimestamp.js#L45

This resulted in "5 minutes ago" -> "5 mins ago" and similar.

Would be useful to have the same the hoist-react based utility.

febbraiod:
Open more for review and discussion than merge

### DIFF
--- a/cmp/relativetimestamp/RelativeTimestamp.js
+++ b/cmp/relativetimestamp/RelativeTimestamp.js
@@ -7,7 +7,7 @@
 
 import {Component} from 'react';
 import PT from 'prop-types';
-import {HoistComponent, LayoutSupport, elemFactory, managed} from '@xh/hoist/core';
+import {XH, HoistComponent, LayoutSupport, elemFactory, managed} from '@xh/hoist/core';
 import {observable, action} from '@xh/hoist/mobx';
 import {box} from '@xh/hoist/cmp/layout';
 import {span} from '@xh/hoist/cmp/layout';
@@ -230,6 +230,6 @@ const getResult = opts => {
     // default FORMAT for seconds.
     if (!elapsedTime && useNowString) return suffix;
 
-    const fmtString = short ? SHORT_FORMAT_STRINGS[unit] : FORMAT_STRINGS[unit]
+    const fmtString = short ? SHORT_FORMAT_STRINGS[unit] : FORMAT_STRINGS[unit];
     return `${prefix}${fmtString.replace('%d', millis)} ${suffix}`;
 };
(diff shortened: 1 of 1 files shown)


cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -31,8 +31,23 @@ const FORMAT_STRINGS = {
...
+    days: '%d days',
+    month: 'a month',
+    months: '%d months',
+    year: 'a year',
+    years: '%d years'
+};
+
 const defaultOptions = {
     allowFuture: false,
+    short: false, // Defaulted based on device in getRelativeTimestamp method
```
febbraiod: So this is because XH is not ready here outside of a component/model. We seem to be doing things here in an old way. And by 'old' I mean a year, software is crazy :).

I could see use putting together a Component Model relationship here, but not sure we want to invest the time in work on a dinky component thats already sort of doing what we want.

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -110,6 +126,7 @@ export const relativeTimestamp = elemFactory(RelativeTimestamp);
  * @param {string} [options.emptyResult] - Returned when timestamp is undefined
  */
 export const getRelativeTimestamp = (timestamp, options) => {
+    defaultOptions.short = XH.isMobile;
```
febbraiod: See note above

lbwexler:
Don -- marking with WIP --- please use going forward if something is not ready for merging!

Thanks!

amcclain:
I would be fine to not have the defaulting based on mobile - honestly I am not sure how much this is about mobile vs. desktop - more about the context in which you are showing the timestamp.  Is it in a toolbar all by itself where you have lots of room (regardless of platform), or are you looking to fit it alongside a number of more important metadata and you want to show it as succinctly as possible.  In the end, the config exists for the developer to choose.

Was that the issue keeping this in WIP status?

febbraiod:
Well the approach I had to take for the defaulting called into question whether the component is an outlier in that it doesn't follow our standard patterns, and whether we should rework it to do so. I haven't spent a ton of time in Hoist React land this year so forgive me if I'm over thinking this.

cmp/relativetimestamp/RelativeTimestamp.js
lbwexler: can we change this to not be a fat arrow function?  typically use regular function declarations for export

febbraiod:
Discussed with Lee, few small changes coming, will look to close today

Issue #1119, opened 2019-05-10
Minor PanelModel config / splitter issues


amcclain:
* We should show a splitter on a Panel with a model configured as `resizable: true` and `collapsible: false`.  Currently we draw it without the splitter but you can weirdly grab an invisible border to resize.
* We should not create a header for that same resizable: true / collapsible: false Panel if it does not have a title, icon, or headerItems of its own.  Currently we do, and you weirdly need to set `showHeaderCollapseButton: false` to suppress.

YanaNirvana:
Merged.  See #1135

PR #1135, opened 2019-05-17
Panel splitter

YanaNirvana:
A `Panel` with configs `resizable: true, collapsible: false` now renders with a splitter.  #1119 

+ A `Panel` with no `icon`, `title`, or`headerItems` will not render a blank header


TomTirapani:
Looks great to me! 👍

Issue #1067, opened 2019-04-11
Support prefix for formatNumber


amcclain:
The specific use-case is for dollar signs - to ensure they end up within () when in ledger mode and take the desired color when showing red/green.

Not sure if there are other use cases or how / if to generalize further but the $ is important. Would it be clear that `prefix` takes the color and is inside ledger ()?  That is the case with `label`

YanaNirvana:
Just so I understand, when `ledger: true`, we want the `prefix` to be inside ledgers; `()` and take on the appropriate red/green color.

But what happens when `ledger: false`?  I'm assuming that we want `prefix` to be after the negative sign, when a number is negative.  e.g., we wouldn't want `$-13.37`.  Instead we want `-$13.37`?

PR #1136, opened 2019-05-20
Format number

YanaNirvana:
`formatNumber` now accepts an optional `prefix` config to be prepended to value:

+ Re: ticket #1067, and its specified use case for dollar signs (`$`)
+ `prefix` before the number, but after the sign (if any)
+  e.g., `-$13.37` or `+$5.20`

### DIFF
--- a/format/FormatNumber.js
+++ b/format/FormatNumber.js
@@ -253,17 +253,20 @@ function fmtNumberElement(v, opts = {}) {
 function fmtNumberString(v, opts = {}) {
     const {ledger, forceLedgerAlign, withSignGlyph, label, labelCls, colorSpec, tooltip, prefix} = opts;
     let str = opts.str;
+    const strOriginal = str;
 
     if (withSignGlyph) {
-        str = signGlyph(v) + '&nbsp;' + str;
+        str = signGlyph(v) + '&nbsp;';
     }
 
     if (isString(prefix)) {
-        if (v < 0) {
-            console.log('v < 0; \n \n this is the str!: ', str);
+        if (strOriginal.startsWith('-') || strOriginal.startsWith('+')) {
+            str += (strOriginal[0] + '$' + strOriginal.substring(1));
         } else {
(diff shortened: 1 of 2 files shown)


format/FormatNumber.js
```diff
@@ -218,7 +220,15 @@ function fmtNumberElement(v, opts = {}) {
         items.push(signGlyph(v, asElement));
     }
 
-    items.push(str);
+    if (isString(prefix)) {
+        if (str.startsWith('-') || str.startsWith('+')) {
+            items.push(str[0], '$', str.substring(1));
```
TomTirapani: Hi Yana.

I think here (and similarly below in `fmtNumberString`), we should ensure we are pushing the `prefix` string onto the `items` array. Here we are pushing a hardcoded `'$'` string, but we may want to use a different string for the prefix (e.g. `'£'`).

YanaNirvana: Hi Tom,

Great catch!  I think I was staring at the ticket for too long and got confused that I hardcoded the `$` in.  😅

Lee just informed me that the tickets I'm working on seem to fall under your expertise, and I'd love to discuss and review code with you from now on (in addition to supervision from Lee).

As I'm still working on this ticket, I'd like to check in with you on a couple things:

1.  I'd like to get symmetry back into `formatNumber` in that we should not support something like: `▴+1.618` That `signGlyph`s should replace all signs (`+` and `-`), not only `-`.  And that this should be the case even if user explicitly sets `withPlusSign: true` What're your thoughts on this?

2. Since `prefix` is to come between a number and its sign, I'm wondering if it's better to separate a number and its sign into two variables instead of the combined `str`.

TomTirapani:
Reviewed with Yana, looks great 👍

amcclain:
@cnrudd wanted to ensure you saw this made it into the latest Hoist - was a ticket I entered after we looked at a client app of yours last time I was in NY.

PR #1137, opened 2019-05-22
Add AsyncUtils with forEachAsync + whileAsync

amcclain:
+ Utils for processing compute-intensive or long running loops in a non-blocking manner.


amcclain:
See exhi/toolbox#193

lbwexler:
Merged tweaked version of this!

Issue #737, opened 2018-10-24
Add support for 'rootSummary' to Grid/GridModel


lbwexler:
Have definite use for a pinned summary row at top, analogous to what we have in HoistSencha's TreePanel.rootSummary.

Certainly AG-Grid has some support for rendering the pinned.  Would like to expose to Grid/GridModel for easy display of a non-scrolling, non-collapsing total row.

lbwexler:
Thought about this a bit more.  Perhaps could have the following semantics:

Add a new property to GridModel
`showSummary`: 'top/bottom/none/[false]

If the store is a TreeStore, and it has a single root node display that as the Total.   Otherwise just display a standard ag-Grid generated total row, with ag-Grid standard aggregation handling.

Need the ability to take the total row in from the store in the case of server-side aggregations, or client-side cube aggregations.

lbwexler:
Closed!

PR #1142, opened 2019-05-28
Grid Summary Row

haynesjm42:
#737 

Support for showing a summary/total row in our Grid component pinned to the top or bottom (uses ag-Grid pinned row feature)

In order to use the summary row the application needs to include a row in the store with an id of "summary" which contains the data to display, and set the new `showSummary` prop on the GridModel. If the store contains hierarchical data, the "summary" record will be excluded from the main record hierarchy (it is stored separately in RecordSet). I went through several iterations on this implementation and landed on this as the best approach. In the future if we end up merging the Aggregation support in Cube/CubeRecord over to Store/Record the interface for the summary support shouldn't need to change and would be able to handle summary record without needing the application to provide the it.

A caveat of using the summary row is that any app-defined pinned rows will be overwritten when the Grid component updates the summary row. An alternative approach to achieving summary rows in ag-Grid would be to use aligned grids, but that seemed like it would be a more complex implementation for our Grid/GridModel, though it has the advantage of not conflicting with pinned rows an application may want to set.

### DIFF
--- a/data/Record.js
+++ b/data/Record.js
@@ -4,7 +4,7 @@
  *
  * Copyright © 2019 Extremely Heavy Industries Inc.
  */
-import {isEqual, isNil, isString} from 'lodash';
+import {isEqual, isNil} from 'lodash';
 import {throwIf} from '@xh/hoist/utils/js';
 
 /**
@@ -53,7 +53,7 @@ export class Record {
     }
 
     get isSummary() {
-        return this.id === 'summary';
+        return this === this.store.summaryRecord;
     }
 
     /**
@@ -75,8 +75,7 @@ export class Record {
      * @param {Record} [c.parent] - parent record, if any.
      */
     constructor({data, raw, store, parent}) {
-        const {idSpec} = store,
-            id = isString(idSpec) ? data[idSpec] : idSpec(data);
+        const id = store.buildRecordId(data);
 
         throwIf(isNil(id), "Record has an undefined ID. Use 'Store.idSpec' to resolve a unique ID for each record.");
 
--- a/data/Record.js
+++ b/data/Record.js
@@ -52,6 +52,9 @@ export class Record {
         return this.store.getChildrenById(this.id, false);
     }
 
+    get isSummary() {
+        return this.id === 'summary';
+    }
 
     /**
      * Construct a Record from a raw source object. Extract values from the source object for all
--- a/data/Store.js
+++ b/data/Store.js
@@ -127,7 +127,7 @@ export class Store {
 
         const oldSummary = this.summaryRecord,
             newSummary = this.getRootSummary(rawData);
-        if (oldSummary && newSummary && summaryRecord.id === this.buildRecordId(newSummary)) {
+        if (oldSummary && newSummary && oldSummary.id === this.buildRecordId(newSummary)) {
             rawData = newSummary.children;
             rawSummaryData = {...newSummary, children: null};
         }
--- a/data/Store.js
+++ b/data/Store.js
@@ -30,13 +30,13 @@ export class Store {
      */
     @observable lastUpdated;
 
-    /** @member {Record|null} - record containing summary data. */
+    /** @member {Record} - record containing summary data. */
     @observable.ref summaryRecord = null;
 
     @observable.ref _all;
     @observable.ref _filtered;
     _filter = null;
-    _treatRootAsSummary = false;
+    _loadRootAsSummary = false;
 
     /**
      * @param {Object} c - Store configuration.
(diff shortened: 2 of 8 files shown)


haynesjm42:
Toolbox example: https://github.com/exhi/toolbox/pull/196

data/Record.js
```diff
@@ -52,6 +52,9 @@ export class Record {
         return this.store.getChildrenById(this.id, false);
     }
 
+    get isSummary() {
+        return this.id === 'summary';
```
lbwexler: I don't think we want to require a special id.   The id's of our root nodes can be significant.  Let's discuss whether it would be ok to assume that a single row with zero siblings is the summary?

haynesjm42: tl;dr - I ran into complications trying to use the shape of the data to identify the summary record, so went down the path of making it explicitly identified.

I was initially handling it the way you suggest, with treating a root node with no siblings as the "summary" record, but that really didn't work out well in all cases.

In the case that the data itself is not hierarchical I didn't want to force apps to add a root node as the means of supplying a summary record, though I could be persuaded on this point.

Also in order to make sure the summary record itself is handled correctly when filtering, I am keeping it completely separated from the main record list/map. This seemed like the cleanest solution for this, and didn't require that additional logic be added in various methods in RecordSet to handle the summary record. To have things work that way it is necessary to know exactly which record is the summary record when the data is processed by RecordSet, and since it is entirely possible that the data being loaded into the store has a single root node which is _not_ a summary record, it is necessary to be explicit when identifying the summary record instead of following some kind of convention based on the shape of the data.

I see a few options here for explicitly identifying the summary record:

1. Have a "special" id we use to identify, but make this something that can be configured on the Store (`summaryIdSpec`?). This would allow the app to keep their root ids. We could default it to `xhSummaryRecordId` or something and have a constant to clean up the code and make it easy for apps to access it. 
2. Instead of using the id, we could look for an `isSummary` prop on the raw data to identify the summary record. This would mean reserving another prop on Record that apps couldn't use for fields, and wanted to avoid doing that, so I went with the id approach.
3. Separate the summary data even more from the other data by having a `setSummaryData` or `loadSummaryData` on Store, and never even keep it with the other records. Would require either having a separate RecordSet for the summary record or some refactoring would be needed to be able to create the actual record without going through a RecordSet.

Please let me know if I am over-complicating things here :) Would be happy to have a simpler means of managing the summary record.

haynesjm42:
The summary row should not be effected by store filtering at all, it is kept separate from the other records for this reason. The summary row will always be visible, even if the store filter would result in 0 records shown.

The same implementation should work fine on mobile, I added an example in the toolbox branch for the mobile tree grid. Were you thinking we needed a different approach for summary rows on mobile?

lbwexler:
John and I went through this in great detail this afternoon.  In particular, he explained to me the rational behind having an explicit "out-of-band" handling of the summary record in Store, vs. just interpreting the single root as the total.

I have to say I wasn't 100% convinced and we agreed to continue thinking about it and get others to chime in.  I am still kind of partial for just looking for a root of any kind.  John mentioned that in the case of flat data, it's strange to introduce the hierarchical data/children thing for the purposes of the total row alone.

It does seem to me that we would never want the filter to whack our total row, so we want storeFilter to also recognize the summary in some way.....

In any case,  I do feel pretty strongly that if we want to identify a total row in the rawData other than having it be the root,  I would prefer a dedicated flag along the lines of "isStoreTotal' versus putting the magic string in the id....

data/Store.js
```diff
@@ -74,25 +83,62 @@ export class Store {
...
-    loadData(rawData) {
+    loadData(rawData, rawSummaryData) {
+        const {_treatRootAsSummary} = this;
+        throwIf(_treatRootAsSummary && rawSummaryData,
+            'Store is configured to treat the root node as summary data but rawSummaryData was provided to loadData!');
+
+        if (this.haveRootSummary(rawData)) {
+            rawSummaryData = rawData[0];
+            rawData = rawSummaryData.children;
+            delete rawSummaryData.children;
```
lbwexler: hmm -- we were trying not to mutate the raw data in this object.  Shouldn't we clone the rawData[0] here?

data/cube/impl/QueryExecutor.js
```diff
@@ -20,7 +16,7 @@ export class QueryExecutor {
     static async getDataAsync(query) {
         const {dimensions, includeRoot, fields, cube, filters} = query,
             cubeRecords = cube.records,
```
lbwexler: should be able to revert this change

Issue #1041, opened 2019-03-28
Add GridModel.cellBorders config


amcclain:
We have a `GridModel.rowBorders` config to turn on/off a standard bottom row border. 

Would like to add a corresponding `cellBorders` config to show Excel-style borders around each cell.

Note:
* Should not conflict with focused cell borders, if enabled.
* Should be sure to avoid double-borders, etc. 
* Can use the same `--xh-grid-border-color` CSS var for the color.

PR #1145, opened 2019-05-29
GridModel.cellBorders

YanaNirvana:
Addresses Issue #1041 & also:
+ moved `showHover` around to be more consistent with configs order


TomTirapani:
Looks great 👍

PR #1146, opened 2019-05-29
LoadingIndicator component + integration with Panel

cnrudd:
This is a new loadingIndicator component that can be put onto a panel just like a mask.  It uses the same PendingTaskModel.

Unlike a mask, the loadingIndicator does not block or obscure the panel contents.  It is a bar that appears in a panel corner (bottom right by default) to let the user know the panel content is loading.

This PR is paired with this [PR in Toolbox](https://github.com/exhi/toolbox/pull/197).


amcclain:
Thanks, Colin - I will take a look today. As mentioned, was just discussing a use case for this is another client app.

desktop/cmp/loadingindicator/LoadingIndicator.js
```diff
@@ -0,0 +1,116 @@
...
+        maxMessageLength: PT.number,
+
+        /** Optional text to be displayed - can also be sourced from bound PendingTaskModel. */
+        message: PT.string,
+
+        /** Optional model for reactively showing the indicator while tasks are pending. */
+        model: PT.instanceOf(PendingTaskModel),
+
+        /** Click handler. **/
+        onClick: PT.func,
```
amcclain: Do we have a use-case for this?  Assume we do (and it does not add any meaningful complexity).  Just wanted to be sure.

mobile/cmp/panel/Panel.js
```diff
@@ -10,6 +10,7 @@ import {castArray, omitBy} from 'lodash';
 import {elemFactory, HoistComponent, LayoutSupport} from '@xh/hoist/core';
 import {vbox, vframe} from '@xh/hoist/cmp/layout';
 import {toolbar} from '@xh/hoist/mobile/cmp/toolbar';
+import {loadingIndicator} from '@xh/hoist/desktop/cmp/loadingindicator';
```
amcclain: We should never have imports from the desktop package into mobile!  

Since the LoadingIndicator is relying on the spinner out of Blueprint, it's also problematic to promote to a cross-platform component as it is.  For this iteration, I would be OK leaving it out of mobile unless we knew we had an immediate use for it.  If/when we do need it, we could look to use a spinner that's not sourced from Blueprint.

desktop/cmp/loadingindicator/LoadingIndicator.js
lbwexler: I would be up with getting rid of this here and for mask (which is where it was copied from).
I think there are issues with its implementation (IIRC, Yana found it not working on mask) and its just cruft.

desktop/cmp/loadingindicator/LoadingIndicator.js
```diff
@@ -0,0 +1,116 @@
...
+
+
+    //------------------------
+    // Implementation
+    //------------------------
+    renderInnerItems() {
+        const {corner, message, showSpinner} = this,
+            spinnerEl = spinner({size: this.spinnerSize});
+
+        if (!message) return [spinnerEl];
```
lbwexler: seems weird that spinner will be shown here, even if use has asked for it not to be with showSpinner= false

desktop/cmp/loadingindicator/LoadingIndicator.js
```diff
@@ -0,0 +1,116 @@
...
+            case 'tr':
+            case 'br': return [msgBox, showSpinner ? spinnerEl : null];
+        }
+    }
+
+    get message() {
+        const {maxMessageLength, message, model} = this.props;
+        return truncate(withDefault(message, model && model.message), {length: withDefault(maxMessageLength, 30)});
+    }
+
```
lbwexler: Would make maxMessageLength a getter, like the other two below, for symmetry

desktop/cmp/panel/Panel.js
```diff
@@ -138,7 +148,17 @@ export class Panel extends Component {
             maskElem = maskProp;
         }
```
lbwexler: Can we DRY up 3) and 4).  e.g. 
`const maskElem = parseLoadDecorator(maskProp, mask )`
`const loadIndicatorElem = parseLoadDecorator(loadIndicatorProp, loadIndicator)`

mobile/cmp/panel/Panel.js
lbwexler: We don't have a mobile spinner-like thing?  Don't we need one?

lbwexler: (but am fine with leaving it off mobile to start)

mobile/cmp/panel/Panel.js
```diff
@@ -10,6 +10,7 @@ import {castArray, omitBy} from 'lodash';
...
 import {mask} from '@xh/hoist/desktop/cmp/mask';
```
lbwexler: YIKES -- we are already importing the desktop mask in to mobile?!?  Need to fix this ASAP

mobile/cmp/panel/Panel.js
TomTirapani: We have a mobile spinner called `progressCircular`, that's being used as part of the mobile mask cmp - could use that if we continue with the mobile route

mobile/cmp/panel/Panel.js
TomTirapani: OH WOW. I'll fix and update in a separate PR

desktop/cmp/loadingindicator/LoadingIndicator.js
cnrudd: If execution has gotten this far, with message = null, then showSpinner must be true.
See line 66.
We still need showSpinner in renderInnerItems to know how to layout a message alone.

desktop/cmp/loadingindicator/LoadingIndicator.js
cnrudd: Done.

desktop/cmp/panel/Panel.js
cnrudd: Done.

desktop/cmp/loadingindicator/LoadingIndicator.js
cnrudd: Removed  onclick prop from mask and loadingIndicator in desktop and mobile.

mobile/cmp/panel/Panel.js
cnrudd: I've made a mobile loadingIndicator component. and switched the mobile panel to use it.   I have not switched the mobile panel to use the mobile Mask.  Tom said he would do that in a separate PR.  I don't want to get in his way on that.

lbwexler:
Thanks for the changes -- am good with this.  Please just make sure ATM has a chance to review and also add a changelog entry!

Issue #688, opened 2018-10-04
Debounce synchronization between GridModel and Grid


lbwexler:
Anselm has noticed that noteAgColumnStateChanged() calls might benefit from being debounced.  I would like to extend the work to generally audit all the reactions and handlers and debounce whatever makes the most sense

See https://github.com/exhi/hoist-react/pull/680

lbwexler:
I am closing this.  It's a vague old ticket, based on a vague sense that things could be better.  We have since worked on a number of grid performance issues, and will continue to jump on them as they arise.

PR #1157, opened 2019-05-31
Grid/Store Selection Optimizations

### LINKED ISSUE #1150: Improve performance of programmatic grid select
When a grid selModel has many rows selected, it gets quite slow. It can take tens of seconds for a thousand rows.

For example, to select all rows:
* This is slow: gridModel.selModel.select(gridModel.store.records, true);
* The native ag-Grid select all is fast: gridModel.agApi.selectAllFiltered();

haynesjm42:
Fixes #1150 

The cause of the poor performance when programmatically selecting several records was that the ag-Grid onSelectionChanged event was being fired for each call we make to `node.setSelected(true)` (on the subsequent frame). This results in the Grid updating the selection model based on the current ag-Grid selection state, which would then trigger another selectionReaction in the grid, etc. So if you are selecting 1000 rows programmatically you would end up with 1000 extra updates of the observable.ref `ids` property of StoreSelectionModel, and 1000 extra calls to the Grid selectionReaction. The selectionReaction would be a no-op because we do check that before updating the ag-Grid state, but we did not have a similar check on the other end before we update our selection model state, and the cost of updating the observable prop and triggering the reaction was causing the very noticeable slow-down.

ag-Grid does not appear to have any kind of `setSelectedRows` method on it's api which would result in the single selection changed event, so it is necessary for us to handle it in our logic.

I made 2 changes to optimize our handling of updating the selection model state. Either would serve to address this specific issue, but I felt both were appropriate changes to make.

1. Debounce (until the next frame) updating of the selection model state when the ag-Grid onSelectionChanged event is fired. This is something we might want to consider doing for every ag-Grid event handler since we always use those events to trigger a full sync of our model with the ag-Grid state, and it is not clear if they may be fired multiple times per change if some kind of bulk operation was performed (such as hiding multiple columns, etc.)
2. Only update the ids prop in StoreSelectionModel if the ids have actually changed. I think this is a sensible optimization and a general best-practice when updating observable.ref properties to avoid triggering reactions unnecessarily.

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -409,6 +409,9 @@ export class GridModel {
         this.applyColumnStateChanges(colStateChanges);
     }
 
+
+    // We debounce this method because ag-Grid will fire the selection changed event for each node
+    // that is being selected
     @debounced(0)
     noteAgSelectionStateChanged() {
         const {selModel, agGridModel} = this;
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -409,13 +409,9 @@ export class GridModel {
         this.applyColumnStateChanges(colStateChanges);
     }
 
-    /**
-     * Sync selectionModel state from the grid's current selection.
-     *
-     * We debounce this method because the implementation of `AgGridModel.setSelectedRowNodeIds()`
-     * selects nodes one-by-one, and ag-Grid will fire a selection changed event for each iteration.
-     * This avoids a storm of events looping through the reaction when selecting in bulk.
-     */
+    // We debounce this method because the implementation of `AgGridModel.setSelectedRowNodeIds()`
+    // selects nodes one-by-one, and ag-Grid will fire a selection changed event for each iteration.
+    // This avoids a storm of events looping through the reaction when selecting in bulk.
     @debounced(0)
     noteAgSelectionStateChanged() {
         const {selModel, agGridModel} = this;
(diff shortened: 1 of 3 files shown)


amcclain:
See #688 for an older, general ticket on debouncing - agreed this still makes sense all around.

Will take a look at this PR shortly in more detail thank you

cmp/grid/GridModel.js
```diff
@@ -408,6 +409,19 @@ export class GridModel {
         this.applyColumnStateChanges(colStateChanges);
     }
```
lbwexler: + I  might put these comments in a block within the method, rather than a jsdoc on the method.
+ Also would support applying to noteAgColumnStateChanged() for symmetry

data/StoreSelectionModel.js
```diff
@@ -76,6 +76,12 @@ export class StoreSelectionModel {
         }).filter(id => {
             return this.store.getById(id, true);
         });
+
+        // Avoid trigger selection reactions unnecessarily by making sure the selected ids have actually changed
```
lbwexler: This comment is unnecessary

lbwexler:
I love it!  And you are using the debounced decorator which puts a huge smile on my face.   Made some minor comments, and would support changing the other handlers to be symmetrical (e.g. debounced(0) method on the model)

cmp/grid/GridModel.js
amcclain: Cool - I had flipped it to a doc comment but don't feel strongly.  I mainly wanted to expand on it a bit as it's the kind of thing I come back to and extra clarity always helps.

These model methods that are called from the component are in kind of an in-between state vis-a-vis public/private, but in general more like a private method for the entire system.  I.e. no reason for app developers to call (arguing for the block comment).

Issue #903, opened 2019-01-02
Add propType entry for comps that accept (and especially require) a model prop


amcclain:
A model is an essential prop for several components, but is not consistently documented within propTypes.  This can lead to non-obvious errors such as NPEs if not provided.

An common spec might be:

`model: PT.oneOfType([PT.instanceOf(FooModel), PT.object]).isRequired`

for a component that requires a model passed in via props to work but will accept either a real instance of its model or a config object for the same.  (Of course other comps such as Panel don't require a model and should leave off the `isRequired` but should still have a propType listed.)

PR #1160, opened 2019-06-03
Documentation on components model #903

YanaNirvana:
addresses issue #903


mobile/cmp/grid/impl/ColChooser.js
```diff
@@ -49,7 +54,7 @@ export class ColChooser extends Component {
                     style: {maxHeight: 16},
                     modifier: 'quiet',
                     omit: !gridModel.stateModel,
-                    onClick: () => model.restoreDefaults()
+                    onClick: model.restoreDefaults
```
lbwexler: yana and I discussed this -- she'll remove, and then we are good to merge

YanaNirvana:
Removed & Ready :D

cmp/ag-grid/AgGrid.js
```diff
@@ -34,7 +34,7 @@ import './AgGrid.scss';
 @LayoutSupport
 export class AgGrid extends Component {
     baseClassName = 'xh-ag-grid';
-    modelClass = AgGridModel;
+    static modelClass = AgGridModel;
```
amcclain: Great catch thank you.

lbwexler:
Great -Atm just mentioned he is already reviewing, so I will let you guys take care of that (and we'll work together to avoid too much doubling up on this stuff :-)

desktop/appcontainer/AppContainer.js
```diff
@@ -55,6 +56,10 @@ export class AppContainer extends Component {
 
     static modelClass = AppContainerModel;
 
+    static propTypes = {
+        model: PT.oneOfType([PT.instanceOf(AppContainerModel), PT.object]).isRequired
+    };
```
amcclain: Can remove this one - the app developer doesn't need to (and indeed can't) pass a model or object to this component.  Instead they provide the *class* within the config passed to `XH.renderApp()` and the toolkit ends up instantiating it with an already-created `AppContainerModel`.  

This is the first example here of where I think having the prop-type doc makes it seem as though you should expect to instantiate the component with props, which ends up being more misleading than helpful.

desktop/cmp/grid/impl/ColChooser.js
```diff
@@ -32,6 +33,10 @@ export class ColChooser extends Component {
 
     static modelClass = ColChooserModel;
 
+    static propTypes = {
+        model: PT.oneOfType([PT.instanceOf(ColChooserModel), PT.object]).isRequired
+    };
```
amcclain: Pls. remove, as above re. implementation classes.  In this case, `Grid` always renders this component and passes it a fully-formed model.

desktop/cmp/mask/Mask.js
```diff
@@ -29,6 +29,9 @@ export class Mask extends Component {
 
     static propTypes = {
 
+        /** A model to which a mask is bound to.  Masks for as long as model remains loading. */
+        model: PT.oneOfType([PT.instanceOf(PendingTaskModel), PT.object]),
```
amcclain: In this case, the Mask will always take an actual `PendingTaskModel` instance - not an object.  The usefulness here is that the model can be bound by other models/code to promises, so  it wouldn't do any good if the component created the model on its own and kept it entirely internal.

Also would fixup the comment a bit - how about `Model to which a mask is bound - mask will be visible whenever model.isPending.`

desktop/cmp/rest/impl/RestGridToolbar.js
```diff
@@ -13,12 +13,18 @@ import {toolbar, toolbarSep} from '@xh/hoist/desktop/cmp/toolbar';
...
 
 @HoistComponent
 export class RestGridToolbar extends Component {
 
     static modelClass = RestGridModel;
 
+    static propTypes = {
+        /** Primary component model instance. */
+        model: PT.oneOfType([PT.instanceOf(RestGridModel), PT.object]).isRequired
+    };
```
amcclain: Pls remove / impl component.

desktop/cmp/tab/TabSwitcher.js
```diff
@@ -27,6 +27,9 @@ export class TabSwitcher extends Component {
     static modelClass = TabContainerModel;
 
     static propTypes = {
+        /** Primary component model instance. */
+        model: PT.oneOfType([PT.instanceOf(TabContainerModel), PT.object]).isRequired,
```
amcclain: Should only include the model class, not an object.  This component is a helper comp that works alongside a TabContainerModel by interacting with its model - it would never instantiate such a model on its own.

mobile/appcontainer/AppContainer.js
```diff
@@ -52,6 +53,10 @@ export class AppContainer extends Component {
 
     static modelClass = AppContainerModel;
 
+    static propTypes = {
+        model: PT.oneOfType([PT.instanceOf(AppContainerModel), PT.object]).isRequired
```
amcclain: Pls remove as with desktop AppContainer comp.

mobile/cmp/button/ColChooserButton.js
```diff
@@ -23,7 +23,7 @@ export class ColChooserButton extends Component {
 
     static propTypes = {
         /** GridModel of the grid for which this button should show a chooser. */
-        model: PT.instanceOf(GridModel).isRequired
+        model: PT.oneOfType([PT.instanceOf(GridModel), PT.object]).isRequired
```
amcclain: Remove object form - comp would not create

mobile/cmp/dimensionchooser/DimensionChooser.js
```diff
@@ -34,7 +34,10 @@ export class DimensionChooser extends Component {
         buttonWidth: PT.number,
 
         /** Width in pixels of the popover menu itself. */
-        dialogWidth: PT.number
+        dialogWidth: PT.number,
+
+        /** Primary component model instance. */
+        model: PT.oneOfType([PT.instanceOf(DimensionChooserModel), PT.object]).isRequired
```
amcclain: Unlike with colChooser, the developer _will_ need to create this component and pass it a model, since it's not rendered directly by grid.  So good to have the prop-type in place, although this is another instance where we only want the class form - not the object.  Idea again here is that an owning/higher-level component would create the model, pass it to this comp, and observe its value to detect changes to dimensions made by the user interacting with this comp.

mobile/cmp/grid/impl/ColChooser.js
```diff
@@ -36,6 +37,10 @@ export class ColChooser extends Component {
 
     static modelClass = ColChooserModel;
 
+    static propTypes = {
+        model: PT.oneOfType([PT.instanceOf(ColChooserModel), PT.object]).isRequired
```
amcclain: Pls remove / impl component.

mobile/cmp/button/ColChooserButton.js
YanaNirvana: Yikes.  what was I thinking?

desktop/cmp/tab/TabSwitcher.js
```diff
@@ -27,6 +27,9 @@ export class TabSwitcher extends Component {
     static modelClass = TabContainerModel;
 
     static propTypes = {
+        /** Primary component model instance. */
+        model: PT.instanceOf(TabContainerModel),
```
amcclain: This should be `isRequired`

mobile/cmp/menu/Menu.js
```diff
@@ -28,7 +28,10 @@ export class Menu extends Component {
         width: PT.number,
 
         /** How to interpret the provided xPos when showing. */
-        align: PT.oneOf(['left', 'right'])
+        align: PT.oneOf(['left', 'right']),
+
+        /** Primary component model instance. */
+        model: PT.oneOfType([PT.instanceOf(MenuModel), PT.object])
```
amcclain: Also `isRequired` - only way to get a menu to be anything

Issue #1162, opened 2019-06-04
Grid - improvements to mobile column chooser


amcclain:
Making several improvements to mobile column chooser:

+ Split the list between displayed and available columns
+ Support drag-and-drop between the two lists, as well as shortcut buttons on each row to add/remove columns.
+ Support drag-and-drop re-ordering of the displayed columns to control actual order in grid.
+ Alpha-sort available columns for easier finding within their list and to make it clear that order of hidden columns not somehow meaningful.
+ Additional user option to pin first column can be turned on via code config. Needs to store pinned state, either in one-off flag for mobile or by resolving #687

Issue #687, opened 2018-10-04
Grid - state management for pinned columns


amcclain:
Columns can be pinned / locked to left/right via config.  Once there is a pinned column in place, the user can by default pin other columns by dragging it to the pinned side.  Even without a pre-existing pinned column, a user can pin a col by dragging to the left edge and waiting a bit.

However we are not tracking pinned state for columns, either at GridModel or GridStateModel level. I attempted to add basic support for this, but was tripped up as the ag-provided state we listen for did not seem to include updated pinning status - i.e. a just pinned or unpinned column did not show as such. Perhaps ag-Grid updates this state later in a further operation after fully resolving column order - I didn't have a chance to look.

Even if/when we have that working, we will might need to think carefully about column state and newly configured columns / column order / etc.

For the moment, I think it is safer to force the `lockedPin` ag column property, which tells the grid to disallow any user-driven pinning.  This does not lose current functionality - cols can't be pinned by the user anyway as gridModel sync immediately unpins them - but it does minimize grid default UI that promises to set pin status and then ends up being confusing/misleading when that fails to work.

shravnk:
See [this branch](https://github.com/exhi/hoist-react/tree/pinnedGridCols). Pinned status does not change onDragStop, but rather when user clicks and holds for one second on the far left/right of the grid. The onColumnPinned hook refers to this event. As far as I can tell it's an ag-grid bug that the pinned status is not stored in the columnState, where it could be accessed by other event handlers.

PR #1163, opened 2019-06-04
Improvements to mobile column chooser

TomTirapani:
See issue #1162 for details. Also adds cross platform state management for pinned columns (#687).

~~See also Toolbox PR: https://github.com/exhi/toolbox/pull/201~~


cmp/grid/columns/Column.js
```diff
@@ -194,7 +194,6 @@ export class Column {
                 resizable: this.resizable,
                 sortable: this.sortable,
                 suppressMovable: !this.movable,
-                lockPinned: true, // Block user-driven pinning/unpinning - https://github.com/exhi/hoist-react/issues/687
```
amcclain: We should probably still lock via this setting when mobile - in that case, we don't want the user-initiated drag-to-edge gesture to interfere with the more managed (and I think appropriate) pin-first-col setting available via the chooser.

(I also tried dragging cols to pin on Chrome w/mobile emulation and it froze up - might just be an emulator thing but was enough weirdness to convince me we don't need this feature on mobile at the moment.)

amcclain:
@haynesjm42 sorry we got cut off earlier - can you look at this tomorrow?  Let us know via IM if you have time (or not)

haynesjm42:
Looking at this now

cmp/grid/Grid.js
```diff
@@ -556,6 +563,13 @@ export class Grid extends Component {
         this.model.agGridModel.agApi.sizeColumnsToFit();
     };
 
+    // Catches column pinning by dragging
+    onColumnPinned = (ev) => {
```
haynesjm42: I think this should be handled a little differently, to be more consistent with how we handle column visibility/position/width changes. For other column changes via drag, we intentionally do not update our GridModel column state until the `onDragStopped` event is fired, which is the best way to determine when the state change has been "committed" when dragging. 

Then in the event handlers for those events we update the GridModel column state only if the event was not triggered by dragging (the logic for that varies, but I think `onColumnVisible` is a good parallel to how the `onColumnPinned` handler should work).

Essentially I think synchronizing the ag-Grid pinned state to our column state should work exactly the same as the column visibility, and go through the same `noteAgGridColumnStateChange` method.

mobile/cmp/button/ColChooserButton.js
```diff
@@ -23,13 +23,16 @@ export class ColChooserButton extends Component {
 
     static propTypes = {
         /** GridModel of the grid for which this button should show a chooser. */
-        model: PT.instanceOf(GridModel).isRequired
+        model: PT.instanceOf(GridModel).isRequired,
+
+        /** Allow pinning the first row **/
+        enablePinFirstRow: PT.bool
```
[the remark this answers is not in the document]
amcclain: The issue in my mind is that we came up with the idea of supporting this as a special flag for mobile only, where it seemed to make sense to have a simpler model for pinning that could be controlled / viewed from the updated column chooser.

I find it difficult to evaluate and balance out competing demands of flexibility / simplicity / consistency between the mobile and desktop toolkits for features like this.

Maybe it would help to enumerate a few things to structure my thoughts...

### What pinned states do we envision being of use?

1. No columns pinned.
2. One column pinned to the left.
    - Most common case I think by a good margin.
3. Multiple columns pinned to the left.
    - Can imagine relatively common case where two columns are pinned - e.g. some kind of minimal status / category column + a primary label / identifier column.
4. Column(s) pinned to the left and right.
    - Do not believe right pinned columns would be common. I can imagine uses for them, but in the real world you do not usually see this.

### How could users pin a column / columns?

1. Use the drag gesture built-in to ag-grid, dragging a column in the grid to either the left or right until the little "pin" icon shows up in the draggable.
    - 👍 Built-in to ag-grid.  Allows quick pinning of any number of columns to either left or right.
    - 👎 Potentially problematic on mobile where dragging can be tricky or less reliable, especially on the problematic browsers we sometimes encounter.
    - 👎 Not easily discoverable for users who don't know about the feature.
    - 👎 Pinned state wouldn't be represented in column chooser without also doing some work there.
2. Toggle a simplified option in the column chooser to pin the first column, covering the most common pinning case (2) above.
    - 👍 Can be presented as a simple, easily understood option to users within the column chooser.
    - 👎 Not compatible with multiple pinned columns, or right pinned columns.
3. Toggle a control on each displayed column in the column chooser turning pinning on/off for that column. 
    - 👍 Could represent pinned state of multiple columns on both left and right.
    - 👍 Could be used alongside dragging option (1).
    - 👎 Most complex to implement. To be clear and usable, would need to account for column order - i.e. what should happen if you could request a column in the middle of the list be pinned? What should happen when you re-order columns?

My motivation for suggesting (2) on mobile as an incremental first step was to minimize complexity and handle the case I think we really need now on mobile, without relying on the drag-to-pin gesture. But maybe it's an uncanny valley that just sets us up for more work in the near future and we should push forwards to (3).

cmp/grid/GridModel.js
```diff
@@ -76,6 +76,8 @@ export class GridModel {
     /** @member {GridGroupSortFn} */
     groupSortFn;
     /** @member {boolean} */
+    enableColumnPinning;
+    /** @member {boolean} */
```
amcclain: Thought about this a bit more, and would like to suggest `enableDragToPin` instead (apologies as I realize I suggested this config name first).

While a bit wordier, I think it's more clear that this config limits the dragging operation only. Columns can still be configured in code to be pinned and, even more so, we're expecting to offer this new mobile chooser option to allow pinning the first column. 

I think that `enableColumnPinning` ends up implying that no pinning will be possible if false, which isn't what we're looking to enforce.

TomTirapani:
Thanks again for taking a detailed look, and for the discussion - really helped simplify the state management side of things!

I've updated this branch with the changes requested and discussed. Please note there is now also a corresponding Toolbox branch: https://github.com/exhi/toolbox/pull/204

TomTirapani:
I changed unpinned to null because this is what is returned from AgGrid's getColumnState(), and thought best to be consistent. The typedef already has the pinned property :)

haynesjm42:
Doh! Was looking at the wrong branch ..

mobile/cmp/grid/impl/ColChooserModel.js
```diff
@@ -60,6 +70,35 @@ export class ColChooserModel {
...
+                shouldPinFirst = false;
+            } else {
+                it.pinned = null;
+            }
+        });
+
+        this.setColumns(columns);
+    }
+
+    onHideBtnClick(colId, hide) {
```
lbwexler: I think that 'event handlers' like this on the model  have a bit of a code smell.  The Model should be independent of the actual affordances in the component, as much as possible.

Is there a method name that makes more sense for this here?

lbwexler:
Realize I am a little late to this party.  First thing I want to say is that it looks awesome.  And amazing that you are able to the DnD with such a reasonable amount of app code!

I made one nitty comment in the code about an event handler on the model, but my more important question is that you have added an optional property to ColChooserModel constructor:

 * @param {boolean} [c.enablePinFirstCol] - show toggle for pinning first column.

When would this get set to false?  This is a purely implementation class, and yet I could not find where it gets set to false in the code.  In general, I think we want to avoid these optional properties in implementation classes -- they are best suited for convenience in application classes....

mobile/cmp/grid/impl/ColChooserModel.js
TomTirapani: Good catch Lee.

I moved the implementation of this handler to the component itself, where it leverages the `moveToIndex()`, `setHidden()` and `updatePinnedColumn()` methods on the model. This provides a nice symmetry with the `onDragEnd` event handler.

TomTirapani:
An important part of this change is that `GridModel.enableColChooser` has become `GridModel.colChooser`. In addition to accepting the boolean as before, you can now pass a config for the `ColChooserModel`. This is how we would set `enablePinFirstCol` to false on the model. The motivation here is the use case of not wanting to allow column pinning at all on a mobile grid (i.e. there's no obvious candidate for a pinned column, or we deem the columns are too wide to support column pinning.

I've updated the CHANGELOG to draw attention to this change. Note that the `enableColChooser` prop is still accepted (albeit with a deprecated warning) to avoid this being a breaking change.

lbwexler:
Ah -- that is sort of what I suspected.  

My problem with this is that the ColChooser classes are now in the '/impl' folder, and marked as 'private'.  If we really want them to be part of the API that applications can examine and use, we need to promote them up a directory, remove the @private marking, and document the classes to indicate how they should be used (e.g. pass the `config` for the model constructor to your gridModel).

A couple of additional thoughts we could discuss?  Happy to forgo all of this, if we have thought about it, or decided it is too much work at this state in the game.

+ I see that this as similar  like `selModel` and `stateModel`, but not exactly.  So understand why we are not calling it `colChooserModel` (we want it to work well as a replacement for `enableColChooser`, for one).  I also suppose if we ever decided to let people to customize the chooser itself, we could allow them to specify `colChooser` as a function that is called with the gridModel itself, when Col Choosing is activated.

+ Seems to me that the enablePinFirstColumn (or some variant of it) is a property of the GridModel itself, not the chooser.

+ Is there any way we could come up with a common colChooserModel for mobile and desktop, w.r.t. to the API?     If necessary, implementation details could be pushed down in to local models within (Maybe more than we want to undertake now -- just want to make sure we considered)

+ Need to

lbwexler:
( apologies for my somewhat incoherent comment above -- it was a crazy morning at home :-))

HOWEVER -- Tom and I had a very productive discussion, and came up with the following proposal:

+ Add a new semantic property `enablePinnedColumns`, that governs whether the GridModel is intended to support pinning.  It's then up to implementations to consult and enforce this as best they can when presented with programmatic changes, configuring built-in ag affordances,  or via custom XH affordances (e.g. in column choosers).

+ leave `enableColumnChooser` as is, and leave our platform Column Choosing classes private and out of the API.  Punting on changes here will avoid API churn and leave us the most flexibility down the road to make potentially bigger changes to accommodate more general grid editing (e.g. view editing).  And we have no client demand driving us to expose a ColumnChooser API now....

Thoughts?

amcclain:
So glad that this got merged - thank you much.

I agree with the tail of the discussion above and had several of the same thoughts over the past few days.  Apologies on my part for any misdirection but I think we landed in a good place.  I was thinking that having control over whether or not pinning was enabled generally was likely important and I am glad we got there.

PR #1165, opened 2019-06-05
Implement Ctrl-A for grid (Fixes #1149)

### LINKED ISSUE #1149: Support Ctrl-A select all rows in grid
ag-Grid supports keyboard shortcuts like Ctrl-C Copy but not Ctrl-A Select-all.

It can be implemented in an application by adding a grid keyDown event handler:
```
keyDown(evt) {
  if (evt.ctrlKey && evt.key == 'a') {
    gridModel.agAPi.selectAllFiltered();
  }
}
```
See related selModel issue #1150.

lbwexler:


cmp/grid/Grid.js
```diff
@@ -597,6 +596,14 @@ export class Grid extends Component {
         return column.isTreeColumn ? node.data[column.field] : value;
     }
 
+    onKeyDown = (evt) => {
+        if (evt.ctrlKey && evt.key == 'a') {
```
haynesjm42: Should we add `this.model.selModel.mode === 'multiple'` to this conditional? Don't think we want Ctrl + a to do anything for a single-mode selection model

data/StoreSelectionModel.js
```diff
@@ -84,6 +84,13 @@ export class StoreSelectionModel {
         this.ids = clearSelection ? ids : union(this.ids, ids);
     }
 
+    /** Select all filtered records. */
+    @action
+    selectAll() {
+        this.select(this.store.records);
```
haynesjm42: Should we check the mode here? Looks like the behavior will be that this would effectively select the first record, which is kind of confusing.  Am torn between whether this should just be a no-op (and a warning logged?) for a 'single' mode sel model, or if we should throw. Seems like a programming error if `selectAll` is called for a non-multiple-mode sel model but throwing might be a little harsh in this case..

lbwexler:
Good Catches John -- fixed.  

Note that I added the additional check for the metaKey for mac users....

Issue #1139, opened 2019-05-24
Add XH.prompt()


TomTirapani:
Just an idea, filing here so I don't forget...

Add an XH.prompt() method, that shows a configurable Message that contains a TextInput. On clicking confirm, the promise resolves with the typed in value.

Motivation here is that:
- Often you don't want to go to whole trouble of setting up a dialog / form to collect some information during a user action.
- window.prompt() is the only native dialog method that doesn't have an XH analogue.
- We had something similar in the Sencha toolkit.

TomTirapani:
See PRS:
https://github.com/exhi/hoist-react/pull/1170
https://github.com/exhi/toolbox/pull/203

lbwexler:
👍

PR #1170, opened 2019-06-06
Add support for XH.prompt()

TomTirapani:
See issue #1139
See Toolbox PR https://github.com/exhi/toolbox/pull/203


core/XH.js
```diff
@@ -329,6 +329,17 @@ class XHClass {
         return this.acm.messageSourceModel.confirm(config);
     }
 
+    /**
+     * Show a modal 'prompt' dialog with a default TextInput, message and default 'OK'/'Cancel' buttons.
+     * Can provide a custom input control with the input prop (true to use default TextInput).
```
lbwexler: The second sentence is confusing here -- need to reserve term 'prop' for react component prop, and try to use the term 'param' or 'option'.  Also, perhaps just 'Applications may also provide a custom HoistInput'.

core/XH.js
```diff
@@ -329,6 +329,17 @@ class XHClass {
...
+     *
+     * @param {Object} config - see XH.message() for available options.
```
lbwexler: Need to actually enhance those docs for XH.message.  That's the main entry point for users/doc readers....

lbwexler:
It seems like if we are promising a custom input, users will also want to set a custom value, and may even want to set their own validation.  (the latter we could say is out of scope, but the former seems critical).

This is very similar to the problem we had with the API for the options dialog.
 
Would we want our 'input' spec specified by the user to potentially also contain a `value`.  

If so we could behind the scenes push the value to our internal form model, and bind a new clone of the input  to the model

TomTirapani:
Hi Lee. I've updated this branch to include support for setting custom `item` (i.e. HoistInput), `value` and `rules` for the prompt input.

core/XH.js
```diff
@@ -305,13 +305,16 @@ class XHClass {
      * @param {string} config.message - message text to be displayed.
      * @param {string} [config.title] - title of message box.
      * @param {Element} [config.icon] - icon to be displayed.
+     * @param {MessageInput|boolean} [config.input] - config for input to be displayed, or true to use default TextInput.
```
lbwexler: wouldn't it be more straightforward to have .message() just require a MessageInput here?   Leave the defaulting to .prompt.  Message() is the full, explicit form

core/appcontainer/MessageModel.js
```diff
@@ -44,6 +56,13 @@ export class MessageModel {
         this.onCancel = config.onCancel;
         this.result = new Promise(resolve => this._resolver = resolve);
 
+        // Extract properties from input
+        if (isPlainObject(this.input)) {
+            const {value, rules} = this.input;
```
lbwexler: Seems like it would be more straighforward and cleaner to just create the form model here in this conditional, no?  Most messages won't need it.  Then we, also don't have to change anything with the FormModel API.

core/appcontainer/MessageModel.js
```diff
@@ -69,6 +92,15 @@ export class MessageModel {
     //-----------------------
     // Implementation
     //-----------------------
+    async doSubmitAsync() {
```
lbwexler: could combine this method and previous method into a single async method.

core/appcontainer/MessageSourceModel.js
```diff
@@ -39,6 +39,11 @@ export class MessageSourceModel {
         return this.message(config);
     }
 
+    prompt(config) {
+        config = defaults({}, config, {confirmText: 'OK', cancelText: 'Cancel', input: true});
+        return this.message(config);
```
lbwexler: here is where we can install the default textInput

desktop/appcontainer/Message.js
```diff
@@ -34,16 +39,33 @@ export class Message extends Component {
...
+                    model.message,
+                    this.getInput()
+                ),
                 toolbar(this.getButtons())
             ],
             ...this.props
         });
     }
 
+    getInput() {
```
lbwexler: should be called renderInput()

desktop/appcontainer/Message.js
```diff
@@ -34,16 +39,33 @@ export class Message extends Component {
...
+            model: formModel,
+            fieldDefaults: {commitOnChange: true, minimal: true, label: null},
+            item: formField({
+                field: 'value',
+                item: withDefault(input.item, textInput({autoFocus: true}))
+            })
+        });
+    }
+
     getButtons() {
```
lbwexler: rename renderButtons

desktop/appcontainer/Message.js
```diff
@@ -55,12 +77,13 @@ export class Message extends Component {
...
                 intent: confirmIntent,
+                disabled: input ? !formModel.isValid : false,
                 onClick: this.onConfirm
             })
         ];
     }
 
-    onConfirm = () =>   {this.model.doConfirm()}
-    onCancel = () =>    {this.model.doCancel()}
+    onConfirm = () =>   {this.model.doConfirm()};
```
lbwexler: these both can be inline

core/appcontainer/MessageSourceModel.js
TomTirapani: Unfortunately, since MessageSourceModel is cross-platform and the TextInputs are platform specific, the only way to do the defaulting here would be to use dynamics.

Instead, I think its simpler to default in the platform-specific Message components. However, I changed it so that all the aspects of the input spec are optional. Passing an empty input spec tells the message to render a form with all the defaults (i.e. a TextInput, no default value and no validation rules).

TomTirapani:
Thanks for taking a detailed look Lee. I've updated this PR with your suggestions, which the exception of where we default the TextInput.

core/appcontainer/MessageModel.js
```diff
@@ -78,3 +102,10 @@ export class MessageModel {
         this.close();
     }
 }
+
+/**
+ * @typedef {Object} MessageInput
+ * @property {HoistInput} [item] - HoistInput to use.
```
lbwexler: I think this is not a HoistInput, but a ReactElement representing a HoistInput

PR #1176, opened 2019-06-07
DimChooser Enhancements(Fixes #1169)

### LINKED ISSUE #1169: DimChooser should support empty value / ungrouped state

lbwexler:
Tried to keep this as simple as possible in order to make it a safe and relatively easily merged enhancement

I think our DimensionChooser is definitely complicated and needs some rework when time allows.

One thing we should definitely do is enable a little clear icon when `enableClear` is true and it is in `styleButtonAsInput` mode.  Couldn't figure out how to style that button safely, but perhaps others that are better at CSS could take a look.  With the button in place (and maybe other changes) we could also remove the `emptyText` label from ever appearing in the history.


lbwexler:
Also fixes #1168 now.

PR #1177, opened 2019-06-07
ag-Grid -> 21.0.0 (Fixes #1164)

### LINKED ISSUE #1164: Update ag-Grid to 21.0.0
Just released - we could test immediately, or see if they do any follow-up patch releases.

https://www.ag-grid.com/ag-grid-changelog/?fixVersion=21.0.0

Some compelling changes:
* Scroll performance "vastly improved"
* More control over react prop change detection to avoid re-renders - https://www.ag-grid.com/react-more-details/#react-row-data-control - as well as a note "declarative column definitions now reactive"
* Icon customization - maybe we can get our standard glyphs in

Note I don't see any tickets re. performance in tree mode - this was a blocker for us with the 20.2.0 upgrade - we should be sure to test to see if performance is OK or if we are still stuck.

haynesjm42:
Looks like the performance issues with delta updates that would remove large amounts of rows has been greatly improved, but is still not as fast as clearing the data first, so I've left our code in which does that.

Performance comparison:
```
Replacing 150000 records

ag-Grid v21.0.0
  6,099ms - no pre-clearing
  4,814ms - with pre-clearing

ag-Grid v20.0.0
  66,917ms - no pre-clearing
  4,642ms - with pre-clearing
```

I didn't see any performance differences with tree grids, in general performance seems comparable to  v20. I think this is fine to merge and try out in some client apps to see if we notice any issues.

Note I didn't explore the new icon customization support in this pr, would be happy to look into that more if we want to do it now.


amcclain:
Tested this out and glad to hear that the performance is looking reasonable.  Agree I don't see any reason to unwind our customization at this point.

However it looks as though keyboard navigation has broken - IIRC we layer that on with some custom handling. There's a breaking change AG-2830 listed in their changelog related to cell navigation - so that seems very likely it.  Can you please take a look?

Would love to try the icon customization to use our standard FA icons while we are at it. Might as well take advantage.  Great if we could get the context menu icons more standard in particular as we have well-defined presets for e.g. copy,

cmp/grid/columns/Column.js
```diff
@@ -222,7 +222,7 @@ export class Column {
         }
 
         if (this.tooltip) {
-            ret.tooltip = isFunction(this.tooltip) ?
+            ret.tooltipValueGetter = isFunction(this.tooltip) ?
```
amcclain: Thanks for this - I was going to suggest it, as I knew we had made then unwound this change.  But then I went to look at their docs and couldn't find the reference again to this change, and tooltips still seemed to be working without....  Were you able to find a pointer on their site to this change?  

(I believe it's the right thing to do based on those recollections - just want to be sure.)

haynesjm42: Didn't see it in the changelog, but noticed some deprecation warnings in the console, and it is marked as deprecated as of v20.1 in the ag-grid source code. A quick search through the ag-grid docs doesn't actually turn up any references to `tooltipValueGetter` at all ... but pretty sure this is correct

PR #1178, opened 2019-06-07
Support for includeChildren and includeParents in Store filtering (fixes #1168)

### LINKED ISSUE #1168: DimChooser throws badly with unknown dim in initialValue

lbwexler:
Some things to think about here:

+ Adding the meta-data to the function itself is a new pattern for us,  but potentially worth considering here?  Could also handle out of band in new flags.

+ 'includeParents' is a bit wonky.  Could be 'includeAncestors' or 'includeParent'?
Trying to get the recursive notion in here...I suppose for symmetry with includeChildren should be just 'includeParent'.

+ Behavior not responsive to changes in props after construction time.  Could change this as well.


data/impl/RecordSet.js
```diff
@@ -63,20 +63,45 @@ export class RecordSet {
...
-        records.forEach(rec => {
-            if (filter(rec)) markPass(rec);
+            isMarked = (rec) => passes.has(rec.id),
+            mark = (rec) => passes.set(rec.id, rec);
+
+        // Pass 1.  Mark all passing records, and potentially their children recursively.
+        // Any row already marked will already have all of its children marked, so check can be skipped
+        const includeChildren = withDefault(filter.includeChildren, false);
+        let markChildren;
+        if (includeChildren) {
```
haynesjm42: Don't think we need to create this func conditionally since we only call it when `includeChildren` is true? Somewhat wasteful to create the func if we don't need it but the code might be slightly cleaner.

data/Store.js
```diff
@@ -217,8 +217,13 @@ export class Store {
         return this._all.rootList;
     }
 
-    /** Filter function to be applied. */
     get filter() {return this._filter}
+    /**
+     * Filter function to be applied.
+     * @param {Function} fn - function taking a Record, and returning a boolean.
+     * @param {boolean} [fn.includeParents] - true to include all ancestors of a passing record (default true)
```
haynesjm42: I think this is neat and encapsulates the props nicely in the function, but seems like this could result in some clunky code in applications trying to use this. I often pass an arrow function or just set the filter to a class member, so if I understand this correctly I would need to do something like this:
```
this.myFilterFn.includeChildren = true;
myStore.setFilter(this.myFilterFn);
```
or 
```
const filterFn = (rec) => rec.passesMyCondition;
filterFn.includeChildren = true;
myStore.setFilter(filterFn);
```
as opposed to:
```
myStore.setFilter(this.filterFn, {includeChildren: true});
```
or
```
myStore.setFilterFn((rec) => rec.passesMyCondition, {includeChildren: true})
```

Is there some magical syntax sugar that would make this cleaner including the props on the function object?

data/impl/RecordSet.js
```diff
@@ -63,20 +63,45 @@ export class RecordSet {
...
-        };
-
-        records.forEach(rec => {
-            if (filter(rec)) markPass(rec);
+            isMarked = (rec) => passes.has(rec.id),
+            mark = (rec) => passes.set(rec.id, rec);
+
+        // Pass 1.  Mark all passing records, and potentially their children recursively.
+        // Any row already marked will already have all of its children marked, so check can be skipped
+        const includeChildren = withDefault(filter.includeChildren, false);
```
haynesjm42: Is the `withDefault` here necessary? `filter.includeChildren` would be falsey when omitted anyways, right?

data/Store.js
lbwexler: John -- thanks for this comment.  It is SO important to actually write the application code that your new API is intended to support.  I totally failed to do that.  I agree the property on the filter is way too clumsy looking.  I'll change to the option you mention

data/impl/RecordSet.js
lbwexler: I prefer withDefault() to be more explicit, and symmetrical with `true` defaults...

data/impl/RecordSet.js
lbwexler: I had the cleaner code, and mucked it up for just the reason you mentioned!  

Thinking it through, it now seems silly (and wrong) to me, given that the cost to setup the filter is going to be small relative to the cost to execute it.

Good catch

lbwexler:
John -- to the big point about 'what is the use case for 'includeParents' --
I thought it might be useful to pluck out records you really care about in certain circumstances -- and it had been working for me fine during testing. 

HOWEVER, its a perfect example of feature creep, and without a clear implementation, and potential problems, we have no reason to waste any more time on it now.  We can revisit if clients ever ask for it.

lbwexler:
Thanks John -- think the latest cut takes care of all issues.

+ Also added new toolbox branch with same name for testing
+ Also realized why i did that conditional bit around includeChildren -- actually trying to avoid generating the childrenMap -- which might be expensive -- if I don't actually need it.

data/Store.js
```diff
@@ -50,6 +51,7 @@ export class Store {
      *      presented to loadData() prior to creating a record from that object.  This function should
      *      return a data object, taking care to clone the original object if edits are necessary.
      * @param {function} [c.filter] - filter function to be run.
+     * @param {object} [c.filterOptions] - options associated with this filter operation. See setFilter() for more details.
```
haynesjm42: Should we add a `@typedef` doc for these filter options since we are expecting the same object here and in setFilter?

Issue #1166, opened 2019-06-05
Turning off tail mode in Admin log viewer reverses order of all lines


amcclain:
Tail mode (default) is intended to load the most recent n lines of the log file and auto-refresh to keep updating and tracking the latest changes to the log file.

Turning off tail mode should load the log starting on the startLine and loading maxLines worth of the log file.  The first line should still be at the top - not sure why it's now flipped around but I don't think it's usable or intentional...

lbwexler:
👍

PR #1181, opened 2019-06-11
tailMode bug fix

YanaNirvana:
Fixes issue #1166: replace `this.startLine` with `parent.startLine` +

* It seems that the server returns an array that depends param `parent.startLine`, but then the client conditionally `.reverse()` the response.  Maybe we can consider removing the work from the client and instead having the server return the array in appropriate order to begin with?   (Yana is unsure whether there are other pieces that depend on that endpoint returning the way it currently does)

### DIFF
--- a/admin/tabs/logging/viewer/LogDisplayModel.js
+++ b/admin/tabs/logging/viewer/LogDisplayModel.js
@@ -72,7 +72,7 @@ export class LogDisplayModel {
             })
             .then(response => {
                 if (!response.success) throw new Error(response.exception);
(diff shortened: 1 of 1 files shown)


lbwexler:
Hi Yana -- it looks like you found the bug in the javascript.  We must have just screwed that up at some point during a refactor.  Very easy to break falsy checks like that in javascript!!

We could consider redoing the client/server contract, but I do recall that it was intentional.  In any case, best to merge this clear bug fix right now!

Issue #1167, opened 2019-06-05
Admin log viewer - add page forward/back buttons


amcclain:
When _not_ in tail mode, the admin log viewer loads the requested log from startLine up to startLine + maxLines.

Would be handy to have a little set of forward/back buttons that could let you skip forward a "page" when in this mode.  

I think the logic would simply be:

Going forward, `startLine -> startLine+maxLines`
Going backward, `startLine -> max(startLine-maxLines, 0)`

YanaNirvana:
A couple thoughts:

1.  If a user wants to click through a "page" they would have to make two clicks: first to toggle out of tail mode (enabling the forward, backward buttons), and then on either forward or backward button.  I'm thinking it might make more sense to enable the buttons, but clicking on either button will take user out of tail mode.  What do you think?

2.  Just want to confirm the behavior of this button.  If a user is looking at the below
      startLine: 1, maxLines: 1000, 
      clicking "forward" button will land them at startLine: 1,001 ?  
      In this situation, when a the total log from server is only 300 lines long, should the button still be enabled, allowing user to jump ahead to 1,001?

amcclain:
Hi - 

1) Agreed re. the two clicks, but I think that's OK.  I was using the log viewer to troubleshoot something, and I had already realized I needed to take it out of tail mode and focus in on a particular line.  Then from there I wanted to walk forwards looking for the next occurrence of an error. I think having the buttons light up when you take it out of tail mode will help keep the interaction between that mode and the line inputs clear.

2) Yes, confirmed that's what I was thinking.

lbwexler:
Hi -- Discussed with @amcclain and @YanaNirvana -- we are going to close this. 
After some great investigation, its pretty clear that there is no low-hanging fruit here and we have to prioritize our time.

PR #1184, opened 2019-06-11
Forward and Backward Buttons in LogView

YanaNirvana:
Addresses issue #1167


amcclain:
I'm gonna close this for now to take it off the board so we can think on it a bit more and either clean up the branches next week or come up with a plan to add those arrows.  Thanks for setting this up and discussing yesterday. 👍

PR #1185, opened 2019-06-11
On mobile, force overflow hidden on the floating top and bottom containers (Fixes #1179)

### LINKED ISSUE #1179: Unwanted vertical scrollbar in mobile grid docked summary row
Noticed this on mobile with a client project when I turned on the new docked summary row.

Screen Shot 2019-06-10 at 11 24 29 AM

To replicate in Toolbox you need to adjust the [primary name column from flex:1](https://github.com/exhi/toolbox/blob/ffdb3af0a52c84716c35226420b1ae3ea1dda418/client-app/src/mobile/treegrids/TreeGridPageModel.js#L45) to some fixed width (I used 200) and then expand a node in the tree to trigger vertical scrolling.

haynesjm42:
ag-grid explicitly sets the overflow to either `scroll` or `hidden` on the floating top and bottom containers based on whether the center viewport has overflow in order to keep the space taken by the scrollbar consistent in all containers. 

https://github.com/ag-grid/ag-grid/blob/fe50182e5542b4616d2ef41a03a38836a83daeae/packages/ag-grid-community/src/ts/gridPanel/gridPanel.ts#L884

I think this change is safe since we don't expect to see any scrollbars in the grid on mobile, so ag-grid forcing them on actually makes the containers unbalanced.

Note that this issue does not appear be happen on an actual mobile device.

### DIFF
--- a/cmp/ag-grid/AgGrid.scss
+++ b/cmp/ag-grid/AgGrid.scss
@@ -5,6 +5,15 @@
   height: auto !important;
 }
 
+.xh-mobile {
+  // ag-grid adds these when the center viewport has overflow, so that the space taken by the scrollbar is consistent in
+  // the top and bottom floating divs. On mobile we do not render scrollbars for the center viewport, so force hiding
+  // in the top and bottom floating divs.
+  .ag-floating-top, .ag-floating-bottom {
+    overflow-y: hidden !important;
+  }
+}
+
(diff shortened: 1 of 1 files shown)


amcclain:
Hi @TomTirapani - as our mobile guru and another resident grid expert could you please review this small change when you have a chance and merge if all looks good to you?  

Thanks much guys

Issue #1194, opened 2019-06-18
Issues with getting children for Record with falsey id


haynesjm42:
If a Record has a falsey id (such as 0) then the children getter will return an empty array because of type coercion when computing the children map

lbwexler:
yikes.  Lets ASAP!

lbwexler:
Great catch!

PR #1195, opened 2019-06-18
Handle updating hierarchical data in Store

### LINKED ISSUE #1193: Store should handle updating hierarchical data
The `updateData` method on Store is currently limited in that you cannot update records in the middle of the hierarchy without also including the data for all of its ancestors. You also cannot add new child records to an existing record conveniently.

Proposed changes:

- `updateData` should make sure that the parentId and tree path are copied over from existing records when updating
- `updateData` should always update children, meaning if children are not included in the updated rawData then they will be removed
- `addData` method to be added to Store and RecordSet to allow adding data as children of an existing record (or as root records if no parent record is provided)

haynesjm42:
Fixes #1193 and #1194 (found while testing these changes)

Modified logic in RecordSet.updateData to properly handle hierarchical data, per proposals in #1193, and added `addData` method to Store and RecordSet

### DIFF
--- a/data/impl/RecordSet.js
+++ b/data/impl/RecordSet.js
@@ -134,8 +134,15 @@ export class RecordSet {
                 newRecord.parent = currRecord.parent;
             }
 
-            const hasChildren = existingRecords.values().find(rec => rec.parentId === currRecord.id);
-            if (!hasChildren) {
+            let hasChildren = false;
+            for (let rec of existingRecords.values()) {
+                if (rec.parentId == currRecord.id) {
+                    hasChildren = true;
+                    break;
+                }
+            }
+
+            if (hasChildren) {
                 const descendantIds = new Set();
                 this.gatherDescendants(currRecord.id, descendantIds);
                 descendantIds.forEach(id => {
--- a/data/impl/RecordSet.js
+++ b/data/impl/RecordSet.js
@@ -126,31 +126,36 @@ export class RecordSet {
             existingRecords = new Map(this.records);
 
         newRecords.forEach((newRecord, id) => {
-            const currRecord = existingRecords.get(id);
+            const existingRecord = existingRecords.get(id);
+            if (existingRecord) {
+                // Since the data we are updating may be from the middle of the hierarchy, we need to
+                // make sure that we copy the parent over for any existing records because we only
+                // include child information in raw data
+                newRecord.parent = existingRecord.parent;
 
-            if (currRecord) {
-                throwIf(newRecord.parent && newRecord.parent.id !== newRecord.parent.id,
-                    'Cannot change parents via updateData!');
-                newRecord.parent = currRecord.parent;
-            }
+                // When updating an existing record we also always update it's children, so we need
+                // to determine if the existing record has children. Avoiding calling Record.children
+                // here so we don't build the children map unnecessarily.
+                let hasChildren = false;
+                for (let rec of existingRecords.values()) {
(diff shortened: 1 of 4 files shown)


data/impl/RecordSet.js
```diff
@@ -127,6 +127,29 @@ export class RecordSet {
...
             const currRecord = existingRecords.get(id);
+
+            if (currRecord) {
+                throwIf(newRecord.parent && newRecord.parent.id !== newRecord.parent.id,
+                    'Cannot change parents via updateData!');
+                newRecord.parent = currRecord.parent;
+            }
+
+            let hasChildren = false;
+            for (let rec of existingRecords.values()) {
```
haynesjm42: Is there a better way to do this with a Map? `forEach` might be cleaner but also cannot early-out..

data/impl/RecordSet.js
```diff
@@ -168,11 +204,11 @@ export class RecordSet {
     computeChildrenMap(records) {
         const ret = new Map();
         records.forEach(r => {
-            const {parentId} = r;
-            if (parentId) {
-                const children = ret.get(parentId);
+            const {parent} = r;
```
haynesjm42: This is the fix for #1194, where the `parentId` was being coerced to a boolean

data/impl/RecordSet.js
```diff
@@ -135,19 +158,32 @@ export class RecordSet {
...
+        });
+
+        return new RecordSet(this.store, new Map([...records, ...newRecords]));
+    }
+
     removeRecords(ids) {
         const removes = new Set();
         ids.forEach(id => this.gatherDescendants(id, removes));
-        return this.applyFilter(r => !removes.has(r.id));
+        return this.applyFilter({fn: r => !removes.has(r.id)});
```
haynesjm42: This fixes a bug introduced with the (unreleased) StoreFilter changes, remove records was not working because it was not conforming to the new signature of `applyFilter`

data/impl/RecordSet.js
```diff
@@ -127,6 +127,29 @@ export class RecordSet {
 
         newRecords.forEach((newRecord, id) => {
             const currRecord = existingRecords.get(id);
+
+            if (currRecord) {
+                throwIf(newRecord.parent && newRecord.parent.id !== newRecord.parent.id,
```
amcclain: Compare to possible `currRecord` here

data/impl/RecordSet.js
amcclain: Is the goal here to detect if there are pre-existing record(s) that think the new record being evaluated is their parent, and remove them if they aren't in the new set? If so, can we look at `currRecord.children`? 

Some tricky logic here, although I'll be honest i am in transit and it's not an ideal place for focusing.  But maybe that compromised state is a useful way to detect where an extra comment or two would help. 😄 In particular the interplay between existing records vs. new ones is difficult for me to keep straight - what side we're looking to clean up and why/when.

haynesjm42: I need to dig back into this but I believe I didn't want to call `currRecord.children` because that would create the children map (if not already created), and didn't want to do that if there was a way to detect this without it, but need to find the balance between performance and readability here.

Will add some comments explaining why we have/need this logic

lbwexler:
John -- can we go over this together first thing tomorrow?  Would like to get it in to Hoist 24 but feel that we need to go over it in person.  (I am also in a compromised state on a train here, but also generally benefitting from getting talked through these!)

PR #1196, opened 2019-06-20
Add Hoist-provided basic AutoRefreshService

### LINKED ISSUE #977: Add AutoRefreshService to  Hoist
Could be linked to a set of xh configs/preferences, and manages the timer to drive the app auto-refresh.  Really analogous to idleService.  Might need a flag on the AppSpec as well.

Note that a more sophisticated mechanism might make use of the fact that the RootRefreshContextModel knows when it was last refreshed manually as well.   Can potentially use a fast timer that examines that date before trying the auto-refresh.  Might avoid extra auto-refreshes when user is actively refreshing manually.

amcclain:
+ Built-in service to trigger app-wide auto-refresh on a configurable interval, subject to soft-config and user preference.
+ Add `AppSpec.clientAppCode` - used by new service to configure per-JS-app intervals, generally provides symmetry with `clientAppName`.
+ Doc comment updates on refresh-related classes made during review.
+ Fixed #977


amcclain:
See matching `autoRefresh` branches in hoist-core and toolbox:

https://github.com/exhi/hoist-core/pull/76
https://github.com/exhi/toolbox/pull/212

PR #1197, opened 2019-06-20
Target sessionExpired check to relative URLs only

### LINKED ISSUE #1189: ExceptionHandler over-eager to show "session expired" message
Our `ExceptionHandler` has some dedicated handling to show a "your session expired please login" message and prompt the user to reload the app anytime it gets an exception back with HTTP 401:

https://github.com/exhi/hoist-react/blob/5783ae1b7ca6308fb4fa344974c1268bac5c7604/core/ExceptionHandler.js#L154

This was intended for a world where the client is only talking to its own server and that server will only return 401s if the user is not logged in overall.  However in a (real) world where we make CORS requests and the client talks to other APIs, this is not appropriate.

We could ensure that the URL is relative/local to the same server, although with proxies a URL might look relative but be remote.  Another option would be to trigger this special handling only for /xh requests, or skip it entirely.

amcclain:
+ Fixes #1189 (or at least hopefully improves it - won't be perfect e.g. if proxies are in play)

### DIFF
--- a/core/ExceptionHandler.js
+++ b/core/ExceptionHandler.js
@@ -161,11 +161,17 @@ export class ExceptionHandler {
     }
 
     sessionMismatch(exception) {
-        return exception && exception.name === 'SessionMismatchException';
+        return exception.name === 'SessionMismatchException';
     }
 
+    // Detect an expired server session for special messaging, but only for requests back to the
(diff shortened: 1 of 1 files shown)


amcclain:
I am happy to test this more if merged on both deployed Toolbox (to ensure we still get the special message when desired) and in a CORS use case I encoutered with a client app (where we don't want this prompt).

Issue #1140, opened 2019-05-24
Selects now don't show all their items when you have filtered


nadkarniarjun:
@lbwexler found this initially:
I noticed that if you select, then blur, and then click the drop down again, you will see the full list once more. 

> Many Selects now don't show all their items when you have filtered 
> This is a Hoist change to <Select>.  Thought we had addressed.  Workaround might be to turn off keyboard querying.  I'll look tomorrow.

PR #1198, opened 2019-06-20
Ensure Select shows all options when refocused

amcclain:
WIP proposal for #1140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## v25.0.0-SNAPSHOT - under development
 
+### 🐞 Bug Fixes
+
+* The `Select` component now properly shows all options when the pick-list is re-shown after a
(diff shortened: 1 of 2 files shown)


nadkarniarjun:
Tested this in Toolbox and confirmed that the fix works.

amcclain:
Thanks Arjun for testing. 

@pauldelano and I ran into this bug yesterday reviewing a client app and it (understandably) surprised him. I would like to proceed with this fix unless anyone has a better approach - @lbwexler do you want to take a look? 

I had a sense we could do better somehow and avoid tracking this extra state, but I don't think it's terrible and believe the fix here is better than the bug 😄

lbwexler:
PROMISE, I will look and merge this this afternoon.

lbwexler:
Sorry for the delay here.  Looks promising!

Issue #1180, opened 2019-06-10
Admin pwd configs shouldn't show hash in rest editor form


amcclain:
For the admin config editor, we send down a digest dash of pwd config values. See:

https://github.com/exhi/hoist-core/blob/develop/grails-app/domain/io/xh/hoist/config/AppConfig.groovy#L74

IIRC there are some reasons for this - not sure if it relates to diffing configs across instances but might be.  

However it's very confusing for admin users, as the digest shows up in the value input when editing / reviewing the config. That hash looks like a plausible password, so users can (and do) think it's the value - not knowing about our special handling - and then attempt to use it in another context.  This can result in account lockouts.

We still *do not* want to send the plaintext password to the admin editor, but we need some special handling to avoid this mistake. 

<img width="485" alt="Screen Shot 2019-06-10 at 2 49 09 PM" src="https://user-images.githubusercontent.com/225806/59229336-f09dc880-8b8e-11e9-99d7-551671b29f8a.png">

amcclain:
Great thanks for assigning Lee - Yana I was going to suggest you could roll on to looking at this as part of your recent Admin client improvements.

This is a funny one for sure in terms of the overall system and special use case around pwd configs. Please reach out if you would like to review at any time or if I can help clarify anything.

amcclain:
Also - the comment about "diffing configs" might not have been obvious - there is the built-in config differ tool. It is very useful and a good tool, but also is something that we've unintentionally broken a few times as it's tucked in a quiet corner of the admin app and easy to overlook.  It also has special handling re. passwords - IIRC the intention is to be able to tell if they are different, but it deliberately does not attempt to sync pwd configs.  We should ensure we understand and document what we support with diffing pwds.

Also - there are some oddities about how that tool behaves in local development mode that I don't want to trip you up - and that we could potentially improve.  When it comes time to look at that part of the ticket please definitely reach out.

PR #1202, opened 2019-06-20
Rest Editor Form does not show password hash anymore

YanaNirvana:
Addresses Issue #1180: Rest Grid is now aware of a 'pwd' type
+ will render an html input of `type="password"` 
+ much thanks to co deving with @haynesjm42


lbwexler:
guys -- this is extremely economical.  Great.  Would love to merge it ASAP.

So for the record, we use 'pwd', to be consistent with our config system types?
(That is, our TextInput takes a type of 'password'.  Blueprint and HTML control takes a type of 'password', but our client-side data package will have a type of 'pwd'.  I think that's ok.)

YanaNirvana:
@lbwexler yes.  the record.valueType = 'pwd', which is what this renderer function relies on: https://github.com/exhi/hoist-react/blob/develop/admin/tabs/general/configs/ConfigModel.js#L125

We can probably change it; we'll just need to make sure the server is aware of that as well

amcclain:
Thanks for getting this done - a great change I am certain will (at least once) prevent people from locking out important service accounts.

I kind of wish we had gone with `password` for the config type all around instead of the abbrev to `pwd` - those super-short abbrevs often seem more confusing than they are worth.  But I don't think it's worth rocking the boat now, and believe we landed in a good place.

PR #1209, opened 2019-06-26
Make sure button child is not null before accessing its props (Fixes #1208)

### LINKED ISSUE #1208: Using 'omit' on a Button child in a ButtonGroupInput causes errors
Because we no longer remove null children in `elem` it is possible for components like ButtonGroupInput to receive `null` children. The components which perform operations on their children need to make sure they are making the proper null checks before attempting to dereference.

haynesjm42:
Handle case where the child Buttons could be null.

Searched through all references to `children` in the codebase, didn't see any other places where we are doing anything other than passing the children through to another component, or similarly null-safe operations.

Fixes #1208

### DIFF
--- a/desktop/cmp/input/ButtonGroupInput.js
+++ b/desktop/cmp/input/ButtonGroupInput.js
@@ -38,6 +38,8 @@ export class ButtonGroupInput extends HoistInput {
         const {children, minimal, disabled, enableClear, ...rest} = this.getNonLayoutProps();
 
         const buttons = castArray(children).map(button => {
+            if (!button) return null;
+
             const {value} = button.props;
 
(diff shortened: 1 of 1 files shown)


haynesjm42:
Note this issue was introduced in v22

amcclain:
Thank you. Merging this now as looks uncontroversial to me and a clear fix  

If this is something a client reported can you please add a changelog entry for the record? Otherwise happy to just include this quietly if we have only run into the bug internally.

haynesjm42:
Ran into it while upgrading a client app from v21 -> v24 - has not been an issue for any deployed applications or been seen by client users.

PR #1218, opened 2019-06-27
Fix DateInput 'showPickerOnFocus'.

### LINKED ISSUE #1212: DateInput 'showPickerOnFocus' is not working
The date picker popover is only showing for a few milliseconds before closing, for some reason the `onInteraction` Popover callback is getting called with `false` as the next state twice immediately after the popover is opened

TomTirapani:
Fixes #1212

### DIFF
--- a/desktop/cmp/input/DateInput.js
+++ b/desktop/cmp/input/DateInput.js
@@ -113,7 +113,8 @@ export class DateInput extends HoistInput {
             layoutProps = this.getLayoutProps(),
             enablePicker = withDefault(props.enablePicker, true),
             enableClear = withDefault(props.enableClear, false),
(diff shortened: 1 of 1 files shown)


amcclain:
@haynesjm42 on second thought I'm going to go ahead and merge this now to get it in - let us know if you are still seeing any issues here.

The unusual waiting, etc. in onBlur that Tom referenced in the fix is something I now remember putting in myself (I think). Still has a "for review..." comment - ultimately still counting on some more reliable replacement date control appearing one day....

haynesjm42:
Just tried it out, and can confirm that the picker does now show up when `showPickerOnFocus` is enabled. However after choosing a date in the picker, the picker popover closes, then re-opens briefly, then closes again.

Issue #1192, opened 2019-06-18
Grid export - scale mismatch when exporting percent values


amcclain:
Configure a grid column showing a percent value with `renderer: percentRenderer()` and `exportFormat: ExportFormats.PCT` - the exported value will display 100x the rendered value.

Not sure if this is possible to workaround - the Excel format makes it clear (within the Excel "format cells" UI) that it will display the value x100 with a percent sign.  Can workaround by providing `exportValue: v => v/100` - but should our percentRenderer match Excel's behavior and also x100 for display?  Would that be more natural in general?  (If so we can close this ticket or modify to make it about the renderer itself.)

lbwexler:
Don had some cycles, and is the original expert on formatting

PR #1225, opened 2019-07-01
Percent formatter (and renderer) will multiply value by 100*,  Fixes: #1192

febbraiod:
Matches Excel behavior

+This will be a breaking change (and clean up to exportValue:  v => v/100 in appropriate cols) in client apps using it; should be added to changelog as such after merge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ### 🎁 Breaking Changes
 
 * Applications that have been using fmtPercent or fmtPercentRenderer methods should adjust to
-  their new behavior. Columns that were previously using `exportValue: v => v/100` as as workaround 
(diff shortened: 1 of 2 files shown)


amcclain:
Hi - can you please include a changelog entry in this pull request?  That way the change and the log entry both go in together - has been working well for these feature branch merges, especially when there is a breaking change.

febbraiod:
Yes sorry, haven't worked in HR world a lot lately. Will do in a bit.

lbwexler:
Should we wait for v. 25 for this?  This is a huge breaking change.

amcclain:
Yes, I think so.  My thinking is that we put together a 24.2 release soon with some of the other changes in the pipeline, then the next time we have another breaking change we want to take to go along with this, we merge for v25.

Fortunately it should be easy to identify and adjust any usages, but I would expect there to be a few across the current crop of apps.

Issue #1216, opened 2019-06-27
Support custom Column sort comparators


haynesjm42:
The only way to achieve this currently is to use `agOptions` to specify a comparator, but it would be nice if we supported `comparator` as a first-level property on our Column config and it was passed Hoist data structures (records, column, store, gridModel, etc.) instead of having to use the ag-Grid RowNode objects if you need to inspect anything other than the value.

Would also want to make sure we pass the `absSort` flag to the comparator.

haynesjm42:
Specific use-case from client here is to have root-level tree rows use a static sorting, while sorting their children alphabetically.

haynesjm42:
So on further investigation with @YanaNirvana we realized that a per-column sort comparator would not be ideal to solve this issue, as a comparator would need to be added for each column.

ag-Grid does provide a `postSort` callback which we can implement to force certain rows to static positions after other sorting has been applied. Unfortunately ag-Grid's api for this requires mutating an array of RowNodes, so the implementation might get a bit messy if we want to add a layer on top of this in our framework. @lbwexler would like to get your thoughts on whether we should add our own `postSort` callback to GridModel. I do think the original request in this ticket is valid and something we should still do though.

YanaNirvana:
Just to collect things together in the same space:

https://www.ag-grid.com/javascript-grid-sorting/#post-sort

(John's keen eye spotted that in their example, they don't return anything-- they mutate in place using `splice`)

lbwexler:
Hi -- This is fine for Hoist 25, which we will look to go out this week.  I discussed with client that is waiting for it.

PR #1226, opened 2019-07-02
GridModel to accept a `defaultCol`

YanaNirvana:
Half address Issue #1216: defaultCol config to be applied to all columns, except where otherwise specified
(co dev with jhaynes)

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -116,7 +116,7 @@ export class GridModel {
     /**
      * @param {Object} c - GridModel configuration.
      * @param {Object[]} c.columns - {@link Column} or {@link ColumnGroup} configs
-     * @param {Object} [c.colDefaults] = Column configs to be set on all columns.  Merges deeply.
+     * @param {Object} [c.colDefaults] - Column configs to be set on all columns.  Merges deeply.
      * @param {(Store|Object)} [c.store] - a Store instance, or a config with which to create a
      *      Store. If not supplied, store fields will be inferred from columns config.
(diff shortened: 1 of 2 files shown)


amcclain:
Thanks for getting this in - appreciate it. 

On reflection, I am not sure about the name `defaultCol` - the sense of that doesn't seem quite right.  How about `colDefaults`?

cmp/grid/GridModel.js
```diff
@@ -191,6 +195,7 @@ export class GridModel {
 
         Object.assign(this, rest);
 
+        this.defaultCol = withDefault(defaultCol, {});
```
lbwexler: In this case, could do the defaulting in the constructor for self-documenting....also more in-line with the other two defaults.

cmp/grid/GridModel.js
```diff
@@ -115,6 +116,8 @@ export class GridModel {
     /**
      * @param {Object} c - GridModel configuration.
      * @param {Object[]} c.columns - {@link Column} or {@link ColumnGroup} configs
+     * @param {Object} [c.defaultCol] = Column configs to be set on all columns.  If a Column has
```
lbwexler: I feel that this comment states the obvious regarding what "wins out".  That's fine, but  I think the less obvious point that we should clarify is that the merging is done "deeply"...  I would not assume that as an API user and would be glad to have it clarified....

lbwexler:
Also agree with Anselm's proposed name change.  This will be a nice pickup!

Issue #1187, opened 2019-06-14
Improve Handling when Grails backend unavailable


lbwexler:
We have two issues here --

1) Our error message is still too wonky ("EnvironmentService blah, blah..") We need to get right to the point about the "UI Server is Unavailable".

2) If you go ahead and try to submit the error, that just fails and annoyingly does nothing, and does not close the dialog.

I think 2 is probably something we can handle with better exception handling generally on that button action.... should never just fail silently like that!  Its just heaping more pain on an already suffering user....

lbwexler:
Yana -- are you working on this yet?  if not, would like to give it to Don

YanaNirvana:
not yet, it's all Don's!

PR #1229, opened 2019-07-02
Clearer error message for case in which grails server is not running, Fixes: #1187

febbraiod:


febbraiod:
Not sure there is anything else to be done here. The submit button will not be shown if the IdentityService does not have a known user, and since this occurs before IdentityService is inited we don't have to worry about getting a second error on an attempted submit.

I tried to think of something that might be less Jargon-y than 'UI server is unavailable' but the only thing I could think of was 'Application is unavailable', but that's so general as to become meaningless and totally unhelpful.

core/XH.js
```diff
@@ -467,7 +467,7 @@ class XHClass {
             try {
                 await this.installServicesAsync(EnvironmentService);
             } catch (e) {
-                throw `Unable to load environment info - is the server running and reachable? (${e.message})`;
+                throw `UI Server is unavailable - please contact support. (${e.message})`;
```
lbwexler: Any reason we aren't using XH.exception() here?

lbwexler: We could do something like XH.exception({
message: 'UI Server is unavailable - please contact support.'
detail: Some details of exception including the actual *URL*
})

Would ideally allow users to copy all details to clipboard, if not mail them  (although not sure we have easy way to do one without the other, and recall, mailing unlikely to work without backend)

febbraiod: Did this. User was always, and still is, able to copy the details to the clipboard, but yes would have to email with another service. No backend to save to, and cannot set up email with only JS.

Screen Shot 2019-07-05 at 9 15 43 AM

febbraiod: The latest:

Screen Shot 2019-07-05 at 10 01 00 AM

PR #1230, opened 2019-07-02
DateInput picker-only mode

### LINKED ISSUE #1220: DateInput component should support picker only mode
We have a client that would like to not enable the direct editing of the text display, and force the user to pick with the picker.

Seemed like a very reasonable request that avoids users getting caught up in the wrong date formatting etc. etc.

febbraiod:
Hoist P/R Checklist
-------------------

Fixes #1220


febbraiod:
There is a corresponding branch in toolbox for demo purpose, but not intended for merge

lbwexler:
This looks good -- glad it was so straightforward.  Sorry I wasn't clearer about the name -- I think I was just trying to bang that ticket in quickly, and was not thinking that would be the name

lbwexler:
Let's go for symmetry with the existing  'enablePicker' and call it 'enableTextInput'?
Obviously a bit weird if both were set false, but think that is going to be the most straightforward.

lbwexler:
I went ahead and changed the name.

Testing out the formatting, I realized that it didn't look quite right.  It really should not be rendered as *disabled*.  That implies its fixed -- Its not!  It just should not be editable with the keyboard, thus encouraging/forcing the user to use the button.

febbraiod:
10-4 will look at the styling vs disabled edit friday

lbwexler:
Looking much better, but still not right.  Right now, if enableTextInput is false, you cannot tab into the control.  Realize this is extremely tricky and sheds light on the fact that our implementation might be seriously wrong for this case.  When enableTextInput is false, I think we just need an independent calendarButton  that is adjacent to a read only Text display, not an embedded icon.  

When they are both enabled, might use this new layout as well, or leave the current setup -- its really a matter of which will be easiest to implement and give a consistent look-and-feel.

desktop/cmp/input/DateInput.js
```diff
@@ -172,22 +179,25 @@ export class DateInput extends HoistInput {
...
                     tabIndex: -1, // Prevent focus on tab
                     onClick: this.onClearBtnClick
                 }),
                 button({
-                    omit: !enablePicker,
+                    className: enablePicker ? '' : 'disabled-picker-icon',
                     icon: Icon.calendar(),
-                    tabIndex: -1, // Prevent focus on tab
-                    onClick: this.onPopoverBtnClick
+                    tabIndex: isPickerOnlyMode ? undefined : -1, // Prevent focus on tab, unless in pickerOnly mode
```
febbraiod: This fixes our focus on tab problem without having to move the picker button out of the disabled input. A little embarrassed at how long this took me to figure out given the incredibly obvious comment that I apparently just didn't read about 100 times I looked at this code. :/

desktop/cmp/input/DateInput.js
```diff
@@ -165,29 +173,34 @@ export class DateInput extends HoistInput {
                 })
             }),
 
+            onClick: !enableTextInput ? this.onOpenPopoverClick : null,
```
febbraiod: Think this needs the isPickerOnlyProp.

PR #1234, opened 2019-07-03
Update react-transition-group to ~4.2.1, Fixes #1154

### LINKED ISSUE #1154: Update react-transition-group to 4.x
We are on 2.8.0 - see https://github.com/reactjs/react-transition-group/blob/master/CHANGELOG.md

febbraiod:


febbraiod:
Only breaking change listed in their [changelog](https://github.com/reactjs/react-transition-group/blob/master/CHANGELOG.md) is:

in environments where esm is supported importing from commonjs requires explicitly adding the .default after require() when resolving to the esm build

which I do not believe effects us.

PR #1235, opened 2019-07-03
Added GridModel exportOption.columns option

pauldelano:


amcclain:
Thanks for getting this in Paul.  Please see my updates where I added a bit of code in GridModel to make this a non-breaking-change for the common use cases.  We don't do this that often, but given that this involves grid and that I would like to release as a minor, non-breaking change I think it is worth it.

I also simplified the logic a bit to extract the exportable columns - was getting tripped up a bit by the recursion and I think / hope this is clearer.  Of course please give a shout if anything looks amiss.

PR #1238, opened 2019-07-05
Ensure loadAsync has proper default loadSpec if passed null, Fixes #1237

### LINKED ISSUE #1237: loadAsync method in loadSupport will not always default the loadSpec as desired
This was discovered while working on:

https://github.com/exhi/toolbox/issues/217

Arguably, toolbox is not using loadSupport correctly, but I think there is a safe change to be made in HR that will avoid this issue.

febbraiod:


febbraiod:
Problem surfaced from calls in toolbox that looked like this:

Screen Shot 2019-07-05 at 3 09 54 PM

lbwexler:
Whoa -- good catch, but I think this is the wrong fix!  The problem is with the usage itself, which I think I may have propogated throughout toolbox at some point.  As Anselm also pointed out to me, this reaction is very problematic.   We never want to randomly pass on the arguments from track, to the run  function if we don't intend to pass them!  In this case, was null, but could have been non-null and would still be very wrong!!

The run argument should be () => this.loadAsync()

lbwexler:
Closing this, as discussed above.  Will comment in toolbox ticket

PR #1243, opened 2019-07-09
Ping url for grails server exception is generated based on env

febbraiod:
+Tweak error message

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -469,11 +469,13 @@ class XHClass {
                 await this.installServicesAsync(EnvironmentService);
             } catch (e) {
                 const loc = window.location,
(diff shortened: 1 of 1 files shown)


core/XH.js
```diff
@@ -468,9 +468,12 @@ class XHClass {
             try {
                 await this.installServicesAsync(EnvironmentService);
             } catch (e) {
+                const loc = window.location,
+                    pingURL = XH.isDevelopmentMode ? `${XH.baseUrl}ping` : `${loc.protocol}//${loc.host}/ping`;
```
amcclain: Will still need to work in baseUrl for the production case - it will be /api/ but that's important as otherwise the link won't pass through to the back-end server.

Believe we can use `${location.origin}${XH.baseUrl}ping` - that looks correct on deployed Toolbox.

PR #1249, opened 2019-07-14
Add GridCountLabel - record + selection count

### LINKED ISSUE #1247: GridCountLabel component
Variation on existing StoreCountLabel, but takes a grid Model and shows selection count in addition to overall record count. 

Might be useful for a filtering indicator as well at some point

amcclain:
+ Fixes #1247
+ Moved StoreCountLabel to /cmp as it is / can be cross-platform. Aliased import from /desktop/cmp to avoid breaking change, allow importing in one line alongside desktop-only StoreFilterField.

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR - https://github.com/exhi/toolbox/pull/223


amcclain:
Wanted to review and ensure we were happy with this overall idea / execution.  If so, we can use this for more admin grids, etc.

Also note decision (happy to discuss!) to move `StoreCountLabel` to /cmp from /desktop but still provide an import alias.

We could have pushed this functionality into StoreCountLabel by providing that with a `StoreSelectionModel` (generic usage) or reading both off of gridModel.  I think it might be more clear / future-proof to have a dedicated comp though in case we want to do anything with filtering, etc.

lbwexler:
One concern I have here es that we are prioritizing a pretty specific display of selection, but have not also dealt with the equally (or more) important case of displaying filter state (e.g. "10 of 1000 records" displayed).

Also a bit wary of adding in the aliases as seems like that could lead to a state where we aren't ever really sure where something actually is.

amcclain:
The filter state is exactly the kind of thing I thought we could add on to this component as we go - when we have a better handle on it vis-a-vis the grid in general. I think it could be a natural enhancement to this component we could make sooner or later.

The main motivation here was that @pauldelano needed a component like this for several client projects, and we had discussed it as a good option for Hoist vs. cloning/modifying StoreCountLabel within apps. If you guys have a chance to touch base and see if this fits the bill, how to proceed that might be the best next step.  I'm of course entirely flexible. 👍

lbwexler:
My concern with this one is that we will end up in the API with two very similar objects due to historical reasons.  The gridModel prop on StoreCountLabel just then becomes really weird.

Would we consider making this a breaking change, and removing support for the gridModel prop on StoreCountLabel?  Setting that prop could yield an exception telling you to change to GridCountLabel instead.

Furthermore, we could remove the Alias on the StoreCountLabel move and just fix two API problems with essentially a single breaking change....All uses of StoreCountLabel have to be adjusted (slightly)....

amcclain:
Sure - sounds like a plan.  I will do that in a little while, when we are sure we're prepping v26.  I think we have at least one 25.1 release to do first.

amcclain:
Updated as described

PR #1255, opened 2019-07-16
Submit *dirty* changes only for record editing.  Fixes 1245

lbwexler:
Hoist P/R Checklist
-------------------


amcclain:
Reviewed and will rebase on now - thanks.  

Was caught up a bit and looked carefully re the state of `currentRecord` in this code - for adds - and also how cloning factored in.  It's a little strange in that, when adding normally, current record is undefined and it seems weird to spread it here.  

But when cloning, currentRecord is the prepared clone rec so the spread does do something - namely, bring along any other fields that weren't included in the formModel.  I think that's OK and probably exactly what we want/need to happen, but remains confusing.

That said, did test carefully and fixes the issue which is what matters here!

PR #1258, opened 2019-07-17
Warn if param passed to loadAsync that does not appear to be a loadSpec

amcclain:
+ This appears to be a common mistake, typically made by binding loadAsync directly to a reaction - it's then passed whatever is being tracked.
+ Won't catch cases where tracked observable is a plain object and is unintentionally passed in, but this seemed better than nothing.

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (N/A)

### DIFF
--- a/core/mixins/LoadSupport.js
+++ b/core/mixins/LoadSupport.js
@@ -52,7 +52,7 @@ export function LoadSupport(C) {
              */
             async loadAsync(loadSpec = {}) {
                 throwIf(
-                    (!isPlainObject(loadSpec)),
+                    !isPlainObject(loadSpec),
                     'Unexpected param passed to loadAsync() - accepts loadSpec object only. If triggered via a reaction, ensure call is wrapped in a closure.'
                 );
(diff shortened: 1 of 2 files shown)


core/mixins/LoadSupport.js
```diff
@@ -50,6 +51,10 @@ export function LoadSupport(C) {
              * @param {LoadSpec} [loadSpec] - Metadata about the underlying request
              */
             async loadAsync(loadSpec = {}) {
+                warnIf(
+                    (!isNil(loadSpec) && !isPlainObject(loadSpec)),
```
lbwexler: it seems to me that we should not be giving isNil a pass?  Won't that just fail badly on line 60?
Its not ok to pass null in here either, and undefined should yield the assignment of the default {}.

amcclain: 👍 Updated, changelog entry added to highlight the stricter behavior.

PR #1265, opened 2019-07-22
copyCell Grid Context Menu Action (Fixes #690)

### LINKED ISSUE #690: Enhance Copy Support in grid

haynesjm42:
Support for new copyCell action which works off the record and column that was clicked on. Uses https://www.npmjs.com/package/clipboard-copy for putting the data into the clipboard programmatically.

Also added changes to use our standard copy icon for the built-in copy actions.

Note that this branch originally contained additional changes to use our own implementation for serializing the grid rows for copying to allow for more flexibility than what ag-grid provides, have preserved that and will explore further in the future, but wanted this feature to be able to get in separately since it has come up multiple times.

Hoist P/R Checklist
-------------------


kit/clipboard-copy/index.js
```diff
@@ -0,0 +1,10 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2019 Extremely Heavy Industries Inc.
+ */
+
+import copy from 'clipboard-copy';
+
+export {copy as copyToClipboard};
```
lbwexler: John -- why do we need this?  Would it be possible to do this with the 'clipboard' package that we already have.....

haynesjm42: I looked at clipboard.js and didn't see any way to trigger the copy to clipboard programmatically. It looked like it always needed to be attached to an element in order to function, which would be a little tricky with what ag-Grid provides us, and would generally require a lot more wiring to function if it is even possible. Unless I am missing something it definitely would not be as simple as the one-liner that the clipboard-copy package provides.

lbwexler:
Could we easily simplify the few places where we are using 'clipboard'?
Uncomfortable with having two libraries that purport to provide the same
exact functionality.  Is it possible you just found something better?

haynesjm42:
Will look into it a bit more to see if we can/should replace clipboard.js with clipboard-copy.

haynesjm42:
So the only usage we have of clipboard.js is the ClipboardButton component. We could replicate the functionality there using the copy-clipboard library but it would be more complicated, as clipboard.js is perfect for that use-case. clipboard.js is excellent if you have an element which you want to trigger copying (or cutting) of another elements content, while clipboard-copy is excellent if you just want to put some arbitrary data into the clipboard. They both serve their use-case very well, but not so well for the other use-case, so they don't really provide the same functionality and I think it makes sense to include both.

lbwexler:
I was able to update the existing classes to use the new promise based library.  I always thought the existing implementation was way too confusing anyway!

Also, please note that I dropped the 'kit' entry.  Seems like overkill here to me -- unless we are installing element factories, or somehow adding value, not sure what the point of a 'kit' entry is....

Let me know your thoughts -- if you are fine with it, ok to merge

haynesjm42:
Looks good to me, and thanks for taking care of eliminating the other library, looks like a good simplification to the ClipboardButton interface. And agree the kit was overkill, not entirely sure what my reasoning was for adding that :)

package.json
```diff
@@ -24,6 +24,7 @@
     "ag-grid-react": "~21.1.0",
     "classnames": "~2.2.6",
     "clipboard": "~2.0.4",
+    "clipboard-copy": "^3.0.0",
```
amcclain: @lbwexler / @haynesjm42  - saw the thread re. avoiding the two libraries, but doesn't look like the old one was actually removed... 😀OK to actually take it out?

lbwexler: great catch on this one.  Removed!

Issue #1228, opened 2019-07-02
JsonInput option for a maximized or "pop-out" view


lbwexler:
Perhaps a modal popup of some sort? Or a grow affordance? 

Would be nice to have the control handle it, so the form containing the control does not need to add special handling to get a usable editor.

I am about ready to poke my eyes out editing these configs in the little window config viewer is giving me....

lbwexler:
its all yours panda

amcclain:
Thanks for filing this - has been on my mental list for some time.  

Agree that ideally this would be a feature nicely encapsulated / managed by the JsonInput itself.  I think a modal pop-out would be most useful with some kind of little way to launch.  

I don't know if it can support some kind of baby toolbar inside of the editor.  If so, I can imagine that hosting a button to pop-out as well as a button to pretty-print (currently supported via a very obscure keyboard shortcut).  I think we would want a prop to suppress this toolbar if not wanted, but could default to on.  I am thinking of something as minimal and unobtrusive as possible, while still being obvious enough that users will find and make use of it.

lbwexler:
This has become a high priority.  We have very lunch json configs to edit in some client apps, and only 5 visible lines to do it in our config editor.

TomTirapani:
https://github.com/exhi/hoist-react/pull/1267

PR #1267, opened 2019-07-22
Add option to show JsonInput in fullscreen maximised popout

TomTirapani:
Also adds button to JsonInput to auto-format content.
Addresses issue  #1228

Hoist P/R Checklist
-------------------


desktop/cmp/input/JsonInput.js
```diff
@@ -47,11 +52,18 @@ export class JsonInput extends HoistInput {
...
-        editorProps: PT.object
+        editorProps: PT.object,
+
+        /** True to show Fullscreen + Auto-format buttons at top-right of input. */
+        showActionButtons: PT.bool
     };
 
     get commitOnChange() {
-        withDefault(this.props.commitOnChange, false);
+        return withDefault(this.props.commitOnChange, false);
```
amcclain: Good catch!

desktop/cmp/input/JsonInput.scss
```diff
@@ -29,6 +29,25 @@
       box-sizing: content-box !important;
     }
   }
+
+  &__action-buttons {
+    position: absolute !important;
+    top: 0;
+    right: 6px;
+    z-index: 3;
```
amcclain: Suggest a comment on how we ended up at 3 here. 😄 

I see this works well when e.g. launched from a dialog. The rules for z-index always escape me - does this end up being relative in some sense (inside a "stacking context")?

TomTirapani: Simply put, that's there to place it "above" the code-mirror row and enable clicking on it. The code-mirror styles use `z-index: 2` internally for the rows of content - without this rule, clicking on the buttons selects the first row of content instead.

Just added a comment to the code to clarify.

lbwexler:
👍 Thanks Tom!

Issue #1264, opened 2019-07-19
NumberFormat colorSpec arg should allow specifying only one case


amcclain:
If supplying `colorSpec` in its object form, you need to provide non-empty values for pos, neg, and neutral.  You might only want to customize the color on one case. 

Think it's a simple matter of removing/relaxing this check - should be OK to return null or empty string to the classlist:

https://github.com/exhi/hoist-react/blob/607732368570389f2253244ef3ab66eec2f6de96/format/FormatNumber.js#L309

Also, docs don't make it clear these should be CSS classes vs. e.g. hex values.

PR #1270, opened 2019-07-22
Number formatters should accept a partial colorSpec

febbraiod:
Hoist P/R Checklist
-------------------

Fixes: #1264

### DIFF
--- a/format/FormatNumber.js
+++ b/format/FormatNumber.js
@@ -43,6 +43,7 @@ const UP_TICK = '▴',
  * @param {string} [opts.labelCls] - CSS class of label <span>,
  * @param {(boolean|Object)} [opts.colorSpec] - show in colored <span>, based on sign of value.
  *      True for red/green/grey defaults, or object of the form {pos: color, neg: color, neutral: color}.
+ *      When passing an object, values should be CSS classes to be applied to the wrapping span.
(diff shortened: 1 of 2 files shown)


format/FormatNumber.js
```diff
@@ -306,10 +307,16 @@ function valueColor(v, colorSpec) {
     const defaultColors = {pos: 'xh-green', neg: 'xh-red', neutral: 'xh-gray'};
     colorSpec = typeof colorSpec === 'object' ? colorSpec : defaultColors;
 
-    if (!colorSpec.pos || !colorSpec.neg || !colorSpec.neutral) {
+    if (!colorSpec.pos && !colorSpec.neg && !colorSpec.neutral) {
```
febbraiod: Is there value to this now? We are not being very strict about this now, what's the harm in just applying no classes at all if your object is missing these? Also if we do think it's worth throwing we should work on the message:

'Invalid color spec: ' + colorSpec is appearing as 'Invalid color spec: [Object object]' int he console.

amcclain: Agreed - don't think we need it!  I will remove and check in a few other cleanups that came to me while waiting here in the Syracuse airport.  (✈️ deferred from NY so have plenty of time to twiddle these things....)

PR #1275, opened 2019-07-23
Cube filter fn

febbraiod:
Far from ready for prime time, but want to show this as a POC and open discussion as to whether or not we think this is worth while (before we invest anymore time is what might turn out to be a hairy change). Please see toolbox branch of the same name.

### DIFF
--- a/data/cube/Query.js
+++ b/data/cube/Query.js
@@ -124,6 +124,6 @@ export class Query {
 
     parseFilters(filters) {
         if (!filters) return null;
-        return filters.map(f => f instanceof ValueFilter ? f : new ValueFilter(f.name, f.values));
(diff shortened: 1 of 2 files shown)


lbwexler:
Hey -- finally had a chance to look at this.

So I think for starters, we might just create a new Filter class called a "FunctionFilter" if our filter config is actually a filter.  

BUT a larger concern I have, however, is with the id generation/parsing code.  It seems like we have a pretty strong understanding that the filters can always be regenerated from a serialized id.....

Do we need to do this?

febbraiod:
Yes that concern ( id generation/parsing) is exactly why I decided not to go any further without this discussion. And no it doesn't seem we NEED it. I just thought it was a good idea before I even looked at the code. The motivating use case in the application has been dealt with in a satisfactory manner without this change

PR #1279, opened 2019-07-25
AgGridModel IsReady checks (Fixes #1232)

### LINKED ISSUE #1232: AgGridModel fails badly with null pointer exception
If you try to use the state getter / setter functions in AgGridModel (e.g. `getFilterState`) before the `agApi` / `agColumnApi` are ready, the app will throw badly. This leads to us often having to perform a null check on these apis prior to attempting to use these methods.

Would be nice if these checks could be contained within the methods, to avoid having to use the boilerplate.

haynesjm42:
We now throw an exception if any of the AgGridModel methods which depend on the ag-Grid api references are called before we have those references (before onGridReady was called).

Also added an `IsReady` property which applications can check before calling any of those methods, and can also add reactions on to be notified when the state can/should be initialized if the application is managing it.

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (https://github.com/exhi/toolbox/pull/230)


cmp/ag-grid/AgGridModel.js
```diff
@@ -68,6 +68,16 @@ export class AgGridModel {
         });
     }
 
+    /**
+     * @returns {boolean} - true if the grid fully initialized and its state can be queried/mutated
+     */
+    // Intentionally not using @computed here because these api references may be updated if the grid
+    // is re-rendered, and @computed would not fire reactions since the return value would not have
+    // changed
+    get isReady() {
```
lbwexler: I don't understand this logic?  Why does someone observing this getter care about the underlying api's?  If you are observing isReady() that's what you are observing.  If you need to observe the API directly, you should do that.

haynesjm42: Well I wanted any reactions on IsReady to be fired whenever the api references changed, because that is what would indicate that the grid state needs to be initialized if it is being managed by something else. It is possible that the grid would be re-initialized by being re-rendered, which means someone managing the state of this grid via this model would need to know about it and set the data/state again. I seem to remember having this same conversion with the initial implementation of this class where I had this same getter.. and we decided to remove it

lbwexler: Yikes.  I also remember this.  Let's discuss.  Don't want to be going in circles here....

lbwexler:
John and I discussed this at length today.  We will make some nice adjustments to actually clean-up the api references when the grid is unmounted.  Then we should be able to go forward with this, with a more reliable isReady() prop that is observable, and toggles as it should.

haynesjm42:
@lbwexler - implemented the changes as we discussed

lbwexler:
👍

PR #1280, opened 2019-07-25
TreeMap

TomTirapani:
Hoist P/R Checklist
-------------------


desktop/cmp/treemap/SplitTreeMap.js
```diff
@@ -0,0 +1,62 @@
...
+@HoistComponent
+@LayoutSupport
+export class SplitTreeMap extends Component {
+
+    static propTypes = {
+        /** Primary component model instance. */
+        model: PT.oneOfType([PT.instanceOf(SplitTreeMapModel), PT.object]).isRequired,
+
+        /** Function to render section titles. Receives section total, and section ['positive', 'negative']. */
+        titleRenderer: PT.func
```
lbwexler: 'sectionTitleRender' might be a clearer name?

desktop/cmp/treemap/SplitTreeMapModel.js
```diff
@@ -0,0 +1,107 @@
...
+@HoistModel
+export class SplitTreeMapModel {
+
+    //------------------------
+    // Immutable public properties
+    //------------------------
+    /** @member {GridModel} */
+    gridModel;
+    /** @member {function} */
+    filter;
```
lbwexler: if we are calling the sections 'posModel' and 'negModel', this implies a metric with a sign that will be different on both sides.    Seems like we either stick with that, and specify a 'sectionMetric' here OR we do something more generic like a "regionFilter" accompanied by a "primaryRegionModel' and a 'secondaryRegionModel'.

desktop/cmp/treemap/SplitTreeMapModel.js
```diff
@@ -0,0 +1,107 @@
...
+            // Only include root records that pass the filter
+            if (it.parent || !this.filter(it.record)) return 0;
+            return it.value;
+        });
+    }
+
+    @computed
+    get negRootTotal() {
+        return sumBy(this.negModel.data, it => {
+            // Only include root records that don't pass the filter
```
lbwexler: could consider using loki here

desktop/cmp/treemap/SplitTreeMapModel.js
```diff
@@ -0,0 +1,107 @@
...
+    } = {}) {
+        throwIf(isNil(gridModel), 'SplitTreeMap requires a GridModel.');
+        throwIf(!isFunction(filter), 'SplitTreeMap requires a filter function.');
+
+        this.config = config;
+        this.gridModel = gridModel;
+        this.filter = filter;
+        this.treeMapModelConfig = treeMapModelConfig;
+
+        if (!['vertical', 'horizontal'].includes(orientation)) {
```
lbwexler: Could just throw here, if simpler....

desktop/cmp/treemap/SplitTreeMapModel.js
```diff
@@ -0,0 +1,107 @@
...
+        return sumBy(this.negModel.data, it => {
+            // Only include root records that don't pass the filter
+            if (it.parent || this.filter(it.record)) return 0;
+            return it.value;
+        });
+    }
+
+    /**
+     * @param {Object} c - SplitTreeMapModel configuration.
+     * @param {Object} c.config - Highcharts configuration object for the managed chart. May include
```
lbwexler: is the 'config' name consistent with something else?  Seems like it would be useful to have highCharts in the name here...

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,250 @@
...
+    _chart = null;
+
+    get data() {
+        // If not bound to a grid, simply use model data
+        const {data, gridModel} = this.model;
+        if (!gridModel) return data;
+
+        // If bound to a grid, mixin selected state from GridModel
+        const selectedIds = gridModel.selModel.ids,
+            {selectionColor} = XH.darkTheme ? DarkTheme : LightTheme;
```
lbwexler: I actually think the selection color on this component is both distracting and problematic.  You are actually losing the ability to visualize this critical color dimension on the one node you presumably have decided that you are interested in examining!  

Could we do a tBook style highlighting of the selected node with a border?

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,238 @@
...
+     * @param {Object} c.config - Highcharts configuration object for the managed chart. May include
+     *      any Highcharts opts other than `series`, which should be set via dedicated config.
+     * @param {Object[]} c.data - Raw data to be displayed.
+     * @param {GridModel} [c.gridModel] - Optional GridModel to bind to.
+     * @param {string} c.labelField - Record field to use to determine node label.
+     * @param {string} c.valueField - Record field to use to determine node size.
+     * @param {string} c.heatField - Record field to use to determine node color.
+     * @param {string} [c.valueFieldLabel] - Label for valueField to render in the default tooltip.
+     * @param {string} [c.heatFieldLabel] - Label for heatField to render in the default tooltip.
+     * @parma {function} [c.filter] - A filter function used when processing data. Receives (record), returns boolean.
```
lbwexler: This filter is confusing seems like an (important) implementation detail.

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,238 @@
...
+        this.valueFieldLabel = valueFieldLabel;
+        this.heatFieldLabel = heatFieldLabel;
+        this.filter = filter;
+        this.tooltip = tooltip;
+
+        this.onClick = withDefault(onClick, this.defaultOnClick);
+        this.onDoubleClick = withDefault(onDoubleClick, this.defaultOnDoubleClick);
+
+        if (!['sliceAndDice', 'stripes', 'squarified', 'strip'].includes(algorithm)) {
+            console.warn(`Algorithm ${algorithm} not recognised. Defaulting to 'squarified'.`);
```
lbwexler: Again, throwing might be simpler.

cmp/ag-grid/AgGridModel.js
```diff
@@ -409,7 +416,7 @@ export class AgGridModel {
     }
 
     getGroupNodePath(node) {
-        throwIf(!node.group, 'Cannot get the path for a non-group node!');
+        throwIf(!node.allChildrenCount, 'Cannot get the path for a non-group node!');
```
amcclain: I see that the two callers to this method both appear to have already run this check.  Wondering what the value is here - especially as we've changed the check to look for children, and I could imagine some valid future usage hitting would-be-group-nodes that happen to not have children due to a filter or similar.  

Would the code below throw? I like these internal assertions where they help catch errors and make dev-time mistakes more clear with helpful exceptions, but not sure that applies here.

cmp/grid/GridModel.js
```diff
@@ -94,6 +94,8 @@ export class GridModel {
     @observable.ref columns = [];
     /** @member {ColumnState[]} */
     @observable.ref columnState = [];
+    /** @member {Object} */
+    @observable.ref expandState = {};
```
amcclain: What's our contract around keeping this expand state accurate and in sync?  I see we call on `expand/collapseAll()` + `onRowGroupOpened` in the Grid. But what about initial setup, or on a single row collapse?

This is another case where performance might be an issue (especially if the above comment would cause us to call more frequently).  Now we're triggering a pretty intensive iteration over the nodes on these common actions. Realize this was added for a reason 😄 but want to ensure we're being as efficient as possible.

desktop/cmp/chart/ChartModel.js
```diff
@@ -14,28 +14,23 @@ import {observable, action} from '@xh/hoist/mobx';
...
 
-    @observable.ref config = {};
-    @observable.ref series = [];
+    @bindable.ref highchartsConfig = {};
+    @bindable.ref series = [];
 
     /**
      * @param {Object} c - ChartModel configuration.
-     * @param {Object} c.config - Highcharts configuration object for the managed chart. May include
+     * @param {Object} c.highchartsConfig - Highcharts configuration object for the managed chart. May include
```
amcclain: 💥 I think we should document in changelog as a breaking change since we're taking breaking changes with this release anyway.

Given that we have relatively few charts in production usage, I would also favor throwing in the ctor vs. swapping the configs to keep things working.  (Typically I lobby for the smoother deprecations when we know we're gonna hit a lot of them, but I think we can be more aggressive here.)

desktop/cmp/chart/ChartModel.js
```diff
@@ -14,28 +14,23 @@ import {observable, action} from '@xh/hoist/mobx';
 @HoistModel
 export class ChartModel {
 
-    @observable.ref config = {};
-    @observable.ref series = [];
+    @bindable.ref highchartsConfig = {};
+    @bindable.ref series = [];
```
amcclain: Don't feel strongly about this, but explicit setters might still have some value if they are an important part of the public API.  WIll make it easier for tooling to find them, and if there are any caveats or notes to document (i.e. can you change anything about the highchartsConfig?) we would have a place to do so.

kit/highcharts/index.js
```diff
@@ -0,0 +1,30 @@
...
+import highchartsHeatmap from 'highcharts/modules/heatmap';
+
+Highcharts.setOptions({
+    global: {
+        useUTC: false
+    },
+    lang: {
+        thousandsSep: ','
+    }
+});
```
amcclain: Great that we moved this here - exactly the kind of thing that's nice to do in `/kit/` index files.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+import {DarkTheme} from './theme/Dark';
+
+import './TreeMap.scss';
+import {TreeMapModel} from './TreeMapModel';
+
+highchartsExporting(Highcharts);
+highchartsOfflineExporting(Highcharts);
+highchartsExportData(Highcharts);
+highchartsTree(Highcharts);
+highchartsHeatmap(Highcharts);
```
amcclain: Should we be doing these in the `/kit/` index file?  Or at least the ones we are also doing in `Chart.js`?  I assume they are initializing / enabling these features/plugins.  Could it be an issue if this is done twice?  

If there is some cost (in terms of running the code or roping additional libraries into the webpack bundle) I can see keeping the two map-related calls here in this component.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+ * Copyright © 2019 Extremely Heavy Industries Inc.
+ */
+import {Component} from 'react';
+import PT from 'prop-types';
+import {Highcharts, highchartsExporting, highchartsOfflineExporting, highchartsExportData, highchartsTree, highchartsHeatmap} from '@xh/hoist/kit/highcharts';
+
+import {XH, elemFactory, HoistComponent, LayoutSupport} from '@xh/hoist/core';
+import {div, box} from '@xh/hoist/cmp/layout';
+import {Ref} from '@xh/hoist/utils/react';
+import {resizeSensor} from '@xh/hoist/kit/blueprint';
```
amcclain: Is this the dep that's keeping this in the desktop package?  Anything else that we think is a real blocker to usage on mobile as well?  

Even if we don't test / recommend up front, it would be a nice win if we could land this in the cross platform `/cmp/` package to set us up for full mobile support now or later. We've seen how nice these maps can look even on phone.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+
+    render() {
+        // Default flex = 1 (flex: 1 1 0) if no dimensions / flex specified, i.e. do not consult child for dimensions;
+        const layoutProps = this.getLayoutProps();
+
+        if (layoutProps.width == null && layoutProps.height == null && layoutProps.flex == null) {
+            layoutProps.flex = 1;
+        }
+
+        this.renderHighChart();
```
amcclain: Is this a no-op on first render as the div has yet to be setup on the DOM?  Just thinking that through - is it the ref getting its value that then triggers it to loop back through?  Might be worth a comment as it seems kind of unusual (although understandable).

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+        this._chart.setSize(width, height, false);
+        this.updateLabelVisibilityAsync();
+    }
+
+    destroy() {
+        this.destroyHighChart();
+    }
+
+    destroyHighChart() {
+        XH.safeDestroy(this._chart);
```
amcclain: This works because Highcharts has very nicely named their cleanup method `destroy()`, which is exactly what we expect / call in our XH method. Wondering though if we should be more explicit and call that method directly - we really intend the XH method to be called on other XH classes, and this might make it more clear that, yes, there really is a destroy method HC provides that we're calling appropriately.
```
if (this._chart) {
  this._chart.destroy();
  this._chart = null;
}
```

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+    getDefaultConfig() {
+        const exporting = {
+            enabled: false,
+            fallbackToExportServer: false,
+            chartOptions: {
+                scrollbar: {enabled: false}
+            },
+            buttons: {
+                contextButton: {
+                    menuItems: ['downloadPNG', 'downloadSVG', 'separator', 'downloadCSV']
```
amcclain: Where/when would these show up?

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+
+        return merge(defaultConf, themeConf, propsConf);
+    }
+
+    getDefaultConfig() {
+        const exporting = {
+            enabled: false,
+            fallbackToExportServer: false,
+            chartOptions: {
+                scrollbar: {enabled: false}
```
amcclain: Does this apply to maps?

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+            tooltip: {
+                enabled: !!tooltip,
+                useHTML: true,
+                padding: 0,
+                shape: 'square',
+                shadow: false,
+                pointFormatter: function() {
+                    if (!tooltip) return;
+                    const {record} = this;
+                    return tooltip == true ? defaultTooltip(record) : tooltip(record);
```
amcclain: Check for `isFunction(tooltip)` likely more clear here.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+    };
+
+    clickHandler(record, e) {
+        const {onClick, onDoubleClick} = this.model;
+        if (onClick && this._clickCount === 1) {
+            onClick(record, e);
+        } else if (onDoubleClick && this._clickCount === 2) {
+            onDoubleClick(record, e);
+        }
+        this._clickCount = 0;
```
amcclain: I would try/finally here to ensure we're always resetting our internal _clickCount as intended, even if app-provided callback were to throw.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+        this._clickCount++;
+        this._debouncedClickHandler(e.point.record, e);
+        if (this._clickCount >= 2) this._debouncedClickHandler.flush();
+    };
+
+    clickHandler(record, e) {
+        const {onClick, onDoubleClick} = this.model;
+        if (onClick && this._clickCount === 1) {
+            onClick(record, e);
+        } else if (onDoubleClick && this._clickCount === 2) {
```
amcclain: I dunno exactly how flushing the debounced is implemented and if we can be sure that's gonna always be synchronous and run this right after click #2 - even if so, do we need to check clickCount == 2 here or should we always call the dblClick handler?

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+
+        const {selectedIds, maxDepth, gridModel} = this.model;
+        let toSelect = [...selectedIds];
+
+        // Fallback to parent node if selection exceeds max depth
+        if (maxDepth && gridModel && gridModel.treeMode) {
+            toSelect = selectedIds.map(id => {
+                const delim = Cube.RECORD_ID_DELIMITER,
+                    parts = id.split(delim),
+                    idDepth = parts.filter(p => p !== 'root').length,
```
amcclain: We always have a root node, right? We could use `idDepth = parts.length - 1` and skip the need for the filter and creation of an interstitial array on this call within a loop.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+    //----------------------
+    // Selection handling
+    //----------------------
+    syncSelection() {
+        if (!this._chart) return;
+
+        const {selectedIds, maxDepth, gridModel} = this.model;
+        let toSelect = [...selectedIds];
+
+        // Fallback to parent node if selection exceeds max depth
```
amcclain: Interesting question if we want to do this....or would it be confusing / set up a bit of a struggle between the select representation across grid and tree....  But I can see why having no selection would be confusing as well, and how selecting the parent would be a natural way to keep the user oriented in the tree relative to the deeper grid selection....

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+                    idDepth = parts.filter(p => p !== 'root').length,
+                    toRemove = idDepth - maxDepth;
+
+                return toRemove < 1 ? id : parts.slice(0, -toRemove).join(delim);
+            });
+        }
+
+        // Update selection in chart
+        this._chart.series[0].data.forEach(node => {
+            node.select(toSelect.includes(node.id), true);
```
amcclain: I always wonder if it would be a worthwhile optimization to get the selected list into a Set and check via `has()` over `includes()`.  It might be. Again, being more paranoid than usual with these nested collection operations.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+                const buffer = 10,
+                    tooSmallWidth = (node.dataLabel.width + buffer) > node.graphic.element.width.baseVal.value,
+                    tooSmallHeight = (node.dataLabel.height + buffer) > node.graphic.element.height.baseVal.value,
+                    style = tooSmallWidth || tooSmallHeight ? {visibility: 'hidden'} : {visibility: 'visible'};
+
+                node.update({dataLabels: {style}}, false, false);
+            }
+        });
+
+        this._chart.redraw();
```
amcclain: Would like to profile this to understand the price we're paying for the (admittedly very nice, probably essential) behavior here.  Could there be a way to do with CSS?

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,297 @@
...
+            heat = record[heatField];
+
+        return `
+            <div class='xh-treemap-tooltip'>
+                <div class='xh-treemap-tooltip__label'>${name}</div>
+                <div class='xh-treemap-tooltip__row'>
+                    <div>${valueFieldLabel || valueField}:</div>
+                    <div>${fmtNumber(value)}</div>
+                </div>
+                <div class='xh-treemap-tooltip__row' ${valueField == heatField ? 'style="display:none"' : ''}>
```
amcclain: Would avoid writing this out at all (including the call to fmtNumber) in the no-op case where we won't be displaying a second row.

desktop/cmp/treemap/TreeMap.scss
```diff
@@ -0,0 +1,48 @@
+.xh-treemap {
+  text {
+    font-family: var(--xh-font-family) !important;
+    font-size: var(--xh-font-size-small-px) !important;
```
amcclain: Any (easy-ish) way to override with a more specific selector and avoid !important?  Thinking it might be something an app would want to override.

Another option is to go ahead and setup the CSS vars for these labels.  We might want to do a pass though that takes the chart into consideration as well, and I'm fine if we're not really ready for that.

desktop/cmp/treemap/TreeMap.scss
```diff
@@ -0,0 +1,48 @@
...
+
+    &.highcharts-point-select {
+      outline-width: 3px;
+      outline-offset: -3px;
+      outline-color: var(--xh-blue-light);
+    }
+  }
+}
+
+.xh-treemap-tooltip {
```
amcclain: These look very nice by default. ✨👍

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,295 @@
...
+     * @param {string} c.heatField - Record field to use to determine node color.
+     * @param {string} [c.valueFieldLabel] - Label for valueField to render in the default tooltip.
+     * @param {string} [c.heatFieldLabel] - Label for heatField to render in the default tooltip.
+     * @param {number} [c.maxDepth] - Maximum tree depth to render.
+     * @parma {function} [c.filter] - A filter function used when processing data. Receives (record), returns boolean.
+     * @param {function} [c.onClick] - Callback to call when a node is clicked. Receives (record, e).
+     *      If not provided, by default will select a record when using a GridModel.
+     * @param {function} [c.onDoubleClick] - Callback to call when a node is double clicked. Receives (record, e).
+     *      If not provided, by default will expand / collapse a record when using a GridModel.
+     * @param {string} [c.algorithm] - Layout algorithm to use. Either 'sliceAndDice', 'stripes', 'squarified' or 'strip'.
```
amcclain: I suggest we provide a link to https://www.highcharts.com/docs/chart-and-series-types/treemap in the top-level class comment, and could refer developers there in this doc string - otherwise these end up being pretty mysterious.  Would also note default of squarified.

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,295 @@
...
+            throwIf(isNil(name), `TreeMap labelField '${labelField}' not found for record ${id}`);
+            throwIf(isNil(value), `TreeMap valueField '${valueField}' not found for record ${id}`);
+            throwIf(isNil(colorValue), `TreeMap heatField '${heatField}' not found for record ${id}`);
+
+            // Create TreeMapRecord
+            const treeRec = {
+                id,
+                record,
+                name,
+                colorValue,
```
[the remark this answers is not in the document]
TomTirapani: This is the data representation intended to be passed directly to Highcharts... `colorValue` is the name that Highcharts needs: https://api.highcharts.com/highcharts/series.treemap.data.colorValue

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,295 @@
...
+        } else {
+            set(expandState, path, true);
+        }
+
+        gridModel.agGridModel.setExpandState(expandState);
+        gridModel.noteAgExpandStateChange();
+    }
+
+    getNodePath(id) {
+        const delim = Cube.RECORD_ID_DELIMITER,
```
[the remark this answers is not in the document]
TomTirapani: Thanks for catching - I had a misunderstanding, and thought that our tree ids would reliably be using this delimiter for drilldown ids. However, upon a closer look, it seems this is something we are doing in Toolbox's PositionService.groovy, and therefore can't be relied on!

Will need to come up with a different way for TreeMap to determine a record's place in the tree - presumably can use xhTreePath for this...

desktop/cmp/treemap/TreeMapModel.js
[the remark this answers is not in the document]
TomTirapani: We could consider renaming `heatField` to `colorField`, if that seems clearer?

desktop/cmp/treemap/TreeMapModel.js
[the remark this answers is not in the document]
TomTirapani: I've updated the 2 places we were using the Cube.RECORD_ID_DELIMITER to use the record's xhTreePath instead - led to much cleaner (and more robust) code. Thanks again for catching!

cmp/ag-grid/AgGridModel.js
```diff
@@ -313,9 +313,16 @@ export class AgGridModel {
...
-            if (!node.group) return;
+            if (!node.allChildrenCount) return;
 
             if (node.expanded) {
+                // Iterate up parents, ensuring each is also expanded
+                let parent = node.parent;
+                while (parent && parent.id !== 'ROOT_NODE_ID') {
+                    if (!parent.expanded) return;
+                    parent = parent.parent;
+                }
```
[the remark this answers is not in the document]
TomTirapani: Thanks for raising the performance consideration. Realised we can use the already partially constructed expandState to check if the parent node is already collapsed, and avoid the extra loop here: https://github.com/exhi/hoist-react/pull/1280/commits/ada5daa77323cfb00b7f4a812e48d4c17499e762

cmp/ag-grid/AgGridModel.js
```diff
@@ -313,9 +313,16 @@ export class AgGridModel {
 
         const expandState = {};
         this.agApi.forEachNode(node => {
-            if (!node.group) return;
+            if (!node.allChildrenCount) return;
```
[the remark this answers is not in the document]
TomTirapani: Yes - `node.group` is true for row groups, but not true for treegrid groups.

cmp/grid/GridModel.js
TomTirapani: `onRowGroupOpened` also triggers on collapse (a bit of a misnomer). I've also since added code to update expandState on data change, since that could lead to stale expandState. The data change reaction should cover the initial setup

amcclain:
Hi @TomTirapani  - looking forwards to catching up on all the latest changes today - thanks for all the ongoing work here.

Anything that you know is clearly outstanding / needs an extra careful look?

TomTirapani:
Not that I can think of.

Tried Highcharts boost but couldn't see any discernible effect.

Also have a local branch trying to use chroma.js and colorbrewer scales, but couldn't quite get it to work and complicated it significantly...

desktop/cmp/treemap/SplitTreeMap.js
```diff
@@ -0,0 +1,88 @@
...
+        let pFlex = 1, sFlex = 1;
+        if (pTotal && sTotal) {
+            pFlex = pTotal / sTotal;
+        } else if (pTotal && !sTotal) {
+            sFlex = 0;
+        } else if (!pTotal && sTotal) {
+            pFlex = 0;
+        }
+
+        return [
```
amcclain: I think we should add some logic here to either round pFlex or compare it against a last-rendered pFlex value to cut down on unwanted jitter and (relatively expensive) resize events.

Currently we're calculating the ratio of the two with full precision based on the totals. This means that even tiny changes to the P&L on one or both sides will cause a resize, even when it's going to be imperceptible. Log the pFlex values and I think this should be clear.

Rounding pFlex to e.g. two dps would likely cut down on a lot of jitter when you had relatively well-balanced maps (i.e. pFlex was close to one).  Comparing it against the last calculated value and simply using that last value when they are e.g. <2% apart or something like that might be a bit more robust.

Here's what the profile looks like with the main data reaction / re-render followed by the debounced resize. I think we can avoid that last big block of work in many cases.

Screen Shot 2019-08-13 at 2 44 09 PM

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,317 @@
...
+                const buffer = 10,
+                    tooSmallWidth = (node.dataLabel.width + buffer) > node.graphic.element.width.baseVal.value,
+                    tooSmallHeight = (node.dataLabel.height + buffer) > node.graphic.element.height.baseVal.value,
+                    style = tooSmallWidth || tooSmallHeight ? {visibility: 'hidden'} : {visibility: 'visible'};
+
+                node.update({dataLabels: {style}}, false, false);
+            }
+        });
+
+        if (this._chart) this._chart.redraw();
```
amcclain: Here's another place where I think we could be a bit more defensive to avoid extra calls to `redraw()`. 

Suggest we check to see if we actually need to make these updates - looks like the node's current visibility can be accessed via `node.dataLabel.styles.visibility`.  If we need to change, we can call `node.update()` and mark a flag set outside the loop to indicate that the loop actually did something. We can then check that before calling `redraw()`.

With a bit of logging and profiling it appears as though this will save us real-world work for some updates.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,317 @@
...
+    }
+
+    //----------------------
+    // Labels
+    //----------------------
+    async updateLabelVisibilityAsync() {
+        if (!this._chart) return;
+
+        // Show / hide labels by comparing label size to node size
+        await forEachAsync(this._chart.series[0].data, node => {
```
amcclain: I don't think we're getting anything from `forEachAsync()` here. Since we need to cap the series length anyway, these are not giant iterations.  Adding a debug tag for logging showed no waits, meaning all this is doing is adding an extra async layer without actually pausing to unblock.  I would remove in this case to simplify.

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,396 @@
...
+        // Select nodes in grid
+        const {selModel} = this.gridModel;
+        if (selModel.mode === 'disabled') return;
+
+        const multiSelect = selModel.mode === 'multiple' && e.shiftKey;
+        selModel.select(record, !multiSelect);
+    };
+
+    defaultOnDoubleClick = (record) => {
+        if (!this.gridModel || !this.gridModel.treeMode || !record.children) return;
```
amcclain: Looks like `children` is an empty array for leaf nodes. Let's check `isEmpty(record.children)` instead to avoid triggering downstream work for the gridModel, etc.

desktop/cmp/treemap/TreeMap.js
```diff
@@ -0,0 +1,317 @@
...
+            className: 'xh-treemap__chart-holder',
+            ref: this._chartElem.ref
+        });
+    }
+
+    //-------------------
+    // Implementation
+    //-------------------
+    renderHighChart() {
+        this.destroyHighChart();
```
amcclain: To ask the question....should we really be destroying the chart on each render cycle? 

I did a bit of experimenting with calling `chart.update()` with the same config (series included) if `_chart` was already defined. That seemed to work and would presumably be more efficient / a better practice if it doesn't have negative side effects...  

Then again, maybe there's no advantage here.  I would be curious to see if there is any broader discussion re. HC and React / render lifecycle.

data/Store.js
```diff
@@ -406,7 +392,7 @@ export class Store {
     }
 
     getRootSummary(rawData) {
-        return this._loadRootAsSummary && rawData.length === 1 && !isEmpty(rawData[0].children) ?
```
haynesjm42: I removed this as it would cause issues (duplicate row added to grid) if the app provided update data with root summary data which had no children (so only the summary data changed). Don't think this extra condition is necessary.

desktop/cmp/treemap/SplitTreeMapModel.js
```diff
@@ -0,0 +1,85 @@
...
+    //------------------------
+    /** @member {Object} */
+    @bindable.ref highchartsConfig = {};
+
+    /**
+     * @param {Object} c - SplitTreeMapModel configuration.
+     * @param {function} c.mapFilter - A filter function used when processing data. Receives (record), returns boolean.
+     *      Records that pass the filter will be placed into the primary TreeMap, and the rest into the secondary TreeMap.
+     *      If not passed, will default to: { return record.valueField >= 0; }
+     * @param {function} [c.mapTitleFn] - Function to render map titles. Receives map name ['primary', 'secondary'] and region TreeMapModel.
```
amcclain: Will this be more convenient if we passed in an `isPrimary` boolean for the first argument?  I think so - would avoid the need to check for that string, and I think the nature of SplitTreeMap means we're gonna keep having two maps so a boolean would work.

Don't feel that strongly about it, but wanted to get that suggestion in.

desktop/cmp/panel/impl/PanelRefreshContextModel.js
```diff
@@ -0,0 +1,51 @@
...
+ */
+import {RefreshContextModel} from '@xh/hoist/core/refresh';
+import {TabRefreshMode} from '@xh/hoist/enums';
+import {loadAllAsync} from '@xh/hoist/core';
+
+/**
+ * @private
+ */
+@RefreshContextModel
+export class PanelRefreshContextModel {
```
amcclain: As discussed on Slack, please remove this new class from the branch - but great that it is on its way! 😄

desktop/cmp/treemap/SplitTreeMapModel.js
lbwexler: We like this suggestion and will make this change now.

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,405 @@
...
+        return field ? field.label : this.valueField;
+    }
+
+    @computed
+    get heatFieldLabel() {
+        const field = this.store.fields.find(it => it.name === this.heatField);
+        return field ? field.label : this.heatField;
+    }
+
+    @computed
```
lbwexler: do we get anything out of the computed for selectedIds and expandState?  Or is this just for consistentcy?

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,405 @@
...
+        if (!this.gridModel || !this.gridModel.treeMode) return {};
+        return this.gridModel.expandState;
+    }
+
+    @computed
+    get hasData() {
+        return !isEmpty(this.data);
+    }
+
+    @computed
```
lbwexler: This seems like an interesting way to recognize an 'error'.  Wondering about the name of this? Would like to discuss at some point

desktop/cmp/treemap/TreeMapModel.js
amcclain: I know we've discussed this before, how if the getter is a trivial transform of an observable the @computed doesn't add that much in terms of efficiency. And that the reaction / component render will detect any observables accessed during its run (however transitively?) and respond to their changes automatically.

That said, I still favor this setup to make it explicit that these are observable getters, without needing to look at their contents too closely to work out if their internals are observables. 

The consistency also seems helpful, and avoids judgement calls around "this implementation used to be trivial. but now has gotten a bit more complicated, should it get the decorator now?"

desktop/cmp/treemap/TreeMapModel.js
amcclain: I had the same thought - definitely an interesting pattern for these components we like to write where they have an error panel that will get rendered if the entire thing is generally not in a working state for whatever reason.  (Something I feel like we could build into more components in a standard way...)

Thought had crossed my mind that `errors` or `errorMessages` might be a bit of a better name and have it always return an array. We could even author a small `ErrorMessages` component that took that array and rendered a list with a standard CSS class...  (Could imagine that getting fancier in some future iteration to truncate messages, optionally disclose exception stacks, etc)

desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -0,0 +1,405 @@
...
+            track: () => [
+                this.store.rootRecords,
+                this.expandState,
+                this.colorMode,
+                this.labelField,
+                this.valueField,
+                this.heatField,
+                this.maxDepth
+            ],
+            run: ([rawData]) => this.data = this.processData(rawData)
```
lbwexler: So interestingly enough, with this setup, we are going to be reprocessing all the data even if we are in error mode?  Its more of a theoretical concern, but does seem wrong that we would still prepare a 5 million node store that we would ultimately not render.  What if our "error" logic was protecting us from other problems that would cause our computation to throw?

A related concern is that we are actually recomputing the color scale whenever we change the expand state.  This seems wrong to me as well....

lbwexler:
So I guess I do have a real concern about how dynamic our coloring scheme is.  It definitely looks nice, but it seems like their are a number of problems:

1) Split TreeMap always needs to use equivalent ranges for both sides.  Right now they are independent, and thats just wrong.  SplitTreeMap's whole point is promising a single integrated maps where analytics are comparable on both sides.
2) I think we generally want the application to set the range that makes sense for the domain - we have seen this in an early client app -- where the viewing of  p&l/risk ratios should be comparable across days.

lbwexler:
Let's deal with the colorization details in another round.
This looks great.  Merging now!

PR #1283, opened 2019-07-28
Replace RSVP-based allSettled() with standard version on Promise.

lbwexler:
Hoist P/R Checklist
-------------------

- [N/A ] Created Toolbox branch / PR (link provided here, or N/A)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,8 +45,8 @@
     "react-transition-group": "~4.2.1",
     "router5": "~7.0.2",
     "router5-plugin-browser": "~7.0.2",
-    "rsvp": "~4.8.4",
-    "store2": "~2.7.0"
+    "store2": "~2.7.0",
+    "core-js": "^3.1.4"
   },
   "peerDependencies": {
     "react": "~16.8.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -45,8 +45,8 @@
     "react-transition-group": "~4.2.1",
     "router5": "~7.0.2",
     "router5-plugin-browser": "~7.0.2",
-    "rsvp": "~4.8.4",
-    "store2": "~2.7.0"
+    "store2": "~2.7.0",
+    "core-js": "^3.1.4"
```
amcclain: Would like to discuss this! We should review in the context of our hoist-dev-utils and the build warnings we've been getting re. the existing `babel-polyfills` being deprecated.  We're supposed to be replacing these with core-js 3 and need to update https://github.com/exhi/hoist-dev-utils/blob/634e4df4b49c5e5212571b031d325939bf5661d4/configureWebpack.js#L154 and potentially other related areas.

Then we should not need to have imports of individual polys like this one as we should be able to configure to handle any usages based on target browsers.

amcclain: See note at top of https://babeljs.io/docs/en/babel-polyfill for the deprecation warning we get at build now.

amcclain:
Closing - included in #1350

Issue #1284, opened 2019-07-29
AutoFocus prop for button + primary dialog button


amcclain:
In general, our button() component would benefit from an autoFocus prop to attempt to grab focus as soon as rendered.  

Similarly, the `XH.message()` dialog methods could take a config to optionally autoFocus the confirm button so the keyboard user can immediately press enter or space to dismiss.

PR #1288, opened 2019-07-30
Allow autoFocus of button, and of input, cancel, or primary button in messages and prompts. (#1284)

cnrudd:
Hoist P/R Checklist
-------------------

- [N/A] Checked for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [x] Reviewed and tested on Mobile
- [x] Created Toolbox branch / PR [https://github.com/exhi/toolbox/pull/238]((https://github.com/exhi/toolbox/pull/238))

This is a non-breaking change that some users would like in a hurry.  Would be good if it could be added to framework in feature release 25.3

This change allows autofocus on any button.
In message, alert, and prompt dialogs, autofocus can put on the input, cancel, or confirm components.

This change also affects the outline that appears on non-input focused components - ie: buttons.  The outline is thinner and not offset as much.

### DIFF
--- a/desktop/appcontainer/Message.js
+++ b/desktop/appcontainer/Message.js
@@ -5,6 +5,7 @@
  * Copyright © 2019 Extremely Heavy Industries Inc.
  */
 import {Component} from 'react';
+import {defaults} from 'lodash';
 
 import {dialog, dialogBody} from '@xh/hoist/kit/blueprint';
 import {HoistComponent, elemFactory} from '@xh/hoist/core';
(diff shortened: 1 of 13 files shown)


desktop/appcontainer/Message.js
```diff
@@ -53,32 +55,37 @@ export class Message extends Component {
...
+        const {formModel, input, autoFocus} = this.model;
         if (!formModel) return null;
         return form({
             model: formModel,
             fieldDefaults: {commitOnChange: true, minimal: true, label: null},
             item: formField({
                 field: 'value',
-                item: withDefault(input.item, textInput({autoFocus: true}))
+                item: withDefault(input.item, textInput({
+                    autoFocus:  autoFocus == 'input' ? true : false
```
amcclain: Why the ternary operator? Could simplify to just `autoFocus == 'input'` - same below in `renderButtons()`.

Here its a bit tricky as any custom input (e.g. a numberInput) won't get focused. That puts me on the fence with supporting input as an option, although I can certainly see the use.  We could clone the element if provided and install as an override - that might be reasonable here.

desktop/appcontainer/Message.js
```diff
@@ -53,32 +55,37 @@ export class Message extends Component {
...
                 intent: cancelIntent,
+                autoFocus:  autoFocus == 'cancel' ? true : false,
                 onClick: () => this.model.doCancel()
             }),
             button({
                 text: confirmText,
                 intent: confirmIntent,
                 disabled: formModel ? !formModel.isValid : false,
+                autoFocus:  autoFocus == 'confirm' ? true : false,
+                // minimal: false,
```
amcclain: I too _sometimes_ want a non-minimal button here...  Makes me wonder if we should have made the config `confirmProps` instead of calling out text and intent...  (And used the same for cancel of course.)

If we did that, we could lose the autoFocus config for messages and simply pass in autoFocus to either button as a prop. Let's discuss quickly tomorrow to see if that feels worth it.  I think it might be - avoids the issue with input above + having the new config with string values vs. true/false as seen elsewhere.

Given that these are pretty commonly used methods, we could look for the dedicated text/intent configs - if found, log a deprecation warning before applying them into the new object-based configs automatically.  That would let us make this change while keeping things non-breaking

desktop/cmp/button/Button.js
```diff
@@ -26,26 +29,29 @@ export class Button extends Component {
...
+            autoFocus,
 
             style: {
                 ...style,
                 ...this.getLayoutProps()
             },
 
             ...rest,
-            className: this.getClassName(minimal ? 'xh-button--minimal' : '')
+            className: this.getClassName(minimal ? `xh-button--minimal${autoFocusClassName}` : autoFocusClassName)
```
amcclain: `getClassName` is setup to accept an arbitrary number of args, some of which can be empty/null, and will call the `classNames` util to concat them safely into a string.  So no need to concat strings here.

amcclain:
Also - before I forget - once we settle on an approach for the messages, would like to use autoFocus for IdleDialog and standard exception dialogs where you need to reload the app.  Those were some use cases I had come across where I wished I had exactly this feature.

desktop/cmp/button/Button.js
cnrudd: Cleaned up.

cnrudd:
I've put in some new commits that switch to using the confirmProps, canceProps and show a deprecation warning.

I switched to using defaultsDeep in MessageSourceModel, and, interestingly, i'm now getting a React key/array warning the first time a dialog loads.  Not sure what is causing this issue.  

But the switch is in, and ready for more review.

amcclain:
Great thank you Colin - will be sure to take a look later this afternoon when I am able and will check out that specific error.  I am happy to look at this with the latest updates. 👍

Issue #1290, opened 2019-07-31
Restore support for direct usage of ag-Grid pinned row data apis


haynesjm42:
Since summary row support was added we no longer support applications using the ag-Grid pinned row apis directly, as we overwrite anything the application may have set. When updating the pinned summary row data we should only add/remove our summary row data and leave any data that was set by the application as-is.

PR #1291, opened 2019-08-01
Grid summary/pinned data fixes (Fixes #1289, #1290)

### LINKED ISSUE #1289: Store summaryData cannot be updated without triggering reactions on records
Store.updateData will always update its RecordSets even if the updated data is empty. This is especially problematic for apps which want to update the summary data when the records change (due to filtering), which results in mobx errors because of unstable reactions.

haynesjm42:
Hoist P/R Checklist
-------------------

Restores support for application usage of the ag-Grid pinned data apis and fixes potential issuers where an application might want to update summary data from a reaction on the Store `records` property.

We now flag the summary record with `xhIsSummary=true` so the Grid can identify it when updating the pinned row data if the summary row is enabled. Grid will extract it from the existing pinned row data and place it back in the appropriate place (top-most for top pinned, bottom-most for bottom pinned). I also moved updating of the pinned data to its own reaction instead of using the data reaction as it was unnecessary to do all the extra work that is done in data reaction and ti seemed like a valid/common use-case where applications would want to update the summary data whenever the Store filter is applied to the `records`, which could/would result in an extra data reaction after the summary data was updated.

In `Store.updateData()` we also now avoid updating our RecordSet references and `lastUpdated` property unless the supplied `rawData` is not empty. This will allow applications to update the summary data with a call like `store.updateData([], newSummaryData)` without causing an infinite loop of reactions.

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -140,6 +140,7 @@ export class Grid extends Component {
         this.addReaction(this.columnStateReaction());
         this.addReaction(this.dataReaction());
         this.addReaction(this.groupReaction());
+        this.addReaction(this.summaryReaction());
 
         this.agOptions = merge(this.createDefaultAgOptions(), props.agOptions || {});
     }
@@ -341,12 +342,25 @@ export class Grid extends Component {
     //------------------------
     // Reactions to model
     //------------------------
+    summaryReaction() {
+        const {model} = this,
+            {agGridModel, store} = model;
+
+        return {
+            track: () => [agGridModel.agApi, store.summaryRecord, model.showSummary],
+            run: ([api]) => {
+                if (!api) return;
+                this.updatePinnedSummaryRowData();
+            }
+        };
+    }
+
     dataReaction() {
         const {model} = this,
             {agGridModel, store} = model;
 
         return {
-            track: () => [agGridModel.agApi, store.records, store.lastUpdated, model.showSummary],
+            track: () => [agGridModel.agApi, store.records, store.lastUpdated],
             run: ([api, records]) => {
                 if (!api) return;
 
@@ -358,7 +372,6 @@ export class Grid extends Component {
 
                         // Load updated data into the grid.
                         api.setRowData(records);
-                        this.updatePinnedRowData();
 
                         // Size columns to account for scrollbar show/hide due to row count change.
                         api.sizeColumnsToFit();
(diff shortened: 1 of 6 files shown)


amcclain:
Thanks, John - I will be sure to review this later this afternoon and we can get it in soon so @johnbc can move forwards with his use-case.

cmp/grid/Grid.js
```diff
@@ -518,20 +531,24 @@ export class Grid extends Component {
...
+            pinnedTopRowData = agGridModel.getPinnedTopRowData().filter(filterSummaryFn),
+            pinnedBottomRowData = agGridModel.getPinnedBottomRowData().filter(filterSummaryFn);
 
         if (showSummary && store.summaryRecord) {
-            const arr = (showSummary === 'bottom') ? pinnedBottomRecords : pinnedTopRecords;
-            arr.push(store.summaryRecord);
+            if (showSummary === 'bottom') {
+                pinnedBottomRowData.push(store.summaryRecord);
+            } else {
+                pinnedTopRowData.splice(0, 0, store.summaryRecord);
```
amcclain: Just use `unshift()` here?

lbwexler:
So john and I discussed many of the same items (including unshift  :-)) earlier on friday.  I usually mention when we do a verbal code review to avoid any double work -- will be more careful.  In this case, the more eyes the better and we certainly missed the issue about the isSummary() getter.   I am all in favor of not duplicating that state.

**I had originally told John to go ahead and remove all the extra code around skipping the change when we pass in updates = [] to updateData**.  But I now think it is important, perhaps for different reason than we had for starting this bit of the project.  Namely, have been seeing in a client app that running our full data reaction -- even when there are no new records -- is actually pretty expensive. We of course need to understand that, but  I think its a common use case, to do empty updates with a streaming api, so we will want to do this simple optimization.

Agree updateData should always change if this method does anything.  John and I also discussed that point

haynesjm42:
Thanks for the careful look guys, implemented the requested changes. Still don't know how I was completely ignorant of the existence of `unshift` ...

lbwexler:
Thanks John.

I agree with Anselm that we should now change that getter on record to look at the new flag we are setting!  Its both self-documenting, and enhances that getter for cases like this.  Once a summary, always a summary!

lbwexler:
ah -- I see John already whacked it.  OK.  I'll just merge -- we can bring it back if we ever need it!

lbwexler:
don't be so hard on yourself john -- I think I ignored it for a few years
myself -- with a name like that, it doesn't sound like the kind of method
you really need for writing applications.....  :-)

amcclain:
Thank you - this looks like it landed in a very good place.

Yes - the key bit about avoiding the full data reaction is key - even though it's resolved to what we would think is a no-op transaction, it is not free.

PR #1296, opened 2019-08-06
Calendar dates (V2)

TomTirapani:
This PR is builds upon the existing PR https://github.com/exhi/hoist-react/pull/1130

The idea was to provide a simple CalendarDate object that will let users manipulate CalendarDates on the client without requiring users to go back and forth between these, moment and JS Dates. The CalendarDate object provides many of the same manipulation / query options as moment dates.

At this stage just looking to see if this a direction we want to continue to pursue, whether we want to look for a third-party lib like (e.g. joda-time), or... something else. My hope is that by leveraging the existing moment methods, the CalendarDate object should be both familiar and stable to use.

See corresponding Toolbox PR: https://github.com/exhi/toolbox/pull/242

Hoist P/R Checklist
-------------------

- [ ] Up to date with `develop` branch as of last change.
- [ ] Ready for review (or add `wip` label).
- [ ] Checked for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [ ] Created Toolbox branch / PR (link provided here, or N/A)


utils/datetime/CalendarDate.js
```diff
@@ -0,0 +1,153 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2019 Extremely Heavy Industries Inc.
+ */
+
+import moment from 'moment';
+
```
lbwexler: This is a really cool idea to simply wrap the string.

desktop/cmp/input/CalendarDateInput.js
```diff
@@ -0,0 +1,61 @@
+/*
```
lbwexler: I wonder if we could avoid duplicating all of our DateInput work, and just add an option to the existing DateInput to get it to bind to a CalendarDate instead of a Date?

TomTirapani:
Thanks for taking a look Lee. Thanks also for the suggesting of adding an option to the existing DateInout to bind to CalendarDates - pushed an update to that effect.

Please let me know if you have time today to go over this discuss approach, any suggestions you  may have or alternative approaches to consider.

TomTirapani:
In a more general vein, how do you feel about the name CalendarDates? Open to suggestions for anything a bit shorter... could consider LocalDates, to mirror the server-side, if that wouldn't be too confusing?

lbwexler:
was thinking the same thing and was going to mention it to you in person!

lbwexler:
Just FYI I have a number of small changes in my sandbox that I will look to get in.  Mainly around using static constructors instead of new.

Really excited to see this get in.

PR #1299, opened 2019-08-07
Tighten up action button sizing/padding in compact grids

### LINKED ISSUE #1287: Action column buttons do not respect compact grid style
We do not use a smaller button for action columns when the Grid is in compact mode, which causes the buttons to overflow the cell (overflow is hidden) and the buttons just look a bit too large compared to the text in the grid. We should use a smaller size for the buttons when the grid is in compact mode.

amcclain:
+ Fixes #1287

Hoist P/R Checklist
-------------------

- [NA] Checked for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [NA] Updated doc comments / prop-types (or N/A)
- [NA] Created Toolbox branch / PR (link provided here, or N/A)

### DIFF
--- a/desktop/cmp/grid/columns/Actions.js
+++ b/desktop/cmp/grid/columns/Actions.js
@@ -68,6 +68,7 @@ export const actionCol = {
             if (disabled) actionButtonEl.setAttribute('disabled', 'true');
             if (!isEmpty(tooltip)) actionButtonEl.setAttribute('title', tooltip);
             if (!isEmpty(intent)) actionButtonEl.classList.add(`bp3-intent-${intent}`);
+            if (gridModel.compact) actionButtonEl.classList.add('xh-record-action-button--compact');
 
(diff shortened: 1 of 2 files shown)


amcclain:
Not promising this is as clean / dynamic as it could be, but wanted to get a fix in as I saw what you meant.  Take a look and let me know if you think this will work?

haynesjm42:
Styling looked good, but noticed it didn't update if compact was changed at runtime, presumably because it would require refreshing the cells via the ag-grid api to force a re-render. Changed things around to just use css to apply the new styling, and also added css to make the icons slightly smaller as well in compact mode which I think makes it look a little more consistent with the other grid text. Let me know what you think about the icon, and whether you think there are any issues with just doing this entirely in css?

amcclain:
Great catch thanks - didn't notice that somehow - the differences were pretty subtle but still definitely worth it.  Thanks for figuring out how to tweak the icon itself.  Merging now. 👍

PR #1308, opened 2019-08-13
Calendar dates v3.  Fixes #925

### LINKED ISSUE #925: Add client-side support for 'value dates'

lbwexler:
Hoist P/R Checklist
-------------------


lbwexler:
Hi -- Think this includes most everything we discussed. Was especially pleased to see that we could alter fetchService.  valueOf() works like a dream.

Note that I went with a `create` name for the main factory.  I think its much better than 'fromIsoString()', but agree this is still somewhat weird, given the memoization. Would be happy to bomb this to, say, `ValueDate.of()` if we think that would be  better (just thought of that.)

Will add a seperate ticket for the enhancement to store defaulting from columns, which I think is orthogonal to this change and should be in a seperate pr.

lbwexler:
should also note that I did not get to test the recalls app -- am running into an unrelated exception ssl exception there that I will look into tomorrow.  Let me know if you guys can try that...

amcclain:
I had checked on this branch at around 8:30 my time here in CA and thought we were done for the day, but I should have checked again a bit later. 😉  

Will take a look at the latest now and try out the recalls app as well.

lbwexler:
great -- am inclined to change LocalDate.create() to LocalDate.of()

Just noticed that this is the factory name on the java class so why not!

-Lee

Issue #1314, opened 2019-08-15
Chart should re-using existing visualization, avoid extra renders


amcclain:
In our TreeMap component, we have reworked to minimize re-renders, having the overall component render() the container (or placeholder/error messages) and then having dedicated more targeted reactions to either create/render the Highcharts viz or reload the existing viz with new series data.

This brought some meaningful performance benefits to TreeMap that should apply to Chart as well.

peterjcoffman:
I found something interesting. My current solution for this issue also solves this one:
https://github.com/xh/hoist-react/pull/1455

peterjcoffman:
WIP PR is here:

https://github.com/xh/hoist-react/pull/1505

This seems to get a decent performance improvement when reloading data, maybe a factor of two.

It also changes some behavior, which may not be desirable. As I mentioned, it will cause charts to keep their zoom level when refreshed. It also seems to prevent charts from displaying an animation when refreshing data, like the one at toolbox.xh.io/app/charts/line

lbwexler:
Great Job on this Peter  👍

PR #1329, opened 2019-08-26
Add ChartModel.highchartsChart - observable pointer to HC instance

### LINKED ISSUE #1326: ChartModel should expose the underlying highcharts component...
...or possibly a set of helpful highcharts API methods.

Original motivation was: I wanted to be able to call methods like setExtremes on charts in a client app in order to respond to a user clicking a range select button. 

Spoke to Lee shortly thereafter regarding the reactive model and how it would be more appropriate to change the state of the highchartsConfig in order to get the chart to render as desired if possible (rather than poking the _hChart).

However, even if that is possible, I need to query the _hChart (getExtremes) to find out if the user has manually zoomed via a drag action so that we can set the new visible range based on that.

At the end of the day, we should at very least be able to use the highcharts api to query the chart for useful info, if not to drive the chart itself. (I think)

amcclain:
+ Fixes #1326

Hoist P/R Checklist
-------------------

- [NA] Checked for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [x] Created Toolbox branch: https://github.com/exhi/toolbox/tree/chartInstanceAccess

### DIFF
--- a/desktop/cmp/chart/Chart.js
+++ b/desktop/cmp/chart/Chart.js
@@ -41,7 +41,7 @@ export class Chart extends Component {
         /** Primary component model instance. */
         model: PT.oneOfType([PT.instanceOf(ChartModel), PT.object]).isRequired
     };
-    
+
     static modelClass = ChartModel;
 
     baseClassName = 'xh-chart';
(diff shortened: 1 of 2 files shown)


amcclain:
That last error sounds like something that *might* change substantially with the refactor suggested in #1314 - where the goal is to avoid destroying and recreating the chart so eagerly....

amcclain:
I will close this and update #1326 with the change to save the reference on the model in general.  Then we can prioritize/schedule the work to tackle both tickets in one go.

PR #1374, opened 2019-09-25
+ Support for `onLoad` directive for 'mask', 'loadingIndicator'

lbwexler:
I am not sure about the name of this directive.  We could go with 'loadSupport'.  Wasn't sure if it would be a feature, or just confusing to use the same name as the actual model property.   For example, would users expect that other props could be bound in that way?  Would we do so in other cases?

One example I have  in mind is the somewhat symmetrical `ref: model.viewRef`.  Users *won't* be able to write `ref: 'viewRef'`.  (Note that ref is special in react, and get's pulled out of the  props, so we could not easily enable that syntax even if we wanted to.

Hoist P/R Checklist
-------------------


lbwexler:
Removed unrelated fix to Panel

lbwexler:
Reviewed and approved by Tom.  Will make some suggested changelog comments on develop

Issue #1365, opened 2019-09-19
ExceptionDialog fails to give indication that message was successfully sent


lbwexler:
Namely, if requireRefresh is `true`, pressing the button to submit feedback does nothing visible.  If requireRefresh is `false`, pressing the button immediately closes the dialog.

Very disconcerting! 

How about just always popup and alert just saying it was sent, and then clear the feedback text?

lbwexler:
Don -- should be a straightforward fix -- not high priority by any means, but would like to just clear if off the docket

lbwexler:
Fixed in PR.  Thanks

PR #1383, opened 2019-09-29
Harden exception report.  Fixes: #1365

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/ExceptionDialogModel.js
+++ b/appcontainer/ExceptionDialogModel.js
@@ -58,16 +58,26 @@ export class ExceptionDialogModel {
     setUserMessage(userMessage) {
         this.userMessage = userMessage;
     }
-    
+
     async sendReportAsync() {
         const {exception, userMessage, options} = this;
 
-        await XH.exceptionHandler.logOnServerAsync({
+        const success = await XH.exceptionHandler.logOnServerAsync({
             exception,
             userMessage,
             userAlerted: true
         });
 
-        if (!options.requireReload) this.close();
+        if (success) {
+            await XH.alert({message: 'Your message has been sent.'});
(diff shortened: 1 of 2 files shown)


lbwexler:
Don -- please review and merge if it seems reasonable to you.  
Ran into this again today, and just wanted to take it off the table for v28.

appcontainer/ExceptionDialogModel.js
```diff
@@ -58,16 +58,27 @@ export class ExceptionDialogModel {
...
-        await XH.exceptionHandler.logOnServerAsync({
+        const success = await XH.exceptionHandler.logOnServerAsync({
             exception,
             userMessage,
             userAlerted: true
         });
 
-        if (!options.requireReload) this.close();
+        if (success) {
+            await XH.alert({title: 'Message Sent', message: 'Your error has been reported.'});
```
febbraiod: Are we sure we don't want to just toast here? Seems like the user has probably already done enough that we don't need to require another click. Realize thats not symmetrical with the code below, but the case there seems to justify the alert.

lbwexler:
Yes, I thought about this as well, but was not sure about what kind of guarantees we could make about the health of our toast infrastructure in this situation (e.g. serious, early exception). 

 Figured since we were already depending on our dialog/alerting, best to avoid any additional dependencies on other parts of our infrastructure.....  Could go either way here....

lbwexler:
Thinking about it some more, Don, I think you are correct -- can you make the change you are suggesting....  Good to avoid being annoying in this situation when our users might already be annoyed....

lbwexler:
Don and I discussed at length and are going to move forward with the version as is.

PR #1384, opened 2019-09-30
Pass on layout and classname props to underlying component more cleanly.

lbwexler:
Hoist P/R Checklist
-------------------


lbwexler:
@atm -- had been thinking that the pattern in this P/R (and related toolbox P/R) is generally cleaner.  Both in terms of how we pass on certain props, and how we declare the props that we *do* want  

For this mini-bombing, just changed the easy ones, to get this pattern in place

amcclain:
Yes, I agree this looks more clear in general and seems like a good way to go. 

Let's discuss whenever you have a chance this AM. I would like to recall why we moved away from this in other contexts (favoring `withDefault()`) just to ensure I'm not missing something.

Would also like to consider zipping through and getting proper prop-types in place for the core layout components that we then mix in when we are doing this.  I think that would also help make it more clear what props are being passed on to what.

PR #1391, opened 2019-10-02
Support for updating a single Record, partial Record data updates (Fixes #1376)

### LINKED ISSUE #1376: Need Record Method to more easily do client-side update.
As far as I can tell, you actually have to write the following code to update a record appropriately on the client:

rec.store.updateData(update:[{id: rec.id, ...rec.raw, fooField: 'newValue'}]);

yikes!

This relates to 
https://github.com/xh/hoist-react/issues/1321

haynesjm42:
Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (https://github.com/xh/toolbox/pull/277)

-----

Added the new method `updateRecordData` to Store. Also made changes to the update object handling to allow for partial data updates so callers only need to include fields which have changed in the update objects. This cleaned up the toolbox tree example considerably.


data/Store.js
```diff
@@ -122,9 +122,13 @@ export class Store {
         throwIf(!isEmpty(other), 'Unknown argument(s) passed to updateData().');
 
         // 1) Pre-process updates and adds into Records
+        const {summaryRecord} = this;
         let updateRecs, addRecs;
         if (update) {
-            updateRecs = update.map(it => this.createRecord(it));
+            updateRecs = update.map(it => {
+                const rec = it.id === summaryRecord?.id ? summaryRecord : this.getById(it.id);
```
amcclain: I might suggest getting `summaryRecord?.id` once before entering the loop - doubt it has a measurable impact, but still seems nicer to do the nullsafe lookup once and then compare to a local cost within the loop itself.

That said, should `getById()` instead be modified to always return the summaryRecord if there is one on the store and the asked-for ID is the summary ID?  Why not?

haynesjm42: I remember almost adding that to `getById()` but not sure why I decided against it, perhaps just to minimize the changes to commonly-used Store methods as I was originally thinking this would be a late merge into v28. Will go ahead and add it, can't think of any reason not to..

haynesjm42:
I believe @lbwexler had some concerns about the partial data updates and whether this change addresses the common use-cases we've run into, so probably want to hold off on merging until we have a chance to discuss further

lbwexler:
yes, would love to discuss with you tomorrow, John.  Thanks

data/Store.js
```diff
@@ -297,6 +297,8 @@ export class Store {
      * @return {Record}
      */
     getById(id, fromFiltered = false) {
+        if (id == this.summaryRecord?.id) return this.summaryRecord;
```
amcclain: Thanks for mixing this in.  Here's one case where I would recommend `===` for some extra safety.  Not sure that we would h😉

PR #1393, opened 2019-10-03
Support for dynamic column and column group header names (Fixes #1389)

### LINKED ISSUE #1389: Support for dynamic column header names
ag-Grid supports dynamic header names via the `headerValueGetter` props. In order to get the header name to update, would need to call `refreshHeader()` on the ag-Grid api. Ideal implementation would accept a function for our `headerName` prop on Column, which would add an autorun on GridModel to automatically call `refreshHeader` on the ag-Grid api as needed. Simpler approach if autorun is not feasible would be to add a `refreshHeaders` to GridModel (or maybe a `refresh` with some config to refresh headers, columns, etc. to wrap up the ag-Grid apis nicely)

haynesjm42:
Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (https://github.com/xh/toolbox/tree/dynamicColumnHeaderNames - example/testing only - not intended to be merged)

-------------------------

Added support for accepting a function for `headerName` in Column and ColumnGroup.

This function will be called by the ColumnHeader and (new) ColumnGroupHeader components while rendering, so changes to any observable properties referenced by the `headerName` function will automatically cause the header to re-render (only the component - no ag-Grid refreshing going on at all).

The caveat to this approach is that we are ignoring the `displayName` which ag-Grid provides to the header component via props when our Column `headerName` is a function. It is not clear if/when this would differ from whatever was set on `headerName` or returned via the `headerValueGetter` on the column definition - couldn't think of anything we typically do with grids where ag-Grid is appending any additional text to the header names we specify. 

This change also necessitated a breaking change to ColumnGroup - we previously did not require the `groupId` and would default it to the `headerName` (we required that you specify at least one of those). Because `headerName` can now be a function, this didn't really work any longer, and was also inconsistent with Column. I made a change so `groupId` is now required (`headerName` was already defaulting to `startCase(groupId)` and does not have a default. App upgrade will just require changing `headerName` to `groupId` for ColumnGroup specs.

ColumnGroupHeader component has been added, and I made sure to support the column group expand/collapse feature of ag-Grid in case apps are using that via `agOptions`. Decided not to go down the rabbit hole of fully supporting that with our Column/GridModel, will save that for when/if we know apps are using it.

#1389


cmp/grid/columns/Column.js
```diff
@@ -27,7 +27,12 @@ export class Column {
      * @param {string} [c.field] - name of data store field to display within the column.
      * @param {string} [c.colId] - unique identifier for the Column within its grid.
      *      Defaults to field name - one of these two properties must be specified.
-     * @param {string} [c.headerName] - display text for grid header.
+     * @param {(Column~headerNameFn|string)} [c.headerName] - display text for grid header. Supports
+     *      both a string value or a function to generate a string. Note that using a function here
+     *      will ignore any ag-Grid functionality for decorating the header name, the return value
+     *      of the function will be used as-is.
+     *      The function should be treated like an autorun - any observable properties referenced
```
lbwexler: "**any subsequent changes** to observable properties....

haynesjm42:
Yes we definitely could make this a non-breaking change, the code would be a bit awkward, but certainly doable. 

With the id/name defaulting breaking change `ColumnGroup` is now more consistent with how `Column` works, but I understand that for groups you usually don't care about the id since we don't deal with column group state at all (yet)

cmp/grid/impl/ColumnGroupHeader.js
```diff
@@ -0,0 +1,72 @@
...
+                e.stopPropagation();
+                props.setExpanded(!impl.isExpanded);
+            }
+        });
+    }
+});
+
+@HoistModel
+class LocalModel {
+    /** @member {ColumnGroup} */
```
lbwexler: would not use this jsdoc for private class

cmp/grid/columns/ColumnGroup.js
```diff
@@ -16,8 +16,13 @@ import {startCase, isEmpty, castArray, clone} from 'lodash';
 export class ColumnGroup {
     /**
      * @param {Object} c - ColumnGroup configuration.
-     * @param {string} [c.groupId] - unique identifier for the ColumnGroup within its grid.
-     * @param {string} [c.headerName] - display text for grid header.
+     * @param {string} c.groupId - unique identifier for the ColumnGroup within its grid.
+     * @param {ColumnGroup~headerNameFn|string} [c.headerName] - display text for column group header.
```
lbwexler: Sorry to be a sticker, but think that if we have a defined type in this file, all this text describing the function can be included in the docstring for that type.  The only description here should be 'display text' or 'display text for this header'

Also, could generalize the description of that type, and use a single 'Column~headerNameFn' as the type for Both ColumnGroup and Column

lbwexler:
How often will we ever have a function for a group's header name?   Think we can just have the groupId default to the headerName if its a string and be done with it.  No need for deprecation, etc.

cmp/grid/impl/ColumnGroupHeader.js
haynesjm42: I've gotten into the habit of adding the `@member` doc for any member variables where the IDE cannot infer the type, I don't think of it so much as a doc but more of a type hint to make the dev experience better, especially when using types from third-party libs like in this case where it is an ag-Grid ColumnGroup. That being said I don't really care too much about it and happy to remove if we are trying to avoid any jsdoc-style  comments in private classes..

cmp/grid/columns/ColumnGroup.js
haynesjm42: There is a slight difference between the two, the Column `headerNameFn` is passed a `column` in the params, ColumnGroup `headerNameFn` is passed a `columnGroup`.

Do we care about that specificity in the param name? Or should we just always pass a `column` prop which could be a Column or ColumnGroup? Or have a single callback doc with both as optional? Couldn't decide when writing this initially, hence the separate callback defs.

cmp/grid/impl/ColumnGroupHeader.js
lbwexler: not a big deal for this PR --- we can discuss our evolving standards on this issue later.

cmp/grid/columns/ColumnGroup.js
lbwexler: In this case, I think you actually get *more* info from using a single doc item for these two things.  It really is the same callback structure, and it actually gives the user more information to see that we think of it as the same object.

Don't know what the magic rule is, but in this case, seems like DRY and clarity both point toward a single doc.

cmp/grid/impl/ColumnGroupHeader.js
amcclain: Yes I occasionally add those @member declarations for the exact same reason, and have found them productive / useful to catch some errors.

Hopefully if we are able to introduce typescript into the equation we will get "real" help with these complex types.

amcclain:
Thank you for the adjustments here, particularly to make it non-breaking, and thanks for merging.

PR #1401, opened 2019-10-09
Clock component

haynesjm42:
Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (https://github.com/xh/toolbox/pull/280)

---------------

Ported the clock component over from hoist-sencha (needed for client app being upgraded to hoist-react)


cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+import {hoistCmp, HoistModel, managed, useLocalModel, XH} from '@xh/hoist/core';
+import {action, observable} from '@xh/hoist/mobx';
+import {MINUTES, ONE_SECOND} from '@xh/hoist/utils/datetime';
+import {Timer} from '@xh/hoist/utils/async';
+import {isEmpty, isNil, isNumber} from 'lodash';
+import moment from 'moment';
+import PT from 'prop-types';
+
+/**
+ * A component to display the current time.
```
amcclain: Would note support for a configurable timezone here in the header comment - it's a special enough feature I think it's worth calling out. We could also note that offsets are loaded from the server, so we don't need to bundle a big TZ database with the client JS code.

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+ *
+ * Automatically updates on a regular interval to stay current.
+ */
+export const [Clock, clock] = hoistCmp.withFactory({
+    displayName: 'Clock',
+    className: 'xh-clock',
+
+    render({timezone, format, updateInterval, prefix, suffix, ...props}) {
+        const impl = useLocalModel(LocalModel);
+        impl.setData({timezone, format, updateInterval, prefix, suffix});
```
amcclain: I see that we're using this convention of having a `setData()` on a local model we configure from props - or at least I see it also exists in `relativeTimestamp`.  What about `configure()` instead, or something more like that?

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+
+    offset;
+    offsetException;
+
+    @observable display = '';
+
+    @managed
+    timer = Timer.create({
+        runFn: () => this.refreshDisplay(),
+        interval: ONE_SECOND
```
amcclain: Should we create this lazily on the first call to `setData()` when we know the actual interval desired vs. also having `ONE_SECOND` here?

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    setData({timezone, format = 'h:mm:ss a', updateInterval = ONE_SECOND, prefix, suffix}) {
```
amcclain: I think I would lose the space before am/pm in the default format - we smoosh them together in our `TIME_FMT` default const.  

(I think that having the seconds here by default _probably_ makes sense, but if we thought they were more likely to be distracting we could just use `TIME_FMT` directly.)

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+    @action
+    refreshDisplay() {
+        if (this.offsetException) {
+            this.display = '';
+            return;
+        }
+
+        const {prefix, suffix, format, offset} = this,
+            time = getAdjustedTime(offset),
+            parts = [time ? moment(time).format(format) : '???'];
```
amcclain: Looks like we will always have a time here so don't think this will fall through.

Should we reserve the `???` display for the (hopefully rare!) case where the client is unable to load the offset and we have an exception, above?  Maybe still pushing on the prefix / suffix?

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+        if (!isEmpty(suffix)) parts.push(suffix);
+
+        const display = parts.join(' ');
+        if (this.display !== display) this.display = display;
+    }
+}
+
+function getAdjustedTime(offset) {
+    const now = new Date();
+    if (isNil(offset) || !isNumber(offset)) return now;
```
amcclain: Why `isNumber()` check here? Seems unlikely...  If we are concerned we could validate before setting the offset in `updateTimezoneAsync()` above.

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+
+        const {prefix, suffix, format, offset} = this,
+            time = getAdjustedTime(offset),
+            parts = [time ? moment(time).format(format) : '???'];
+
+        if (!isEmpty(prefix)) parts.unshift(prefix);
+        if (!isEmpty(suffix)) parts.push(suffix);
+
+        const display = parts.join(' ');
+        if (this.display !== display) this.display = display;
```
amcclain: Feel badly that I don't know this for sure, but won't mobx make this a no-op anyway if the two (primitive) strings are the same?

cmp/clock/Clock.js
lbwexler: I think this is a pattern we are going to see a lot.  Could be something like "noteRenderedProps" or "setLatestProps".

Something about the fact that this is called on every render. 

 We could even make this -- *gulp* -- a new optional built-in lifecycle method, if useLocalModel was provided with props....   (not sure about that -- but something worth thinking about ).

cmp/clock/Clock.js
```diff
@@ -0,0 +1,135 @@
...
+
+    /** Frequency (ms) for updating the displayed text */
+    updateInterval: PT.number
+};
+
+@HoistModel
+class LocalModel {
+    timezone;
+    format;
+    updateInterval;
```
lbwexler: Don't think we need to store an extra copy of this state.

cmp/clock/Clock.js
lbwexler: In general, would like to see these defaults applied on the render method of the component itself.  This is a self documenting defaulting pattern that works pretty well.  No reason to make API users dive into this implementation code to see these defaults.

amcclain:
I would like to cut a 28.1 release today if at all possible - if you guys have time this AM, we could review this for inclusion in that, as well as any other outstanding items we wanted to get in....

I will be in the office at the top of the hour....

amcclain:
I am updating this now with some of the comments above and will get a new version posted today that I think we can go ahead and merge.

amcclain:
I made some updates here and am going to go ahead and merge this.  Can't say that it's perfect / ideal but don't want to have it hanging out there any longer, and this is clearly not a critical component.

As they say, the clock is ticking....

PR #1404, opened 2019-10-11
Handle status 205 correctly in fetchJson

cnrudd:
Hoist P/R Checklist
-------------------

- [X] Ready for review (or open as a draft PR / add `wip` label).
- [X] Reviewed for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)

As per https://httpstatuses.com/205   205 should also return null when using fetchJson.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "fast-deep-equal": "~2.0.1",
     "filesize": "~4.2.1",
     "highcharts": "~7.2.0",
+    "http-status-codes": "~1.3.2",
     "lodash": "~4.17.15",
     "lodash-inflection": "~1.5.0",
     "mobx": "~5.14.0",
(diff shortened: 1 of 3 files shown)


amcclain:
Thanks for this fix!

As I was looking up the new status code - I can remember approx. two of them 😉 - I thought "there is probably some enum we could use for this...."  

The "http-status-codes" lib seems to be the popular choice. I added if as a small additional dep - maybe overkill - but maybe helpful in some other contexts.

(If you guys think this is silly of course we can unwind - I would then just add comments on what the codes mean wherever we find ourselves using them.)

cnrudd:
It would be a breaking change, but it might be useful if the framework returned
```
{
 status: XXX,
body: ...,
}

from the fetchJSON method.  It would save having to return {success: true} for rest form edits.

The benefit becomes more apparent I guess in high volume situations where bits transferred add up.

PR #1407, opened 2019-10-15
Use lint-staged + husky to lint changes prior to git commit

amcclain:
+ We could add these libs to dev-utils if we think we will use them consistently, although they would need to be enabled within each app / project as we see here.
+ Add .jsx extension to our on-demand `yarn lint` command for completeness / as a template for apps that might have JSX files.

Hoist P/R Checklist
-------------------

- [NA] Reviewed for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [NA] Updated doc comments / prop-types (or N/A)
- [NA] Created Toolbox branch / PR (link provided here, or N/A)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,11 +8,19 @@
   "author": "Extremely Heavy Industries, Inc.",
   "license": "UNLICENSED",
   "scripts": {
-    "lint": "yarn run eslint ."
+    "lint": "eslint --ext .js,.jsx ."
   },
   "sideEffects": [
     "./static/polyfills.js"
   ],
+  "lint-staged": {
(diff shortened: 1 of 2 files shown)


amcclain:
@lbwexler I think this is worth trialing in hoist-react and toolbox (I didn't push a TB branch but it's the same change). 

Seems to play well with committing files through IJ, which will stage them when you go to commit. If the linting fails on any changed file it will show an error notification and the commit won't have gone through.

If this is annoying / onerous in general we don't need to take it, but I would like to try and I think it's a popular / common approach.

PR #1408, opened 2019-10-15
Stylelint

amcclain:
Hoist P/R Checklist
-------------------

- [NA] Reviewed for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)


amcclain:
I don't necessarily feel that the stylelint default choices are strictly better than what we had in place, but also was happy to take them as standards and just roll with it.  I think having consistency is worth it more than anything. Hopefully you don't mind the changes the linter made to the styles and taking these new conventions going forward.

I did disable the `no-descending-specificity` rule as it seemed overly fussy to me and there were a decent number of breaks (and no auto-fix).  This rule says that if you have two selectors targeting some class or element, they should be listed in the file in order of increasing specificity. I can't say it seemed to add much, and some of the changes made the rules less readable to me.  Happy of course to tweak other rules.

If we decide we want to use this more widely, we could publish a Hoist stylelint ruleset as a package - just as we do with https://github.com/xh/eslint-config - but I wasn't in any rush to jump to that. 

Opened up related, although not actually dependent, PR on Toolbox @ https://github.com/xh/toolbox/pull/282

.stylelintrc
```diff
@@ -0,0 +1,21 @@
+{
+    "plugins": [
+        "stylelint-scss"
+    ],
+    "extends": [
+        "stylelint-config-standard"
+    ],
+    "rules": {
+        "at-rule-no-unknown": null,
```
amcclain: We use some SASS-specific directives that are flagged as unrecognized by stylelint out of the box - they made the deliberate decision to only include pure CSS @ rules.  So this rule is turned off here, and instead we use the stylelint-scss plugin and enable the `scss/at-rule-no-unknown' rule it provides below to offer the same coverage, just with built-in awareness of standard sass functions, etc.

.stylelintrc
```diff
@@ -0,0 +1,21 @@
...
+        "no-descending-specificity": null,
+        "property-no-unknown": [
+            true,
+            {"ignoreSelectors": [":export"]}
+        ],
+        "scss/at-rule-no-unknown": true,
+        "selector-pseudo-class-no-unknown": [
+            true,
+            {"ignorePseudoClasses": ["export"]}
```
amcclain: The two rule customizations related to `export` cover an edge-case usage we have in https://github.com/xh/hoist-react/blob/develop/desktop/cmp/grid/columns/Actions.scss where we do some fancy stuff to make a value importable from JS.

TomTirapani:
Overall, looks great - and like you say, the main benefit here is to get some kind of standard in place.

The only thing I can think of is to perhaps set `except: ["after-single-line-comment"]` for this rule: https://stylelint.io/user-guide/rules/rule-empty-line-before . I find the new empty lines between comments and the style makes it a bit less clear what the comment refers to.

PR #1414, opened 2019-10-17
Immutable Record

haynesjm42:
Record immutability is now enforced by freezing the Record object, and deep freezing it's data, immediately after construction.

Also added the ability to update Record fields (instead of just raw data) via Store methods, and tracking of when a Record is "dirty". Per discussion with @lbwexler - updating Record fields puts the Record into a dirty state. Whenever the Record raw data is updated, we assume it is the pristine version of the Record, and the Record will no longer be dirty.

To accommodate inline grid editing now that Records are immutable I've added `editable` and `updateValueFn` as Column configs. We install a default `updateValueFn` which will automatically call `Store.updateRecordFields()` with the new value after inline editing (via ag-Grid `valueSetter` callback)

As part of these changes it might make sense to add inline editing support to RestGrid. I've added basic support to the RestGridPanel in Toolbox for testing/demonstration purposes, but could pretty easily see how this could fit well into RestGrid as an alternative to the form-based editing. Would probably want to implement our own text/number/date/boolean cell editors if we do that.

https://github.com/xh/toolbox/tree/recordUpdate

----------------------------

Hoist P/R Checklist
-------------------

- [ ] Up to date with `develop` branch as of last change.
- [ ] Ready for review (or open as a draft PR / add `wip` label).
- [ ] Reviewed for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [ ] Created Toolbox branch / PR (link provided here, or N/A)


cmp/grid/columns/Column.js
```diff
@@ -346,6 +364,12 @@ export class Column {
         const sortCfg = find(this.gridModel.sortBy, {colId: this.colId});
         return sortCfg ? sortCfg.comparator(v1, v2) : agUtils.defaultComparator(v1, v2);
     };
+
+    defaultUpdateFieldFn = ({value, record, store, field}) => {
+        const data = {};
```
lbwexler: minor point, but can be done as a one-liner with `const data = {(field):value}`

data/Record.js
```diff
@@ -20,29 +21,44 @@ export class Record {
     id;
     /** @member {(string|number)} */
     parentId;
-    /** @member {Store} */
-    store;
+    /** @member {Record} */
+    originalRecord;
```
lbwexler: might default original to point to `this`, given its name, and then use a === check for isDirty().

data/Record.js
```diff
@@ -114,24 +130,31 @@ export class Record {
...
+
+        forEach(data, (v, key) => {
+            Object.defineProperty(this, key, {
+                get: () => this.data[key],
+                set: () => {
+                    throw XH.exception(`Cannot set read-only field '${key}' on immutable Record. Use Store.updateData() or Store.updateFields().`);
+                }
+            });
         });
+        deepFreeze(this.data);
```
lbwexler: Why the deepFreeze, and the special property definition?

cmp/grid/GridModel.js
```diff
@@ -400,8 +400,8 @@ export class GridModel {
     }
 
     /** Update the underlying store. */
```
lbwexler: could consider whether to trampoline the additional new store methods in this way as well

cmp/grid/columns/Column.js
```diff
@@ -126,6 +126,8 @@ export class Column {
         exportWidth,
         excludeFromExport,
         tooltip,
+        editable,
```
lbwexler: needs some docs here to help evaluated these new args

data/Record.js
```diff
@@ -20,29 +22,57 @@ export class Record {
...
+    get isDirty() {
+        return this.originalRecord !== this;
+    }
+
+    /** @returns {boolean} */
+    get isModified() {
+        return !this.isNew && this.isDirty;
+    }
+
+    /** @returns {Object|null} */
```
lbwexler: perhaps make this a method, since it is doing a fair bit of work and allocation?

data/Record.js
```diff
@@ -114,24 +144,29 @@ export class Record {
...
-            );
-            this[name] = f.parseVal(data[name]);
+        this.originalRecord = withDefault(originalRecord, this);
+        this.raw = raw;
+        this.data = data;
+        this.store = store;
+        this.xhTreePath = this.parent ? [...this.parent.xhTreePath, id] : [id];
+        this.xhIsSummary = isSummary;
+
+        forEach(data, (v, key) => {
```
lbwexler: this suggests a new syntax for working with records, no?

data/Record.js
```diff
@@ -146,10 +181,17 @@ export class Record {
...
             this.store == rec.store &&
-            this.store.fields.every(f => f.isEqual(this[f.name], rec[f.name]))
+            equal(this.data, rec.data)
         );
     }
 
+    isFieldDirty(name) {
+        if (!this.isDirty) return false;
+
+        const value = this.data[name], originalValue = this.originalRecord.data[name];
```
lbwexler: could we use getters on record here?  also, why are removing the field.isEqual check?

data/Store.js
```diff
@@ -138,25 +150,26 @@ export class Store {
...
-                summaryUpdateRec = this.createRecord(rawSummaryData);
+                summaryUpdateRec = this.createRecord({...this.summaryRecord.raw, ...rawSummaryData}, null, true);
             }
         }
 
         // 3) Apply changes
         let didUpdate = false;
         if (!isEmpty(updateRecs) || (addRecs && addRecs.size) || !isEmpty(remove)) {
-            this._all = this._all.updateData({update: updateRecs, add: addRecs, remove: remove});
+            this._original = this._original.loadRecordTransaction({update: updateRecs, add: addRecs, remove: remove});
```
lbwexler: if original and all were the same when this method was called, aren't we doing double the work, and saving an extra representation?

lbwexler: We could consider not allowing the loading of a transaction for a dirty recordset.

haynesjm42:
Opening new PR for latest changes

PR #1420, opened 2019-10-18
Add Hoist TimeInput - WIP

febbraiod:
Hoist P/R Checklist
-------------------

- [X] Ready for review (or open as a draft PR / add `wip` label).
- [ ] Reviewed for breaking changes (add `breaking-change` label + CHANGELOG if so, or N/A)
- [X] Created Toolbox branch / PR (https://github.com/xh/toolbox/pull/284)


febbraiod:
Opening for discussion. This would serve my purposes for the client app that needs it, but I know that these inputs are always more than meets the eye. Also didn't take on the idea of LocalTime, which makes some sense to me especially to avoid code like:                             

minTime: moment().set({'hour': 8, 'minute': 0, 'second': 0, 'millisecond': 0}).toDate()

But not sure exactly what that Class should look like and isn't necessary for my current needs. Can we maybe take this as an intermediate step to please the client and iterate from here?

One important thing to note is I didn't see an obvious way to get this input to display a null value. As we can see in the toolbox example, when the input is bound to a null value it displays the minTime. Also forces out of bounds values inbounds, which we were getting away from in the dateInput. Again for my current use case, this is irrelevant as it is intended to have no max/min

PR #1424, opened 2019-10-23
Pass component props to model createFn

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+
+## v29.0.0-SNAPSHOT - Unreleased
+
+### 🎁 New Features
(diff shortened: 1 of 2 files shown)


amcclain:
What is the use-case again here?

I don't want to block this, but I worry it further confuses the story around models vs. components vs. model configs vs. component props.

amcclain:
On further thought, I would definitely like to see and discuss examples and determine if this is the right path forward before merging.  Will need to be convinced.

If a component can create a model it should be able to do so independently.  Otherwise the model should be configured and given to it by the parent component.

Further conflating the component props and the model config seems like an anti-pattern to me in an area where we are looking to get greater clarity and separation of purpose.

Issue #1425, opened 2019-10-24
Fix context menu displayFn call when right click on grid row that does not have a store record 


cnrudd:
As reported by client:
If a user right clicks on grouped line popup, app returns error in console and displays standard context menu.
![image](https://user-images.githubusercontent.com/2573928/67502906-ac4e2900-f654-11e9-9942-a1ec19290b77.png)

cnrudd:
Closed.  Not a framework issue.  Applications using displayFn need to null check `record`.

PR #1426, opened 2019-10-24
Check that record exists before calling displayFn (#1425)

cnrudd:
Hoist P/R Checklist
-------------------

This fixes issue #1425, reported by client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v29.0.0 - TBD
+
+### 🐞 Bug Fixes
+
+* In grid context menus, check that record exists before calling displayFn (#1425)
+
+
+[Commit Log](https://github.com/xh/hoist-react/compare/v28.1.1...v29.0.0)
+
 ## v28.1.1 - 2019-10-23
 * Fixes a bug with default model context being set incorrectly within context inside of `Panel`.
 
--- a/data/RecordAction.js
+++ b/data/RecordAction.js
@@ -107,7 +107,7 @@ export class RecordAction {
             disabled: this.disabled || !this.meetsRecordRequirement(recordCount)
         };
 
-        if (this.displayFn) {
+        if (this.displayFn && record) {
             return {
                 ...defaultDisplay,
                 ...this.displayFn({
--- a/data/RecordAction.js
+++ b/data/RecordAction.js
@@ -107,7 +107,7 @@ export class RecordAction {
             disabled: this.disabled || !this.meetsRecordRequirement(recordCount)
         };
 
-        if (this.displayFn && record) {
+        if (this.displayFn && this.meetsRecordRequirement(recordCount)) {
             return {
                 ...defaultDisplay,
                 ...this.displayFn({


haynesjm42:
I'm not sure that we want to do this, as it is perfectly acceptable to have a RecordAction intended only for group rows, and hidden for leaf rows. I would say that instead it is up to the application to handle a null `record` passed to the `displayFn` and `actionFn`, as would be the case for group rows in a Grid.

Perhaps the expected default behavior for our Grid would be to now show a context menu at all when right-clicking on a group row? Would this make more sense as a GridModel config instead?

It is kind of weird though that we call it RecordAction, and it could be called without a `record` or `selectedRecords`... perhaps we could consider renaming the class to Action (should not really be used by apps, they should just be dealing with configs for RecordActions).

cnrudd:
@haynesjm42 That's a good point.

But if the action specifies recordsRequired >= 1, a developer might be thinking reasonably if he does not expect the displayFn to run if that requirement is not met.
I have updated my change so that it does:
`        if (this.displayFn && this.meetsRecordRequirement(recordCount)) {
`

This handles this case well:
https://github.com/xh/toolbox/blob/develop/client-app/src/desktop/common/SampleGridModel.js
    viewDetailsAction = {
        text: 'View Details',
        icon: Icon.search(),
        tooltip: 'View details on the selected company',
        recordsRequired: 1,
        displayFn: ({record}) => ({tooltip: `View details for ${record.company}`}),
        actionFn: ({record}) => this.showInfoToast(record)
    };

cnrudd:
The inconvenience of doing this:
`if (this.displayFn && this.meetsRecordRequirement(recordCount)) {`
though, is that it doesn't give the developer the chance to HIDE the menu item when the record requirement is not met, since that is configurable (if I am reading the code correctly) only when displayFn is run.

We might replace this:
`            disabled: this.disabled || !this.meetsRecordRequirement(recordCount)`
with this:
`            hidden: this.hidden || !this.meetsRecordRequirement(recordCount)`

But maybe a discussion about the benefits of hide vs disable on a menu list has already been had and I did not hear the winning argument that favored disable.

Also worth noting that in tree grids, the group row has a record.   Is it just in rows that are created by AG-Grid for Grouped columns that the row does not have a record?

haynesjm42:
Well we do already default the action to be disabled if it does not meet the record requirement here: https://github.com/xh/hoist-react/blob/2b863142ee6313a4dfcf64065320620731f729ba/data/RecordAction.js#L107

We do clearly document that the `record` may not be present when calling the `displayFn` or `actionFn` - I don't think this is a framework bug at all but an application bug, our expectation has always been that if an app has implemented their own `displayFn` or `actionFn` then they need to gracefully handle the case that the `record` may not be present, if their usage of the action makes that possible. If they specify `recordsRequired` as true or > 1 then we already take care of disabling that action for them (so they don't need to handle it in their `actionFn`)

haynesjm42:
But just having the `hidden` defaulted instead of `disabled` still doesn't solve the issue that the app `displayFn` would need to handle the lack of a `record` in the params.

Also remember that RecordActions are not just used in context menus, also buttons (action columns and RestGrid toolbar buttons) - disabling in those cases is much more preferable. I don't remember if we discussed that default in particular when this was being worked on. I think we were seeing the `displayFn` as more of an advanced use-case, and decided it was fine to require the developer to handle the case of missing the record.

For Tree grids, there are no group rows, just parent rows. Every row has data (a record) so this is not an issue. Group rows in ag-grid have no actual backing data, they only consist of aggregates calculated on the fly (by ag-Grid) from the leaf rows under the group whereas parent rows in a tree grid have actual data to render.

cnrudd:
@haynesjm42 

re: _Also remember that RecordActions are not just used in context menus, also buttons (action columns and RestGrid toolbar buttons) - disabling in those cases is much more preferable._
Thanks, definitely the argument for hiding rather than disabling.

re:_I think we were seeing the displayFn as more of an advanced use-case, and decided it was fine to require the developer to handle the case of missing the record._
Ok, I understand that thinking.  I agree this change is not necessary, and I understand that if displayFn is an escape hatch to more control, then it is buyer beware, like agOptions.  

If everyone agrees, I can just close this PR and write back to the client that we have given their suggestion ample consideration and I think I have enough information now to explain why we will not be making any changes here.

cnrudd:
Hi, 

I discussed with Lee.   We'll close this PR without merging, not make any changes to the framework.  I will edit Toolbox to handle the null record case, and let the client know about our decision.

PR #1457, opened 2019-11-26
Json input context model fix (Fixes #1456)

### LINKED ISSUE #1456: JsonInput does not receive model from context
This is due to the way that context works for class components. It is passed to the constructor of the class, and needs to be passed along to the super class as well, which we are not doing in JsonInput or HoistInput.

Since we are not passing it along, and the `model` is accessed in the constructor of JsonInput, we are attempting to resolve it from context, but since we do not have a context we end up setting the internal `_model` to `null` and will never attempt to re-resolve it on subsequent access (though the reactions being added in the constructor would not work anyways)

It does not look like any other components have this issue, though worth adding the passthrough to any class components which are flagged to support model from context just to be safe (HoistInputs)

haynesjm42:
Small change to pass the `context` through to `super` in the constructors for our HoistInput components which have constructors. 

See https://github.com/facebook/react/issues/13944

The issue was only occurring for JsonInput, since the other components were not referencing their `model` in the constructor, but seemed safest to add everywhere.

In general any HoistComponent which has `supportModelFromContext = true` and has a constructor should be doing this. Not sure how/if we should communicate that in the changelog?

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v29.0.0-SNAPSHOT - under development
 
+### 🐞 Bug Fixes
+
+* Fixed issue where JsonInput was not receiving its `model` from context ([#1456](https://github.com/xh/hoist-react/issues/1456))
+
 ### 📚 Libraries
 
 * @blueprintjs/core `3.19 -> 3.20`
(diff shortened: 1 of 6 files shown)


lbwexler:
This is really a great find john.  Thank You. 

`supportModelFromContext` is really more of a legacy flag designed to just help us keep HoistInputs playing nicely with Context models.  I don't expect we will publicize it much.  Passing context to the constructor of a component is now really just a best practice, just like passing props to the constructor, and not sure Hoist needs to document that.  Hopefully we can just lead by example.

Issue #1474, opened 2019-12-13
Add StoreFilterField to mobile toolkit


TomTirapani:
Alternatively, make it cross-platform - its only desktop dependency is the TextInput itself, all else can be shared,

This is needed for a client mobile app

PR #1476, opened 2019-12-13
Added StoreFilterField to mobile toolkit

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/1474
See Toolbox PR: https://github.com/xh/toolbox/pull/303

Hoist P/R Checklist
-------------------


lbwexler:
Tom -- Thanks for this!

Two comments:

1)  If the API is really intended to be the same let's go ahead and create a single wrapper component in the shared directory.  It can house the documentation etc, and simply delegate to `Impl` components in the respective packages.  TabContainer is a great example of this.

2) The new StoreFilterModel you propose here is really an implementation model.  (we arguably could have built it differently, as the public model interface, but we did not).  It seems to me, we have two choices:

a) Rework the API, so that we expose it as a public entry point to manage the control, and apps can actually specify it.  This would mean nesting all the current props in a "model" config, and would be 
a pretty big breaking change.

b)  Leave the API for the control as is, and simply put this model in an 'impl' folder and perhaps call it StoreFilterFieldImplModel.  We could potentially even put the hook to create it in the wrapper function's render method and than pass it explicitly to the implementation components as a prop.


I vote for b) for this change.  Seems least disruptive and adds the cross-platform support we need.

PR #1485, opened 2019-12-18
Add tooltipElementRenderer Column config (Fixes #1459)

### LINKED ISSUE #1459: Support for Column tooltip element renderers
ag-Grid supports tooltip components, so we should be able to provide a similar API to our Column `elementRenderer` for tooltips (and similar implementation for the anonymous component classes in the column definition).

Not sure what the Column config should be. We have `tooltip` which accepts a `boolean` or a function, so we could have `tooltipElement` to accept a function which returns a React Element?

For the cell renderer configs, we have `renderer` and `elementRenderer`. Not sure if there is a way for us to name the configs consistently. Should we do `tooltipElementRenderer` for symmetry? (though we have no `tooltipRenderer`...)

haynesjm42:
#1459 

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR (https://github.com/xh/toolbox/pull/308)

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -76,7 +76,7 @@ export class Column {
      *      with ExportFormat.LONG_TEXT to enable text wrapping.
      * @param {(boolean|Column~tooltipFn)} [c.tooltip] - 'true' displays the raw value, or
      *      tool tip function, which is based on AG Grid tooltip callback.
-     * @param {Column~tooltipElementRendererFn} [c.tooltipElementRenderer] - function which returns a React component.
+     * @param {Column~tooltipElementFn} [c.tooltipElement] - function which returns a React component.
      * @param {boolean} [c.excludeFromExport] - true to drop this column from a file export.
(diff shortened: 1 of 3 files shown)


cmp/grid/columns/Column.js
```diff
@@ -76,6 +76,7 @@ export class Column {
      *      with ExportFormat.LONG_TEXT to enable text wrapping.
      * @param {(boolean|Column~tooltipFn)} [c.tooltip] - 'true' displays the raw value, or
      *      tool tip function, which is based on AG Grid tooltip callback.
+     * @param {Column~tooltipElementRendererFn} [c.tooltipElementRenderer] - function which returns a React component.
```
lbwexler: John -- Bummer that we can't roll this in to `tooltip`?

If we do need a second argument, can we just call it `tooltipElement` and have the element be a `tooltipElementFn` -- seems more analogous to tooltip, and less likely to confuse with the renderer itself.

PR #1488, opened 2019-12-18
Use agGrid's aggData

peterjcoffman:
Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/GridExportService.js
+++ b/svc/GridExportService.js
@@ -182,8 +182,10 @@ export class GridExportService {
     getCellData(gridModel, record, column) {
         const {field, exportValue} = column;
 
+        // If ag-grid calculated aggregates, it stores them in aggData.
+        const aggData = gridModel.agApi.getRowNode(record.id).aggData;
+        let value = (aggData && !isNil(aggData[field])) ? aggData[field] : record[field];
         // Modify value using exportValue
-        let value = record[field];
         if (isString(exportValue) && record[exportValue] !== null) {
             // If exportValue points to a different field
             value = record[exportValue];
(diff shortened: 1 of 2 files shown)


lbwexler:
Tom, John -- Please have a look at this one with Peter

Its an important one, I think.

svc/GridExportService.js
```diff
@@ -182,8 +182,10 @@ export class GridExportService {
     getCellData(gridModel, record, column) {
         const {field, exportValue} = column;
 
+        // If ag-grid calculated aggregates, it stores them in aggData.
+        const aggData = gridModel.agApi.getRowNode(record.id).aggData;
```
haynesjm42: Can we throw this logic into an `else` block below? I don't think we need to do this unless there isn't an `exportValue` configured

TomTirapani: Currently, we're looking up aggData for every cell in the grid.

To help minimize the number of calls to the agApi, we could get the `aggData` once in `getRecordRow()` above and pass it into `getCellData()`.

We could also check if `treeMode && record.children.length` to avoid getting it for leaf nodes / flat grids.

peterjcoffman: I've put in those two optimizations.

lbwexler: Regarding Tom's optimization about TreeMode -- wouldn't that prevent this from working when we are just using agGrid and grouping with non-full-width group rows?

peterjcoffman: Can we define an aggregation function on something that isn't a TreeGrid?

lbwexler: My bad here -- Haynes straightened me out -- yes we can create a purely "grouping" based aggregation, but our export is not going to see it because our export relies on records -> Different problem!

Issue #1292, opened 2019-08-01
Method to scroll programmatically selected grid row into view


febbraiod:

lbwexler:
[not a UI affordance, I think, but a method to call on GridModel]

lbwexler:
👍

PR #1492, opened 2019-12-20
Add a GridModel method to scroll to selection

peterjcoffman:
Relates to issue 
https://github.com/xh/hoist-react/issues/1292

I'm not sure that the use of selModel.singleRecord is correct. If the user selects multiple records, there are various ways we might want to handle that.

Hoist P/R Checklist
-------------------

https://github.com/xh/toolbox/pull/309
This Toolbox branch has a button with a `null` position, i.e. it will scroll as little as possible. I've tested with the other three position options.

> change. This helps keep the commit log clean and easy to scan across releases. PRs containing a single commit should be *rebased* when possible.


peterjcoffman:
I have not yet tested this on mobile.

lbwexler:
Thanks for grabbing this Peter.  A couple of thoughts here:

1) I think `ensureSelectionVisible` might be more appropriate name.   It aligns nicely with the underlying API, and also gets at what we are trying to do. (`Ensure` is a great name for software API's.)

2) Should we skip the position argument?  The ticket doesn't call for that, I am not sure when an application would actually use it, and furthermore it makes a solution for "multi-selection" more complex (see 3).

3) For multi-selection, one thing I have done in the past is first ensure the "last" row is visible and then ensure the "first" row is visible.  This maximizes the chances that you will bring the *entire* selection into view, but favors the beginning of the selection if the selection is larger than the visible area.  You might need to pause a tick in-between the two -- you'll need to test.  alos, hopefully the selected rows are sorted -- if not, things might be more complicated for you to find the first and last, but not impossible.

peterjcoffman:
@lbwexler I have made those changes.

Issue #1458, opened 2019-11-29
Select input performance issues with large option lists


haynesjm42:
The performance of the Select input component becomes very poor with large option lists. Both opening the list itself when clicking on the select, and attempting to navigate the list when open, become very unresponsive. 

Have not done extensive testing to see how many options it can handle before the performance issues become noticeable, but it does perform very badly with ~2500 options.

It looks like this is an issue in general with react-select, and recommended solutions are to provide a custom Menu component and truncate the list of options being rendered (forcing user to type in text to filter the list). Also other optimizations such as providing a custom option component to remove mouse event handlers (and some other things) to speed up user interaction on the option list itself might be useful/sufficient.

Relevant react-select issue with several proposed solutions: https://github.com/JedWatson/react-select/issues/3128

amcclain:
Thanks for filing - this was one of the lingering issues left in #733 but deserves its own ticket so we can get more visibility on it.

IIRC this was the one real regression we had in behavior over the prior Blueprint implementation, where I believe we had some solution in place for a brief period of time before it became clear we wanted to jump to another component for a variety of other reasons.

I think it's reasonable to assume (as a general principle) that a 1000+ option dropdown would be minimally useful (without typing to filter) even if performance wasn't a problem. That points to an approach that would truncate the list automatically at a preset limit, potentially inserting a final (unselectable) option such as "Not all options displayed, enter text to filter..." so it was clear to the user who did scroll to the bottom of the list that they were not seeing everything. 

This would of course require `enableFilter: true`. Otherwise we might need to resort to just printing a dev-time console warning or similar... 

But there are a lot of pointers on that busy issue, including some like the above.....

haynesjm42:
I wonder if we would get acceptable performance by replacing the menu component with one that uses a Grid/DataView to take advantage of ag-Grid row virtualization. For basic lists which just show the label, using a standard Grid component should result in great performance. In the case that the option element is customized, we'd need to use a DataView, which would potentially have performance issues on scroll while the grid is destroying/creating react components, but I bet it would still perform much better than the default menu in react-select

haynesjm42:
I have a select with about 330 options in it where the mouse hover and scrolling delays are very noticeable (though not nearly as bad as it is with ~2000 options). Pretty disappointing that the out-of-the-box performance of react-select can't handle a few hundred options..

peterjcoffman:
I found a library that aims to improve performance. Maybe we don't want to pull in another dependency, but it's interesting.

[...]

PR #1502, opened 2019-12-27
Add option to use react-windowed-select

peterjcoffman:
Hoist P/R Checklist
-------------------

https://github.com/xh/toolbox/pull/317


peterjcoffman:
To improve performance of Select on large lists, per this issue:
https://github.com/xh/hoist-react/issues/1458

lbwexler:
@amcclain -- any concerns here?  I would like to merge this and have us all start kicking the tires.  As @haynesjm42 can attest, this is a pretty serious bug

kit/react-select/index.js
```diff
@@ -5,10 +5,10 @@
  * Copyright © 2019 Extremely Heavy Industries Inc.
  */
 import {elemFactory} from '@xh/hoist/core';
-import Select from 'react-select';
-import AsyncSelect from 'react-select/async';
-import AsyncCreatable from 'react-select/async-creatable';
-import Creatable from 'react-select/creatable';
+import Select from 'react-windowed-select';
+import AsyncSelect from 'react-windowed-select';
```
lbwexler: This can all be on one line now

package.json
```diff
@@ -63,7 +63,8 @@
     "regenerator-runtime": "~0.13.3",
     "router5": "~7.0.2",
     "router5-plugin-browser": "~7.0.2",
-    "store2": "~2.10.0"
+    "store2": "~2.10.0",
+    "react-windowed-select": "^2.0.1"
```
amcclain: It looks like we should *not* be specifying `react-select` as a direct dependency ourselves - i.e. we should remove it from this list. This new library specifies it as a dep (and notably not a peer-dep) with a particular version.  We only want one copy of `react-select` in our project and it should be the version this new library will work with.

Also, we want to use the `~` semver qualifier for our deps - this allows auto-upgrades of patch versions only, not minor versions.  (We have seen that minor versions are more likely to break, despite what the authors might claim with their version number choice.)

Finally, these deps should remain in alpha-order, and any time we change package.json we need to run `yarn` and check-in the modifications to `yarn.lock` as part of the change.

kit/react-select/index.js
amcclain: Don't we need to export `WindowedSelect` and use that in our `Select` component for us to get any advantage from this change???

As per their readme they re-export all the exports from react-select unmodified, but then also export their custom component. Apologies if I am missing something obvious here 😄

amcclain:
I would also ask that you please:
* Add a Toolbox example with a large number of generated options (1000?  What kind of item count do we think will still be reasonable in terms of performance?)
* Test the mobile implementation (uses the same library)
* Add a note in the Select component header re. the use of `WindowedSelect`
* Add a CHANGELOG entry 

Those will let us tick off a few more of our PR checklist items.  Thanks 🙏

peterjcoffman:
In the previous state, this was broken for everything except 'normal' selects- create and async did not work at all. Now they should work, I believe.

peterjcoffman:
>     * Add a Toolbox example with a large number of generated options (1000?  What kind of item count do we think will still be reasonable in terms of performance?)

The branch https://github.com/xh/toolbox/tree/bigSelect has a new select with 5000 items on app/forms/inputs.

peterjcoffman:
I am not sure menuIsOpen works correctly, i.e. the onKeyDown logic at Select:240-Select:249 does not seem to work properly.

peterjcoffman:
I have found out in more detail what is wrong with `menuIsOpen`.

react-windowed-select does not export a creatable or async `WindowedSelect`. It exports only a 'basic' `WindowedSelect`. Therefore, I have to make Creatable and Async `WindowedSelect` myself.

I do this here:

```

const Select = WindowedSelect;
const AsyncSelect = (makeAsyncSelect(Select));
const AsyncCreatable = (makeCreatableSelect(AsyncSelect));
const Creatable = (makeCreatableSelect(Select));
```

`makeAsyncSelect` and `makeCreatableSelect` are functions exported from react-select which add creatable or async functionality to an existing select.

The problem is that there is an additional step in creating an async or creatable Select. Take a look at the very bottom of this file:
https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Creatable.js

```
const SelectCreatable = makeCreatableSelect<ElementConfig<typeof Select>>(
  Select
);

export default manageState<ElementConfig<typeof SelectCreatable>>(
  SelectCreatable
);
```

`SelectCreatable` is wrapped in a [StateManager](https://react-select.com/props#statemanager-props), which is what handles the [controllable props](https://github.com/JedWatson/react-select#controllable-props). We do not have access to the `manageState` function, so we cannot wrap our Selects in a `StateManager` and cannot get its functionality.

This breaks at least the functionality of `menuIsOpen`. On `Select.js`, the following branch (lines 243-247) will never be taken on a creatable or async Select:
```
    const {menuIsOpen} = this.reactSelectRef.current ? this.reactSelectRef.current.state : {};
    if (menuIsOpen && (e.key == 'Escape' || e.key == 'Enter')) {
        console.log('menu is open');
        e.stopPropagation();
    }
```
It may also break the functionality of `onChange`, `onInputChange`, and [the other things listed as controllable props](https://github.com/JedWatson/react-select#controllable-props). I believe this won't be a problem, since we already handle `onChange`, `value`, `onInputChange`, and `inputValue` ourselves.

peterjcoffman:
Something else noteworthy: the library's handling of the StateManager is also buggy! Compare the code for Async and the code for Creatable:

https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Creatable.js
```
const SelectCreatable = makeCreatableSelect<ElementConfig<typeof Select>>(
  Select
);

export default manageState<ElementConfig<typeof SelectCreatable>>(
  SelectCreatable
);
```
It first makes the Select creatable, then wraps it in a StateManager.

https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Async.js
```
const SelectState = manageState<ElementConfig<typeof Select>>(Select);

export default makeAsyncSelect<ElementConfig<typeof SelectState>>(SelectState);
```
It first wraps the Select in a StateManager, then makes it async.

They do not do the same thing. And the implementation for Async is broken! Add a print statement inside the branch on line 243-247:
```
                const {menuIsOpen} = this.reactSelectRef.current ? this.reactSelectRef.current.state : {};
                if (menuIsOpen && (e.key == 'Escape' || e.key == 'Enter')) {
                    console.log('menu is open')
                    e.stopPropagation();
                }
```
The expected behavior is that hitting Escape while the Select is open will cause 'menu is open' to be printed to the console. This works as expected for a normal Select and a creatable Select, but not for an async Select.

peterjcoffman:
Lee and I discussed this, and we believe it's appropriate to just swallow the keydown events even if the menu is closed. This means we no longer need to care about whether or not the menu is open, so my concerns are probably not a problem.

peterjcoffman:
@amcclain 
>Test the mobile implementation (uses the same library)
I have tested the mobile implementation using the Selects available in the mobile version of Toolbox. However, I do not know how to test the Async or Creatable selects on mobile.

lbwexler:
Peter, Colin, and I discussed.  There are issues with using this for the asyn/ createable/async creatable cases, and known bugs around monitoring the menu open state, that will effects how we handle key presses.

Shame to let this cause problems for the majority of simple selects by trying to force its use automatically.

Safer to expose this as an option -- ` useWindowedSelect` that can be opt-in for apps.  We should document it as having known limitations, and primarily for non-async non-creatable cases.

lbwexler:
I think this is ok for merging now.  Its a very safe change now, that should have no impact, unless the options is set.  Toolbox example has some issues we could go over on monday

lbwexler:
After discussing with Anselm late on Friday, we decided to go with a safer, and more minimal change that only effects basic selects.  We can always revisit, and remove this restriction, if needed and if react-windowed-select fixes its bugs.

Also tested large option lists on mobile.  Weirdly, performance for large option lists seems to be fine, so not adding the prop to the mobile component at this time.

PR #1505, opened 2019-12-30
Reuse chart viz

peterjcoffman:
Hoist P/R Checklist
-------------------


peterjcoffman:
There is one remaining bug here that I cannot figure out how to fix.

You can see that I added something to the default highChartsConfig:

```
            navigator: {
                adaptToUpdatedData: false
            }
```

If you remove that change, you will notice a slight problem with zooming. When loading the line chart example on toolbox, the zoom level will slightly change on the first data reload.

How it looks at first:
image
After reloading data:
image
It is slightly zoomed in. It might be hard to see in a screenshot, but it's obvious in a browser.

Since setting `adaptToUpdatedData: false` fixes this, I think it's a problem with the navigator. But I don't know the root cause and I don't know of a better way to fix it.

peterjcoffman:
I have here a minimal example of the bug I was seeing:

https://jsfiddle.net/a1co7k82/

peterjcoffman:
The zooming bug is fixed fixed. Turns out there is, by default, some padding on the max and min side of the x axis. This padding is respected when making a chart but ignored when changing zoom on an existing chart, so the zoom jumps because the padding is turned off when re-zooming.

I simply removed the padding.

Bottom line: this seems to save about 70 ms on loading data into the LineChart Toolbox example (on my machine) and gets rid of that annoying behavior where zoom was reset on reload.

Issue #1441, opened 2019-11-07
RelativeTimestamp improvements


amcclain:
Two proposed improvements for RelativeTimestamp - the component and its public backing function:

1. Accept a param/prop for `referenceTimestamp` - default `Date.now()` but could be given another static timestamp to compute and display the relative diff between two explicit dates.
2. Use `moment.from()` utility function to generate relative string. This can replace bespoke internal logic and is presumably a better / more robust implementation. @pauldelano also observed that the current impl and moment use different rounding rules, outputting different results for certain diffs.

pauldelano:
On #1 - Do we need 2 props for the now timestamp? One for a static value and other with a bind variable?

lbwexler:
I am not sure about the value of making the change to use lodash, unless we have known bugs/issues. Using lodash would be a breaking change, as it would not the various options that we provide on (e.g. `suffix/prefix/epsilon/short` etc.)

lbwexler:
Also, note that we describe the current "timestamp" property as the "reference time" in the documentation, so will very easy to get confused as to which is which when we allow specifying two times.  We should just be sure to clean up the documentation and be consistent and clear with whatever we do.

peterjcoffman:
If two times are specified, it will not be necessary to periodically refresh the display. So we can turn off the timer in that situation.

peterjcoffman:
I'm not sure how to handle the API change on getRelativeTimestamp. I don't see a good way to add a second timestamp parameter, other than putting it in the options parameter (where I don't think it logically belongs).

lbwexler:
I think it does logically belong as an option.  Its not required, and
frequently won't be set, and secondary to the main timestamp argument.

-Lee

amcclain:
> I am not sure about the value of making the change to use lodash, unless we have known bugs/issues. Using lodash would be a breaking change, as it would not the various options that we provide on (e.g. `suffix/prefix/epsilon/short` etc.)

I don't feel strongly about swapping to moment (note [this is the method I was looking at](https://momentjs.com/docs/#/displaying/from/), but I think the only option we would be unable to easily support would be the `short` flag. The `from()` util takes a second argument that removes the default "in/ago" text around the time units, which would allow us to surround them again with our suffix configs. Likewise the `nowEpsilon` could continue to be evaluated early and `nowString` returned if within that limit.  The minor code cleanup within this file was appealing, but the consistency with moment as our reference date formatting library was even more so.

[...]

PR #1512, opened 2020-01-03
Improve relative timestamp

peterjcoffman:
Hoist P/R Checklist
-------------------


peterjcoffman:
Relates to: https://github.com/xh/hoist-react/issues/1441

lbwexler:
Peter and I are in progress on this -- discussed a number of issues, including better defaulting of suffices when relative is set, and the need to rename 'nowEpsilon' to 'epsilon' and `nowString` to 'equalString' (with our without deprecation support)

lbwexler:
Peter -- can you please indicate how the length of these strings compare with the existing version (with "short" option) and the moment version?  

Need to know if we can remove short option or need to keep the custom logic to continue supporting it.

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -98,151 +97,56 @@ class LocalModel {
...
+
     const defaultOptions = {
             allowFuture: false,
             short: XH.isMobile,
-            futureSuffix: 'from now',
-            pastSuffix: 'ago',
-            nowString: null,
-            nowEpsilon: 30,
-            emptyResult: ''
+            futureSuffix: options.relativeTo ? `from ${fmtCompactDate(options.relativeTo)}` : 'from now',
```
lbwexler: seem like we should use 'after XXX' as the default when relativeTo is set.

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -98,151 +97,56 @@ class LocalModel {
...
-    )(opts);
+    return doFormat(opts);
 }
 
-
 //------------------------
 // Implementation
 //------------------------
-const getNow = opts => ({...opts, now: Date.now()});
+const doFormat = opts => {
```
lbwexler: hoist style not to use fat arrow function for top level functions in a file.  Use `function doFormat() {}' instead.

peterjcoffman: Done.

cmp/relativetimestamp/RelativeTimestamp.js
peterjcoffman: Done.

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -98,151 +97,56 @@ class LocalModel {
...
-    const {isInvalid, prefix} = opts;
-    if (isInvalid) return opts;
-    opts.prefix =  prefix ? prefix + ' ' : '';
-    return opts;
-};
-
-const getResult = opts => {
-    const {isInvalid, elapsedTime, millis, unit, useNowString, prefix, suffix, short} = opts;
-    if (isInvalid) return '[???]';
+    const suffix = isFuture ? opts.futureSuffix : opts.pastSuffix;
```
lbwexler: Need a comment here

lbwexler:
@amcclain -- let me know if you want to review this, or go over it.  I am happy to merge.  I think we landed in a very good place.  Please see the table for how things will change with this release.

thanks @peterjcoffman

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -83,166 +83,71 @@ class LocalModel {
...
-
 }
 
 
 /**
  * Returns a string describing the approximate amount of time between a given timestamp and the
  * present moment in a friendly, human readable format.
  *
- * @param {(Date|int)} timestamp - Date or milliseconds representing the starting / reference time.
+ * @param {(Date|int)} timestamp - Date or milliseconds representing the starting time / time to compare.
```
amcclain: I see that we had `int` before but please swap to `number` as I don't believe `int` is a valid jsdoc type (please correct me if I am wrong)

cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -83,166 +83,71 @@ class LocalModel {
...
-
-const getResult = opts => {
-    const {isInvalid, elapsedTime, millis, unit, useNowString, prefix, suffix, short} = opts;
-    if (isInvalid) return '[???]';
-
-    // if elapsedTime was normalized to 0 (smaller than nowEpsilon) then return the nowString
-    // if specified, otherwise return the default FORMAT for seconds.
-    if (!elapsedTime && useNowString) return suffix;
+    let ret = (elapsed > 10 * SECONDS && elapsed < 60 * SECONDS) ?
+        '<1 minute' :  // override weird moment default in this range.
```
amcclain: Not sure what the weird default is, but I wonder if it would be more robust to look for an replace that string instead of keying off of these time ranges?  I suppose both could change with future moment releases, but replacing the weird string might be more robust (and make it more clear what we are swapping out for "<1 minute")

peterjcoffman: The default is 'a few seconds'. This is undesirable, since displaying 'a few seconds' for 40 seconds is strange.

lbwexler: 1) yep -- that lame comment was mine -- lets fix the comment to specify the problem we have with using a 'few seconds'.

2) Let's extend the start time of that interval, to, say 15 seconds.  Think 'a few seconds' is very nice, and appropriate for slightly larger numbers.  But I strongly agree that it is weird for our users when you get deep into the minute.....

peterjcoffman: @lbwexler You already merged this branch. I have pushed my changes, but I'm not sure what to do with them.

PR #1519, opened 2020-01-06
Dynamic empty text

peterjcoffman:
Hoist P/R Checklist
-------------------

https://github.com/xh/toolbox/tree/dynamicEmptyText
Not suggesting we pull these changes in, but might be useful to see.

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -4,6 +4,7 @@
  *
  * Copyright © 2019 Extremely Heavy Industries Inc.
  */
+import {observer} from 'mobx-react';
 import {isValidElement, createRef} from 'react';
 import PT from 'prop-types';
 import {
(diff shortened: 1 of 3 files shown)


cmp/grid/Grid.js
```diff
@@ -216,7 +216,11 @@ class LocalModel {
             rowDeselection: true,
             getRowHeight: () => this.rowHeight,
             getRowClass: ({data}) => model.rowClassFn ? model.rowClassFn(data) : null,
-            overlayNoRowsTemplate: model.emptyText || '<span></span>',
+            noRowsOverlayComponentFramework: hoistCmp({
+                render() {
+                    return model.emptyText;
+                }
+            }),
```
amcclain: Here I would suggest collapsing this to one line with the more minimal syntax `hoistCmp(() => model.emptyText)`

peterjcoffman:
> @lbwexler what do you think about the use of `@bindable` here w/o an explicit setter? I'm on the fence re. adding a setter - more boilerplate code, but easier to discover given current (lack of) tooling around those annotations and IDE/doc hints.

Why doesn't bindable define a 'proper' ES6-style setter?

lbwexler:
Looking at this with Peter now.  We discussed the fact that we use setXXX() and at this point are unlikely to be able to change that.

lbwexler:
Regarding Anselm's question -- would be happy to use an explicit setXXX method rather than bindable in this case, both for the new option and the existing showSummary().

On a related note,  groupBy should be @observable.ref.

peterjcoffman:
@lbwexler I fixed those few things.

PR #1529, opened 2020-01-10
Add support for grouped options

peterjcoffman:
Hoist P/R Checklist
-------------------


desktop/cmp/input/Select.js
```diff
@@ -342,21 +344,27 @@ export class Select extends HoistInput {
...
+        const options = srcIsObject ? src.options : null;
+
         throwIf(
-            srcIsObject && !src.hasOwnProperty(valueField),
-            `Select options/values provided as Objects must define a '${valueField}' property.`
+            srcIsObject && !(src.hasOwnProperty(valueField) || options),
+            `Select options/values provided as Objects must define a '${valueField}' property or a sublist of options.`
         );
 
         return srcIsObject ?
```
febbraiod: Something about doing this check 3 times bothers me. I might try and see if I could check it once and use a helper. May or may not be clearer. Something like:

return srcIsObject? this.ensureValidOptionObj(src) : {label: src != null ? src.toString() : '-null-', value: src}

peterjcoffman: I tried to simplify this slightly. However, I think we might want to keep the throwIf at the top outside of any branches. I think we usually have our validation logic right at the top.

febbraiod: I can buy that validation should be done at the top of the method, however, we aren't doing any validation for the string version of these opts. So this method has gotten hairy just for the sake of the object version and I really think can be cleaner. I opened a PR with my suggestion. Obviously I'm not final arbiter here, but think my version is worth a look.

PR #1539, opened 2020-01-14
Impersonation config

peterjcoffman:
Client-side changes of this issue:

https://github.com/xh/hoist-core/issues/96

Hoist P/R Checklist
-------------------


appcontainer/ImpersonationBarModel.js
```diff
@@ -46,6 +46,7 @@ export class ImpersonationBarModel {
         if (this.isOpen) {
             this.hide();
         } else {
```
lbwexler: why wouldn't this go with the other throwIf in show()??

peterjcoffman: It probably should.

appcontainer/ImpersonationBarModel.js
```diff
@@ -33,6 +33,7 @@ export class ImpersonationBarModel {
     @action
     show() {
         throwIf(!XH.identityService.canImpersonate, 'User does not have right to impersonate.');
+        throwIf(!XH.getConf('xhEnableImpersonation', false), 'Impersonation is disabled for this app.');
```
amcclain: Thanks for getting this change in! 

What do you think about baking this into `XH.identityService.canImpersonate`?  It seems to me that getter should take the config into account on its own. 

That would also prevent `ImpersonationBar` from rendering entirely and installing its hotkey handler. It's a little mixed how it works now (i.e. if you are not a Hoist admin, bar never renders, but if you and impersonation disabled, bar renders but throws if you attempt to open.

Even more importantly, it would stop mobile from displaying a non-functional button in the app menu - we need to fix that regardless:

https://github.com/xh/hoist-react/blob/19234a5a213b460af2d3ce34e44d113658904fa0/mobile/cmp/header/AppMenuModel.js#L65

(Mobile adds a button as you don't have hotkeys)

lbwexler:
thanks for looking.  Had considered that, but not fully grasped the
implications -- clearly the right thing to do

-Lee

appcontainer/ImpersonationBarModel.js
peterjcoffman: And the logic for whether or not impersonation is allowed should be baked into identityService.

amcclain:
You bet!

Also, sorry to be a pain (I guess it's my job 😄) but Peter, this is one reason why we have that item on the PR checklist for mobile. We need to be sure we don't start just checking those off (as tempting as it might be) without actually exercising the relevant feature / component while at least emulating mobile in the browser. 

In this case, that would have involved first remembering how to impersonate on mobile, then disabling via the new config and ensuring that you could no longer impersonate - and that the app properly reflects that by not offering a button to do so that quietly throws. Mobile so often has these little differences and gotchas - it's a constant process of reviewing what could be different. 

But again - one more tweak and we should be good to go with a nice additional bit of control over this powerful feature.

PR #1554, opened 2020-01-16
Move config check to identityService

peterjcoffman:
Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/IdentityService.js
+++ b/svc/IdentityService.js
@@ -89,7 +89,7 @@ export class IdentityService {
 
     /** Can the underlying user impersonate other users? */
     get canImpersonate() {
-        return this._authUser.isHoistAdmin && XH.getConf('xhEnableImpersonation');
+        return this._authUser.isHoistAdmin && XH.getConf('xhEnableImpersonation', false);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


svc/IdentityService.js
```diff
@@ -89,7 +89,7 @@ export class IdentityService {
 
     /** Can the underlying user impersonate other users? */
     get canImpersonate() {
-        return this._authUser.isHoistAdmin;
+        return this._authUser.isHoistAdmin && XH.getConf('xhEnableImpersonation');
```
amcclain: We want to add in the second argument here with the default of `false` for this config.  Otherwise an app that takes the client-side hoist-react update but has yet to get the hoist-core update (or create the config manually) will start throwing. That would be easy to fix (just add in the config) but given what a big release this is, I would like to avoid throwing developers off with an avoidable exception.

We want to be sparing in our use of the second argument to provide in-code defaults, but I think it's justified in this case where we have a clear / safe default choice and it's a simple boolean. The in-code defaults are less compelling when they are things like timeouts or other magic numbers.

lbwexler: agree -- meant to mention that to you peter when I added before.  Just re-added and merging.  

Thanks for the careful look atm

PR #1555, opened 2020-01-16
Add an `appMenuButtonPosition` param

peterjcoffman:
Hoist P/R Checklist
-------------------


amcclain:
Hi Peter - this is looking good to me.  I had been wondering if we should use the new `appMenuButtonPosition` prop to replace `hideAppMenuButton` - i.e. you could pass a documented value of `'none'` to the new prop.

However I think it is best to keep them distinct as you have now - backwards compatible and maintains the `hide` prop consistent with the other buttons. So, I vote for this change as you have it.

* Can you please add support on mobile?
* No need for a Toolbox branch or demo on this one - I want to keep Toolbox with the default behavior (and do *not* want to add the complexity of an option or anything like that).
* Once mobile done and branch caught up with develop, I think we would be good to go. Anything else WIP on this?

desktop/cmp/appbar/AppBar.js
```diff
@@ -37,7 +37,7 @@ export const [AppBar, appBar] = hoistCmp.withFactory({
             hideAppMenuButton,
             className,
             appMenuButtonOptions = {},
-            appMenuButtonPosition = 'left'
+            appMenuButtonPosition = 'right'
```
amcclain: Oh good catch!

peterjcoffman:
>     * Can you please add support on mobile?

That was a little tricky, but it's done.

TomTirapani:
Great job getting the mobile version to work!

desktop/cmp/appbar/AppBar.scss
```diff
@@ -6,6 +6,10 @@
  */
 
 .xh-appbar {
+  .xh-app-menu {
+    margin-right: var(--xh-pad-px);
```
amcclain: Let's have this apply only when the button is actually on the left - otherwise we introduce some extra margin in the default position. You should be able to use the `:first-child` selector for that to get it done without needing to apply more classes in the component.

mobile/cmp/toolbar/Toolbar.scss
```diff
@@ -11,7 +11,7 @@
     white-space: nowrap;
     flex-shrink: 0;
 
-    &:not(:first-child):not([class*='xh-no-pad']) {
+    &:not(:first-child):not([class*='xh-no-pad']):not([class*='xh-mask']) {
```
amcclain: What does this change do?

desktop/cmp/appbar/AppBar.js
```diff
@@ -86,5 +94,8 @@ AppBar.propTypes = {
     hideAppMenuButton: PT.bool,
 
     /** Options to pass to the AppMenuButton. */
-    appMenuButtonOptions: PT.object
+    appMenuButtonOptions: PT.object,
```
amcclain: Arg - this is pre-existing and not the direct target for this change, but let's take the opportunity to improve....  Can we please also make the breaking changes to:
* Rename `appMenuButtonOptions` -> `appMenuButtonProps` - this is more clear and matches what we have on mobile.
* Ensure that `AppMenuButton` takes all Button props in addition to its own special ones.  It should extract its special props and then pass any others onto the button target for its popover.  I wanted to set a custom icon for the button but could not.  (Specifically I wanted to take one of the example Toolbox apps and drop its primary app icon, replace by moving appButton to left and using the app-specific icon there instead.)
  * Note you can spread `Button.propTypes` onto `AppButton.propTypes` in terms of documenting that it takes any valid button prop.

mobile/cmp/toolbar/Toolbar.scss
peterjcoffman: Since we moved the appMenu from AppContainer to AppBar, the menu is now a descendant of the toolbar. This means it was getting a left margin by default, which is obviously not desirable on the mask.

Before this change:
image
After this change:
image

amcclain: Ah great - thanks!

PR #1558, opened 2020-01-17
Allow user to rename tabs in DashContainer

TomTirapani:
+ User can rename tabs by double-clicking on the title when ViewSpec.allowRename = true.
+ Make DashViewModel's icon and title distinct observables from viewState

See Toolbox PR: https://github.com/xh/toolbox/pull/327/files

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dash/DashViewModel.js
+++ b/desktop/cmp/dash/DashViewModel.js
@@ -5,9 +5,8 @@
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
 import {HoistModel, managed} from '@xh/hoist/core';
-import {bindable, observable, action} from '@xh/hoist/mobx';
+import {bindable} from '@xh/hoist/mobx';
 import {throwIf} from '@xh/hoist/utils/js';
-import {pickBy, isNil} from 'lodash';
 
 import {DashRefreshContextModel} from './impl/DashRefreshContextModel';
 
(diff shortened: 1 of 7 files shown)


desktop/cmp/dash/DashViewModel.js
```diff
@@ -65,4 +66,30 @@ export class DashViewModel {
...
+
+    /**
+     * Overwrite viewState, preserving any existing title or icon
+     */
+    @action
+    setViewState(viewState) {
+        const {title, icon} = this.viewState ?? {},
+            newState = {title, icon, ...viewState};
+        this.viewState = pickBy(newState, v => !isNil(v));
+    }
```
lbwexler: my only concern here tom is:  what if I actually want to store null as a key value in my own custom state -- I should be able to do that

lbwexler: did we consider storing title and icon *next* to viewState?  Then if we had a widget that has no viewState (but the user set the title and icon) we would have no viewState.

TomTirapani: Good idea - I agree that at this point, now we are passing the viewModel to the view, it makes sense to separate title and icon into their own distinct things. Think it will make it cleaner all round. Will have a go at that.

TomTirapani:
Updated this PR to split a DashViewModel's icon and title from its view state, allowing them to be set separately.

Issue #1537, opened 2020-01-14
DockViewModel content config should accept a function


amcclain:
Should be consistent with `TabModel` and accept a closure that returns a React element.

amcclain:
@lbwexler we might want to roll this ticket into any review / standardization of our tab/dock/dash containers and their associated models and configs (as per Slack earlier today).

PR #1562, opened 2020-01-17
Tighten/Clean-up handling of 'content' param in containers

lbwexler:
+ Make content arg consistent across tab/dock/dash containers
+ Apply additional props consistently regardless of how content specified
+ DRY for all implementations

Fixes: https://github.com/xh/hoist-react/issues/1537
Toolbox PR:  https://github.com/xh/toolbox/pull/325

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/react/ReactUtils.js
+++ b/utils/react/ReactUtils.js
@@ -8,21 +8,24 @@
 import {isValidElement, createElement, cloneElement} from 'react';
 import {throwIf} from '../js';
 
+/**
+ * Return the display name for either a class-based or functional Component.
+ * @param obj
+ * @return {string}
+ */
 export function getReactElementName(obj) {
-    return obj.type.name || obj.type.displayName;  // Support for class-based and functional cmps, respectively
+    return obj.type.name || obj.type.displayName;
 }
 
 /**
- * Create a react element from either a HoistComponent, or a function returning an element.
+ * Create a React element from either a HoistComponent or a function returning an element.
  *
- * Used by the TabContainer, DashContainer, and DockView APIs to process the application
- * 'content' parameters provided to them for their tabs and views.
+ * Used by the TabContainer, DashContainer, and DockView APIs to process the 'content' configs
+ * provided to them for their tabs and views.
  *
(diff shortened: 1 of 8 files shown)


utils/react/ReactUtils.js
```diff
@@ -5,6 +5,40 @@
...
+ *
+ * @param {(Object|function)} content - HoistComponent or function returning a React element. If a
+ *      function, it may be an 'elemFactory' (as created by elemFactory()), or any function that
+ *      returns a React element. In either case, the function will be called with no arguments.
+ * @param {Object} [addProps] -- optional additional props to apply to the element.
+ *      These will override any existing props placed on the element, and should be used with care.
+ */
+export function elementFromContent(content, addProps) {
+
+    // Note: Would be more general to look for a *react* component.
```
amcclain: Should we do this now? 😃

utils/react/ReactUtils.js
```diff
@@ -5,6 +5,40 @@
...
+
+/**
+ * Create a React element from either a HoistComponent or a function returning an element.
+ *
+ * Used by the TabContainer, DashContainer, and DockView APIs to process the 'content' configs
+ * provided to them for their tabs and views.
+ *
+ * @param {(Object|function)} content - HoistComponent or function returning a React element. If a
+ *      function, it may be an 'elemFactory' (as created by elemFactory()), or any function that
+ *      returns a React element. In either case, the function will be called with no arguments.
```
amcclain: Why do we call the function with no args, then clone the element to addProps? Why not call the function with addProps in the first place?

lbwexler: The reason for not specifying the “additional” props as arguments to that function, is that the application author of the function should not have to think about those props at all and mixing them in.  The API for the application is to just produce an element.

utils/react/ReactUtils.js
lbwexler: its actually surprisingly tricky with functional components and class components.  We have a clear way to identify a HoistComponent, and that's the way the code currently works.

I guess I just put the comment there because I felt guilty.  :-)

PR #1564, opened 2020-01-17
Added `requiredIndicator` to FormField

peterjcoffman:
Hoist P/R Checklist
-------------------

https://github.com/xh/toolbox/pull/326

### DIFF
--- a/desktop/cmp/form/FormField.js
+++ b/desktop/cmp/form/FormField.js
@@ -65,9 +65,7 @@ export const [FormField, formField] = hoistCmp.withFactory({
                     item: ' ' + requiredStr,
                     className: 'xh-form-field-required-indicator'
                 }) : null;
-
-        defaultProp('requiredIndicator', props, formContext, '*');
-
+    
         // Child related props
         const child = getValidChild(children),
             [stableId] = useState(XH.genId()),
--- a/desktop/cmp/form/FormField.js
+++ b/desktop/cmp/form/FormField.js
@@ -209,7 +209,7 @@ FormField.propTypes = {
      */
     readonlyRenderer: PT.func,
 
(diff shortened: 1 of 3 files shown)


desktop/cmp/form/FormField.js
```diff
@@ -205,7 +207,10 @@ FormField.propTypes = {
      * Optional function for use in readonly mode. Called with the Field's current value
      * and should return an element suitable for presentation to the end-user.
      */
-    readonlyRenderer: PT.func
+    readonlyRenderer: PT.func,
+
+    /** The indicator to display next to a required field. Defaults to span(' *'). */
```
lbwexler: + Think we should just say that it defaults to '*' -- here and in the formField.  Don't need to give the technical details of how we accomplished the spacing, etc.. 

+  This should be of type PT.element here.

peterjcoffman: >This should be of type PT.element here.

I agree, but that breaks. If you give it an element in fieldDefaults, the page will fail to load.

desktop/cmp/form/FormField.js
```diff
@@ -59,7 +59,9 @@ export const [FormField, formField] = hoistCmp.withFactory({
             notValid = model?.isNotValid || false,
             displayNotValid = validationDisplayed && notValid,
             errors = model?.errors || [],
-            requiredStr = (isRequired && !readonly) ? span(' *') : null;
+            requiredIndicator = (isRequired && !readonly) ?
+                defaultProp('requiredIndicator', props, formContext, ' *') :
```
lbwexler: + need the space to be applied to any app applied string.  App should not have to worry about that.

+ More Importantly, see Anselm's point implementing the overrideable styling.

PR #1567, opened 2020-01-21
Grid treeMode indent fix

peterjcoffman:
Hoist P/R Checklist
-------------------


cmp/ag-grid/AgGrid.scss
```diff
@@ -67,11 +67,11 @@
       background-color: var(--xh-grid-group-bg);
     }
 
-    .group-header-icon-expanded {
+    .ag-group-expanded {
       margin: 0 var(--xh-pad-half-px);
```
peterjcoffman: There are two usages of 'ag-group-expanded' in this file, both of which seem to be doing the same thing (i.e. wiping out the left margin). Is this redundant?

amcclain: I don't think so - this one here is adding a left margin - we have a 0 top/bottom and a half-pad left/right margin being applied on this line.

Below we have the rule nested under ` .xh-grid--hierarchical` - see the comment on line 213.

cmp/ag-grid/AgGrid.scss
```diff
@@ -214,12 +214,14 @@
     // this will cause horizontal mis-alignment between expandable and non-expandable sibling rows
     // (and in general tighter padding preferred here vs. full-width groups).
     .ag-row {
-      .group-header-icon-expanded {
+      .ag-group-expanded {
         margin: 0 var(--xh-pad-half-px) 0 0;
+        width: 10px;
```
amcclain: The comment mentions removing a margin, but then we are putting in a hard-coded pixel width here.  We will have to review, but on the face of it I don't understand what we're looking to achieve here.  The comment and the code change don't seem to match.

In general we are also looking to avoid the hard-coded use of specific sizes wherever possible, in favor of our standard CSS vars for padding.

peterjcoffman: The 'expanded' and 'contracted' icons are different sizes. This means that expanding or contracting a row will cause the row to move around. The comment was not updated to reflect this.

I certainly agree with avoiding hard-coded magic numbers.

peterjcoffman:
> Noticed that contents of full-width group rows shift a bit now when expanding/collapsing (_not_ talking about tree grids but e.g. Toolbox SampleGrid w/group by enabled)

I think I have that fixed, but in doing so found another small problem: the last row shifts vertically a small amount when toggled. To see this:

1. Go to the Standard grid in Toolbox.
2. Group by Win/Lose
3. Collapse both groups.
4. Toggle the second group open or closed repeatedly. The text in the row will shift around.

I *think* this applies only to the last row on any normal grid with grouping enabled.

The same problem may also apply to Tree grids, but I'm not sure whether or not I actually see movement or am just fooling myself. It certainly isn't as obvious.

PR #1569, opened 2020-01-21
Don't swallow exceptions thrown by async Select option processing

haynesjm42:
Small change to make sure we log any exceptions in the Select `queryFn` or while processing the options - was at a loss as to why my select wasn't working, turned out I needed to add the `valueField` and `labelField` props, but the exception we throw was being swallowed by higher level code.

Hoist P/R Checklist
-------------------

- [N/A] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.

### DIFF
--- a/desktop/cmp/input/Select.js
+++ b/desktop/cmp/input/Select.js
@@ -436,7 +436,8 @@ export class Select extends HoistInput {
 
                 // But only return the matching options back to the combo.
                 return matchOpts;
-            });
+            })
+            .catch(e => console.error(e));
     };
 
     loadingMessageFn = (params) => {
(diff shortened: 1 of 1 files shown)


desktop/cmp/input/Select.js
```diff
@@ -436,7 +436,8 @@ export class Select extends HoistInput {
 
                 // But only return the matching options back to the combo.
                 return matchOpts;
-            });
+            })
+            .catch(e => console.error(e));
```
lbwexler: john -- minor thing, but should we generally re-throw here as a matter of practice?

haynesjm42: Good call, yes I think we should do that, even though it seems to get swallowed by higher level code in this case

PR #1573, opened 2020-01-21
Add Column.enableDotSeparatedFieldPath

### LINKED ISSUE #1556: Column option to suppress splitting field names containing dots
We just added support in `Column` to process a field name containing `.` into a path array, and then use that path in its default value getter to extract values from nested sub-objects.

https://github.com/xh/hoist-react/blob/263fb281b717399ae62ac24d387371bbdc955018/cmp/grid/columns/Column.js#L145

I had been thinking that we might need the option to turn this on/off in case we had field names with dots in them - i.e. that we should implement our version of ag-Grid's `suppressFieldDotNotation` prop. But then I figured, "let's wait until we see a real case of this in the wild before getting fancy."

Well, today I ran into exactly such a case. We are dynamically constructing a grid model for a client project based on metadata returned from a back-end API, and the field defs provided there can contain dots. The project is on Hoist v28.2, so we popped in the ag prop and all was well, but we will need a solution to move to v29.

amcclain:
+ Fixes #1556

Hoist P/R Checklist
-------------------


amcclain:
I did not create a TB branch for this, but I did run/test locally with some mocked up changes to SampleGrid + its data to confirm.  Happy to review that - still don't think it's something we would include there.

lbwexler:
Agree with the sense of this and the prop name.  

My concern is about the code in GridModel, where we automatically generate a store, including filling in fields.  Should we create a synthetic field in this case based on the *first* part of that path, if we have a `arrayPath`?

amcclain:
Yes, I think that we should provided that we can implement it without too much complexity. It's definitely a gotcha now, where to use something like `country.name` you would have to add `country` as a field in your store config (and it might be the only one if all other fields you care about are referenced directly from columns).

I will look at it as this has become my little creature. 😄  Let's see if we can layer it onto this PR - if it gets too complex I can ticket and we could address separately.

amcclain:
Support for auto-configured / dot-separated store fields added in 9960dd26a50d6bce4c6ee74cf35d2376e95a8497

Issue #1586, opened 2020-01-23
TextInput clear button is visible when value is empty


haynesjm42:
The clear button on TextInput is currently visible even if the value is empty, because the code is explicitly checking for `null` to determine if it should be rendered. We should instead just hide if the value is empty since there is nothing to clear and apps may use an initial value of `null`, `undefined` or an empty string and would expect the same behavior from the input control.

PR #1587, opened 2020-01-23
Hide TextInput clear button when value is empty

haynesjm42:
Fix for #1586 

Initially thought this issue was just with StoreFilterfield, but turns out TextInput would keep the clear button around even after deleting the text manually, which seemed wrong. I opted not to change anything else about the value we set on clear (`null`) or the initial value for StoreFilterField (empty string) - we might want to evaluate what the "empty" value for a TextInput should be, and whether we want to provide an `emptyValue` prop which we would use for the clear button logic, but wanted to keep this change minimal and low-risk.

Hoist P/R Checklist
-------------------


lbwexler:
John and I discussed this on slack.  There was definitely an intention to keep null and empty string separate, and not to make any assumptions about empty string.

Now this may have been a wonky decision.  The other option is to go ahead and try to conflate the two, but this brings up tricky questions as well

lbwexler:
John -- can you let me know your thoughts about my change?  

With the change to having the explicit edit to the empty string taking the value to null, we favor null over the somewhat degenerate empty string.  But I think that makes sense as null is the canonical cleared value across all the controls.

Your Icon visibility change still useful, I think, in that it now matches with the visibility of the placeholder when value is empty string.  These should be consistent -- you can't clear the empty string!

PR #1591, opened 2020-01-23
Allow elements as parameters to getRelativeTimestamp

peterjcoffman:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,7 +115,7 @@ Note the following more specific changes to these related classes:
 * `Chart` and `ChartModel` have been moved from `desktop/cmp/charts` to `cmp/charts`.
 * `StoreFilterField` have been moved from `desktop/cmp/store` to `cmp/store`.
 * The options `nowEpsilon` and `nowString` on `RelativeTimestamp` have been renamed to `epsilon` and
-  `equalString`, respectively.
+  `equalResult`, respectively.
 * `TabRenderMode` and `TabRefreshMode` have been renamed to `RenderMode` and `RefreshMode` and moved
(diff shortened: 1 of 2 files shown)


lbwexler:
Maybe `emptyDisplay` and `equalDisplay`.?

Also, no need to call out {Element | string}.  The auto-coercion that happens from a string to an element is built-in react behavior that is well-understood...

peterjcoffman:
>Maybe emptyDisplay and equalDisplay.?

We certainly can do that, but that would make this a breaking change.

We're already changing `nowString` to `equalString` in Hoist 29. Changing that to `equalResult` is not a breaking change, but changing `emptyResult` would be.

haynesjm42:
> Also, no need to call out {Element | string}. The auto-coercion that happens from a string to an element is built-in react behavior that is well-understood...

I had requested that change because we do `{Element|string}` elsewhere and wouldn't want the caller to think they _have_ to provide an element to what appears to be a formatter function..

lbwexler:
Got it re: Element | string.  Sorry to muddy the waters there.  We can redo that globally at anytime (or not)

amcclain:
Do we have a clear decision on who is approving this and if / when we are moving forwards?

haynesjm42:
I don't think so - I think I would prefer the display logic and customization to be in the component, instead of the `getRelativeTimestamp` function, it just doesn't feel right to have the function doing that.

Also if we aren't worried about having a green "OK" for the "equals" state in a certain client app, we can ignore this for now

lbwexler:
I would like to continue to be involved with this -- John can you and I discuss today?  This warrants a discussion.

PR #1593, opened 2020-01-23
Updated so key navigation respects selectable row attribute and row order

pauldelano:
Hoist P/R Checklist
-------------------


lbwexler:
Looking at this now

lbwexler:
Paul and Peter and I discussed this.  Paul will tighten this up a bit, but it looks very good.

As I discussed with Paul -- we impose this code on ALL ag-Grids (not just Grid).  So we need to be careful.  We could potentially make this custom key handling optional, or wait until a client application requires it to be.

PR #1595, opened 2020-01-24
Fix overflow for dash container titles

TomTirapani:
This was trickier than anticipated to get working without totally breaking GoldenLayout!

Note: GoldenLayout automatically drops out tabs when the container shrinks. To get the undesirable overflow behaviour that prompted this ticket, you need just one tab whose title is wider than its container.

Before:
Screenshot 2020-01-24 at 15 34 02

After:
Screenshot 2020-01-24 at 15 32 56

Also includes styling tweaks to the title input to better suit dark mode

Hoist P/R Checklist
-------------------

### DIFF
--- a/kit/golden-layout/styles.scss
+++ b/kit/golden-layout/styles.scss
@@ -11,6 +11,7 @@
   background: var(--xh-bg-alt);
 
   .lm_header {
+    display: flex;
     background: rgba(0, 0, 0, 0.06);
 
     .xh-dash-container-add-button {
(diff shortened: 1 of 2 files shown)


kit/golden-layout/styles.scss
```diff
@@ -23,11 +24,23 @@
...
+    .lm_controls {
+      position: relative;
+    }
+
     .lm_tab {
       display: flex;
       align-items: center;
       margin: 2px 0 0 2px;
       padding: 4px 8px;
+      max-width: calc(100% - 25px);
```
amcclain: Is this 25px the width of the add button?  Maybe pop it into a SASS variable to self-document it in the two places we are using.  Should we also set on the button itself to make explicit?

TomTirapani: Kind of - it's whats needed to allow for the add button, but its actually that width plus a bit of margin to balance out the right padding added to the header by GL.

Still, will put it in a var to help document it a bit

PR #1598, opened 2020-01-27
Mask login panel while submitting

amcclain:
+ This is almost always instant during local development, but can take a bit of time on a slower connection.
+ Mask should also help to avoid double-submits, which can result in app init being called twice.

Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/login/LoginPanelModel.js
+++ b/appcontainer/login/LoginPanelModel.js
@@ -4,40 +4,46 @@
  *
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
-import {XH, HoistModel} from '@xh/hoist/core';
-import {observable, computed, bindable} from '@xh/hoist/mobx';
+import {XH, HoistModel, managed} from '@xh/hoist/core';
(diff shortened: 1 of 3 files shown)


amcclain:
I didn't think a changelog entry was necessary here, but could add if desired.

TomTirapani:
Looks good - clear improvement to actually wait for the login request.

My only thought is if we should take this opportunity to use a FormModel and field validation here, so we get the red validation around the missing fields? These login prompts predate that functionality.

It's not that I think users are going to be confused about why the simple login form is disabled, but more about providing a feature rich first impression.

lbwexler:
Tom -- I think its a great suggestion, but I am going to merge this now in any case
+ Its really independent of this fix.
+ We are actually looking to tear this whole login out (for non-admins), so don't want to spend anytime on this right now.

amcclain:
Thanks for merging.  I am hopeful this will also avoid those occasional double-service-init exceptions we see.  I think those might have been triggered by multiple clicks getting through here and running the submit code twice.  Will keep my eye out to see if they ever recur.

PR #1601, opened 2020-01-28
Upgrade ag grid

peterjcoffman:
Hoist P/R Checklist
-------------------


amcclain:
I believe that we should look at their new flex implementation and see if we can remove our custom handling around flex columns.

See https://github.com/xh/hoist-react/blob/a0538de977b083760c545fcbc75f9aba1b51caea/cmp/grid/columns/Column.js#L369 where we currently take our formerly unique `flex` config on Column and translate it into an extremely large width.  This ends up paired with multiple calls to `sizeColumnsToFit()` to create the effect of flex columns taking up as much room as they can (but not actually taking up the width that's been specified):

https://github.com/xh/hoist-react/blob/a0538de977b083760c545fcbc75f9aba1b51caea/cmp/grid/Grid.js#L470

Would need to look carefully as these two lines of code are definitely not the entire implementation of this feature. Would also want to understand how the new ag-grid impl differs from the scheme we have in place currently.

lbwexler:
Moving Anselm's comment to new ticket

PR #1612, opened 2020-01-29
Update to reflect ag-Grid Module system

peterjcoffman:
In version 22.0, ag-Grid was broken into a larger number of modules so that applications only need to ship with the ag-Grid features they will actually use. We now have to import any modules we will use and register them with ag-Grid.

This branch should keep almost identical behavior to what we had before. It simply imports and registers _all_ modules. This means we will get no benefit from the new modularization of ag-Grid, but our grids will still work as expected. 

Hoist P/R Checklist
-------------------


lbwexler:
Hi -- this looks great, and is definitely the first step to fixing the regression.  Will merge.  I will also create a ticket for taking advantage of the modules.

lbwexler:
(actually seems like its gonna be a stretch to remove features, given that we are promising our clients access to the full power of raw ag-grid)

lbwexler:
Peter:  noticed this comment on ag site:  

> If you decide to use @ag-grid-enterprise/all-modules then you do not need to specify @ag-grid-community/all-modules too. @ag-grid-enterprise/all-modules will contain all Community modules.

`

PR #1614, opened 2020-01-29
Add support for multiple Grid sizing modes

### LINKED ISSUE #1433: Enhance grid sizing options, esp. need for "tiny" mode.
Several clients have indicated that they would like to see a *more* compact "extra compact" mode in the grid.  

Right now we have the boolean property `compact` on gridModel.  I think the most straightforward would be to deprecate and introduce a new `sizingMode` or `density` property with values `default`|`compact`|`tiny`

Things to consider:  
- Do we need to/can we  extend this mode to the styling of the app as a whole at this time?
- Can we maximize the ability of clients to tweak the mode with styling?
- Backward compatibility?

### LINKED ISSUE #1609: Grid compact mode should have more compact headers
The header height doesn't actually change, which seems a bit inconsistent with the rest of the grid, I think we should have slightly shorter headers in compact mode, and should expose the heights of the various types of headers like we do for row height so applications can customize if they want to.

haynesjm42:
Fixes #1609 
Fixes #1433 

Adds support for a new `sizingMode` config on `GridModel` (really `AgGridModel`) which can be set to 'large', 'standard', 'compact', or 'tiny'. I did this is a non-breaking way to preserve support for the `compact` config which is likely used heavily. It is a very easy breaking change to handle with a mass rename, but need this merged ASAP so not sure we want to include a breaking change in a minor release ...

The 'standard' and 'compact' modes should look the same as before this change, aside from slightly shorter headers for compact mode.

This change also exposes the ability for apps to customize the grid header height for each of these modes via new `AgGrid.HEADER_HEIGHTS` property.

This is also setup in such a way that an app could provide any value they want to `sizingMode` and add a property to `AgGrid.ROW_HEIGHTS` and `AgGrid.HEADER_HEIGHTS` for that value, and also implement their own `xh-ag-grid--my-custom-sizing-mode` css overrides. Wasn't sure how/if that should be documented in the changelog.

Toolbox PR: https://github.com/xh/toolbox/pull/341

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/ag-grid/AgGridModel.js
+++ b/cmp/ag-grid/AgGridModel.js
@@ -77,11 +77,9 @@ export class AgGridModel {
         this.addReaction({
             track: () => this.sizingMode,
             run: () => {
-                if (this.agApi) {
-                    this.agApi.resetRowHeights();
-                    this.agApi.setHeaderHeight(this.headerHeight);
-
-                    // TODO: Do we need to set all of the different type of headers here?
+                const api = this.agApi;
+                if (api) {
+                    api.resetRowHeights();
                 }
             }
         });
--- a/cmp/ag-grid/AgGridModel.js
+++ b/cmp/ag-grid/AgGridModel.js
@@ -63,7 +63,7 @@ export class AgGridModel {
         showCellFocus = false
     } = {}) {
         if (compact) {
-            console.warn('The \'compact\' has been deprecated in favor of the more powerful \'sizingMode\' config.');
+            console.warn('The \'compact\' config has been deprecated in favor of the more powerful \'sizingMode\' config.');
             sizingMode = 'compact';
         }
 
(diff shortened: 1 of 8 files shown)


cmp/ag-grid/AgGridModel.js
```diff
@@ -45,32 +46,41 @@ export class AgGridModel {
 
     /**
      * @param {Object} [c] - AgGridModel configuration.
-     * @param {boolean} [c.compact] - true to render with a smaller font size and tighter padding.
+     * @param {string} [c.sizingMode] - one of large, standard, compact, tiny
```
lbwexler: We could potentially make an enumeration here and do some validation.

Not sure about the idea of supporting arbitrary custom modes --- that seems confusing to me.

lbwexler:
I went over with John in detail.  Assume we can still tweak the final sizing before release, but think the approach is solid and clever -- backward compatible and a nice expansion of functionality with minimal complexity.

We can also think about whether we want this to be enumeration, but merging now so we can start working with it.

PR #1618, opened 2020-01-30
Add support for grouped data views

peterjcoffman:
Hoist P/R Checklist
-------------------

https://github.com/xh/toolbox/pull/354
The roadmap change will take advantage of this as well.


amcclain:
Thanks for keeping this up-to-date @peterjcoffman - I'm going to aim to review and merge this today.

amcclain:
@lbwexler this I believe is ready for review here in hoist-react.

Looking at the TB PRs to determine what is the correct matching PR and what state that's in currently. Please stand by.

lbwexler:
Looks good and realize I am late to  this party.  Will make some small tweaks, but my small concerns is the following:

1) Should itemRenderer and groupRowRenderer be `itemElementRenderer` and `groupRowElementRenderer`?  Very wordy, but more consistent with the underlying name on Column, and would also allow us to provide html based renderers for both of these in the future.  Might be important for performance in large data views.

2) Don't we need a reaction to agRowHeight that mirrors the one we have on rowHeight?

3) We could potentially implement the agOptions computed state/getter and the local reaction abouve on a LocalModel within the DataView class.  Just generally trying to leave application related state on the API model, and implementation details on LocalModels closer to the component, and invisible to the application.

amcclain:
Thanks for looking, @lbwexler.  

1. Should we go one step further and support both, for full consistency with Column and to enable HTML based renderers with this change?
2. Did you mean `groupRowHeight`? I see you added that to the reaction - looks like a good fix, although TBH it seems like a bit of overkill to have that be bindable - but still likely best for symmetry. 
3. I haven't done much with the model+localModel pattern, but I think I know what you mean and happy to have it here if we want to recommend as a best practice - that will also bring this component inline with how `Grid` manages its implementation. 

@peterjcoffman could you please take a look at these points today so we can wrap this up? Let's let Lee confirm his thoughts on (1) though before proceeding (renaming vs. renaming + adding new renderer configs).

amcclain:
I'd also ask that we please document the `groupRowRenderer` config with an `@callback` typedef, so the developer knows what they are getting passed to their function.

peterjcoffman:
>     1. Should we go one step further and support both, for full consistency with Column and to enable HTML based renderers with this change?

We certainly can. Right now we have exactly one use case, so I'm not sure what the correct behavior is.

There are actually several different possible things we could expose. We can also expose only an `innerRenderer`, which allows for less customization but also lets ag-Grid handle things like expanding/collapsing groups.

https://www.ag-grid.com/javascript-grid-provided-renderer-group/

peterjcoffman:
I added some slightly better documentation and changed the naming.

I do not know where the callback documentation should live. Right now it's at the very bottom of Grid.js

peterjcoffman:
I added some slightly better documentation and changed the naming.

I do not know where the callback documentation should live. Right now it's at the very bottom of Grid.js

peterjcoffman:
I have tried out moving the agOptions getter to a new local model. Here is the code:

https://github.com/xh/hoist-react/pull/1673

lbwexler:
closing this, in favor of updated PR:

https://github.com/xh/hoist-react/pull/1673

PR #1620, opened 2020-01-31
Add default Autosize Columns option

peterjcoffman:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   * Header and row heights are configurable for each via the `HEADER_HEIGHTS` or `ROW_HEIGHTS`
     properties of the `AgGrid` component
 * The Grid context menu now contains an `Autosize Columns` option by default.
+* `GridModel` now exposes an `autoSizeColumns` method.
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   * Each of these sizing modes has their own css variables for applications to override as needed
   * Header and row heights are configurable for each via the `HEADER_HEIGHTS` or `ROW_HEIGHTS`
     properties of the `AgGrid` component
+* The Grid context menu now contains an `Autosize Columns` option by default.
 
 ### 🐞 Bug Fixes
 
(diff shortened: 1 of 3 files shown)


cmp/grid/GridModel.js
```diff
@@ -649,6 +650,10 @@ export class GridModel {
         return c.children ? new ColumnGroup(c, this) : new Column(defaultsDeep({}, c, this.colDefaults), this);
     }
 
+    autoSizeAllColumns() {
```
haynesjm42: I think we should consider changing this to `autoSizeColumns` which can accept a single or list of `Column` objects or column ids, and it can default to `this.getLeafColumns()`.

Pretty sure we want to use `getLeafColumns()` here instead of the `columns` list as that contains column groups, so we are probably skipping autosizing of their child columns - unless ag-grid is handling that for us when we ask to autosize a column group.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   * Each of these sizing modes has their own css variables for applications to override as needed
   * Header and row heights are configurable for each via the `HEADER_HEIGHTS` or `ROW_HEIGHTS`
     properties of the `AgGrid` component
+* The Grid context menu now contains an `Autosize Columns` option by default.
```
haynesjm42: Would also note the new GridModel method as well in the changelog

cmp/grid/GridModel.js
```diff
@@ -650,8 +650,8 @@ export class GridModel {
         return c.children ? new ColumnGroup(c, this) : new Column(defaultsDeep({}, c, this.colDefaults), this);
     }
 
-    autoSizeAllColumns() {
-        this.agColumnApi.autoSizeColumns(this.columns.map(col => col.colId));
+    autoSizeColumns(colIds = this.getLeafColumns().map(col => col.colId)) {
+        this.agColumnApi.autoSizeColumns(colIds);
```
haynesjm42: Lets also add `colIds = castArray(colIds);` here so I can call it like `gridModel.autoSizeColumns('myColumnId')`

Also we need to add docs for this new public method

PR #1629, opened 2020-02-05
Form context validation

peterjcoffman:
Hoist P/R Checklist
-------------------


desktop/cmp/form/FormField.js
```diff
@@ -47,8 +47,13 @@ export const [FormField, formField] = hoistCmp.withFactory({
...
         // Resolve FieldModel
-        const formContext = useContext(FormContext),
-            formModel = formContext.model;
+        const formContext = useContext(FormContext);
+        throwIf(
+            !formContext || (Object.entries(formContext).length === 0),
+            'Form field could not find valid FormContext. ' +
+            'Possibly caused by using an HTML form rather than a Hoist form.'
+        );
+        const formModel = formContext.model;
```
lbwexler: + . Doesn't seem like the Form or FormModel is necessary to Render (or should not be difficult to make it so). Think it a gentler development ex

+ Error message should be clearer, reference the actual package e.g.  "Make sure you are using a Hoist form ('@xh/hoist/cmp/form/Form')  and not an Html Form ('@xh/layout/Form`).

peterjcoffman: I addressed those two things.

amcclain: Minor style point. would recommend using backticks here for the multi-line string with inner quotation marks.

amcclain: No need to go back to change this now, but the idea with the backticks is that you don't need to concat with `+` - you can include newlines in backtick quoted strings, so multi-line is cleaner.

PR #1630, opened 2020-02-05
Add new props for validation tooltip

peterjcoffman:
This change introduces two new props, `tooltipPosition` and `tooltipBoundary`.

Additionally, we now default `boundary` to `viewport` rather than `scrollParent`. Validation tooltips may now extend beyond the limits of their parent, which is desirable for modal dialogs. Additionally, the `scrollParent` default was confusing because the behavior of tooltips would change depending on how scrolling is handled in the application.

As an example, I created a modal dialog with scroll bars. Before the change in this PR, the tooltip automatically moves to the left of the field to avoid leaving the modal dialog:

image

After the change, it extends beyond the boundaries of the dialog:

image

Hoist P/R Checklist
-------------------


peterjcoffman:
I am not sure the 'tooltipBoundary' works as expected.

desktop/cmp/form/FormField.js
```diff
@@ -46,6 +46,8 @@ export const [FormField, formField] = hoistCmp.withFactory({
 
     render({model, className, field, children, info, ...props}) {
 
+        console.log(props);
```
TomTirapani: This isn't needed :)

desktop/cmp/form/FormField.js
```diff
@@ -81,7 +83,9 @@ export const [FormField, formField] = hoistCmp.withFactory({
             labelAlign = defaultProp('labelAlign', props, formContext, 'left'),
             labelWidth = defaultProp('labelWidth', props, formContext, null),
             label = defaultProp('label', props, formContext, model?.displayName),
-            commitOnChange = defaultProp('commitOnChange', props, formContext, undefined);
+            commitOnChange = defaultProp('commitOnChange', props, formContext, undefined),
+            tooltipPosition = defaultProp('tooltipPosition', props, formContext, 'right'),
```
TomTirapani: Think the position should default to 'auto', which is what it was before. That means app devs can take this change without worrying about changes

peterjcoffman: The position used to default to 'right'. See FormField 108-117:

        if (minimal) {
            childEl = tooltip({
                target: childEl,
                targetClassName: `xh-input ${displayNotValid ? 'xh-input-invalid' : ''}`,
                targetTagName: !blockChildren.includes(getReactElementName(child)) || child.props.width ? 'span' : 'div',
                position: 'right',
                disabled: !displayNotValid,
                content: getErrorTooltipContent(errors)
            });
        }

TomTirapani: Ah! In that case, please ignore my miscomprehension

lbwexler:
Great change -- Thank you

PR #1640, opened 2020-02-07
More robust scrollable css for mobile panel.

cnrudd:
Hoist P/R Checklist
-------------------

I found that when two grids were the only items in the panel, and the grids are set to use domLayout='autoheight'', the panel's xh-vframe does not allow the content to expand to its full height, unless display = 'block' instead of 'flex'.

I think it makes sense that if we want a container to be scrollable, then we don't need a display:flex setting on it to try to adjust child item heights so that they fit inside the container.

I can demo the issue on my client app, if anyone wants to see before merging this.


TomTirapani:
Thanks for catching - flexing the contents of a scrollable container doesn't really make sense. Can't imagine a reason not to take this change.

The only thing I would say is that perhaps we should switch the wrapper in the component rather than override the css - if we're neither hiding overflow nor flexing, it begs the question why use a vframe at all?

cnrudd:
> Thanks for catching - flexing the contents of a scrollable container doesn't really make sense. Can't imagine a reason not to take this change.
> 
> The only thing I would say is that perhaps we should switch the wrapper in the component rather than override the css - if we're neither hiding overflow nor flexing, it begs the question why use a vframe at all? I think changing the inner component to a div with `overflow-y: auto` will make the intention clearer.

@TomTirapani Hi Tom, I've implemented your suggestion.

Issue #1530, opened 2020-01-10
Build view tracking into TabContainerModel


amcclain:
We have a pattern across several applications where we want to log a track statement when a user swaps to a particular tab, typically the top-level tabset in an app.

Would be convenient if this were baked into `TabContainerModel` - could take a `track` config that would accept true to track with defaults, or perhaps an object that could be used to set the category or customize the track message.

PR #1655, opened 2020-02-12
Tab container model tracker

petradish:
Hoist P/R Checklist
-------------------

This PR addresses issue https://github.com/xh/hoist-react/issues/1530
- Added a 'track' parameter to TabContainerModel that tracks tab views in the Admin activity logger.
- Made a branch in toolbox to refactor the main AppModel to include this new feature: https://github.com/xh/toolbox/tree/tabContainerModelTracker

It shouldn't be a breaking change as the default value of track is set to false.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ Previously, it was overridden by any `className` passed in via optional params.
   'lift', 'fade', or 'none'. These values are passed to the underlying onsenNavigator component.
   ([#1641](https://github.com/xh/hoist-react/pull/1641))
 * `AppOption` configs now accept an `omit` property for conditionally excluding options.
+* `TabContainerModel` now accepts a `track` property for tracking tab views in admin track log. The
+  default value is set to `false`.
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 * `StoreFilterField` accepts all other config options from `TextInput` (e.g. `disabled`, `textAlign`,
   and `spellCheck`).
 * Clicking on a summary row in `Grid` now clears its record selection.
+* `TabContainerModel` now accepts a `track` property for tracking tab views in admin track log. The
+  default value is set to `false`.
 
 ### 🐞 Bug Fixes
 
@@ -58,8 +60,6 @@
   'lift', 'fade', or 'none'. These values are passed to the underlying onsenNavigator component.
   ([#1641](https://github.com/xh/hoist-react/pull/1641))
 * `AppOption` configs now accept an `omit` property for conditionally excluding options.
-* `TabContainerModel` now accepts a `track` property for tracking tab views in admin track log. The
-  default value is set to `false`.
 
 ### 🐞 Bug Fixes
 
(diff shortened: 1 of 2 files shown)


cmp/tab/TabContainerModel.js
```diff
@@ -87,6 +94,20 @@ export class TabContainerModel {
...
         }
+
+        if (track) {
+            this.addReaction({
+                track: () => this.activeTab,
+                run: (activeTab) => {
+                    XH.track({
+                        msg: `Viewed ${activeTab.title}`,
+                        category: 'Tab',
+                        // route data could be useful for non-top-level tabs.
```
lbwexler: use a local variable route for this expression repeated three times

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ Previously, it was overridden by any `className` passed in via optional params.
   'lift', 'fade', or 'none'. These values are passed to the underlying onsenNavigator component.
   ([#1641](https://github.com/xh/hoist-react/pull/1641))
 * `AppOption` configs now accept an `omit` property for conditionally excluding options.
+* `TabContainerModel` now accepts a `track` property for tracking tab views in admin track log. The
+  default value is set to `false`.
```
amcclain: Need to move this up to the snapshot version (common issue when a PR stays open across releases).

cmp/tab/TabContainerModel.js
```diff
@@ -87,6 +94,21 @@ export class TabContainerModel {
 
             this.forwardRouterToTab(this.activeTabId);
         }
+
+        if (track) {
+            this.addReaction({
+                track: () => this.activeTab,
+                run: (activeTab) => {
+                    const route = activeTab.containerModel.route;
```
amcclain: This can just be `const {route} = this`, correct? I think that's a bit clearer.

cmp/tab/TabContainerModel.js
```diff
@@ -87,6 +94,21 @@ export class TabContainerModel {
...
+                    XH.track({
+                        msg: `Viewed ${activeTab.title}`,
+                        category: 'Tab',
```
amcclain: I am sure this is what we had in Toolbox, but having 'Tab' as the category seems a bit awkward now that we're building it in.  How about:
```
{category: 'Navigation', message: `Viewed ${activeTab.title} tab`, ....}
```

Also note that `msg` is the old param that we're trying to stamp out - trackService supports it as a fallback, but let's swap to `message` which we're looking to use consistently on the client.

cmp/tab/TabContainerModel.js
```diff
@@ -48,6 +51,8 @@ export class TabContainerModel {
      *      be route-enabled, with the route for each tab being "[route]/[tab.id]".
      * @param {string} [c.switcherPosition] - Position of the switcher docked within this component (or 'none').
      *      Valid values are 'top', 'bottom', 'left', 'right', 'none'.
+     * @param {boolean} [c.track] - Enables tracking of tab views in admin logger under category
```
amcclain: I would say "True to enable activity tracking of tab views" - that uses the same language as the admin client itself, vs "admin logger".

PR #1660, opened 2020-02-14
Include exception message in "Failed to initialize services" exception dialog

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -23,6 +23,7 @@ import {
     WebSocketService
 } from '@xh/hoist/svc';
 import {throwIf, withShortDebug} from '@xh/hoist/utils/js';
+import {p} from '@xh/hoist/cmp/layout';
 import {camelCase, flatten, isBoolean, isString, uniqueId} from 'lodash';
 import ReactDOM from 'react-dom';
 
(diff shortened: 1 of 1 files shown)


amcclain:
Hi @TomTirapani - would you happen to have a screenshot of what this looks like in practice? 

I've been thinking that this dialog needs a bit of work, as we're showing it to the user when the app has failed to start (meaning things are in a bad state, and they are probably not pleased to see it).

For that reason, I agree it's great/important to provide more visibility into what has gone wrong. I just want to ensure it's not even more confusing/distressing appearing to the user. I suppose it depends on how reasonable the exception messages are that we're producing?

TomTirapani:
Screenshot 2020-02-14 at 12 30 41
Here's a demo from Toolbox. FYI, this is an interim solution - Lee, John and myself were having a detailed look at the app startup failures and have some ideas we're going to write up and offer for wider discussion

amcclain:
Cool - thanks! Seems like a clear improvement regardless. (And when the app is not going to start, ultimately the more information that we can float up and make visible, the better...)

PR #1680, opened 2020-02-20
debounce support for reactions

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/logging/viewer/LogViewerModel.js
+++ b/admin/tabs/logging/viewer/LogViewerModel.js
@@ -90,7 +90,7 @@ export class LogViewerModel {
                 this.file = rec?.data?.filename;
                 this.loadLog();
             },
-            debounce: 300
+            debounce: {interval: 300, leading: true}
         };
     }
 
--- a/admin/tabs/logging/viewer/LogViewerModel.js
+++ b/admin/tabs/logging/viewer/LogViewerModel.js
@@ -90,7 +90,7 @@ export class LogViewerModel {
                 this.file = rec?.data?.filename;
                 this.loadLog();
             },
-            delay: 300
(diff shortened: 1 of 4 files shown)


admin/tabs/logging/viewer/LogViewerModel.js
```diff
@@ -90,7 +90,7 @@ export class LogViewerModel {
                 this.file = rec?.data?.filename;
                 this.loadLog();
             },
-            delay: 300
+            debounce: {interval: 300, leading: true}
```
amcclain: Do you think it would be worth supporting a boolean variant that would default to these options if `true`?

It feels like this is often what we would want - some consistent interval that's short enough to be responsive yet long enough to make a difference. I feel like 300ms has been our go-to number since the ExtJS days. Defaulting `leading: true` might be a bit more opinionated, but at the same time since all these reactions are async anyway I have some sense that might be useful. Feel like we are often looking to debounce clusters (as in > 2) of calls when we specify a debounce.

Given all that, supporting a simple boolean to turn on a Hoist default for reaction debouncing w/o needing to come up with the magic number each time seems like it might be worth the incremental complexity.  Just an idea though - happy to merge as-is - great catch all around on delay-is-not-debounce.

lbwexler: I am inclined not to be too magical here -- I feel that requiring specifying the number encourages the user to actually consider, and play with the GUI and feel whats rights.  A default value may deiscourage that.  right value  gonna depend on so many factors.   Also -- if we had a magical constant -- could we ever change it safely in the future, or would we always be "stuck" with it

lbwexler:
Reviewed by ATM and Peter

PR #1690, opened 2020-02-20
Fix menu when there are no available widgets to add.

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dash/impl/DashContainerContextMenu.js
+++ b/desktop/cmp/dash/impl/DashContainerContextMenu.js
@@ -64,6 +64,11 @@ function createMenuItems(props) {
             menuDivider({title: 'Add'}),
             ...addMenuItems
         );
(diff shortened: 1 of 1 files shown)


TomTirapani:
Looks good, but I'm wondering if we even need this now we use a context menu rather than the '+' button? I think the ticket was added when we were using the '+' button - its drop-down menu will be empty, but the button still visible. On the other hand, our context menu also contains 'Remove', 'Refresh' etc., so will never be "empty".

I think it is better UX to simply omit the 'Add' section than to insert a dummy item. Of course, happy to be convinced otherwise! :)

TomTirapani:
.... and as soon as I clicked "Comment", I remembered that if you click on the empty space you get the menu without 'Remove', 'Refresh' etc. In that place, makes perfect sense! :)

Merging

Issue #1711, opened 2020-02-25
Admin config grid slow w/large JSON values


amcclain:
We have an app that has some very large JSON configs. Believe the Admin client config grid / panel is getting very slow as ag-Grid renders the entire JSON string in the column, even though you only see a fraction of it.

Consider truncating via custom value getter - could also apply to prefs.

PR #1713, opened 2020-02-25
Truncate JSON column values in Admin Config + Preference grids

petradish:
Hoist P/R Checklist
-------------------

This PR addresses issue https://github.com/xh/hoist-react/issues/1711
 - Admin config model now truncates json type values to 50 characters with getValueFn, improving speed of cell render.

### DIFF
--- a/admin/tabs/general/configs/ConfigModel.js
+++ b/admin/tabs/general/configs/ConfigModel.js
@@ -16,6 +16,7 @@ import {
     RestStore
 } from '@xh/hoist/desktop/cmp/rest';
 import {ConfigDifferModel} from './differ/ConfigDifferModel';
+import {truncate} from 'lodash';
 
 @HoistModel
 @LoadSupport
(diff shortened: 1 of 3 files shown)


amcclain:
Great - thanks for jumping on this.  I pulled your branch and it definitely resolved the performance issue I was seeing.

A few requests - 
* Realized there was already a custom renderer for password masking. Seems like we should be able to use the same approach to handle both cases in one function.  I know I pointed to the value getter (not sure why now) but it seems like the renderer should also work just as well and is a bit more standard.  That way the exports would also be complete, which is good.
* No need to be quite so aggressive on 50 chars - the performance issues I saw were due to JSON blobs with thousands of lines 😄.  I say let's truncate on 500 chars instead.
* Let's apply the same approach to the preference grids, which might also have big JSON blobs.

Thank you!

amcclain:
Edit - re. value getter / renderer - need to ensure the tooltip also gets the truncated value!  That might be a big part of what is making it all so slow.

petradish:
Yes, tooltip is also grabbing the truncating value with the value getter. I was also wondering whether there was a performance difference between using getValueFn versus applying another function call on the renderer. I guess there isn't one?
Also, I still think we should define two separate functions but I'll call them with && in the renderer and tooltip!

Issue #1513, opened 2020-01-03
FileIcon Component


haynesjm42:
The current `fileIcon` helper function should be refactored into an actual component, and support the following options:
* Standard coloring based on extension
* Standard tooltips based on extension

Also should make sure that we allow overriding any of the icon props if an app wants to use a different icon variant.

Also important that the root of this component is just the icon, no wrapper divs/spans, so it can be passed to `convertIconToSvg` without issues

PR #1726, opened 2020-02-27
FileIcon addition as Element; asSvg prop added to Icon factories

petradish:
Hoist P/R Checklist
-------------------

This PR addresses https://github.com/xh/hoist-react/issues/1513
- New fileIcon Element and Component created to set default configuration (tooltip name and color) for file extensions.
 - Default configs are overridable by passing in add'tl props
- asSvg prop added to all Icon elements -- no longer require the function convertIconToSvg to render icons within AgGrid renderers.


cmp/grid/Grid.js
```diff
@@ -349,7 +343,7 @@ class LocalModel {
 
             let icon = displaySpec.icon;
             if (isValidElement(icon)) {
```
lbwexler: This seems wrong to me.  Think you need to test this.

lbwexler: (Probably want to leave this one as is.)

desktop/cmp/grid/columns/Actions.js
```diff
@@ -70,7 +69,7 @@ export const actionCol = {
             if (!isEmpty(tooltip)) buttonEl.setAttribute('title', tooltip);
             if (!isEmpty(intent)) buttonEl.classList.add(`bp3-intent-${intent}`);
 
-            buttonEl.innerHTML = convertIconToSvg(icon);
```
lbwexler: again problematic here -- my guess is that you need to revert this code to get the html you need here.

icon/Icon.js
```diff
@@ -915,43 +925,66 @@ export const deserializeIcon = function(iconDef) {
  * @param {String} extension
  * @return {Element}
  */
-export const fileIcon = function(extension) {
+
+export const [FileIcon, fileIcon] = hoistCmp.withFactory({
```
lbwexler: Think we can skip the component, since we have Icon.fileIcon()

lbwexler:
Note: this is no longer a breaking change if we leave convertIconToSvg() in place, also, I am suggesting we don't need the component FileIcon/fileIcon, if we have the elementFactory Icon.fileIcon already in place.

Flagging @amcclain for your thoughts -- 

We have the strange situation where the hoist API passes these "Icons" around as react elements. We always create them with our icon element factories, e.g. . Icon.fooBar(), but there is nothing in the API that really requires that.  For the most part, they are just rendered in react, and could in fact be arbitrary react elements.  But in certain circumstances -- e.g. creating menus, or passing to Golden Layout, we need to get the svg representation, and make a hacky dive into the elements property to pull out the underlying 

Seems like there are two alternatives to how to better document/improve this.

1) Have an explicit Icon Object as the type of our various API properties, rather than Element (either our own, or FontAwesome's).  This would be something that could always be guaranteed to be turned in to either an react element, or SVG.  This might be more backward compatible than we think, but might also be confusing when users interact directly with say, blueprint, and need to call an additional .element() method on objects.  

2)Double down on our current assumption. Just improve our documentation regarding "Icon".  Maybe add a JSDoc for `IconElement` to indicate the special type of element we are passing around.    Add a throwIf() 

3) Stay flexible and be robust. Allow users to pass any React element as an icon.  If we ever need to get an SVG out of it, and can't, log an error and return some default icon....

cmp/grid/Grid.js
petradish: It works in toolbox!

desktop/cmp/grid/columns/Actions.js
petradish: This also works if RecordAction takes an icon that has config {asSvg: true}. Not sure how to document that change more clearly (and also why this is a breaking change).
See RecordAction.js line 46: ` // @param {Element} [c.icon] - icon to be displayed. Always specify with config {asSvg: true}.`

lbwexler:
John nice idea about adding the method to the element, but not gonna work.  React elements are immutable and frozen solid like the lake you were on yesterday

styles/XH.scss
```diff
@@ -98,6 +98,16 @@ body.xh-app {
 
   .xh-yellow-light { color: var(--xh-yellow-light); }
 
+  .xh-fileIcon-excel { color: #0e7b41; }
```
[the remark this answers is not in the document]
petradish: Thanks for the catch on this and the word .doc extension class name! I am not sure when a color warrants it's own variable. I'll leave it up to the big cheeses for that decision.

TomTirapani:
Took a look at this - think it's a great change. I think we should make sure we highlight the new `asSvg` option in the CHANGELOG, and encourage devs to use that instead of `convertIconToSvg()`

lbwexler:
So John and Tom and I discussed this, and were all leaning toward option 3) above.  It seems like its a pretty important/useful feature that generally our 'icon' props accept elements are are not in anyway tied to Font-Awesome or our own classes.  The element expectation makes the most sense in general.

The new asSvg option on our Elem factory is a nice to  

Will catch this up, and enhance the changelog a bit and look to merge.  Tom will also review.

styles/XH.scss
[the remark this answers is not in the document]
lbwexler: lets just leave them hard-coded but please fix the camel case

icon/Icon.js
lbwexler: let's remove this component -- users can just use the elem factory in their render method.

desktop/cmp/grid/columns/Actions.js
lbwexler: Ah -- I see, but then we are lying with jsdoc {Element} declaration on line 46 -- its not an element, its a string of html.   This is a tricky bit, lets dicsuss tommorrow.

PR #1727, opened 2020-02-27
Add headerAlign config on Column

peterjcoffman:
Hoist P/R Checklist
-------------------


peterjcoffman:
Does this need a Toolbox branch? As usual, I can't think of anything sensible.

lbwexler:
A toolbox example would have been nice for sure, but can see how it would take time to setup something clean to the user that actually allows us to toggle the column API.

In this case, will go ahead and merge because it seems correct and quite safe.

amcclain:
Sweet and simple - thanks for this.  No, I don't think we need a TB example in this case - if one comes up naturally though we have it to use.

PR #1728, opened 2020-02-27
Change default selectOnFocus behavior in Select

peterjcoffman:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 
 [Commit Log](https://github.com/xh/hoist-react/compare/v30.0.0...develop)
 
+### ⚙️ Technical
+
+* `Select` will now default to selecting contents on focus if in filter or creatable mode.
+
 ## v30.0.0 - 2020-02-29
 
 ### 🎁 New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,9 @@
 * Enhancements to form validation to improve handling of asynchronous validation.  Individual rules
 and constraints are now re-evaluated in parallel, allowing for improved asynchronous validation.
 
-[Commit Log](https://github.com/xh/hoist-react/compare/v30.0.0...develop)
-
-### ⚙️ Technical
-
 * `Select` will now default to selecting contents on focus if in filter or creatable mode.
+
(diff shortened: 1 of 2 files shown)


lbwexler:
I like this one alot, in theory.  @haynesjm42  -- do you feel so strongly that this is always what you need, to the degree that we should change the default?   Sometimes aren't you just further drilling down?

You pretty often want it in a TextField as well, but we did not make it the default.  Don't feel strongly here....

haynesjm42:
Just tried this out and works great. To @lbwexler's point about whether this should be on all the time, I'm not 100% sure, I could go either way.  Might make sense for this to be off by default, as it was selects with very large option lists where you rarely scroll through the list which prompted the ticket, pretty easy to know if you want this enable and the auto-selection of the text could get distracting on selects with smaller option lists.

peterjcoffman:
@haynesjm42 If you want to enable this on a case-by-case basis, I think it would make more sense to just use the existing `selectOnFocus` prop. This ticket is changing the default. If you're happy with an opt-in behavior, that is already available.

haynesjm42:
Ok I am going back to my original stance here, if you are setting up a select which you can filter, then you are expecting there to be enough options in the list that the user would want to be able to narrow it down by typing in the input. Same with allowing create I think. I like this as default behavior, but will leave it to others to make the final call.

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 
 [Commit Log](https://github.com/xh/hoist-react/compare/v30.0.0...develop)
 
+### ⚙️ Technical
+
+* `Select` will now default to selecting contents on focus if in filter or creatable mode.
```
amcclain: This needs to get shifted up before the commit log link. Agree I think Technical is the correct category for it and that it's worth a one-liner in the changelog.

amcclain:
I like this change and agree it's a better default for when the control is setup for the user to type in input.

It might have the positive side-effect of encouraging users to turn off filter mode for small select lists where they just want users to choose an existing option. I try to remember to do that, as I think having the interactive cursor in those small lookups is non-ideal.

CHANGELOG.md
lbwexler: done

lbwexler:
Thanks for this peter, and thanks for the strong advocacy john :-)

PR #1730, opened 2020-02-27
Relative Timestamp fixes

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/relativetimestamp/RelativeTimestamp.js
+++ b/cmp/relativetimestamp/RelativeTimestamp.js
@@ -133,7 +133,7 @@ function doFormat(opts) {
     const {prefix, equalString, epsilon, allowFuture, short} = opts,
         diff = opts.relativeTo - opts.timestamp,
         elapsed = Math.abs(diff),
-        isEqual =  elapsed < epsilon * SECONDS,
+        isEqual = elapsed <= (epsilon ?? 0) * SECONDS,
         isFuture = !isEqual && diff < 0;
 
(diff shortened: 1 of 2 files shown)


cmp/relativetimestamp/RelativeTimestamp.js
```diff
@@ -133,7 +133,7 @@ function doFormat(opts) {
     const {prefix, equalString, epsilon, allowFuture, short} = opts,
         diff = opts.relativeTo - opts.timestamp,
         elapsed = Math.abs(diff),
-        isEqual =  elapsed < epsilon * SECONDS,
+        isEqual = elapsed <= (epsilon ?? 0) * SECONDS,
```
haynesjm42: Why the `?? 0` here? We default it to 10 above, do we expect someone to pass `null`?

lbwexler:
Particularly concerned about the 'undefined'
yes, easy to do when you have the code (like our client app) that sometimes
passes one, sometimes, doesn't.

haynesjm42:
But we default it to `10` in `getRelativeTimestamp`, so if no `epsilon` is passed then we are covered there, not sure this is a case we really need to worry about? It's not going to hurt anything of course, but should we at least be defaulting to `10` instead of `0` to be consistent?

lbwexler:
I want someone who tried to turn 'epsilon off' with an explicit setting to null or undefined to get zero.  That is wha they are expecting

Issue #1720, opened 2020-02-26
Add RefreshMode support to mobile navigator


TomTirapani:
When refreshing a mobile app, I would expect pages deeper in the stack to wait until they become active again to refresh. Currently, they all refresh at once.

peterjcoffman:
Do mobile pages have a RefreshMode at all? It looks to me like RefreshMode is only used on TabContainerModel and DashContainerModel.

TomTirapani:
No, they don't. I'm saying they should - will update the title to clarify

TomTirapani:
PR here: https://github.com/xh/hoist-react/pull/1732

PR #1732, opened 2020-02-28
Add `RefreshMode` and `RenderMode` support to mobile navigator.

TomTirapani:
Addresses issue #1720

See Toolbox PR: https://github.com/xh/toolbox/pull/375

Hoist P/R Checklist
-------------------


mobile/cmp/navigator/NavigatorPageModel.js
```diff
@@ -34,26 +40,55 @@ export class NavigatorPageModel {
         return stringify({id, props}, qsOpts);
     }
 
+    get renderMode() {
+        return this._renderMode || this.navigatorModel.renderMode;
```
lbwexler: we should prefer nullish coalesce ('??') here (and I realize this code is copied from some similar objects)

lbwexler:
This is great, and looks a lot like TabModel.  Would we consider pushing that analog even further in the naming?  Specifically, would we consider renaming NavigatorPageModel to just `PageModel` and have the `routes` argument to NavigatorModels ctor renamed  to `pages`;

TomTirapani:
Thanks for taking a look. Great idea about the naming - will make those updates. My main concern there is that this makes it a breaking change, but it will be quite obviously broken and fairly trivial to update apps.

lbwexler:
could potentially maintain "routes" as a deprecated name.   But I think
could also sell this as an enhancement worth retyping that one key.  Don't
think anything else likely to be breaking.

TomTirapani:
Note that navigator already exposes an observable `pages[]`, which is distinct from routes. (routes being the "possible" pages and how they map to Router5, and pages being the instances of the pages currently in the stack)

TomTirapani:
I've implemented the refactor as discussed yesterday with Lee. Please see new Toolbox PR required for this branch: https://github.com/xh/toolbox/pull/375

Issue #1717, opened 2020-02-26
Down Caret in Select is a bit bulky


lbwexler:
Seems like the down caret  affordance in the Select is a bit wide, taking up precious space in crowded toolbars.  If possible would like to slim it down, giving some more room for content.

amcclain:
Note we do have a `Select.hideDropdownIndicator` prop if you want to zap it completely (as we've done in the past)

TomTirapani:
Took a quick go at this. Managed to reduce the width a bit, but I'm not sure how much more we can without it looking wonky. I think in practice we may be more inclined to just hide the indicator when we deem it problematic.
https://github.com/xh/hoist-react/pull/1740

lbwexler:
I think this is a material improvement and looks better too.  Thanks

PR #1740, opened 2020-03-02
Make dropdown indicator in Select a bit less bulky.

TomTirapani:
This change reduces the dropdown width from 31px > 20px. The idea was to make it look consistent with our other right-align input actions, like clear.

Screenshot 2020-03-02 at 10 24 34

vs. current prod:

Screenshot 2020-03-02 at 10 24 46

Don't know if there's too much more we can actually do here. See issue: #1717

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/Select.js
+++ b/desktop/cmp/input/Select.js
@@ -6,7 +6,7 @@
  */
 import {HoistInput} from '@xh/hoist/cmp/input';
 import {box, div, hbox, span} from '@xh/hoist/cmp/layout';
-import {elemFactory, HoistComponent, LayoutSupport, hoistCmp} from '@xh/hoist/core';
+import {elemFactory, HoistComponent, LayoutSupport} from '@xh/hoist/core';
 import {Icon} from '@xh/hoist/icon';
 import {
     reactAsyncCreatableSelect,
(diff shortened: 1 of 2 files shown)


desktop/cmp/input/Select.js
```diff
@@ -257,8 +264,7 @@ export class Select extends HoistInput {
         if (props.hideDropdownIndicator) {
             rsProps.components = {
```
lbwexler: Since we are redoing this code to always install `components`  think it would be much simpler to just consult this prop on line 226 with a ternary.

lbwexler: (and of course we need to spread in ...rsProps.components  on line 226 as well -- that would have been a regression here when the dropdownIndicator was not hidden)

desktop/cmp/input/Select.js
```diff
@@ -547,3 +553,7 @@ export class Select extends HoistInput {
 
 }
 export const select = elemFactory(Select);
+
```
lbwexler: At the risk of contradicting myself, vis a  vis cost of components, I will say that I would just inline this particular framework component above. If not, at least minimize by removing its model and observability.  

This is a framework component, we will have many of them in the world, and might as well keep it tight!  Also, should be able to get a better display name from the Icon component itself over a new anonymous component (although we may have to fix that)

PR #1741, opened 2020-03-02
Form fix

lbwexler:
Note:  A side-effect of this change is that rules on a field are no longer observable and can no longer be changed after construction.

It did not seem like we had a public way of doing this (e.g. addRule) -- an application would have had to be re-writing this property directly in an action.  But I wasn't sure.  Dynamic rules were certainly something we had considered in the past.

Hoist P/R Checklist
-------------------

- [] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.

### DIFF
--- a/cmp/form/field/BaseFieldModel.js
+++ b/cmp/form/field/BaseFieldModel.js
@@ -97,7 +97,7 @@ export class BaseFieldModel {
             this.computeValidationAsync();
         });
         this.addAutorun(() => {
-            if (this.isDirty) this.displayValidation(false);
+            if (this.isDirty && !this.isValidationPending) this.displayValidation(false);
         });
     }
 
@@ -261,7 +261,7 @@ export class BaseFieldModel {
         return this
             .evaluateAsync(this.rules)
             .thenAction(errors => {
-                if (runId == this._validationRunId) {
+                if (runId === this._validationRunId) {
                     this.errors = errors;
                 }
                 return this.validationState;
--- a/cmp/form/field/BaseFieldModel.js
+++ b/cmp/form/field/BaseFieldModel.js
@@ -52,7 +52,9 @@ export class BaseFieldModel {
     @observable _disabled;
     @observable _readonly;
 
-    // An array with the result of evaluating each rule (null for not complete, or list of errors)
+    // An array with the result of evaluating each rule.  Each element will be array of strings
+    // containing any validation errors for the rule.  If validation for the rule has not
+    // completed will contain null
     @observable _errors;
 
     @managed
--- a/cmp/form/field/BaseFieldModel.js
+++ b/cmp/form/field/BaseFieldModel.js
@@ -204,7 +204,7 @@ export class BaseFieldModel {
     /** Validation state of the field. */
     @computed
     get validationState() {
-        return this.deriveValidationState()
+        return this.deriveValidationState();
     }
 
     /** @member {boolean} - true if this field is confirmed to be Valid. */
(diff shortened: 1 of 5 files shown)


lbwexler:
@amcclain -- adding you as I know you are deep in the think of forms.  Perhaps we could go over this change, and discuss any issues you have discovered in your recent client work using these forms.

cmp/form/field/BaseFieldModel.js
```diff
@@ -138,9 +140,16 @@ export class BaseFieldModel {
...
+    @computed
+    get errors() {
+        return compact(flatten(this._errors));
+    }
+
+
     /** @member {String[]} - all validation errors for this field and its sub-forms. */
     get allErrors() {
-        return this.errors || [];
+        return this.errors;
```
amcclain: Are these intentionally the same, i.e. are we maintaining both of these getters to avoid a breaking API change? The different docs make me think there should be a difference.

cmp/form/field/BaseFieldModel.js
```diff
@@ -162,9 +171,9 @@ export class BaseFieldModel {
 
         // Force an immediate 'Unknown' state -- the async recompute leaves the old state in place until it completed.
         // (We want that for a value change, but not reset/init)  Force the recompute only if needed.
```
amcclain: This comment is a bit difficult to grok - I admittedly can't follow what it's saying.

cmp/form/field/BaseFieldModel.js
```diff
@@ -48,9 +46,15 @@ export class BaseFieldModel {
      */
     @observable validationDisplayed = false;
 
+    //----------------------
+    // Implementation State
+    //----------------------
     @observable _disabled;
     @observable _readonly;
 
+    // An array with the result of evaluating each rule (null for not complete, or list of errors)
```
amcclain: Could we clarify the contents here a bit more?  `null` if the rule has not run, a `string` error if there's a problem, but what kind of element if the rule passes OK?  Empty string?  (If that's the case seems a bit weird TBH, although since it's not the public API I guess that's fine.)

cmp/form/validation/Rule.js
```diff
@@ -31,16 +31,12 @@ export class Rule {
      * Compute current set of errors (if any) for this rule
      */
```
amcclain: Could we please add the return type?

cmp/form/validation/constraints.js
```diff
@@ -23,8 +23,8 @@ import moment from 'moment';
 export const required = ({value, displayName}) => {
     if (
         isNil(value) ||
-        (isString(value) && value.trim().length == 0) ||
-        (isArray(value) && value.length == 0)
+        (isString(value) && value.trim().length === 0) ||
+        (isArray(value) && value.length === 0)
```
amcclain: Is the IDE inserting these `===` everywhere?  I don't really mind them, but curious as I've been seeing more and more of these diffs.

cmp/form/field/BaseFieldModel.js
lbwexler: Hi -- there was always an observable `errors` property,  just now implemented as a computed getter for technical reasons.  Differs from `allErrors` as documented, the former includes sub-form errors for this field as well as errors on
the field itself.

cmp/form/validation/constraints.js
lbwexler: no, but it is recommending them, and I am adding manually as appropriate.

cmp/form/field/BaseFieldModel.js
amcclain: But `allErrors` just returns `errors`....

lbwexler: Not in the sub-class SubformsFieldModel

PR #1742, opened 2020-03-02
Don't track width of non-resizable columns

peterjcoffman:
Hoist P/R Checklist
-------------------


peterjcoffman:
I currently remove the width of non-resizable columns in two places:

1. in `columnReaction`, i.e. remove width before saving to local storage.

2. in `updateGridColumns`, i.e. remove width after loading from local storage but before loading into Grid

This is redundant. It is needed to do this check in `updateGridColumns` because there may be old, malformed preferences in local storage once we deploy this change, but if we check the width after loading the state there is no need to check it before saving it.

However, I thought it best to avoid saving bad data even if it won't do any harm.

cmp/grid/GridStateModel.js
```diff
@@ -189,9 +191,24 @@ export class GridStateModel {
...
+     * @returns A copy of columnState with width removed from non-resizable columns.
+     */
+    removeWidthOfResizable(columnState) {
+        const cols = this.gridModel.getLeafColumns();
+
+        return columnState.map(state => {
+            const col = this.gridModel.findColumn(cols, state.colId);
+            return col.resizable ? state : omit(state, 'width');
+        });
+    }
```
amcclain: A few small comments here:
* Believe this method name is backwards - we are stripping width from *non*-resizable cols 😄 
* But a more general name would be better anyway - `preprocessColumnState` or `cleanColumnState`.  It could then both remove any non-existent cols and strip the widths. That would have the added benefit of it ensuring it was internally defending against missing columns vs. relying on the caller calling it after stripping null cols.
* I use the full jsdoc comment style heavily in code I write (as you might have noticed...) but only for public methods on a class. Our standard is to avoid it for implementation comments, precisely because it makes them look like public APIs. (I also primarily add those docs for the types, FWIW, which we don't have set here anyway.)

peterjcoffman: I have updated this to reflect your comments.

amcclain:
Thanks for updating this.  

I need to look carefully and really think this through - that will take a little bit of time for me to get to.  This is a complex model w.r.t. the different places and modes in which it both reads and sets state, and I want to ensure I have a good model in my head for how it fits together, and when/where we should be processing that state.

Will loop back here when I can - thanks again for getting this setup for review.

Issue #1736, opened 2020-02-29
Enhancement to "autosize columns" functionality


lbwexler:
+ Need to clarify/simplify best practices for doing this programmatically.  Specifically, when is agGrid ready, after being presented with new data?  How to deal with data in virtualized rows?

+ Need an affordance to allow users to do this on mobile.  Specifically, we discussed:
  - double-tap on header to resize individual column. 
  - button or option on column chooser to resize columns.

amcclain:
@cnrudd also reports that the following CSS override improves the reliability of autosizing in some client apps, specifically that it helps ensure fewer values remain truncated:

```
.with-autosize-columns {
    .xh-ag-grid {
        .ag-row {
            .ag-cell {
                overflow: unset;
                .ag-cell-wrapper {
                    .ag-group-value {
                        overflow: unset;
                    }
                }
            }
        }
    }
}
```

He suggested investigating with ag-grid to see if this points to a bug or similar there.

TomTirapani:
Taking a look at the mobile affordances for this.

The double tap / click on header works well, and I think would be a clear win for both mobile and desktop.

Adding a button the mobile column chooser isn't quite as good - it seems that in practice, due to a combination of small screens and column virtualisation, it doesn't actually have much of an effect. Newly added columns are likely to be off-screen, and I think maybe adding the affordance would be a case of over-promising. Perhaps best to just require the user to double-tap the columns they want to resize, where it is guaranteed they won't be virtualised?

TomTirapani:
Of course, if we could find someway to make it work for virtualised columns / rows, then that's a different story :)

lbwexler:
Seems like adding this affordance would be a clear quick win.  Great idea
Tom.

-Lee

TomTirapani:
Opened this PR: https://github.com/xh/hoist-react/pull/1757 for the column header double-click functionality

lbwexler:
@TomTirapani - know we have added the double tab affordance here, and also just recently added a button affordance as discussed throughout a client app.

Would it be useful to add a "drop-in" button to the framework -- ala ColumnChooserButton.

TomTirapani:
Sure thing, seems like an easy win - pr here: https://github.com/xh/hoist-react/pull/1863

lbwexler:
Thanks Tom -- Think we fully addressed this ticket for now!

PR #1757, opened 2020-03-09
Autosize columns on column header double-click / double tap.

TomTirapani:
See issue #1736

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnHeader.js
+++ b/cmp/grid/impl/ColumnHeader.js
@@ -5,6 +5,7 @@
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
 
+import {XH} from '@xh/hoist/core';
 import {div, span} from '@xh/hoist/cmp/layout';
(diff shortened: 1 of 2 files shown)


cmp/grid/impl/ColumnHeader.js
```diff
@@ -121,24 +126,50 @@ class LocalModel {
     }
 
     onClick = (e) => {
```
lbwexler: should we be more explicit about what handlers we expect to be registered/running in a mobile context, and what in a desktop context?  Just looking at this code,  you might think there is some risk of these things getting called twice.

cmp/grid/impl/ColumnHeader.js
```diff
@@ -152,4 +183,9 @@ class LocalModel {
             return {colId, sort: 'asc', abs: false};
         }
     }
+
+    autosize() {
+        if (!this.gridModel) return;
+        this.gridModel.autoSizeColumns(this.colId);
```
lbwexler: convenient place to this.gridModel?.autoSizeColumns()

PR #1766, opened 2020-03-13
Use Icon.x() for select clear indicator

amcclain:
Hoist P/R Checklist
-------------------


amcclain:
Hi @TomTirapani - can you take a look at this whenever you have a few spare minutes and let me know if you think it's an improvement?

If you add `enableClear: true` to the Toolbox > Forms > FormModel text and select fields it lets you compare the two and see how I was trying to match them better.

The new X icon is at once more consistent but also undeniably larger. Made me wonder if we need/want to use the large font size for the BP controls. but I really didn't want to revisit that now.

I don't feel strongly about this at all but it was bugging me on an app form I was building how different the clear indicators were across the input types.

desktop/cmp/input/Select.js
```diff
@@ -515,6 +516,18 @@ export class Select extends HoistInput {
             () => Icon.chevronDown({className: 'xh-select__indicator'});
     }
 
+    // As per example @ https://react-select.com/components#replaceable-components
```
lbwexler: just curious about the asymmetry with our custom dropdown indicator above.    Are we potentially missing something there with respect to refs, and additional props?

amcclain: I had the same question - I can tell you the clear indicator required the innerProps to be spread on the div as that's what wired up the correct / provided click handler.

lbwexler:
I think its fine -- looks like the Dropdown example they show is actually
simpler and does not need that ref.

Will merge now

TomTirapani:
Looks great, much better than the inconsistent tiny one before :)

I noticed that the vertical alignment of the clear indicator was a bit 'off', because of the outer wrapping div - pushing a tiny styling tweak to line it up better with the other text and the dropdown indicator.

Before:
Before

After:
After

Issue #1769, opened 2020-03-13
HeaderItems in collapsible panels collapse the panel on double click


petradish:
Header items in collapsible panels seem to inherit the onDoubleClick toggle function from the panelHeader component.
If a button or column chooser is double clicked, the panel collapses immediately. (The column chooser onRowDoubleClicked still preserves its functionality of moving columns, however, before collapsing the panel.)

PR #1770, opened 2020-03-13
Stopped event propagation in panel header items

petradish:
Co-dev: John and Petra
This PR addresses issue https://github.com/xh/hoist-react/issues/1769

Hoist P/R Checklist
-------------------

- [N/R] Added CHANGELOG entry, or determined not required.
- [N/R] Updated doc comments / prop-types, or determined not required.
- [N/R] Created Toolbox branch / PR, or determined not required.

### DIFF
--- a/desktop/cmp/panel/impl/PanelHeader.js
+++ b/desktop/cmp/panel/impl/PanelHeader.js
@@ -45,8 +45,13 @@ export const panelHeader = hoistCmp.factory({
                             item: title
                         }) :
                         filler(),
-                    ...(!collapsed ? headerItems : []),
-                    collapseButton({panelModel})
+                    hbox({
+                        className: 'xh-panel-header__items',
+                        items: [...(!collapsed ? headerItems : []),
+                            collapseButton({panelModel})
+                        ],
+                        onDoubleClick: (e) => e.stopPropagation()
+                    })
                 ],
                 onDoubleClick
             });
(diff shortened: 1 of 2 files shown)


desktop/cmp/panel/impl/PanelHeader.js
```diff
@@ -45,8 +45,13 @@ export const panelHeader = hoistCmp.factory({
                             item: title
                         }) :
                         filler(),
-                    ...(!collapsed ? headerItems : []),
-                    collapseButton({panelModel})
+                    hbox({
```
lbwexler: Why is the double click on the grid column chooser even getting here?  
Does that mean we are doing something wrong in our handling of the  column chooser or the popover? 

Seems to me that:

-If the column chooser is using the double click, it should certainly consume it. 
-The popover should not be sending it

petradish: I think the issue is that all items in the panel header are valid targets for the panel header's `onDoubleClick` event handler, including items currently passed in through the `headerItems` prop. So no matter what item lies there in headerItems, whether it's something static like text or dynamic like a button or a column chooser, we don't want `onDoubleClick` to propagate to the parent. So we are now wrapping the headerItems in its own div and stopping any double click events from toggling the panel.

lbwexler:
Discussed with John and Petra.  We will merge this improvement to the PanelHeader, but also make another ticket to handle the fact that Grid ColumnChooser panel is leaking these events in the first place.

PR #1771, opened 2020-03-14
Add HoistAppModel.preAuthInitAsync()

amcclain:
Hoist P/R Checklist
-------------------


core/ExceptionHandler.js
```diff
@@ -150,8 +150,8 @@ export class ExceptionHandler {
         ret.showAlert = ret.showAlert ?? (!isAutoRefresh && !isFetchAborted);
         ret.requireReload = ret.requireReload ?? false;
 
-        ret.title = ret.title ?? (ret.showAsError ? 'Error' : 'Message');
-        ret.message = ret.message ?? e.message ?? e.name ?? 'An unknown error occurred.';
+        ret.title = ret.title || (ret.showAsError ? 'Error' : 'Message');
+        ret.message = ret.message || e.message || e.name || 'An unknown error occurred.';
```
lbwexler: uugh -- great fix

core/HoistAppModel.js
```diff
@@ -33,6 +33,14 @@ export function HoistAppModel(C) {
              */
             async initAsync() {},
 
+            /**
+             * Hoist will call this method early in the initialization sequence, prior to user
+             * authentication. This means that several core services (identity, configs, prefs)
+             * will *not* be available, but it provides the app a hook to do early service
+             * initialization or other work to support flows such as OAuth.
+             */
+            async preAuthInitAsync() {},
```
lbwexler: Is throwing/rejecting here an appropriate response, if there is an auth problem?  Or should the app let the framework fail by the server on line 512 below?

Seems like it might be worth being a bit more explicit about any responsibilities for this method if we are not on the happy path...

svc/IdentityService.js
```diff
@@ -73,6 +73,11 @@ export class IdentityService {
      * and refreshes the application to present a login panel.
      */
     async logoutAsync() {
+        try {
```
lbwexler: I find the mix of async/exception styles in this method a little bit confusing.   Also, why should't an app exception during logout just be surfaced to the user in the same way a framework exception would be?  Seems like we add the call to logoutAsync to the top of the promise chain.

PR #1779, opened 2020-03-17
Do not save/restore width of non-resizable columns

### LINKED ISSUE #1718: GridState should not restore width of !resizable columns
I don't think that grid state should remember or apply widths to columns marked as `resizable: false`.

I had a grid with both grid state enabled and an action column, sized via `calcActionColumnWidth()` and taking `resizable: false` from the actionCol defaults.

I removed an action - a feature went away in this case - but the column stayed at the old width, which was confusing and then impossible for me as the user to fix except by resetting my grid state.

amcclain:
+ Fixes #1718

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridStateModel.js
+++ b/cmp/grid/GridStateModel.js
@@ -4,9 +4,9 @@
  *
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
-import {XH, HoistModel} from '@xh/hoist/core';
-import {cloneDeep, debounce, find, remove, isUndefined} from 'lodash';
+import {HoistModel, XH} from '@xh/hoist/core';
 import {start} from '@xh/hoist/promise';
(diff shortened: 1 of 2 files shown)


cmp/grid/GridStateModel.js
```diff
@@ -189,8 +176,37 @@ export class GridStateModel {
...
 
     //--------------------------
-    // Helper
+    // Other Implementation
     //--------------------------
+    cleanColumnState(columnState) {
+        const {gridModel} = this,
+            gridCols = gridModel.getLeafColumns();
+
+        let cleanedState = [...columnState];
```
lbwexler: nit -- no need to clone here, with the filter below.

lbwexler:
Great change.  Made some minor tweaks as I was reviewing

PR #1784, opened 2020-03-18
Support menuDividers in appMenuButton.extraItems, Fixes #1628

### LINKED ISSUE #1628: Support menuDividers in appMenuButton.extraItems
Currently there is no way to put a menu divider into the set of extra items passed to `appBar.appMenuButtonProps.extraItems` - this only accepts plain objects and passes each to the `menuItem` elemFactory.  

At the least `appMenuButton` should check to see if any extra item is a component vs. plain object and handle accordingly. 

Better still would be to support the string token `'-'` as a shortcut to add dividers, although I don't know if we do this consistently elsewhere.

febbraiod:
Hoist P/R Checklist
-------------------


desktop/cmp/button/AppMenuButton.js
```diff
@@ -19,15 +19,32 @@ export const [AppMenuButton, appMenuButton] = hoistCmp.withFactory({
...
+            if (it.type.name === 'MenuDivider') {
+                // Remove starting separators
+                if (idx === 0) return false;
+
+                // Remove consecutive separators
+                const prev = idx > 0 ? arr[idx - 1] : null;
+                if (prev?.type.name === 'MenuDivider') return false;
+            }
+            return true;
+        });
```
febbraiod: This will take care of manually added consecutive dividers, but the way the code is for the hard coded items below, it's possible that all of the menuItems in a given section will be 'omit'ted and will then result in double dividers. Do we want to move away from using 'omit' and manually build the items to avoid this?

TomTirapani: Why not concatenate extra items and the default items below first, and then run the code to remove consecutive separators on that?

TomTirapani: i.e. exactly what you suggested :)

febbraiod: Yeah will also require the extra step of not including 'omitted' items, since we wont get that magic here

desktop/cmp/button/AppMenuButton.js
```diff
@@ -17,74 +17,119 @@ export const [AppMenuButton, appMenuButton] = hoistCmp.withFactory({
     className: 'xh-app-menu',
 
     render(props) {
-        let {className, hideOptionsItem, hideFeedbackItem, hideThemeItem, hideAdminItem, hideImpersonateItem, hideLogoutItem, extraItems, ...rest} = props;
-        extraItems = extraItems ?
-            [...extraItems.map(m => menuItem(m)), menuDivider()] :
-            [];
+        let {className, hideOptionsItem, hideFeedbackItem, hideThemeItem, hideAdminItem, hideImpersonateItem, hideLogoutItem, extraItems, ...rest} = props,
```
TomTirapani: I think if we default `extraItems` to an empty array, we can skip the ternary logic below and always include them + menu divider (with the knowledge the divider will be removed if no items provided).

TomTirapani:
Looks good, but I think if `extraItems` includes omits we can still end up with consecutive dividers.

`extraItems` seems to be configs, and we convert them to menuItems before adding. If we also made the default items into configs, we could do something like this:

```
[...extraItems, ...defaultItems]
.filter(it => !it.omit)
.map(it => ... convert to menu item / divider)
.filter(it => ... remove dupe dividers)
```

TomTirapani:
Also, we have similar 'remove duplicate divider' logic in ContextMenu.js. Perhaps we could extract that to a util and reuse?

febbraiod:
Good Idea, the remove dupe logic in CM doesn't account for separators being passed as elements, so world be good to have a util that knows about both and use that in both places. TY

febbraiod:
Turns out context menus don't support menuDividers as items, so that isn't a concern but util probably still worth it for DRY purposes?

TomTirapani:
Expanded the scope of this to standardise how we handle separators and everywhere (App Menu, ContextMenu, Grid's context menu and Toolbars)

lbwexler:
Can I review and merge this?

lbwexler:
Reviewed,  looks like a great change

I think those utility functions arguably don't really below in 'utils/js'?   Think they are perhaps closer to 'utils/react'.

amcclain:
Hi @lbwexler and @febbraiod  - 

Is there an element of this that is a breaking change?  I am getting a new error in my dev tools console - see warning and app menu code snippet below - 

Screen Shot 2020-03-23 at 9 11 32 AM

Screen Shot 2020-03-23 at 9 11 56 AM

febbraiod:
Hi. Just reviewing the latest on this change at this very second. Glad to looks at what is going on here.

amcclain:
Fixed in #1789 and merged just now - thanks @febbraiod !

PR #1786, opened 2020-03-18
Consistent ClassName placement on Panel dom elements

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 * `GridStateModel` no longer saves/restores the width of non-resizable columns.
   [#1718](https://github.com/xh/hoist-react/issues/1718)
+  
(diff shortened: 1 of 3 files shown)


lbwexler:
Tom -- Let me know if you think this is reasonable, or risky in some way given potential existing application overrides? 

I think this is an example of us trying to being overlying frugal with avoiding extra dom elements.  We should not change the structure of the dom tree more than needed between different component  configurations

TomTirapani:
Looks good, and definitely makes sense to use a similar DOM structure / class names across the different configurations.

I think this change should be documented in the CHANGELOG - Panel is such a core component, and I think there are quite a few cases where we're relying on the old structure / class names for styling. We should also audit hoist-react / toolbox styling to fix any broken styles there.

Issue #1310, opened 2019-08-13
PanelModel should delay loadAsync() calls when collapsed


inewell:

amcclain:
Seems like this could be a `PanelRefreshContextModel` similar to https://github.com/exhi/hoist-react/blob/develop/cmp/tab/impl/TabRefreshContextModel.js

lbwexler:
Seems like we just want to add support for refreshMode to PanelModel?   Hoping this will fall right into place!

TomTirapani:
Added support for RefreshMode to PanelModel. Also DRYed up our duplicate RefreshContextModels across the various APIs that use RefreshMode.

See PRs:
https://github.com/xh/hoist-react/pull/1787
https://github.com/xh/toolbox/pull/382

PR #1787, opened 2020-03-19
PanelModel supports RefreshMode.

TomTirapani:
Also, all models which implemented RefreshMode use the new RefreshModeContextModel.

See Issue: https://github.com/xh/hoist-react/issues/1310
See Toolbox PR: https://github.com/xh/toolbox/pull/382

Hoist P/R Checklist
-------------------


desktop/cmp/panel/PanelModel.js
```diff
@@ -47,6 +51,13 @@ export class PanelModel {
     /** Is this panel currently resizing? */
     @observable isResizing = false;
 
+    /**
```
lbwexler: no need for comment

desktop/cmp/panel/impl/ResizeContainer.js
```diff
@@ -37,12 +38,15 @@ export const resizeContainer = hoistCmp.factory({
         const cmp = vertical ? vbox : hbox,
             maxDim = vertical ? 'maxHeight' : 'maxWidth';
 
-        return cmp({
-            ref,
-            className,
-            flex: 'none',
-            [maxDim]: '100%',
-            items
+        return refreshContextView({
```
lbwexler: My only concern here is that this refreshContext only exists for a resizable/collapsible panel. Maybe that is what we want, but we got burned by overly branching the structure between the two types of panel before.....

lbwexler:
This is really awesome, and a nice DRY up.  A couple of thoughts:

1)  Would it be worth calling this a LazyRefreshContextModel?  Realize it is not always lazy, but that is the default, and is its main reason for existence.  Could also be called a ManagedRefreshContextModel or an InactiveContentRefreshContextModel.

2)

desktop/cmp/panel/impl/ResizeContainer.js
TomTirapani: Good point - will move to the main panel component, would be less surprising

TomTirapani:
Updated this branch with comments from the code review - thanks!

Will also add the Toolbox example shortly

lbwexler:
Found some issues in this we need to work out.  Seems like getting some strange behavior with registration of models within this context.  For example, notice the refresh buttons in toolboxes toolbars no longer seem to work

lbwexler:
Update -- not seeing any strange behavior -- that was a user error.  Have been discussing with Tom, however, and have the following thoughts:

1) Definitely not worth the complexity of all of the additional refresh context models to insert these for non-collapsible panels.

2) Might not even be worth it in the case of collapsible panels.  We need to discuss.

lbwexler:
I would be inclined to merge this now.  I think its a nice cleanup/standardization.  We don't have a strong example yet of the use of lazy refresh in the collapsed panel, but it seems analogous to TabContainer and its nice to bring these all in to line.

TomTirapani:
Great. I've got a Toolbox example in the Portfolio app cooking - will link here once available

PR #1790, opened 2020-03-23
WIP - Empty group values

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -14,6 +14,7 @@ import {
     find,
     get,
     isArray,
+    isEmpty,
     isFinite,
     isFunction,
     isNil,
@@ -101,6 +102,8 @@ export class Column {
      * @param {Column~setValueFn} [c.setValueFn] - function for updating Record field for this
      *      column after inline editing.
      * @param {Column~getValueFn} [c.getValueFn] - function for getting the column value
+     * @param {string} [c.emptyGroupedValue] - placeholder for rows with empty
+     *      values used to allow ag-grid to group these rows
      * @param {boolean} [c.enableDotSeparatedFieldPath] - true (default) to enable configuration
      *      of field name as a dot-separated path - e.g. `'country.name'` - where the default
      *      `getValueFn` will expect the field to be an object and render a nested property.
@@ -153,6 +156,7 @@ export class Column {
         editable,
         setValueFn,
         getValueFn,
+        emptyGroupedValue,
         enableDotSeparatedFieldPath,
         agOptions,
         ...rest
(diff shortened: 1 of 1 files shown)


cmp/grid/columns/Column.js
```diff
@@ -385,8 +391,11 @@ export class Column {
             setElementRenderer(class extends Component {
                 render() {
                     const agParams = this.props,
-                        {value, data: record} = agParams;
-                    return elementRenderer(value, {record, column: me, gridModel, agParams});
+                        {value, data: record} = agParams,
+                        isGrouped = gridModel.groupBy.includes(colId);
+
+
+                    return isGrouped && emptyGroupedByValue && isEmpty(value) ?  elementRenderer(emptyGroupedByValue, {record, column: me, gridModel, agParams}) : elementRenderer(value, {record, column: me, gridModel, agParams});
```
febbraiod: Not working yet

cmp/grid/columns/Column.js
```diff
@@ -453,6 +462,24 @@ export class Column {
         if (isArray(fieldPath)) return get(record.data, fieldPath);
         return record.data[fieldPath];
     };
+
+    generateEmptyGroupGetValueFn(getValueFn) {
+        return (params) => {
+            const v = getValueFn(params),
+                isGrouped = params.gridModel.groupBy.includes(this.colId);
+            return isGrouped && isEmpty(v) ? this.emptyGroupedValue : v;
```
febbraiod: IsEmpty(v) or v == null?

febbraiod:
Pausing work on this at Lee's request for Anselm and him to review.

lbwexler:
Thanks Don!

John -- Can you look at this one and give some thoughts?  

Is the cost-benefit worth it on this one?

haynesjm42:
How often have we run into this specific use-case where you can't just deal with this in `processRawData` because you absolutely can't have your nil/empty field values be set to a placeholder value, and don't want to see that placeholder in normal columns, only in group rows?

What the app would need to do to handle this specific case themselves really isn't that bad, and we could just add an example of dealing with unbalanced groups to toolbox.
- Add a `getValueFn` on your column which returns your placeholder value when the field value is empty - this will be used for grouping _and_ for rendering in the group cell for the autoGroupColumn
- Add a `renderer` to your column which checks if the value is equal to your placeholder value, and return null or an empty string in that case, otherwise just return the value

This is pretty consistent with what ag-grid recommends for dealing with unbalanced groups: https://www.ag-grid.com/javascript-grid-grouping/#unbalanced-groups

That being said, I think Don's solution is creative and if we can handle this case better than ag-grid, then great - but I do get concerned about adding additional complexity to our Column, and am especially concerned about adding any additional processing to our renderers - we should definitely make sure we are only using the more complex renderers when we need them.

Another potential solution to this would be for us to reverse how we are handling the rendering - we could convert nil values to an empty string in the default `getValueFn` and then also install an innerRenderer for the autoGroupColumn which converts the empty value into the `emptyGroupedValue` in the GridModel

lbwexler:
Thanks for the hard work here Don.

I am going to agree with  John's Assesment here -- I think the risk is that the cure here is worse than the disease.  It complicates a pretty complicated story already.  Let's let apps deal with this by pre-processing their nulls, either before presenting to the grid, or before loading the store its

PR #1794, opened 2020-03-26
Consolidate exports to top-level index files

amcclain:
I find that when I go to import `uses` or `creates` in my IDE (IntelliJ), the IDE adds the import from `@xh/core/modelspec`.

That is fine and works, of course, but I would rather it be imported from `@xh/core` directly to reduce the overall number of import lines, and because we had wanted to standardize on requiring a single level of package nesting for most/common imports. If we only export the inner packages from the top-level index file, the IDE auto-creates the import statements as desired.

This would be a breaking change, of course, and would increase the size of these top-level index files. But it would also let us remove all the nested index files, and in the end I think it would encourage the import statements we want and potentially give us a bit more flexibility in the future to restructure nested packages with fewer breaking changes.

Thoughts?

Hoist P/R Checklist
-------------------


lbwexler:
Totally agree that those internal index files were not intended for app-level imports.  Might we use a convention of, say,  _index.js ?  Just something to consider.

amcclain:
Thanks for taking a look.  I will enter a ticket for the comprehensive change

We can see if the _index.js approach looks helpful. I think it might just be a bit more confusing than it's worth (i.e. not immediately to a developer looking over the code why sometimes we have _index and other times index), but if the listings in the top-level index.js files get outta control that sounds like the best way to wrangle it.

TomTirapani:
THANK YOU for this!!

lbwexler:
agree with all that.  If we organize the top-level index file by
sub-directory, I will be happy!

amcclain:
Hi - sorry, why did we close this and delete the branch? I had asked Petra to continue working on this PR off of the branch/ticket.  It had been opened as a draft, so work was still ongoing.

amcclain:
Nevermind I see she opened a new one entirely.

Issue #1806, opened 2020-03-30
Unsized DockViews should elide their titles to fit inner content width


TomTirapani:
If a DockView is sized naturally (i.e. based on inner content dimensions, without explicit size), it currently derives its width from the widest of the title and the content:

<img width="961" alt="Screenshot 2020-03-30 at 17 34 04" src="https://user-images.githubusercontent.com/3017757/77937736-ac5def00-72ac-11ea-8434-207407b8b1bb.png">

Reported by @petradish, this issue is apparent in a client app, where the DockView are narrow but can have quite long titles. Ideally, the title should elide when it is longer than the inner content would allow.

(Note that we currently support a `collapsedWidth` argument with allows the title to elide when docked - but this doesn't apply when it's not docked.)

TomTirapani:
This one was a bit tricky to figure out - had to use absolute positioning to ensure the size of the header counted for the flexbox calculations, but only when not collapsed.

https://github.com/xh/hoist-react/pull/1807

PR #1807, opened 2020-03-30
Unsized DockViews elide their titles to fit inner content width.

TomTirapani:
See https://github.com/xh/hoist-react/issues/1806

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dock/impl/Dock.scss
+++ b/desktop/cmp/dock/impl/Dock.scss
@@ -39,13 +39,18 @@
   pointer-events: auto;
 
   &__header {
-    align-items: center;
+    position: relative;
     height: var(--xh-title-height-px);
-    padding: var(--xh-title-pad-px);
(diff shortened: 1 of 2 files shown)


amcclain:
That's an interesting trick - thanks for figuring it out, and seems like it might be useful in the future for similar kinds of considerations. Any chance you have a link of any kind that explains the approach? Would love to bookmark myself.

I tested this out in my current client project and it worked well. At first I wasn't sure how often we would have "naturally sized" docked views, but then I realized this was actually convenient in my use case where I have a generic `AppModel.showDockedView` method that could then not need to worry about sizing and have sizing set instead in CSS or by the child component.

Merging now - thanks!

TomTirapani:
Sure thing - here's the part of the spec that states absolute positioned children do not participate in flex layout: https://www.w3.org/TR/css-flexbox-1/#abspos-items

PR #1809, opened 2020-03-30
Dialog

### LINKED ISSUE #861: Window / Dialog component - reduce/remove direct use of Blueprint dialog
We make frequent and direct use of the Blueprint Dialog out of `kit`.  There are some concrete issues with this:
* In general we want to avoid needing to go into kit for core components, as it reduces our flexibility to swap out third-party library comps for alternate implementations.
* Can't use these same components cross-platform.
* BP dialogs make use of inner header/footer components / classes which I find somewhat ambiguous (are they required? do we use them consistently?) and have some styling / sizing differences from our other comps (namely toolbars and Panel headers).
* No support moving / resizing / maximizing.

We could consider:
* Lightweight wrapper around Dialog to avoid the kit imports, or...
* Wholesale replacement of Dialog - could be done via addition of floating / positioned support to Panel.

lbwexler:
This PR is a WIP based on Colin's original PR.

Todo:
+ styling - Smooth out default styling, ensure styling prepared for client override
+ styling -  Fix up styling for all internal dialogs created by AppContainer.
+ API -- confirm default values on model and component.
+ API -- do we want to break out width and height to more closely match other apis?
+ Check that change to default context in HoistCmp makes sense.
+ Test Refresh behavior.
+ Explicit support for z-index in API?

See Manual Tests in Admin -> Tests 

Hoist P/R Checklist
-------------------


lbwexler:
Parting ways with this unmergable  pre-typescript artifact.  Don't see any real value here

PR #1815, opened 2020-03-31
Allow Icons as rightElements in inputs, Fixes #1803

### LINKED ISSUE #1803: NumberInput.rightElement prop unusable
I wanted to use the documented `NumberInput.rightElement` prop to display a right-aligned icon (in my case, a percent sign as my input was holding a rate), but the element did not render as expected. Using a simple text string also did not render properly.

It appears as though this element was getting a zero default width. Attempting to set a width just made things worse. (To replicate, try to do something useful with this prop 😄)

febbraiod:
Hoist P/R Checklist
-------------------

- [NR] Updated doc comments / prop-types, or determined not required.
- [NR] Reviewed and tested on Mobile, or determined not required.
- [NR] Created Toolbox branch / PR, or determined not required.

### DIFF
--- a/cmp/input/HoistInput.scss
+++ b/cmp/input/HoistInput.scss
@@ -12,3 +12,9 @@
     }
   }
 }
+
+.icon-as-right-element {
+    .bp3-input-action {
+        width: 30px !important;
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1279,6 +1279,10 @@ leverage the context for model support discussed above.
   property. The implementation of these properties was generally broken. Use `leftSorted` and
   `rightSorted` instead.
 
+### 🐞 Bug Fixes
+ 
+* Inputs that accept a rightElement prop will now properly display an Icon pass as that element
+
 #### Mobile
 
 * Mobile `Page` has changed - `Pages` are now wrappers around `Panels` that are designed to be used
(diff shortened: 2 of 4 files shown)


cmp/input/HoistInput.scss
```diff
@@ -12,3 +12,9 @@
     }
   }
 }
+
+.icon-as-right-element {
```
febbraiod: Not in love with this name by any stretch

lbwexler:
Thanks Don!  Tom -- can you review and merge?

amcclain:
Hi - I looked at this for a while on the plane as I had been curious what was going on.

See https://github.com/xh/hoist-react/commit/7251f2152c5f0b7eb06483dfb0ebef3ebc8b4c71 for an alternate approach - it led me down a bit of a rabbit hole with how elements are sized and ultimately this bit of code in the BP input that sets up the padding added directly to the input element - https://github.com/palantir/blueprint/blob/0c7c2090fcf0e3e65ee5ab434d623df53912e9ee/packages/core/src/components/forms/inputGroup.tsx#L158

TomTirapani:
Thanks for this Don! Looks good, and solves the issue. :)

However, I think it would be better to take Anselm's alternative mentioned above - the change is more minimal, doesn't require our inputs to check the type of element passed and gets to the root of the issue w.r.t. BP's sizing calculations (and supports any width for icon, should we want to).

I'll close this in favour of that PR, make a couple of tweaks there and merge.

PR #1818, opened 2020-04-01
Fix 'empty flash' due to delayed rendering of framework components.

TomTirapani:
Don't trigger Grid.dataReaction until the ColumnHeaders have been rendered. This allows AgGrid to keep the loading overlay displayed until framework components have been rendered, and prevents a 'flash' that can occur during grid initialisation.

You can see this in action by opening the 'Complex Docked View' with CPU 6x slowdown, and comparing against the same in https://toolbox-dev.xh.io/app/containers/dock

Hoist P/R Checklist
-------------------


cmp/grid/GridModel.js
```diff
@@ -697,6 +699,17 @@ export class GridModel {
         if (colIds.length) this.agColumnApi.autoSizeColumns(colIds);
     }
```
lbwexler, line 701: A bit unfortunate that we don't have a local implementation model where we can hang implementation state like this.

lbwexler:
Seems like a clever workaround.  Wondering if there is a way we can introduce a LocalGridModel to keep this implementation state?  Might be useful going forward to adopt that pattern of seperating the API model from any implementation related state.

cmp/grid/GridModel.js
```diff
@@ -697,6 +699,17 @@ export class GridModel {
...
 
+    /**
+     * Set a flag acknowledging that a framework component has been mounted.
+     * We debounce here catch the end of a collection of components mounted at
+     * the same time (i.e. column header initialisation)
+     */
+    @debounced(50)
+    @action
+    noteFrameworkCmpMounted() {
+        this.frameworkCmpsMounted = true;
```
lbwexler, line 710: why do we want/need to wait/delay until they are all mounted?  Doesn't this just become a no-op for the subsequent calls?

TomTirapani:
Thanks for taking a look Lee - push a couple of cleanups.

Regarding the local model - Grid actually already has a LocalModel, which I agree would be a better candidate to keep this state. The problem is making the Grid's LocalModel accessible to the ColumnHeader. We may be able to put a reference to it on the GridModel and access it that way, but then its no longer really a LocalModel...

lbwexler:
Tom and I just had a convo about this -- we'll move the workaround to the empty text component we already have in the local model, and thereby isolate the fix a bit more to the grid implementation.  
Conceptually it makes sense to have that component ready before rendering!  Even before this agGrid version I had noticed that you would get the default text rendering for a bit even if you had turned off emptyText -- this will fix that.

Issue #1805, opened 2020-03-30
Consolidate exports to top-level index file


amcclain:
We export Hoist dependencies via index.js files in our package/directory tree. We maintain these files at almost all levels of the tree, but expect developers / IDEs to (ideally) import them from the highest level at which they are available.  (This depth of this level varies based on the tree - it's a bit deeper for e.g. `@xh/hoist/desktop/cmp/panel` vs. `@xh/hoist/core`.)

So we intend imports like:
`import {hoistCmp, uses} from '@xh/hoist/core';`

However, we also include index.js files deeper in the tree, so IDEs will find these and suggest imports like:
```
import {hoistCmp} from '@xh/hoist/core';
import {uses} from '@xh/hoist/core/modelspec';
```

This works just fine, but exposes our internal package structure more than intended, and results in more import statements than we really need.

I propose we add index.js files only at the levels of the package tree from which we want apps to import, and have those index files reach into subdirectories as required. See #1794 for an example change to handle the case above. These files already exist - they are the "first" files in their respective trees - we just want to add to them to include all the references from the nested files directly.

Note that on the PR @lbwexler suggests we could use a convention such as `_index.js` files for deeper levels. I'd like to first see if the top-level imports really need that to stay manageable before introducing another artifact, but if we find them getting way too unwieldy to maintain that sounds like the approach to use.

PR #1819, opened 2020-04-01
Consolidate exports

petradish:
This PR addresses: https://github.com/xh/hoist-react/issues/1805
- Moved exports from subdirectories 1 or 2 levels deeper up to top-level export indices, namely: these 4 packages: `@xh/react/core`, `@xh/react/data`, `@xh/react/cmp/grid`, and `@xh/desktop/cmp/grid`
- Updated import statements within hoist-react
- Ran a global optimize imports on hoist-react (sorry for so many file changes!)
- Toolbox branch that updates import statements: https://github.com/xh/toolbox/pull/391
Hoist P/R Checklist
-------------------

- [ N/A ] Updated doc comments / prop-types, or determined not required.


lbwexler:
Couple of places the style-sheets got re-ordered within a file (e.g. cmp/ag-grid/index.js).

Need to make sure this is safe, or revert.  Think that the order of style sheet definition can be signifigant.

lbwexler:
I looked at this and the toolbox branch as closely as I could, and it looks ok.  Definitely some nice pickups and consistency. 

Ideally we would have done the sorting/optimization seperate from removing the index files -- it would have been easier to verify the latter, and just trust intellij for the former.  But not worth redoing now.  

My main concern is with the css imports.  Besides the fact that I like them seperate, there is the real concern that re-ordering them alphabetically will somehow will have a semantic effect that we don't intend.  We should consider keeping the same order, and restoring them to a location below the javascript imports.

(Note this could be an issue with re-ordering anything in kit that might transitively import css, although not sure we need to worry about that).

PR #1824, opened 2020-04-07
Use Column.getValueFn to retrieve cell values for Copy Cell action

### LINKED ISSUE #1823: Copy Cell not working on Column based on field:'id'
From a Client:

> When we try to use the ‘Copy Cell’ option in the context menu, instead of copying the content of the cell it copies ‘undefined’.  This happens specifically for cells belonging to a column with a ‘field’ if ‘id’.
> 
>  
> ‘Copy Cell’ works for all other columns.
> Copying the whole row also gets the values correctly.

It looks to me as if agGrid's native copy-cell implementation is not actually using the column 'getValueFn' but instead trying to pull it right out of the raw data.

This is presumably an agGrid error, but it does again the raise the question as to whether we should go ahead and insert the id directly into our record data.  That would workaround this agGrid and just generally may be what users expect.

haynesjm42:
Fixes #1823 

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/contextmenu/StoreContextMenu.js
+++ b/desktop/cmp/contextmenu/StoreContextMenu.js
@@ -67,7 +67,17 @@ export class StoreContextMenu {
                     hidden: !gridModel,
                     recordsRequired: true,
                     actionFn: ({record, column}) => {
-                        if (record && column) copy(record.data[column.field]);
+                        if (record && column) {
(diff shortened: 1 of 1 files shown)


lbwexler:
Merging this to fix this bug, but we will add a new ticket. It should be much easier to get the column value out of a record!

lbwexler:
Thanks!

PR #1837, opened 2020-04-15
Enhance record data with id

lbwexler:
Please also merge related:

https://github.com/xh/toolbox/pull/393/files

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v33.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+* The object returned by the `data` property on `Record` now includes the record `id`.  This will
+  allow for convenient access of the id with the other field values on the record.
+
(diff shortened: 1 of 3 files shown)


haynesjm42:
Curious what specific use-case prompted this change? I know we have talked about it in the past and decided not to include `id` in the `data` - was it the inconsistency with the `updateData` or `modifyRecords` calls which want an object where the `id` is mixed in with the `data`?

If you really want an `id` in your data, why don't we just allow you to have an `id` field? I see we are still reserving that field name in `Store.parseFields()`, though I don't think we really have any reason to reserve that now.

I'm hesitant to add `id` - which is a Hoist concept - to the `data` - which is the application-defined field values processed from the application raw data. That being said I have no problems if the lack of `id` has just become an annoyance and inconvenience in practice - I haven't used the new stuff much with CRUD type stores yet which I would assume is where the `id` becomes more important to application code.

PR #1840, opened 2020-04-16
Styling fixes to ag-grid popups and column set filter

petradish:
+ Changed z-index of ag-popups to appear over bp-dialogs
+ Applied padding to select all checkbox option in col set filter
Co-dev: Anselm and Petra

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/ag-grid/AgGrid.scss
+++ b/cmp/ag-grid/AgGrid.scss
@@ -334,6 +334,17 @@
   .ag-bl-overlay {
     display: none;
   }
+
+  // Adds padding to select all checkbox option in column set filter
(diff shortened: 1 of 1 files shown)


TomTirapani:
Looks good.

My only concern is if we ever decided to use an agGrid popup in the context of a mobile dialog (i.e. if we replace the mobile column chooser with a grid implementation) - Onsen currently uses `z-index: 20001` for its dialogs. But perhaps that's more of an issue with Onsen, not sure it actually needs to be that high and may be worth seeing if we can normalise the z-indices across toolkits.

PR #1848, opened 2020-04-23
Use `xhExportConfig` config to manage export toasts.

TomTirapani:
See issue: https://github.com/xh/hoist-core/issues/74
See corresponding hoist-core pr: https://github.com/xh/hoist-core/pull/137

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/GridExportService.js
+++ b/svc/GridExportService.js
@@ -37,7 +37,8 @@ export class GridExportService {
 
         if (isFunction(filename)) filename = filename(gridModel);
 
-        const exportColumns = this.getExportableColumns(gridModel, columns),
+        const config = XH.configService.get('xhExportConfig'),
+            exportColumns = this.getExportableColumns(gridModel, columns),
             records = gridModel.store.rootRecords,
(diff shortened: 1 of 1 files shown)


lbwexler:
This looks great, but without a handling of the when the config is not present, will require a more modern version of hoist core to be present that guarantees the presence of the config.

Its easy enough to add the key on the fly of course, but this looks like it has the potential to be a silent export killer.

Maybe wait until hoist core is in the wild for a bit more before merging this one?  there is no rush.

lbwexler:
I will go ahead and merge this, and add to the changelog a mention of it, so app updaters can throw in the config if needed.

TomTirapani:
Thanks for merging Lee. I also expected us to wait until hoist-core was out (especially in client apps) before merging this.

Perhaps we could quickly some fallback handling where we access these config props, so it won't fail if the key is missing.

TomTirapani:
Added a quick PR here to not break exports if they run don't yet have the hoist-core update:
https://github.com/xh/hoist-react/pull/1850

PR #1851, opened 2020-04-24
Add Icon.placeholder()

amcclain:
+ Returns empty div with fa-fw className applied, as well as any optional FA sizing class.
+ Used to take up space where an icon would be - e.g. when aligning menuItems with/without icons.

Hoist P/R Checklist
-------------------

### DIFF
--- a/icon/Icon.js
+++ b/icon/Icon.js
@@ -47,7 +47,7 @@ export const Icon = {
      *      for ag-Grid.
      * @param {...*} [c.rest] - Additional props to be passed directly to underlying
      *      component.  These arguments are ignored when asHtml = true.
-     * @returns {(Element| string)}
+     * @returns {(Element|string)}
      */
     icon({
         iconName,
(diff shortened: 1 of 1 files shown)


amcclain:
FYI - I didn't think this really warrants a changelog entry, but figured it was worth getting in.

This appears to handle the simple case I needed. Let me know pls if you see any issues or need for improvement on it. Thanks!

TomTirapani:
Looks great to me - just as discussed :)

lbwexler:
Hi -- My concern is that it looks like it will work exactly like the other factory functions on Icon, 
namely, that it will take an 'asHtml' property, but that it doesn't. 

Surprise FA doesn't have something built-in that we could use.

amcclain:
Well, I don't think they do.

If you think this is an important case to support the html option will add now.  I don't want to keep a one-off workaround in the app.

amcclain:
FWIW it's also documented as only supporting the option it supports but i will add asHtml

lbwexler:
I also spent some time looking at FA, and was considering how to use transparency to get the same efffect. 

Think this is a big improvment, and great to have.

PR #1872, opened 2020-05-01
Fix issue with incorrect re-use of parent pointers in Store.loadData

lbwexler:
Hoist P/R Checklist
-------------------

- [x ] Caught up with `develop` branch as of last change.

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -20,7 +20,7 @@ import {
     isString,
     remove as lodashRemove
 } from 'lodash';
-import {warnIf, withDefault} from '../utils/js';
+import {warnIf} from '../utils/js';
 import {Field} from './Field';
 import {RecordSet} from './impl/RecordSet';
 import {Record} from './Record';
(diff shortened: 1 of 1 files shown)


data/Store.js
```diff
@@ -697,15 +702,7 @@ export class Store {
...
 
         // Note idSpec run against raw data here.
-        const id = this.idSpec(raw),
-            rec = this.getById(id);
-
-        if (rec) {
-            // We are creating a record for an update. Lookup our parent record and determine if
-            // this is the the summary record based on the current state (if not provided).
-            parent = withDefault(parent, rec.parent);
-            isSummary = withDefault(isSummary, rec.id === this.summaryRecord?.id);
```
haynesjm42, line 707: I think we also need to add this check up in the `updateData` method now like are doing for `parent` to ensure we are setting this flag if the summary record is provided in the update data?

lbwexler: see later in the updateDate method where we intentionally identify and recreate the summary record, after doing a merge with the existing one, and at that point only set the flag

PR #1873, opened 2020-05-01
Allowing grid to reset without a StateModel

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Virtual column rendering has been disabled by default, as it offered a minimal performance benefit
   for most grids while compromising autosizing. See new `GridModel.useVirtualColumns` config, which
   can be set to `true` to re-enable this behavior if required.
+* All `GridModel`s can now be reset to their code-prescribed defaults via the column chooser reset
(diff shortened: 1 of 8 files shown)


febbraiod:
Briefly reviewed with Lee. He wants to think about the approach.

lbwexler:
Tom -- can you take a look and give your thoughts on this?  I was hoping for something simple that would just get us the restore ability in all grids.

TomTirapani:
Looks good to me.

Did we consider saving the entire constructor params as e.g. `this.initialConfig`? Would mean we wouldn't need the `saveInitialConfigs()` method, and may be useful generally to know how a model was configured.

I think Ext had a pretty baked in concept of keeping `initialConfigs` around.

lbwexler:
This one makes me a little nervous -- also feel like its a bit unfortunate that we don't have a way to 'clear' this config when are saving it in the dashboard, as we do in some client apps -- this changes would just write out the default, I think.

lbwexler:
Looks like we forgot the mobile  col chooser here.

febbraiod:
Glad to fix? Are we running with this approach?

PR #1874, opened 2020-05-04
Autosize columns

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -199,15 +199,32 @@
 }
 
 .xh-grid-autosize-cell {
-  top: var(--xh-grid-tree-indent-px);
   padding: 0 var(--xh-grid-cell-lr-pad-px);
-  font-family: var(--xh-grid-font-family);
   font-size: var(--xh-grid-font-size-px);
-  -webkit-font-smoothing: subpixel-antialiased;
+  font-family: var(--xh-grid-font-family);
   visibility: hidden;
   display: none;
 
+  // This property is not actually used in rendering, but is provided here
+  // for later interpretation by GridAutosizeService.getIndentation()
+  left: var(--xh-grid-tree-indent-px);
+
   &--active {
     display: inline-block;
+  }
+
+  &--large {
+    font-size: var(--xh-grid-large-font-size-px);
+    padding: 0 var(--xh-grid-large-cell-lr-pad-px);
+  }
+
+  &--compact {
+    font-size: var(--xh-grid-compact-font-size-px);
+    padding: 0 var(--xh-grid-compact-cell-lr-pad-px);
+  }
+
+  &--tiny {
+    font-size: var(--xh-grid-tiny-font-size-px);
+    padding: 0 var(--xh-grid-tiny-cell-lr-pad-px);
   }
 }
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -193,6 +193,11 @@
   position: absolute;
 }
 
+.xh-grid-autosize-canvas {
+  position: absolute;
+  display: none;
+}
+
 .xh-grid-autosize-cell {
   padding: 0 var(--xh-grid-cell-lr-pad-px);
   font-family: var(--xh-grid-font-family);
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -195,7 +195,7 @@
 
 .xh-grid-autosize-cell {
   // Add a slight padding buffer for security against ellipses
-  padding: 0 calc(var(--xh-grid-cell-lr-pad-px) + 2px);
+  padding: 0 calc(var(--xh-grid-cell-lr-pad-px) + 2px); // Todo: Add buffer to number in JS, make an option
   display: inline-block;
   font-family: var(--xh-grid-font-family);
   font-size: var(--xh-grid-font-size-px);
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -194,11 +194,14 @@
 }
 
 .xh-grid-autosize-cell {
-  // Add a slight padding buffer for security against ellipses
-  padding: 0 calc(var(--xh-grid-cell-lr-pad-px) + 2px); // Todo: Add buffer to number in JS, make an option
-  display: inline-block;
+  padding: 0 var(--xh-grid-cell-lr-pad-px);
   font-family: var(--xh-grid-font-family);
   font-size: var(--xh-grid-font-size-px);
   -webkit-font-smoothing: subpixel-antialiased;
   visibility: hidden;
+  display: none;
+
+  &--active {
+    display: inline-block;
+  }
 }
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -199,6 +199,7 @@
 }
 
 .xh-grid-autosize-cell {
+  top: var(--xh-grid-tree-indent-px);
   padding: 0 var(--xh-grid-cell-lr-pad-px);
   font-family: var(--xh-grid-font-family);
   font-size: var(--xh-grid-font-size-px);
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -193,6 +193,9 @@
   position: absolute;
 }
 
+//-----------
+// Autosize
+//-----------
 .xh-grid-autosize-canvas {
   position: absolute;
   display: none;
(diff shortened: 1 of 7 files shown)


cmp/grid/columns/Column.js
```diff
@@ -258,6 +260,15 @@ export class Column {
         this.setValueFn = withDefault(setValueFn, this.defaultSetValueFn);
         this.getValueFn = withDefault(getValueFn, this.defaultGetValueFn);
 
+        this.autoSizeOptions = {
+            enabled: withDefault(this.resizable, true),
```
lbwexler, line 264: Could this default also be linked to `!this.elementRenderer`?

cmp/grid/GridModel.js
```diff
@@ -695,9 +696,19 @@ export class GridModel {
...
-            return col && !col.flex;
+            return col && col.resizable && !col.hidden && !col.flex;
         });
-        if (colIds.length) this.agColumnApi.autoSizeColumns(colIds);
+        if (colIds.length) {
+            start().then(async () => {
+                this.agApi.showLoadingOverlay();
+
+                const colStateChanges = XH.gridAutosizeService.autoSizeColumns({gridModel: this, colIds});
+                this.applyColumnStateChanges(colStateChanges);
```
lbwexler, line 706: should we log 'colStateChanges'  on debug?

cmp/grid/GridModel.js
```diff
@@ -695,9 +696,19 @@ export class GridModel {
     autoSizeColumns(colIds = this.getLeafColumns().map(col => col.colId)) {
```
lbwexler, line 696: (Realize this change is not related to this PR, but I find the expression in the signature too complicated).  We should have colIds default to undefined, and then have the method do this defaulting on the first line.  Also add a docstring.

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,143 @@
...
+import {throwIf, stripTags} from '@xh/hoist/utils/js';
+import {isFunction, isNil, isFinite, sortBy, reduce} from 'lodash';
+
+@HoistService
+export class GridAutosizeService {
+
+    _canvasContext;
+    _cellEl;
+
+    autoSizeColumns({gridModel, colIds = []}) {
```
lbwexler, line 18: Need docstring on method and/or on this class.   I think it does make sense to make this a service -- its quite similar to GridExportService, and you could also imagine applications replacing it or enhancing it in some way,

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,143 @@
...
+                records = [...store.records, store.summaryRecord],
+                column = gridModel.findColumn(gridModel.columns, colId),
+                {autoSizeOptions, field, getValueFn, renderer, elementRenderer} = column,
+                {enabled, sampleCount, buffer, minWidth, maxWidth} = autoSizeOptions,
+                useRenderer = isFunction(renderer);
+
+            if (!enabled) return;
+
+            // Columns with element renderers are not supported
+            if (elementRenderer) return;
```
lbwexler, line 47: would we consider falling back to the native agGrid resizing if someone really wanted to resize a column with an elementRenderer. (see my comment above -- might be useful for columns that have fixedWidths that result from elementRenderer)

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,143 @@
...
+            });
+
+            // 2) Use a canvas to estimate and sort by the pixel width of the string value.
+            // Strip html tags but include parentheses / units etc. for renderers that may return html,
+            const sortedValues = sortBy(Array.from(values), value => {
+                return isNil(value) ? 0 : this.getStringWidth(stripTags(value.toString()));
+            });
+
+            // 3) Extract the sample set of longest values for rendering and sizing
+            const longestValues = sortedValues.slice(Math.max(sortedValues.length - sampleCount, 0));
```
lbwexler, line 65: Tom and I discussed this, but still wondering if specifying sampleCount as a percentage might be more appropriate?  Or (not to overdesign this) but also allow specifying either a number like '1%' or 10. (in the latter case, would be sure to want a minimum)

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,143 @@
...
+
+            // 4) Render to a hidden cell to calculate the max displayed width
+            const result = reduce(longestValues, (currMax, value) => {
+                const width = this.getCellWidth(value, useRenderer) + buffer;
+                return Math.max(currMax, width);
+            }, 0);
+
+            if (isFinite(minWidth) && result < minWidth) return minWidth;
```
lbwexler, line 73: lodash max handles undefined nicely, can simplify these two statements nicely

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,143 @@
...
+            // Strip html tags but include parentheses / units etc. for renderers that may return html,
+            const sortedValues = sortBy(Array.from(values), value => {
+                return isNil(value) ? 0 : this.getStringWidth(stripTags(value.toString()));
+            });
+
+            // 3) Extract the sample set of longest values for rendering and sizing
+            const longestValues = sortedValues.slice(Math.max(sortedValues.length - sampleCount, 0));
+
+            // 4) Render to a hidden cell to calculate the max displayed width
+            const result = reduce(longestValues, (currMax, value) => {
```
lbwexler, line 68: lodash has a maxBy that would work great here as well

TomTirapani: I tried maxBy, but couldn't get it to work with 2 loops as it would return the *value* that had the max width rather than the max width itself

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,182 @@
...
+        const longestValues = sortedValues.slice(Math.max(sortedValues.length - sampleCount, 0));
+
+        // 4) Render to a hidden cell to calculate the max displayed width
+        let result = reduce(longestValues, (currMax, value) => {
+            const width = this.getCellWidth(value, useRenderer) + indentation + buffer;
+            return Math.max(currMax, width);
+        }, 0);
+
+        result = max([result, minWidth]);
+        result = min([result, maxWidth]);
```
amcclain, line 108: There is lodash `clamp` in case you would like to use that.

cmp/grid/Grid.scss
```diff
@@ -192,3 +192,22 @@
...
+  position: absolute;
+  display: none;
+}
+
+.xh-grid-autosize-cell {
+  top: var(--xh-grid-tree-indent-px);
+  padding: 0 var(--xh-grid-cell-lr-pad-px);
+  font-family: var(--xh-grid-font-family);
+  font-size: var(--xh-grid-font-size-px);
+  -webkit-font-smoothing: subpixel-antialiased;
```
amcclain, line 206: I would be pretty surprised if this font-smoothing rule made a difference (although certainly could be wrong).  I would vote for removing it in the interest of simplicity, unless we have reason to believe it's important.

cmp/grid/Grid.scss
```diff
@@ -192,3 +192,22 @@
...
   position: absolute;
 }
+
+.xh-grid-autosize-canvas {
+  position: absolute;
+  display: none;
+}
+
+.xh-grid-autosize-cell {
+  top: var(--xh-grid-tree-indent-px);
```
amcclain, line 202: Why `top` here for the indent?  I am sure this is as intended, but seems worth a comment as it threw me for a loop (I would have assumed `left` since we're all about measuring width / indenting)

cmp/grid/GridModel.js
```diff
@@ -692,12 +693,26 @@ export class GridModel {
...
+        if (colIds.length) {
+            start().then(async () => {
+                this.agApi.showLoadingOverlay();
+
+                const colStateChanges = XH.gridAutosizeService.autoSizeColumns({gridModel: this, colIds});
+                this.applyColumnStateChanges(colStateChanges);
+
+                console.debug('Columns autosized:', colStateChanges);
+
+                await wait(100);
```
amcclain, line 712: Could we comment this await? Is it here to allow the grid to complete its render/paint before removing the mask, or something else?

cmp/grid/columns/Column.js
```diff
@@ -258,6 +260,15 @@ export class Column {
         this.setValueFn = withDefault(setValueFn, this.defaultSetValueFn);
         this.getValueFn = withDefault(getValueFn, this.defaultGetValueFn);
 
+        this.autoSizeOptions = {
+            enabled: withDefault(this.resizable && !this.elementRenderer, true),
+            sampleCount: 10,
+            buffer: 5,
+            minWidth: this.minWidth,
+            maxWidth: this.maxWidth,
```
amcclain, line 268: I like that these can be controlled for autoSize, separately from the limits imposed on the column for manual resizing. 👍

cmp/grid/columns/Column.js
```diff
@@ -258,6 +260,15 @@ export class Column {
         this.setValueFn = withDefault(setValueFn, this.defaultSetValueFn);
         this.getValueFn = withDefault(getValueFn, this.defaultGetValueFn);
 
+        this.autoSizeOptions = {
+            enabled: withDefault(this.resizable && !this.elementRenderer, true),
+            sampleCount: 10,
+            buffer: 5,
```
amcclain, line 266: Calling this `bufferPx` or `extraPx` might be a bit more clear.

svc/GridAutosizeService.js
TomTirapani: In my testing, `clamp` only seems to work if *both* the lower and upper bound are defined, which we can't guarantee

cmp/grid/Grid.scss
TomTirapani: We're not *actually* using it directly - this is a dummy style rule, used only as a way to trampoline the `--xh-grid-tree-indent-px` value in a way that it can be brought into JS via `window.getComputedStyle()` (it uses a calc internally). Will comment as such to explain.

svc/GridAutosizeService.js
amcclain: Ah yes it's weirdly (to me) asymmetrical where the lower bound is optional but the upper bound is not.

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,285 @@
...
+                const record = store.getById(node.data?.id);
+                if (record) records.push(record);
+            });
+        } else {
+            records = [...store.records];
+        }
+
+        if (store.summaryRecord) {
+            records.push(store.summaryRecord);
+        }
```
amcclain, line 57: Suggest we include only when `GridModel.showSummary`

svc/GridAutosizeService.js
```diff
@@ -0,0 +1,285 @@
...
+
+        return ret;
+    }
+
+    //------------------
+    // Implementation
+    //------------------
+    autoSizeColumn(gridModel, records, colId) {
+        const column = gridModel.findColumn(gridModel.columns, colId);
```
amcclain, line 71: This was always a weird signature for `gridModel.findColumn` - not sure we can change it now, but it would have been nice if `colId` were the first param and there was a second optional param that let you specify a subset of columns to search (but defaulted to all the gridModel's columns).

PR #1886, opened 2020-05-07
autosizeColumns options param, allow autosize without overlay

haynesjm42:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -12,7 +12,16 @@ import {ColChooserModel as DesktopColChooserModel} from '@xh/hoist/dynamics/desk
 import {ColChooserModel as MobileColChooserModel} from '@xh/hoist/dynamics/mobile';
 import {action, observable} from '@xh/hoist/mobx';
 import {start, wait} from '@xh/hoist/promise';
-import {apiRemoved, debounced, deepFreeze, ensureUnique, errorIf, throwIf, warnIf, withDefault} from '@xh/hoist/utils/js';
+import {
+    apiRemoved,
+    debounced,
+    deepFreeze,
+    ensureUnique,
+    errorIf,
+    throwIf,
+    warnIf,
+    withDefault
+} from '@xh/hoist/utils/js';
 import equal from 'fast-deep-equal';
 import {
     castArray,
@@ -24,6 +33,7 @@ import {
     findLast,
     isArray,
     isEmpty,
+    isFunction,
     isNil,
     isPlainObject,
     isString,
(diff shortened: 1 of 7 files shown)


cmp/grid/GridModel.js
```diff
@@ -228,6 +241,7 @@ export class GridModel {
 
         contextMenu,
         useVirtualColumns = false,
+        autosizeOptions = {showMask: true},
```
lbwexler, line 244: Don't think we want to default it like this -- need to merge in that default in the code, so changing another option does not step on this one

cmp/grid/GridModel.js
```diff
@@ -716,14 +731,16 @@ export class GridModel {
...
      * This method will ignore hidden columns, columns with a flex value, and columns with
      * autosizing disabled via the autosizeOptions.enabled flag.
      */
-    autosizeColumns(colIds) {
-        colIds = colIds ?? this.getLeafColumns().map(col => col.colId);
+    async autosizeColumnsAsync(options = {}) {
+        options = {...this.autosizeOptions, ...options};
 
+        const {columns} = options;
+        let colIds = isFunction(columns) ? columns() : columns ?? this.getLeafColumns().map(col => col.colId);
```
lbwexler, line 743: wouldn't it be simpler/safer to have this function just be a predicate function that would be applied to a leaf column and return true if the column should be resized?  

We always want to operate on some subset of the leaf columns

cmp/grid/GridModel.js
```diff
@@ -732,26 +749,28 @@ export class GridModel {
...
 
     //-----------------------
     // Implementation
     //-----------------------
-    hoistAutosize(colIds) {
-        start(async () => {
-            this.agApi?.showLoadingOverlay();
-            // Wait to allow mask to render before starting potentially compute-intensive autosize.
-            await wait(100);
+    async hoistAutosize(colIds, {showMask}) {
```
lbwexler, line 767: naming convention: probably should be internalAutosizeAsync at this point, given that the experimental bit is going away.

cmp/grid/impl/ColumnHeader.js
```diff
@@ -196,11 +196,12 @@ class LocalModel {
             if (isFinite(currIdx)) idx = (currIdx + 1) % availableSorts.length;
         }
 
+
         return availableSorts[idx];
     }
 
     autosize() {
-        this.gridModel?.autosizeColumns(this.colId);
+        this.gridModel?.autosizeColumns({colIds: this.colId});
```
lbwexler, line 204: Need to fix name of method and argument

PR #1897, opened 2020-05-11
Autosize distributes extra space

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -814,8 +814,8 @@ export class GridModel {
                 await wait(100);
             }
 
-            const flexColumns = this.getVisibleLeafColumns().filter(it => it.flex);
-            if (!flexColumns.length && fillMode !== 'disabled') {
+            const hasFlexColumn = this.getVisibleLeafColumns().some(it => it.flex);
+            if (!hasFlexColumn && fillMode !== 'disabled') {
                 const colStateChanges = XH.gridAutosizeService.fillColumns(this, colIds, fillMode);
                 this.applyColumnStateChanges(colStateChanges);
                 console.debug('Column widths filled via GridAutosizeService', colStateChanges);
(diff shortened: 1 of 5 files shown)


cmp/grid/GridModel.js
```diff
@@ -810,6 +810,15 @@ export class GridModel {
 
             if (!isEmpty(agSizable)) {
                 this.agColumnApi?.autoSizeColumns(agSizable);
+                // Wait to allow ag-grid autosize to propagate before potentially filling width.
+                await wait(100);
+            }
+
+            const flexColumns = this.getVisibleLeafColumns().filter(it => it.flex);
```
lbwexler, line 817: consider using 'some()' here vs. filter(), just to get the boolean you want

cmp/grid/GridModel.js
```diff
@@ -810,6 +810,15 @@ export class GridModel {
 
             if (!isEmpty(agSizable)) {
                 this.agColumnApi?.autoSizeColumns(agSizable);
```
lbwexler, line 812: We should discuss how import this fallback is, and if is worth the complexity, if there should be a flag for it, etc.

Additional complexity needed here with the wait() brings this question to the fore

cmp/grid/GridModel.js
```diff
@@ -789,7 +789,7 @@ export class GridModel {
     //-----------------------
     // Implementation
     //-----------------------
-    async internalAutosizeAsync(colIds, {showMask}) {
+    async internalAutosizeAsync(colIds, {showMask, fillMode}) {
         await start(async () => {
             if (showMask) {
```
lbwexler, line 794: I think at this point we probably want to move the entire body of this method (with perhaps the exception of the masking) in to an async method on the service.   Easier to have the "brains" of the operation in one place -- e.g. for testing, or swapping in alternative approaches -- and I don't think given the name of the service, it is that terrible for it to actually operate on the the gridModel and the agAPI

svc/GridAutosizeService.js
```diff
@@ -302,4 +340,74 @@ export class GridAutosizeService {
...
+                ret[colId] += extraWidth;
+                widthAdded = true;
+            });
+
+            // If we couldn't add any width this iteration (e.g. all columns are at their max widths),
+            // we must break to prevent an infinite loop.
+            if (!widthAdded) break;
+        }
+
+        return keys(ret).map(colId => {
```
lbwexler, line 373: consider lodash map, which will operate on an object

lbwexler:
Tom -- hope the code re-org makes sense to you?   Just wanted to split up that big file and it seems like the measuring code for a single column shipped off nicely to a calculator object.

PR #1907, opened 2020-05-15
Dimension chooser can save history and value independently, Fixes #1898

### LINKED ISSUE #1898: DimensionChooserModel should allow saving only the history
Have a use-case where I need to share the dimension chooser history among multiple controls, but _not_ the value. Currently the built-in `preference` support in DimensionChooserModel saves both the value and history. 

Also would be ideal if the history changes could be synchronized between the different DimensionChooserModel instances. Perhaps this would require that we make prefs observable though?

Am not sure the best way to implement this, we could accept a spec for the `preference` config which could indicate which parts of the model state should be persisted to the pref. e.g. `{preference: {mode: 'historyOnly'}` - or we could just have a flag?

Alternatively we could just allow the app to take full control of the history by making it observable and providing a setter, and the ability to set the initial history (like we do for `initalDimensions`. That way the app could save and restore that history however it wanted and ignore the value

febbraiod:
Hoist P/R Checklist
-------------------


lbwexler:
Sorry if I wasn't clear here -- was hoping for simplicity to have the preference keep the same format, as being able to hold both.    This maximizes backward compatibility without the need to call it out explicitly, and its totally valid to continue to store both in the bits in the same preference

cmp/dimensionchooser/DimensionChooserModel.js
```diff
@@ -82,6 +88,7 @@ export class DimensionChooserModel {
         this.maxDepth = maxDepth;
         this.enableClear = enableClear;
         this.preference = preference;
+        this.historyPreference = withDefault(historyPreference, preference);
```
lbwexler, line 91: This bit of defaulting can be done right in the constructor -- you can refer to previous arguments when defining subsequent ones -- very cool feature of javascript

PR #1910, opened 2020-05-16
Persistence API

lbwexler:
See Related Toolbox PR: https://github.com/xh/toolbox/pull/400

Hoist P/R Checklist
-------------------


cmp/grid/impl/GridPersistenceModel.js
```diff
@@ -5,100 +5,76 @@
...
-     * Attach this object to a GridModel and have it observe / apply changes to tracked properties.
-     * @param {GridModel} gridModel
-     */
-    init(gridModel) {
+    path;
+
+    @observable.ref
+    state;
+
+    constructor(
```
TomTirapani, line 27: We should document what these arguments are

svc/PrefService.js
```diff
@@ -101,6 +101,16 @@ export class PrefService {
         this.pushPendingBuffered();
     }
 
+    /**
+     * Restore a preference to its default value.
+     * @param key
+     */
+    unset(key) {
+        // TODO: round-trip this to the server as a proper unset?
```
amcclain, line 109: Looks like we have a method on PrefService.groovy that's ready to use, although we would need to expose via a controller and coordinate with a Hoist-Core release. To be honest I am not super-motivated to add that to the mix right now.

cmp/grid/impl/GridPersistenceModel.js
amcclain: I would agree - even for the private class, it's nice to have the ctor and its args documented and helps I think with additional IDE type inference when looking at / working on the class.

cmp/grid/GridModel.js
```diff
@@ -1003,8 +1002,7 @@ export class GridModel {
 
 
     createChooserModel() {
-        const Model = XH.isMobileApp ? MobileColChooserModel : DesktopColChooserModel;
-        return this.markManaged(new Model(this));
+        return XH.isMobileApp ? new MobileColChooserModel(this) : new DesktopColChooserModel(this);
```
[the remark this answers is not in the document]
lbwexler: I think that we just adopted the markManaged pattern from stateModel and selectionModel, where we had support for the user passing a model in.  I don't think it predated it.

lbwexler:
Thanks Guys -- trying to put this to bed today!

Would like to discuss, but what would you think about modifying things such that we have a single PersistOptions bundle of @typedef.  We would continue to pass it around as "persistWith"

```
{
    provider: PersistenceProvider.
    path: [defaults to undefined and determined by component] 
    buffer: [defaults to 250, or defaults to undefined determined by component]
     ...rest: additional component specific arguments (namely for grid).
}
```

It cleans up a lot, although it does mean more verbose specifications, when you only want to give a key i.e.
```
new GridModel({
         persistWith: {provider: {prefKey: 'foobar'}}
})

The other option is that we continue to have `persistWith` and `persistOptions` as separate things.  But this will be a little clunkier when you actually do have both, which I think may be more common now given that we will need to use 'path' to intelligently combine stuff.  Also thinking about something like dim chooser, which I would like to modify to have the following:

Also, we could rename `provider: PersistenceProvider` to `source:PersistenceSource` for more brevity, if we thought that might help.  Somewhat of an orthogonal change...

Perhaps best to just discuss this in person!?

File: cmp/grid/impl/GridPersistenceModel.js [hunk 4]
```diff
@@ -189,14 +161,26 @@ export class GridStateModel {
...
     }
 
-    @debounced(500)
-    saveStateChange() {
-        const {state} = this;
+    @action
+    patchState(updates) {
+        this.state = {...this.state, ...updates};
+    }
+
```
TomTirapani, line 168: Should we consider a straightforward `setState()` method, which simply replaces the entire state blob?

cmp/grid/impl/GridPersistenceModel.js
```diff
@@ -189,14 +161,26 @@ export class GridStateModel {
         return ret;
     }
 
-    @debounced(500)
-    saveStateChange() {
-        const {state} = this;
+    @action
+    patchState(updates) {
+        this.state = {...this.state, ...updates};
```
TomTirapani, line 166: Is there anyway for old, unused state to be cleared out? For example, if we turn off `persistColumns`, it looks like columns will forever still be kept around as part of the state? (not saying we didn't have this situation already)

lbwexler:
great idea Tom -- Anselm and I were just thinking same!

Sounds like consensus!

-Lee

lbwexler:
Made the changes you suggested -- think its all ready for re-review.  Putting together an example now, probably in our example apps

cmp/grid/GridModel.js
```diff
@@ -219,11 +221,13 @@ export class GridModel {
         treeMode = false,
         showSummary = false,
         selModel,
-        stateModel = null,
         emptyText = null,
         sortBy = [],
         groupBy = null,
 
+        persistWith,
+        persistOptions,
```
TomTirapani, line 229: Do we still require separate options here? I thought we were folding persistOptions into persistWith?

persist/Persist.js
```diff
@@ -0,0 +1,97 @@
...
+ * Decorator to make a class property persistent, syncing its value via a configured
+ * `PersistenceProvider` to maintain and restore mutable values across browser sessions.
+ *
+ * This decorator may be used on any @observable or @bindable class property which is a primitive.
+ * It will initialize the observable's value from the class's default `PersistenceProvider` and will
+ * subsequently write back any changes to the property to that Provider. If the Provider has not
+ * yet been populated with a value, or an error occurs, it will use the value set in-code instead.
+ *
+ * This decorator should always be applied "before" the mobx decorators, i.e. closest to definition,
+ * second in file line order: `@bindable @persist fooBarFlag = true`
```
TomTirapani, line 21: Curious... what happens if this ordering is not adhered to? Does it break in an obvious way, or an "invisible" way?

I expect this to be a common source of bugs if the break is quite subtle. If the break is hard to spot, could we somehow detect and throw?

persist/provider/DashViewProvider.js
```diff
@@ -0,0 +1,40 @@
...
+ * Copyright © 2020 Extremely Heavy Industries Inc.
+ */
+
+import {PersistenceProvider} from './PersistenceProvider';
+import {throwIf} from '../../utils/js';
+
+/**
+ * PersistenceProvider that stores state within a DashView.
+ */
+export class DashViewProvider extends PersistenceProvider {
```
TomTirapani, line 14: Minor pedantry (and *not* suggesting we change anything), but writing / reading to DashView state isn't *really* persistence... :) don't have any better suggestions.

lbwexler:
not subtle, I think, but think this is detectable.  Will look at it.

PR #1913, opened 2020-05-18
WIP - ExceptionDialog will include a button to end impersonation if appropriate, Fixes: 

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/appcontainer/ExceptionDialog.js
+++ b/desktop/appcontainer/ExceptionDialog.js
@@ -53,19 +53,18 @@ export const exceptionDialog = hoistCmp.factory({
 //--------------------------------
 const bbar = hoistCmp.factory(
     ({model}) => toolbar(
+        button({
+            omit: !XH.identityService.isImpersonating,
+            icon: Icon.impersonate(),
+            text: 'End Impersonation',
+            onClick: () => XH.identityService.endImpersonateAsync()
+        }),
         filler(),
         button({
             icon: Icon.search(),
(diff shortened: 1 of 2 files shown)


desktop/appcontainer/ExceptionDialog.js
```diff
@@ -60,6 +60,13 @@ const bbar = hoistCmp.factory(
             onClick: () => model.openDetails(),
             omit: !model.options.showAsError
         }),
+        button({
+            omit: !XH.identityService.isImpersonating,
+            icon: Icon.impersonate(),
+            text: 'End Impersonation',
+            minimal: false,
+            onClick: () => XH.identityService.endImpersonateAsync()
+        }),
```
febbraiod, line 69: I have my doubts about this. The change stems from a client app with a very specific permissioning system that does not utilize Hoist Roles. When impersonating a user an admin can get locked out via an exception thrown by an interceptor with no way to get out of it (other than close the browser). The point is are we making a change to Hoist that is too focused on one specific use case. Not sure we have another alternative in app.

Then again, maybe if you are impersonating and get an error, it's a reasonable thing to want to have the choice to end impersonation there.

lbwexler: discussed with Don -- think this is great, and a useful failsafe for any exception that might occur during initial appload due to impersonation.  Don will move to left to get it out of the way of the "normal" app.

PR #1916, opened 2020-05-21
Add device detection to XH

TomTirapani:
Adds UA based `isPhone` & `isTablet` flags to XH. Intended to complement the app set `isMobile` flag, to support device-specific logic in both desktop and mobile apps.

Tested:

+ Mac OSX 10.15.4
  + Chrome
  + All Chrome emulations
  + Safari

+ iPhone 11 Pro running iOS 13.4.1
  + Chrome
  + Safari
  + Client browsers

+ iPad Pro (9.7-inch) running iPadOS 13.5
	+ Chrome
	+ Safari
	+ Client browsers

Please note that iOS Safari enables "Request Desktop Website" by default on iPad (default false on iPhone), which breaks the tablet detection. I don't think this is an issue - I think we want to respect that setting - but we may want to bear that in mind in our communications.

Hoist P/R Checklist
-------------------


core/XH.js
```diff
@@ -122,6 +124,8 @@ class XHClass {
     getUsername()               {return this.identityService ? this.identityService.getUsername() : null}
 
     get isMobile()              {return this.appSpec.isMobile}
+    get isPhone()               {return this.uaParser.getDevice().type === 'mobile'}
```
lbwexler, line 127: My only concern is that with having these named as siblings we are some how implying they come from the same place, where isMobile obviously comes from our hard-coded spec rather than the UA.

Its almost like the new ones could be isPhoneDevice() and isTabletDevice().

Ideally, we could have named the former 'isMobileApp()', but that ship has sailed.

Thoughts?  Not sure there is a great answer, and certainly understand wanting these succict and on XH.

TomTirapani: Yeah, I had a similar concern... I think the suggestion to change to `isPhoneDevice` and `isTabletDevice` makes sense.

I don't think it so bad to also change to `isMobileApp` at this point... we could set it up with a deprecation warning for now, and make the proper breaking change in next release. Should be easy enough to update apps, and is used pretty sparingly.

TomTirapani: Renamed to `isPhoneDevice` and `isTabletDevice`

lbwexler:
co-dev: Tom and Lee

PR #1921, opened 2020-05-21
Add device/window info to App Loaded tracking entry

haynesjm42:
Example from my desktop:

image

Example from my Pixel 2:

image

Hoist P/R Checklist
-------------------


admin/tabs/activity/tracking/ActivityDetail.js
```diff
@@ -54,11 +54,11 @@ const detail = hoistCmp.factory(
                 }),
                 jsonInput({
                     omit: !data.data,
-                    value: data.data,
-                    disabled: true,
-                    height: 100,
+                    value: JSON.stringify(JSON.parse(data.data), null, 2),
```
lbwexler, line 57: We don't have a mode by which jsonInput will pretty print automatically.

utils/js/BrowserUtils.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Extract information about the client browser window and screen
+ */
+export function getClientDeviceInfo() {
+    const data = pick(window, 'screen', 'devicePixelRatio', 'screenX', 'screenY', 'innerWidth', 'innerHeight', 'outerWidth', 'outerHeight');
+    if (data.screen) {
+        data.screen = toPlainObject(data.screen);
+        if (data.screen.orientation) {
+            data.screen.orientation = toPlainObject(data.screen.orientation);
+            delete data.screen.orientation.onchange;
```
amcclain, line 12: What happens if some other unexpected function or not-easily-serializable / recursive reference gets through into this mix? Could this ever throw or does toPlainObject() protect well against that kind of thing?  Not sure how critical it is given that it's called from within the tracking reaction, but since it would run on every launch wanted to ask the question.

Otherwise though I think this is a great change and I can see it being quite useful - thanks for putting it out there!  Will be a further prompt to make that `data` block within track statements something you can filter or slice/dice on..one day...

lbwexler: I really agree with the concern -- it would be a HUGE bummer if some perfectly functional set of Hoist apps were no longer able to track their usage due to a browser upgrade -- and required an upgrade to continue doing so.

haynesjm42: Definitely not something we would want to happen, am thinking about how to change this to be 100% safe for any changes to the browser apis in the future.. would just picking the properties we want out of these objects work? Should be pretty save to assume the props we care about remain serializable, since they are all numbers

admin/tabs/activity/tracking/ActivityDetail.js
haynesjm42: Not that I could find, seems like what you would usually want as default behavior - Filed #1923

PR #1935, opened 2020-05-27
Add Store.recordIsFiltered()

amcclain:
+ JsDoc and method param name updates for clarity. (I found "fromFiltered" a bit ambiguous - i.e. could be incorrectly interpreted as "from the records that have been filtered OUT".)

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -647,9 +647,7 @@ export class Store {
      */
     recordIsFiltered(recOrId) {
         const id = recOrId.isRecord ? recOrId.id : recOrId,
-            inStore = !!this.getById(id, false),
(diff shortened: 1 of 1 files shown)


amcclain:
This method would be of use in a client application where we have a global search feature that can direct a user to a particular grid that has a StoreFilterField wired up. We want to check to see if we need to clear that filter before attempting to select/show the result in the grid.

It might be weird to return the same value `false` if the record is either not filtered or simply not in the store at all.  I considered returning `null` to represent not-in-store, if that would be helpful.

Even if we don't take this new method, I would suggest the other method param name update + jsdoc tweaks for clarity.  Happy to discuss of course.

PR #1941, opened 2020-05-29
Use `preventDefault()` in panel dragger event listeners to prevent overscroll on tablets

TomTirapani:
Note that in order to preventDefault, events listeners must be added to the element directly using a ref (React events are `passive: true`)

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/panel/impl/dragger/DraggerModel.js
+++ b/desktop/cmp/panel/impl/dragger/DraggerModel.js
@@ -27,8 +27,8 @@ export class DraggerModel {
         this.panelModel = panelModel;
         this.throttledSetSize = throttle(size => panelModel.setSize(size), 50);
 
-        // We use a ref to allow us to add listeners directly to the element. This allows
-        // us to mark events as non-passive, which in turn allows us to preventDefault()
+        // Add listeners to el to ensure we can get non-passive handlers than can preventDefault()
+        // React synthetic touch events on certain browsers (e.g. airwatch) don't yield that
         this.addReaction({
             track: () => this.ref.current,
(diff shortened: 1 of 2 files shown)


desktop/cmp/panel/impl/dragger/DraggerModel.js
```diff
@@ -23,6 +26,21 @@ export class DraggerModel {
...
+        this.addReaction({
+            track: () => this.ref.current,
+            run: (current) => {
+                if (current) this.addListeners(current);
+            }
+        });
+    }
+
+    addListeners(el) {
+        el.addEventListener(XH.isDesktop ? 'dragstart' : 'touchstart', e => this.onDragStart(e));
```
lbwexler, line 41: Are we sure that the passive false was needed here, or is it just enough to forego the react "synthetic" events?

Just trying to be careful on this one here, as the change currently effects all of our code base here in order to fix the Airwatch problem we are seeing at a client

lbwexler:
For reference -- somewhat concerned about using the extra options argument where it might not be supported.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners

desktop/cmp/panel/impl/dragger/DraggerModel.js
TomTirapani: Yes - most browsers default `touchmove` (and scroll events) to passive in order to improve performance. Without setting `{passive: false}`, the call to preventDefault() is ignored and you get warnings in the console.

PR #1942, opened 2020-05-29
WIP - Json input fixes. Fixes #1922, Fixes #1923, Fixes #1940

### LINKED ISSUE #1922: Issues with Fullscreen/format buttons in JsonInput
These buttons depend on focus on the input, but when it is disabled clicking on the input doesn't focus it. I was able to get the buttons to appear briefly by right clicking on the input in some cases.

We might consider using `editorProps: {readOnly: true}` to handling disabling a JsonInput. That makes it so you can still select text in the input and focus on it, but cannot edit the contents. Having a JsonInput that you can't select the contents of is usually pretty frustrating and not what you would expect to happen.

### LINKED ISSUE #1923: JsonInput should auto-format the first time it renders data
Or we should at least provide an option to do this automatically. Usually don't want unformatted json in the input when viewing or editing.

### LINKED ISSUE #1940: JSONInput/CodeInput cannot toggle it's disabled/enabled state
This is true at least for a form model that is disabled or enabled. But I don't think we'd see any difference no matter how we toggle this bit of state.

Other other inputs will use this disabled prop more directly and what the input re-renders this prop will be reset and the expect behavior will occur. However in this case the method that sets this property on the codeMirrorEditor does not get call but once when the textArea is converted to a codeMirrorInstance. (https://github.com/xh/hoist-react/blob/6091a001ed2e694513e83281e10c369d59a5551c/desktop/cmp/input/CodeInput.js#L241).

That state is now its default and it remains that way even if the prop changes.

Behavior can be observed in toolbox's inputs panel.

febbraiod:
Hoist P/R Checklist
-------------------


desktop/cmp/input/CodeInput.js
```diff
@@ -215,11 +223,13 @@ export class CodeInput extends HoistInput {
...
         }
 
+        const val = this.tryPrettyPrint(editor.getValue());
+        editor.setValue(val);
         return editor;
     }
 
     createDefaults() {
-        const {disabled, mode, linter, autoFocus} = this.props;
+        const {disabled, mode, linter, autoFocus} = this.props;  // TODO: None of these will get reset on prop change
```
febbraiod, line 232: This PR contains the fix to ensure that disabled can be toggled. Should we do the same for these other options or are assuming that the other three are truly more like immutable configs on the input.

desktop/cmp/input/CodeInput.js
```diff
@@ -215,6 +223,8 @@ export class CodeInput extends HoistInput {
             editor.setSize(width, height);
         }
 
+        const val = this.tryPrettyPrint(editor.getValue());
```
lbwexler, line 226: Not 100% sure about this feature -- Doesn't it cause the control to indicate its value changed? 

Does this mean that just loading one of these things in a form will immediately dirty the form when it is shown?

Maybe we want to thing twice about monkeying with the value.

desktop/cmp/input/CodeInput.js
```diff
@@ -238,7 +248,7 @@ export class CodeInput extends HoistInput {
             },
             foldGutter: true,
             scrollbarStyle: 'simple',
-            readOnly: disabled ? 'nocursor' : false,
+            readOnly: disabled,
```
lbwexler, line 251: what's the deal with the 'nocursor' string from before.  Is that still of concern?

febbraiod: That was what was prevent John H from selecting the text or focusing the input. It's like an even more extreme version of 'disabled'

lbwexler: sounds good

desktop/cmp/input/CodeInput.js
lbwexler: Please remove this to a separate PR so we can merge this fix?  Thanks!

lbwexler: I took it out!

Issue #1943, opened 2020-06-01
Admin Activity Tab - follow-up improvements


amcclain:
Hi - here are some suggestions from me re. the newly upgraded Activity tab - looking good, will be great to push a bit farther this week.

### Grid
- [x] Make Track + Username columns wider
- [x] Rename Day > "Day / Range" - the range is pretty heavyweight in terms of how long the field gets / how wide the col needs to be, but it does seem like it might be useful...
- [x] Rename Avg Elapsed (ms) -> Elapsed (ms) to get header to fit, drop commas from number format
- [x] Hide Impersonating and Data columns by default
- [x] Consider adding small detail panel showing all fields of selected row instead of needing to go into (somewhat bad looking) modal.
- [x] Expand tree row on double-click (I still expect tree grids to always do this)
- [x] Make sure (default hidden) 'data' and 'message' both have an autosizeMax (Activity)
- [x] Make sure 'error' has an autosizeMax (Client Error)

### Query
- [x] Make default grouping Date > User > Category > Message
- [x] Query start date should go back one month by default (unless we find that's too slow, would hope not)
- [x] Reset query button should reset all fields and be to the right of all of them.
- [x] Maybe remove refresh button. We have app-level refresh in top toolbar, but I see how some users might want it to ensure query update is run (vs. pressing enter key)

### Chart
- [x] Use ButtonGroup (instead of tabs) to switch chart type between Bar & Line
- [x] Use Checkboxes or switches to control which series on chart - entries, unique users, average elapsed. Any combo could be shown, first two could share a Y axis.
- [x] Could tie the series on/off controls to the plotline color w/a colored dot or square next to label, avoiding need for chart legend.
- [x] Drop X axis label "Days"

lbwexler:
Like all of these changes.  Need to make sure (default hidden) 'data'  and 'message' both have an autosizeMax too!

febbraiod:
Like the idea of a small detail panel, but think we'll want to keep the dialog as the data column can have a lot in it, and the jsonEditor in the dialog seems like the best way to display it to me. I guess we'll see.

febbraiod:
Not sure about sharing axes here. Users and Logs for example could have a wide difference (other slices might be even worse)

<img width="1680" alt="Screen Shot 2020-06-08 at 12 21 09 PM" src="https://user-images.githubusercontent.com/15745614/84055360-b7a35a00-a982-11ea-83e0-d095374b4cb9.png">

[...]

PR #1951, opened 2020-06-08
Admin Activity Tab - follow-up improvements, Fixes: #1943

febbraiod:
Hoist P/R Checklist
-------------------


admin/tabs/activity/tracking/detail/ActivityDetailPanel.js
```diff
@@ -0,0 +1,32 @@
...
+    model: uses(ChartsModel),
+    render({model}) {
+        if (isEmpty(model.dimensions)) return null;
+        return panel({
+            items: hframe(
+                activityDetailTable(),
+                chartsPanel()
+            ),
+            model: {
+                defaultSize: 380,
```
febbraiod, line 26: Size chosen to help accommodate activityDetailTable (which still doesn't always fill the space nicely) perhaps there is a way to better style the table so that this can be taller?

admin/tabs/activity/tracking/ActivityModel.js
```diff
@@ -208,10 +207,13 @@ export class ActivityModel {
...
-            track: () => this.getParams(),
+            track: () => {
+                const {startDate, endDate, username, msg, category, device, browser} = this.formModel.values;
+                return [startDate, endDate, username, msg, category, device, browser];
+            },
             run: () => this.loadAsync(),
-            equals: comparer.structural,
-            fireImmediately: true
+            fireImmediately: true,
+            delay: 1
```
febbraiod, line 216: Here because the buttons that adjust the date fields can change both and cause a double network call.

lbwexler: we want to use debounce, not delay I think

admin/tabs/activity/tracking/ActivityPanel.js
```diff
@@ -38,32 +42,65 @@ export const activityPanel = hoistCmp.factory({
...
+                }),
+                formField({
+                    field: 'startDate',
+                    label: null,
+                    item: dateInput({...dateProps})
+                }),
+                Icon.caretRight(),
+                formField({
+                    field: 'endDate',
+                    label: null,
```
lbwexler, line 59: The label: null is a bit tedious, isn't it.

admin/tabs/activity/tracking/charts/ChartsPanel.js
```diff
@@ -7,32 +7,69 @@
...
-            bbar: bbar(),
-            model: {
-                defaultSize: 500,
-                side: 'bottom',
-                prefName: 'xhAdminActivityChartSize'
-            }
+            items: [
+                chart({
+                    model: model.categoryChartModel,
+                    omit: isTimeSeries
```
lbwexler, line 30: This would be much more straightforward with a ternary on the model and a single .  But I wonder if you would need to also assign the key of the component to the id of the model

febbraiod: I tried to do just that that but ran into issues. 

It seemed that the model switching wasn't being accepted by the chart for some reason. (It would just stay with the type of chart I had originally designated, even if the render function ran IIRC) Perhaps the chart itself wasn't rerendering despite the switch? Anyway I immediately knew this would work so I didn't really did any further.

lbwexler: yes, need to just assign the key to the component to get it to re-render when the model pulled out from under it.  but this is fine as well

amcclain:
I am going to pick this up today and had some ideas I wanted to try off of the latest - will report back with something (I hope) to review together.

amcclain:
Picking this effort up in #1976

PR #1952, opened 2020-06-08
Set --panel-bg in Hoist-React Admin app.

cnrudd:
Fixes a styling issue in the Admin app that Lee noticed on the IdleDialog.
If an App does not specify a --panel-bg color, an App's Panel background will fall back to 'transparent'

See 
https://github.com/xh/hoist-react/blob/develop/desktop/cmp/panel/Panel.scss#L3 
and
https://github.com/xh/hoist-react/blob/develop/styles/vars.scss#L229

### DIFF
--- a/admin/App.scss
+++ b/admin/App.scss
@@ -5,6 +5,10 @@
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
 
+body {
+  --panel-bg: var(--xh-bg);
+}
+
 .xh-admin-app-frame {
   .bp3-tabs.bp3-vertical > .bp3-tab-list {
     min-width: 130px;


amcclain:
I don't recall, but I think it's worth making that change to have --panel-bg == --xh-bg and we can look out for anything that breaks.

cnrudd:
> I don't recall, but I think it's worth making that change to have --panel-bg == --xh-bg and we can look out for anything that breaks.

@amcclain Do you mean we should do this?:
In this file: https://github.com/xh/hoist-react/blob/develop/styles/vars.scss#L229
replace  `--xh-panel-bg: var(--panel-bg, transparent);`
with `--xh-panel-bg: var(--panel-bg, --xh-bg);`

amcclain:
Yes - done in 0fd88e28bf4543b62bf8b6c243f54fc4a4e600e3 - I honestly don't know why it was how it was. I will keep an eye out for any problems!

Issue #1950, opened 2020-06-08
Use of ag-Grid autoGroupColumn throws in Hoist ColumnHeader


amcclain:
The code within (https://github.com/xh/hoist-react/blob/develop/cmp/grid/impl/ColumnHeader.js)[ColumnHeader] expects an `xhColumn` to be available, i.e. it expects that all headers/columns are created via the Hoist Column class, which passes a reference to its instance into our customized header component via `headerComponentParams`.

However if a developer is using agOptions to configure ag-provided (non-full-width) grouping, the grid will configure an auto grouping column that will not have an `xhColumn` property, but still get our customer header component as per the `frameworkComponents` we specify in `Grid.js`.

That custom header comp will then throw an NPE from https://github.com/xh/hoist-react/blob/bd46bb3efec58f7fd93cca0e9e41db17a2e3e3aa/cmp/grid/impl/ColumnHeader.js#L226

lbwexler:
Hi Colin -- Can you take a look at this one when you get a chance?

Thanks,

Lee

PR #1958, opened 2020-06-10
issue #1950 - ag-Grid-AutoColumn no longer throws, can now be sorted.

cnrudd:
This resolves the [issue](https://github.com/xh/hoist-react/issues/1950) identified by [John](https://github.com/johnbc) in which an ag-Grid-AutoColumn causes the browser to throw a JS error.

This change also brings back support for sorting that agGrid has in its ag-Grid-AutoColumn.  See the demo at https://www.ag-grid.com/javascript-grid-grouping/#configuring-auto-column

I was not able to bring back the rendering of a 'field' in the ag-Grid-AutoColumn.  It would be nice to bring that back.  Maybe John can test this branch and see if that works in his app, or someone with more experience coding our grid can point me to where the ag-Grid-AutoColumn's 'field' prop might be getting ignored/removed.

This toolbox branch: https://github.com/xh/toolbox/tree/agGridAutoColumnSupport has a new
ag-Grid-AutoColumn on the Orders grid of the Portfolio demo app, defined by: 
```
                    autoGroupColumnDef: {
                        suppressMenu: false,
                        filter: true,
                        headerName: 'B/S',
                        minWidth: 200,
                        field: 'trader'
                    }
```
 This makes it easier to test this change.  It might be good to keep at least one ag-Grid-AutoColumn around in toolbox so that we can see which of its features we have preserved as we continue to change our Grid and GridModel.

Hoist P/R Checklist
-------------------


cmp/grid/impl/ColumnHeader.js
```diff
@@ -107,8 +108,8 @@ class LocalModel {
     _lastTouchStart = null;
     _lastMouseDown = null;
 
-    constructor({gridModel, xhColumn, column: agColumn, enableSorting}) {
-        this.gridModel = gridModel;
+    constructor({gridLocalModel, xhColumn, column: agColumn, enableSorting}) {
+        this.gridModel = gridLocalModel.model;
```
amcclain, line 112: So this was broken before, but we weren't paying a price for it that we noticed?

Why not have the grid pass the actual `GridModel` in as a prop to the header for use directly here? I realize it's internal so am not stressed about it, but having any component need to be aware of another components localModel seems like an anti-pattern, even if they are part of a family.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ wide variety of enterprise software projects. For any questions regarding this c
     `DimensionChooserModel`, and `DashContainerModel`.
     configurable via the new `persistWith` option on those classes.
 * `Store` gets new `clearFilter()` and `recordIsFiltered()` helper functions.
+* Hoist now supports sorting on agGrid group columns.
```
amcclain, line 30: I would suggest moving under bugfixes.

PR #1968, opened 2020-06-19
Refactor of Persistence to support dynamic persistWith on instances

lbwexler:
This change was devised by John and I, as we tried to use the Persistence layer in the dash.  It add more flexibility for defining your persistWith for objects dynamically per instance, is more consistent with what we do for the components themselves, and gives the app much greater opportunity for re-use.  I think it also cleans things up by providing a PersistSupport mixin, 
which is a well-defined and documented host for the functionality:

In particular:

+ A dedicated PersistSupport annotation that mirrors our other mixins.
+ A new method markPersistent() that is useful. for dynamically wiring your persistent properties within the constructor.  Objects supporting dynamic persistence would use this *instead* of the annotation.
+ Reverse the order that we apply the decorators to a more natural order. 
+ Persistence is now part of the core:  less imports!

One question that I would like to revisit is the naming of the annotation and method.  Would like them to be named as similarly as possible, and also would be nice to have symmetry with @managed.  Was thinking of the following:

@persisted/markPersisted()
@persistent/markPersistent()
@persist/markPersistent().  [current]
@persistent/makePersistent()

etc., etc.

Hoist P/R Checklist
-------------------


core/mixins/PersistSupport.js
```diff
@@ -0,0 +1,144 @@
...
+             * directly on the property declaration itself.  Use this method in the general case,
+             * when you need to control the timing,
+             */
+            markPersistent(property, options = {}) {
+                const providers = this._xhPersistenceProviders = this._xhPersistenceProviders ?? {};
+
+                // Read from and attach to Provider.
+                // Fail gently -- initialization exceptions causes stack overflows for MobX.
+                try {
+                    const persistWith = {path: property, ...options, ...this.persistWith};
```
haynesjm42, line 47: Looks like `options` and `this.persistWith` need to be swapped here

lbwexler:
Reviewed by John

PR #1970, opened 2020-06-22
Display Timezones in Admin 

lbwexler:
NOTE: This change needs the related Hoist-Core update to see the timezones, but should be backward compatible.

Hoist P/R Checklist
-------------------


admin/tabs/general/about/AboutPanel.js
```diff
@@ -41,30 +43,33 @@ function renderTables() {
...
-                row('Environment', svc.get('appEnvironment')),
-                row('Database', svc.get('databaseConnectionString')),
-                row('DB User / Create Mode', `${svc.get('databaseUser')} / ${svc.get('databaseCreateMode')}`),
+                row('App Name / Code', `${get('appName')} / ${get('appCode')}`),
+                row('Environment', get('appEnvironment')),
+                row('Database', get('databaseConnectionString')),
+                row('DB User / Create Mode', `${get('databaseUser')} / ${get('databaseCreateMode')}`),
+                row('App Timezone', fmtTimezone(get('appTimezone'), get('appTimezoneOffset'))),
+                row('Server Timezone', fmtTimezone(get('serverTimezone'), get('serverTimezoneOffset'))),
+                row('Client Timezone',  fmtTimezone('Unknown', (new Date()).getTimezoneOffset()*MINUTES*-1)),
```
amcclain, line 52: Note we could use `Intl.DateTimeFormat().resolvedOptions().timeZone` if we wanted the display the "Americas/Los Angeles" form of the client TZ alongside the more descriptive zone names for server/app.

(Also note the capital Z in that API.)

admin/tabs/general/about/AboutPanel.js
```diff
@@ -24,15 +26,15 @@ export const aboutPanel = hoistCmp.factory(
 );
 
 function renderTables() {
-    const svc = XH.environmentService,
-        startupTime = svc.get('startupTime'),
+    const get = (str) => XH.environmentService.get(str),
```
amcclain, line 29: 👍  😃

cmp/layout/Tags.js
```diff
@@ -42,4 +42,5 @@ export const textarea = elemFactory('textarea');
 export const th = elemFactory('th');
 export const thead = elemFactory('thead');
 export const tr = elemFactory('tr');
-export const ul = elemFactory('ul');
\ No newline at end of file
+export const ul = elemFactory('ul');
+export const b = elemFactory('b');
```
amcclain, line 46: Nit - could we use `strong` instead, or `span({className: 'xh-bold'})`?  

(See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b re `b` as tag of last resort.)

admin/tabs/general/about/AboutPanel.js
lbwexler: went with the capital Z on server and client -- think that makes sense(although our current Clock component is using the small Z as the JavaScript Date API....)

Fortunately, moment uses `tz` escaping the issue largely!

cmp/layout/Tags.js
lbwexler: 10-4

PR #1987, opened 2020-07-07
Pref differ

lbwexler:
Hoist P/R Checklist
-------------------


lbwexler:
Hi Colin -- thanks for taking a careful look!

None of these are regressions from the current config differ, but the javascript error is easily reproducible, and is something we should understand, even if it is not causing any obvious issues -- I will take a quick look at that and see if I can address.

I don't agree about calling [ ] equal to [].  We are just looking for absolute equality of the raw data in the primitive form that it is living in the database in.  Don't want to make this thing smarter than it needs to be

lbwexler:
Worked around that error (and cleaned up the initial state of the dialog when opened the second time).  

But we do have an annoying issue where a gridmodel that is cleared/loaded just before its grid is unmounted is going to cause problems.  We can file a seperate bug on this.

PR #1999, opened 2020-07-14
Track app version/build/href with initial load

amcclain:
+ In particular, we don't otherwise/generally track the app version with each entry, but including that with the initial load bundle seemed expedient.

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/XH.js
+++ b/core/XH.js
@@ -691,12 +691,11 @@ class XHClass {
                     case AppState.RUNNING:
                         XH.track({
                             category: 'App',
-                            msg: `Loaded ${this.clientAppName}`,
+                            msg: `Loaded ${this.clientAppCode}`,
                             elapsed: now - loadStarted - loginElapsed,
                             data: {
                                 appVersion: this.appVersion,
                                 appBuild: this.appBuild,
-                                clientAppCode: this.clientAppCode,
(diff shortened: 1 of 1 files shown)


amcclain:
Two small notes:

- I didn't think these new bits of data belonged within the `getClientDeviceInfo()` util method, so I inlined them here. 
- I poked around in `window` within my browser to see if there was anything else that seemed like a slam dunk to include. You will see I added the href as I could see that being useful for cases where we had e.g. an app reachable at a variety of routes, or via a mobile + internal URLs, or HTTPS or not, and knowing exactly what the URL was in the browser might shed some light on something. If you can think of anything else that would clearly be high value to include, this seems like a good way to do it. But I wasn't looking to pile on too much!

lbwexler:
Looks great -- one nit:

Wasn't sure about clientAppCode -- should it be just used in the msg instead of clientAppName?  

Seems like we want to be able to change the name at any time without screwing up our tracking -- the code should be more stable.

amcclain:
Certainly could - I don't feel strongly about it.  Not sure how stable/consistent the clientAppCode values are across apps - my sense is that we would likely get things like "app" and "desktop" - which would be fine, and maybe even more useful than the name. I can make that tweak.

PR #2002, opened 2020-07-15
Filter model

lbwexler:
Please see Toolbox PR: https://github.com/xh/toolbox/pull/419

Hoist P/R Checklist
-------------------


data/filter/Filter.js
```diff
@@ -0,0 +1,85 @@
...
+     * @param {*} [c.value] - value to use with operator in filter
+     */
+    constructor({
+        field,
+        operator,
+        value
+    }) {
+        throwIf(!isString(field), 'Filter requires a field');
+        throwIf(
+            !isString(operator) || !VALID_OPERATORS.includes(operator),
```
lbwexler, line 33: why do we need the first clause of this check on line 33

cmp/filter/FilterFieldModel.js
```diff
@@ -0,0 +1,201 @@
...
+    // Immutable properties
+    filterOptionsModel = null;
+    limit = null;
+    maxHistoryLength = null;
+
+    @managed provider;
+
+    /**
+     * @param c - FilterFieldModel configuration.
+     * @param {(FilterOptionsModel|Object)} filterOptionsModel - FilterOptionsModel, or config to create one.
```
lbwexler, line 31: first param should indicate that it is part of c

cmp/filter/FieldFilterSpec.js
```diff
@@ -0,0 +1,90 @@
...
+    /** @member {string} */
+    field;
+
+    /** @member {string} */
+    displayName;
+
+    /** @member {string} */
+    fieldType;
+
+    /** @member {*[]} */
```
lbwexler, line 28: is this jsdoc syntax correct?

cmp/filter/FieldFilterSpec.js
```diff
@@ -0,0 +1,90 @@
...
+import {fmtDate} from '@xh/hoist/format';
+import {throwIf, stripTags} from '@xh/hoist/utils/js';
+import {isString, startCase} from 'lodash';
+
+/**
+ * Defines the filter options available for a given store field, and also
+ * provides useful metadata for including these options in UI affordances.
+ * Typically generated from a Store via @see FilterOptionsModel
+ */
+export class FieldFilterSpec {
```
lbwexler, line 17: Is this intended to be immutable?

cmp/filter/FilterFieldModel.js
```diff
@@ -0,0 +1,296 @@
...
+        filterModel,
+        filterOptionsModel,
+        limit,
+        persistWith,
+        maxHistoryLength = 5
+    }) {
+        throwIf(!filterModel, 'Must provide a FilterModel (or a config to create one)');
+        throwIf(!filterOptionsModel, 'Must provide a FilterOptionsModel (or a config to create one)');
+
+        if (isPlainObject(filterModel)) {
```
lbwexler, line 50: nice!  Can use ternary operator here and below for clarity

cmp/filter/FilterFieldModel.js
```diff
@@ -0,0 +1,296 @@
...
+
+        this.limit = limit;
+        this.maxHistoryLength = maxHistoryLength;
+
+        // Read state from provider -- fail gently
+        if (persistWith) {
+            try {
+                this.provider = PersistenceProvider.create({path: 'filterField', ...persistWith});
+                const state = this.provider.read();
+                if (state?.history) this.history = state.history;
```
lbwexler, line 70: should we store/rehydrate value as well?

TomTirapani: Honestly, the persistence isn't really in place yet... that's on the list of things to discuss

data/cube/Query.js
```diff
@@ -54,9 +57,12 @@ export class Query {
         this.cube = cube;
         this.fields = this.parseFields(fields);
         this.dimensions = this.parseDimensions(dimensions);
-        this.filters = this.parseFilters(filters);
         this.includeRoot = includeRoot;
         this.includeLeaves = includeLeaves;
+
+        this._filterModel = new FilterModel({
```
lbwexler, line 63: probably should be @managed.

data/cube/Query.js
```diff
@@ -18,15 +18,21 @@ export class Query {
...
-    /** @member {StoreFilter[]} */
-    filters;
     /** @member {boolean} */
     includeRoot;
     /** @member {boolean} */
     includeLeaves;
     /** @member {Cube} */
     cube;
 
+    @managed
```
amcclain, line 28: We need to apply `@ManagedSupport` to the class for this to have an effect, right?

I looked briefly to see if we could warn in the `@managed` decorator if the target wasn't marked as `isManagedSupport` but I believe that the class member decorators are being evaluated before the top-level decorator - or at least, when I logged a warning it looked to me like _all_ uses were warning with no targets having the flag set when I was attempting to run that check...

amcclain:
Just realized an issue with the admin-specific refactoring of the activity viewer to do all the filtering using our beautiful new client-side model - we have a hard-coded limit of 10k rows returned from the server:

https://github.com/xh/hoist-core/blob/750b1bd72d72f95d899ca36f00ea2c904c421d79/grails-app/controllers/io/xh/hoist/admin/TrackLogAdminController.groovy#L23

For a busy app with a lot of tracking, this means it could be impossible to search for some kind of activity across a large enough date range to be useful. It also doesn't help that we don't indicate to the user that the rows have been truncated.

We can discuss a few options:
- Restore a mix of server-side querying + client-side filtering. We could even explore the advanced use case of observing the FilterModel filters to extract some of them and pass to the server while letting others be applied locally. 
  - Would need to get the filter options to know about the server-side facets, given that they could not be extracted from a complete set of local data....
  - Local `FilterModel` could still apply all the filters to the cube - the ones that have already been applied by the server would be no-ops since all records would pass. Might reduce complexity a tiny bit - i.e. we could get away _without_ a mechanism to declaratively tell the client-side model to ignore certain filters.
- Raise the limit / make the limit a user-adjustable parameter.
- Show a clear warning indicator when the results have been topped.

data/Store.js
```diff
@@ -78,20 +81,32 @@ export class Store {
...
         this.idSpec = isString(idSpec) ? (data) => data[idSpec] : idSpec;
         this.processRawData = processRawData;
+        this.filterModel = this.parseFilterModel(filterModel);
         this.lastLoaded = this.lastUpdated = Date.now();
         this._loadRootAsSummary = loadRootAsSummary;
 
         this.resetRecords();
-        this.setFilter(filter);
+
+        warnIf(!isNil(filter), "The 'filter' config has been deprecated. Use 'filterModel' instead");
```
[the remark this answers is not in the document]
TomTirapani: Think that's a great idea. I've added it to this branch, and used it here.

Also opened this PR to add it separately: https://github.com/xh/hoist-react/pull/2021

amcclain:
I've noticed that the filters will sometimes reorder themselves after entry (let me know if I imagined that but I am pretty confident it happens, not sure how to reliably repro). 

It would be good from a usability perspective if we could avoid that - I think it would be a common-enough scenario that a user enters a filter and then decides they immediately want to remove it by hitting the backspace key. If the filter you've just entered is not the last filter in the list, you can end up deleting another filter unintentionally.  (Plus it's just a bit jumpy.)

I am not worried about ensuring the order is stable across persistence - just when the user is actively editing. Of course also open to discussing counter-arguments if we think it's more important to reorder to keep filters together by field or similar.

amcclain:
This is a bigger topic and I want to look over it a bit more / can try to come up with better-structured thoughts - but I wanted to get it into this thread in case it's on other people's minds.

Do we have any opportunities to better consolidate field metadata by encouraging developers to lean more heavily on the `Field` class configuration - thinking most specifically about the label, but could imagine a future where renderers, export info, or other metadata also can be attached to `Field`.

For the labels alone, ideally a central config on Field could be used to (optionally) populate/default:
* `Column.headerName`
* `FilterOptionSpec.displayName`
* `DimChooserModel.dimension.label`
* `FieldModel.displayName` < in forms package, in some near-future world of inline grid editing where we are setting up FormModels for rows under edit.

This flips the script a bit, where for most use cases we have the grid columns hold the most information about a field, and then are auto-creating the store fields based on that.  But in a new world where we need the type information on field to do the filtering well, the dev will be in the business of specifying those store fields more completely, and it would be really great to avoid then having to still specify the same info several other places.

amcclain:
It would be a small incremental improvement if we could avoid the requirement that the user type spaces around the operators.  Right now if I type `elapsed>` in the filter chooser, I don't get any suggestions - need to type `elapsed >` with the space to proceed. I can imagine users getting caught / confused by this.

amcclain:
Update for you @TomTirapani and anyone else following this PR.  @lbwexler and I spent a while going over the new system and had some updates we want to make - notably around a few carefully-chosen breaking changes to rationalize some names, commit to `Field.displayName` as the best place to set the user-friendly name, etc. 

I am going to transcribe my rough notes and either add them in a structured form to this PR and/or make some directly. We can touch base tomorrow (Tuesday) re. time/work management for this week to get it done, balancing with client projects, etc.  More to follow.

cmp/filter/impl/FilterChooserFieldSpec.js
```diff
@@ -154,11 +154,26 @@ export class FilterChooserFieldSpec {
...
+    }
+
     parseOperators(operators) {
-        operators = operators ?? this.filterType === 'value' ? ['=', '!=', 'like'] : ['>', '>=', '<', '<=', '=', '!='];
+        operators = operators ?? this.getDefaultOperators();
         return operators.filter(it => FieldFilter.isValidOperator(it));
     }
 
+    getDefaultOperators() {
+        if (this.fieldType === FieldType.BOOL) return ['=', '!='];
```
amcclain, line 173: Supporting e.g. `!= true` would allow a user to filter on records with false OR null values in the field, but I wonder if that's something that users would expect.  As opposed to only providing `=` along with the true/false value suggestions.  

Or to put it another way - should we drop `!=` to simplify or is that functionality useful (or does it seem otherwise more consistent with both operators)?

amcclain:
Additional items for review - discussed on Slack today after the latest round of updates, but good I think to record here so we can revisit in a more managed fashion:

- [ ] Avoid rebuilding `FilterFieldChooserSpec` objects and eagerly extracting unique values from the store on every store update. This could potentially be a lot of work for a large / rapidly updating store maintaining value suggestion lists for field filters the user never has cause to activate.  Review if the spec could take the store reference and lazily build its value list when called for, ideally with an optimization to avoid rebuilding if the data has not changed.
- [ ] Review options for more flexible date parsing and relative date filters (e.g. dateCreated > startOfYear).  The latter suggestion should almost certainly be a follow-on ticket for later review after we have some integrations up and running, but we might need to look sooner at how we could handle date/time inputs w/o needing to install custom `valueParser` functions in multiple places.

cmp/filter/FilterChooserModel.js
```diff
@@ -0,0 +1,606 @@
...
+     * @param {(FilterModel|Object)} c.filterModel - FilterModel, or config to create one.
+     * @param {Store} c.store - Store to use for Field resolution as well as extraction of available
+     *      Record values for field-specific suggestions. Note that configuring the store here does
+     *      NOT cause that store to be automatically filtered or otherwise bound to the FilterModel.
+     * @param {(string[]|FilterChooserFieldSpecConfig[])} [c.fieldSpecs] - specifies the Store
+     *      Fields this model will support for filtering and customizes how their available values
+     *      will be parsed/displayed. Provide simple Field names or `FilterChooserFieldSpecConfig`
+     *      objects to select and customize fields available for filtering. Optional - if not
+     *      provided, all Store Fields will be included with options defaulted based on their type.
+     * @param {string} [c.valueSourceRecords] - determines the set of Store Records used to extract
```
amcclain, line 81: This config was named `storeMode` in an earlier iteration, which I found to be a bit too general / non-descriptive. However I am not wild about my new name (it does imply that it would be an actual collection of records...) and am open to any suggestions for improvement.

I also am not personally clear on when you should override the default - we should provide guidance on when you would choose one over the other.

cmp/filter/FilterChooserModel.js
```diff
@@ -0,0 +1,606 @@
...
+ *     class. Identical to `FilterChooserFieldSpecConfig` but with resolved `Field` instance.
+ * @property {Field} field
+ * @property {string} displayName
+ * @property {string[]} operators
+ * @property {boolean} suggestValues
+ * @property {[]} values
+ * @property {FilterOptionValueRendererCb} [valueRenderer]
+ * @property {FilterOptionValueParserCb} [valueParser]
+ * @property {*} [exampleValue]
+ */
```
amcclain, line 585: I fully admit that maintaining/typing this intermediate object is inelegant.  This could be cleaned up if we decided to create `FilterChooserFieldSpec`s once and then let them internally update in response to store changes. We keep these processed specs around in this model so we can pre-process once and then re-create as needed.

And this is also a case where the typing is not for public documentation but really helped me stay sane and keep track of what these objects have - especially with 'field' which can reference both field names and Field instances throughout the codebase.

cmp/store/impl/StoreFilterFieldImplModel.js
```diff
@@ -115,38 +117,41 @@ export class StoreFilterFieldImplModel {
...
         if (!initializing && this.onFilterChange) this.onFilterChange(newFilter);
 
         // Only respect the buffer for non-null changes.  Allows immediate initialization and quick clearing
         if (!initializing && newFilter) {
             this.bufferedApplyFilter();
         } else {
-            this.applyFilter();
+            // start() here required to resolve a "Cannot update during an existing state transition"
+            // React error. Further investigation required.
+            start(() => this.applyFilter());
```
amcclain, line 154: Thanks for the note here - don't have any immediate answer and did not look at this in any detailed way, but calling out via PR comment so we can track and resolve.

data/cube/Query.js
```diff
@@ -73,18 +81,32 @@ export class Query {
         return new Query(conf);
     }
 
+    /** @return {string} */
+    static filterAsString(filter) {
+        const {field, operator, value} = filter;
+        return `${field}${operator}[${castArray(value).join('||')}]`;
+    }
```
amcclain, line 88: Does this belong here or is this something more general ?  Why not instance method on FieldFilter?

data/cube/View.js
```diff
@@ -207,7 +192,8 @@ export class View {
         appliedDimensions = {...appliedDimensions};
         return map(groups, (groupLeaves, val) => {
             appliedDimensions[dimName] = val;
-            const id = parentId + Cube.RECORD_ID_DELIMITER + ValueFilter.encode(dimName, val);
+            const filter = new FieldFilter({field: dimName, operator: '=', value: val});
```
amcclain, line 195: Do we need to provide `fieldType`?

data/filter/FieldFilter.js
```diff
@@ -0,0 +1,153 @@
...
+    get isFieldFilter() {return true}
+
+    /** @member {string} */
+    field;
+    /** @member {string} */
+    operator;
+    /** @member {(*|[])} */
+    value;
+    /** @member {FieldType} */
+    fieldType;
```
amcclain, line 30: Arg it really bugs me that we need to carry the type here but I can see how the requirement to serialize / deserialize and also keep this general makes it really tricky to maintain a link to a real `Field` instance or a `Store` that could provide one from the name.  ☹️

cmp/filter/FilterChooserModel.js
```diff
@@ -0,0 +1,606 @@
...
+        throwIf(!filterModel, 'Must provide a FilterModel (or a config to create one).');
+        throwIf(!store, 'Must provide a Store to resolve Fields and provide value suggestions.');
+        throwIf(!['filtered', 'all'].includes(valueSourceRecords), `Invalid valueSourceRecords config '${valueSourceRecords}'.`);
+
+        this.filterModel = isPlainObject(filterModel) ? this.markManaged(new FilterModel(filterModel)) : filterModel;
+        this.store = store;
+        this._rawFieldSpecs = this.parseRawFieldSpecs(fieldSpecs);
+
+        this.addReaction({
+            track: () => this.store.lastUpdated,
```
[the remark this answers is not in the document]
amcclain: This almost certainly could/should get broekn out into its own ticket

admin/tabs/activity/tracking/ActivityTrackingModel.js
```diff
@@ -42,79 +48,106 @@ export class ActivityTrackingModel {
...
+                    field: 'elapsed',
+                    valueRenderer: (v) => {
+                        return fmtNumber(v, {
+                            label: 'ms',
+                            formatConfig: {thousandSeparated: false, mantissa: 0}
+                        });
+                    }
+                },
+                {
+                    field: 'dateCreated',
```
lbwexler, line 131: why not have user filter on localDate 'day' instead here?

admin/tabs/activity/clienterrors/ClientErrorsModel.js
```diff
@@ -178,21 +231,22 @@ export class ClientErrorsModel {
...
 
     getParams() {
-        return {
-            startDate: this.startDate,
-            endDate: this.endDate,
-            username: this.username,
-            error: this.error
-        };
+        const {startDate, endDate} = this;
+        return {startDate, endDate};
```
lbwexler, line 245: This can actually be a one liner: 
`return {startDate, endDate} = this`;

TomTirapani: FYI, I found this on SO which says doing it as a one-liner creates global vars:
https://stackoverflow.com/questions/41794086/es6-destructuring-within-a-return-statement

lbwexler:
great catch tom!

lbwexler:
Have been tweaking some of the FilterModel stuff and working with Tom on that, but wanted to float a couple of  API questions:

 + What do folks think of changing FilterChooserModel.limit to FilterChooserModel.maxResults?
+ Any ideas for alternatives to FilterChooserModel.valueSourceRecords ?  I find this name to be confusing.

lbwexler:
Also -- Am inclined to *not* deprecates the `Store.setFilter` method and the `filter` argument to its constructor.  

Its a very natural way to set actual filters on a store, if you don't care about observing the model, replacing filters, etc. etc.

Can support setting multiple filters of any type on the Store, and seems like it could remain a first class member of the API.

amcclain:
> * What do folks think of changing FilterChooserModel.limit to FilterChooserModel.maxResults?

I agree that would be more clear. Would `maxSuggestions` be even more accurate / specific in a useful way? Is that what we're talking about?

> * Any ideas for alternatives to FilterChooserModel.valueSourceRecords ?  I find this name to be confusing.

Thinking more about this - do we expect this to be a binary choice - will there be some third mode/source in the future?  If it could be a boolean arg we might have more options for better names.

But I confess I still don't have a mental model of when we use one vs. another - if someone has time to chat about it would be helpful.

TomTirapani:
> What do folks think of changing FilterChooserModel.limit to FilterChooserModel.maxResults?

Agree with this. I'm inclined to `maxResults` rather than `maxSuggestions`, since I think its a more accurate name. 'Suggestions' is used in the API to mean the autofill syntax suggestions, whereas this config is used to limit the number of *value* suggestions when the '=' operator is used for facet-type searching.

> Any ideas for alternatives to FilterChooserModel.valueSourceRecords ? I find this name to be confusing.

Can't imagine any third mode off the top of my head, so perhaps a boolean arg would be better. Perhaps `filterRecordValues`?

The use case for this flag is to allow the dev to control whether or not applied filters restrict the suggested values. I recall we discussed wanting to keep the control separate from the grid, so you could pick a value that's not necessarily in the current results (although I'm afraid I can't remember the specifics).

> Am inclined to not deprecates the Store.setFilter method and the filter argument to its constructor.

I also agree - at this point, I think we can unwind those deprecation messages.

PR #2004, opened 2020-07-15
Allow columns property on ExportOptions to take a function

lbwexler:
+ Note that for consistency, changing the semantics such that we always return explicitly enumerated columns
 from columns, via array or function.

+ This also fixes an important bug in the sorting of ExportOptions that was recently introduced and has not yet been released

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/GridExportService.js
+++ b/svc/GridExportService.js
@@ -112,6 +112,8 @@ export class GridExportService {
     // Implementation
     //-----------------------
     getExportableColumns(gridModel, columns) {
+        if (isFunction(columns)) return columns(gridModel);
+
         const toExport = castArray(columns),
             includeAll = toExport.includes('ALL'),
             includeViz = toExport.includes('VISIBLE');
@@ -122,11 +124,9 @@ export class GridExportService {
         }).filter(col => {
             const {colId, excludeFromExport} = col;
             return (
-                !excludeFromExport &&
-                (
-                    includeAll ||
-                    toExport.includes(colId) ||
-                    (includeViz && gridModel.isColumnVisible(colId))
+                toExport.includes(colId) ||
+                (!excludeFromExport &&
+                    (includeAll || (includeViz && gridModel.isColumnVisible(colId)))
                 )
             );
         });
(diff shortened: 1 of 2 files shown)


svc/GridExportService.js
```diff
@@ -112,6 +112,8 @@ export class GridExportService {
     // Implementation
     //-----------------------
     getExportableColumns(gridModel, columns) {
+        if (isFunction(columns)) return columns(gridModel);
```
TomTirapani, line 115: The function form is documented as returning column Ids, but that wouldn't work here, since `getExportableColumns` return column instances. I think the function form *should* return ids, but we've have to do a bit of work here to get the instances from the GridModel.

svc/GridExportService.js
```diff
@@ -275,8 +275,9 @@ export class GridExportService {
...
  * @property {string} [options.type] - type of export - one of ['excel', 'excelTable', 'csv'].
- * @property {(string|string[])} [options.columns] - columns to include in export. Supports tokens
- *      'VISIBLE' (default - all currently visible cols), 'ALL' (all columns), or specific
- *      colIds to include (can be used in conjunction with VISIBLE to export all visible and
- *      enumerated columns).
+ * @property {(string|string[]|function)} [options.columns] - columns to include in export. Supports
+ *      tokens 'VISIBLE' (default - all currently visible cols), 'ALL' (all columns), or specific
+ *      column IDs to include (can be used in conjunction with VISIBLE to export all visible and
+ *      enumerated columns). Also supports a function taking the GridModel, and returning an array
+ *      of column IDs to include.
```
TomTirapani, line 282: See comment above

svc/GridExportService.js
```diff
@@ -122,11 +124,9 @@ export class GridExportService {
...
             const {colId, excludeFromExport} = col;
             return (
-                !excludeFromExport &&
-                (
-                    includeAll ||
-                    toExport.includes(colId) ||
-                    (includeViz && gridModel.isColumnVisible(colId))
+                toExport.includes(colId) ||
+                (!excludeFromExport &&
+                    (includeAll || (includeViz && gridModel.isColumnVisible(colId)))
```
TomTirapani, line 129: 👍

svc/GridExportService.js
lbwexler: great catch -- fixed this, and tested properly.

Note that the testing also helped me find and fix an unrelated problem in the new sorting code that you should also review as well

lbwexler:
CR:  Colin & Tom

PR #2014, opened 2020-07-23
Add valid email constraint

petradish:
Addresses https://github.com/xh/hoist-react/issues/2013
Added toolbox example, which replaces previous email validation: https://github.com/xh/toolbox/pull/417

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 🎁 New Features
+
+* Added a `validEmail` constraint for forms.
+
 ## v35.2.0 - 2020-07-21
 
 ### 🎁 New Features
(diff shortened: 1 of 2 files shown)


amcclain:
The comments in the linked stack overflow mention some problems / valid emails that are rejected here - there appear to be some really weird ones I am not worried about but `username+something@gmail.com` is also called out as being rejected, which does look like a very plausible address (+ in the first part).  Did / could you please test that one in particular?

petradish:
Hey, Anselm -- that email does pass validation with this regex test. I tested a bunch of whacky ones (lots of symbols, which most email providers would also not allow anyway) and they still passed validation. Emails like petra.@xh.io would fail validation tho (can't have a special char like . before the @)

amcclain:
Great thanks much for checking and humoring me here.  Those massive regexes always a bit of a black hole - a big chunk of this one also appears to be related to handling IP addresses in the domain part, but I am not inclined to touch it!

If we run into problems (unlikely) we could also look at a library like https://github.com/scottgonzalez/sane-email-validation - but let's run with this for now, thanks again!

PR #2016, opened 2020-07-27
Rationalize and improve Button styles

amcclain:
Hoist P/R Checklist
-------------------


amcclain:
Hi @TomTirapani - this started with a "small" change Petra and I went to make to get our desktop buttons to respect our CSS "intent" variables when in minimal/outlined mode.  Turned into a giant saga to try to reconcile the blueprint button overrides that had built up and get some kind of control over all the variations on the coloring that can happen between the button states, light/dark mode, and the intents.

If you have a chance to look at this and give it a sanity check I would appreciate it.  We absolutely have a big muddle of styles and approaches when it comes to these colors now, and I've introduced the use of HSL colors to add to that muddle, but this ended up being my best shot at getting a set of variables together that could be adjusted in a maybe-rational way.

Didn't touch mobile - need to add a changelog and ensure that we didn't break anything there - but maybe we could use the same vars in another iteration (does NOT need to be now, mainly want to get this merged safely) to support more customization options for mobile as well.  Thanks!

desktop/cmp/button/Button.scss
```diff
@@ -5,16 +5,376 @@
...
+.xh-app.xh-dark.bp3-dark {
+  .xh-button.bp3-button {
+    border-radius: var(--xh-border-radius-px);
+    font-family: var(--xh-button-font-family);
+    font-size: var(--xh-button-font-size-px);
+    white-space: nowrap;
+
+    // Shaded/filled-in (non-minimal) mode.
+    &.xh-button--standard {
+      padding: var(--xh-pad-half-px) var(--xh-pad-px);
```
TomTirapani, line 22: Curious why standard has a wider horizontal padding than outlined? I can understand the rationale for reduced padding in minimal mode, but I would expect outlined and standard to take the same width.

desktop/cmp/button/Button.scss
```diff
@@ -5,16 +5,376 @@
...
+    }
+
+    &.bp3-disabled {
+      opacity: 0.6;
+    }
+
+    &--autofocus-enabled {
+      // Hoist calls FocusStyleManager.onlyShowFocusOnTabs(), which by default hides focus indicators
+      // unless a component has been tabbed into via keyboard. Override here to always show a focus
+      // indicator for an autoFocus button so it's clear that it has focus.
```
TomTirapani, line 38: Nice 👍

desktop/cmp/input/ButtonGroupInput.js
```diff
@@ -65,23 +72,26 @@ export class ButtonGroupInput extends HoistInput {
...
+                },
+                // Workaround for https://github.com/palantir/blueprint/issues/3971
+                key: `${active} ${value}`
             });
         });
 
         return buttonGroup({
             items: buttons,
             ...buttonGroupProps,
+            minimal: withDefault(minimal, outlined, false),
```
TomTirapani, line 94: What's going on here? Do 'outlined' buttons need to be in a 'minimal' buttonGroup?

styles/vars.scss
```diff
@@ -46,6 +46,92 @@ body {
...
+  --xh-intent-warning-trans2: hsla(var(--xh-intent-warning-h), var(--xh-intent-warning-s), var(--xh-intent-warning-l3), var(--xh-intent-a2));
+
+  --xh-intent-danger-darkest: hsl(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l1));
+  --xh-intent-danger-darker: hsl(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l2));
+  --xh-intent-danger: hsl(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l3));
+  --xh-intent-danger-lighter: hsl(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l4));
+  --xh-intent-danger-lightest: hsl(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l5));
+  --xh-intent-danger-trans1: hsla(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l3), var(--xh-intent-a1));
+  --xh-intent-danger-trans2: hsla(var(--xh-intent-danger-h), var(--xh-intent-danger-s), var(--xh-intent-danger-l3), var(--xh-intent-a2));
+
```
TomTirapani, line 134: This is a lot of code, but HSL-adjusted color variations are great to have :)

Wonder if we can write a SASS function to create these for a given color? (not saying we do that now)

desktop/cmp/input/ButtonGroupInput.js
amcclain: Yes, it's a great question. 

BP supports the minimal modifier on ButtonGroup but not outlined. Prior to this change, you could set `outlined:true` on the _child_ buttons and it would work, but you could not set it on the group itself. Seemed nice to make it more consistent since we're in the business of unpacking these props anyway and applying them to children ourselves.

BUt I need to look again though at what this specific default is doing - if anything - it might not be required.

styles/vars.scss
amcclain: I completely agree - a lot of code, pretty maddening.  The combinatory effects between:

+ 5 intent-styles (four intents + no-intent) x
+ three buttons states (normal/hover/pressed) x
+ two themes

made for a lot of variations, with some surprising differences in how BP was choosing to make colors get lighter/darker between the states and themes.

Unfortunately the SASS color functions don't work with CSS vars since the SASS functions are run prior the vars being resolved at runtime in the cascade.  A (big) part of me thinks that we might have been better using SASS all the way, like we did in Sencha world, so we could use those functions and the nicer variable syntax. But then again the CSS vars still somehow seem like a newer / more standards-compliant solution, and don't require recompilation which is really nice in terms of tweaking in the dev tools, etc.  Tradeoffs for sure!

PR #2029, opened 2020-08-05
Defend against accidental double-submit of login

amcclain:
+ Debounce LoginPanelModel.submitAsync (I think this is the real fix) and also disable / modify login button text to indicate login in progress.
+ Await XH.completeInitAsync() to avoid taking down login panel mask before the app has re-rendered with its primary interface.
+ Should avoid the rare-but-persistent double init exception (manifests as an exception re. re-init of IdentityService).

Hoist P/R Checklist
-------------------


amcclain:
Didn't think this needed a CHANGELOG entry but happy to add if anyone thinks otherwise.

lbwexler:
👍 looks like this solidifies things in a couple of ways....

Issue #2027, opened 2020-08-04
Column headers should show full text in title tip when elided


amcclain:
If no specific tooltip is configured for a grid column header, and if the header text is elided due to it being longer than the column is wide, it would be a usability win if the full header text was automatically made available in a title tip on hover.

cnrudd:
Here is a good starting point on how one might implement this: https://stackoverflow.com/a/52930448/1893251
And a good discussion about other approaches in that same SO article.

Issue #2006, opened 2020-07-16
Some string values can cause URIError: URI malformed error when using  tooltipElement propery in grid column.


johnbc:
For example, a column like 
```
{
                field: 'UserComment',
                headerName: 'User Comments',
                width: 200,
                tooltipElement:  (v)=>  div({style: {width: '400px', border: '1px outset grey', padding: '5px'}, item: v })
}
```

Will throw a **URIError: URI malformed** error while trying to activate the tooltip if the tooltip value contains a % sign.

This issue appears to be related to the usage of decodeURIComponent(value) this line in Columns.js

https://github.com/xh/hoist-react/blob/9a0537450cf7d99fd56b21760502585e6e1faf37/cmp/grid/columns/Column.js#L387

I'm not clear on the exact flow here, but potentially passing escape(value) instead of the raw value will fix the issue.

lbwexler:
John - saw that you added this code in December, including a very specific comment about the encoding/decoding.  Can you take a look.  

Would we be better off passing our tooltipElement function the raw value to be rendered?  We can get it, using the record, which we have, and the value getter.  

Is the concern for grouped rows for which we have no record?  Not sure there is an expectation to get a tooltip on those.

PR #2031, opened 2020-08-06
Elided header tooltip + bonus bugfix!

cnrudd:
This resolves: https://github.com/xh/hoist-react/issues/2027

This PR brings in a small code change to automatically show the full headerName value in a browser "title" attribute tooltip if the headerName is elided in a Grid column header.

If the column width is expanded enough so that the headerName is no longer elided, the title tip no longer shows.

This logic does not run if the developer has already specified a `headerTooltip` on the column.

This PR also fixes a minor bug: prior, if a developer had specified a tooltipElement and a headerTooltip on the same column, mousing over the column header triggered a JS error.
This PR fixes this by passing the headerTooltip as the value to the tooltipElement when mousing over header.
See agGrid docs https://www.ag-grid.com/javascript-grid-tooltip-component/#header-tooltip-with-custom-tooltip  where they detail that this is what they expect a tooltipElement/Component to do.

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.
- [x] Updated doc comments / prop-types: not required
- [x] Created Toolbox branch / PR: https://github.com/xh/toolbox/tree/elidedHeaderTooltip

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -394,7 +394,7 @@ export class Column {
                     }
 
                     const record = api.getDisplayedRowAtIndex(rowIndex).data,
-                        value = record.get(colDef.colId);
+                        value = record.data[colDef.field];
                     return me.tooltipElement(value, {record, column: me, gridModel, agParams});
                 }
             };
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -19,6 +19,7 @@ import {
     isString,
     startCase
 } from 'lodash';
+import {div} from '@xh/hoist/cmp/layout';
 import {Component} from 'react';
 import {ExportFormat} from './ExportFormat';
 import {GridSorter} from '../impl/GridSorter';
@@ -377,15 +378,23 @@ export class Column {
         } else if (this.tooltipElement) {
             ret.tooltipValueGetter = ({value}) => value;
             ret.tooltipComponentFramework = class extends Component {
-                getReactContainerClasses() {return ['xh-grid-tooltip']}
+                getReactContainerClasses() {
+                    const agParams = this.props,
+                        {location} = agParams;
+                    return location === 'header' ?
+                        ['ag-tooltip'] :
+                        ['xh-grid-tooltip'];
+                }
                 render() {
                     const agParams = this.props,
-                        {rowIndex, api, colDef} = agParams,
-                        {headerTooltip} = colDef,
-                        overHeader = isNil(rowIndex),
-                        value = overHeader ? headerTooltip : agParams.value,
-                        record = overHeader ? null : api.getDisplayedRowAtIndex(rowIndex).data;
+                        {rowIndex, api, colDef, location} = agParams;
 
+                    if (location === 'header') {
+                        return div(colDef.headerTooltip);
+                    }
+
+                    const value = agParams.value,
+                        record = api.getDisplayedRowAtIndex(rowIndex).data;
                     // ag-Grid encodes the value, so we decode it before passing to the renderer
                     return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
                 }
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -380,8 +380,11 @@ export class Column {
                 getReactContainerClasses() {return ['xh-grid-tooltip']}
                 render() {
                     const agParams = this.props,
-                        {value, rowIndex, api} = agParams,
-                        record = api.getDisplayedRowAtIndex(rowIndex).data;
+                        {rowIndex, api, colDef} = agParams,
+                        {headerTooltip} = colDef,
+                        value = isNil(rowIndex) ? headerTooltip : agParams.value;
+
+                    const record = api.getDisplayedRowAtIndex(rowIndex)?.data;
 
                     // ag-Grid encodes the value, so we decode it before passing to the renderer
                     return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -384,9 +384,7 @@ export class Column {
                         {headerTooltip} = colDef,
                         overHeader = isNil(rowIndex),
                         value = overHeader ? headerTooltip : agParams.value,
-                        record = overHeader ? null : api.getDisplayedRowAtIndex(rowIndex).data
-
-                    const record = api.getDisplayedRowAtIndex(rowIndex)?.data;
+                        record = overHeader ? null : api.getDisplayedRowAtIndex(rowIndex).data;
 
                     // ag-Grid encodes the value, so we decode it before passing to the renderer
                     return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -382,7 +382,9 @@ export class Column {
                     const agParams = this.props,
                         {rowIndex, api, colDef} = agParams,
                         {headerTooltip} = colDef,
-                        value = isNil(rowIndex) ? headerTooltip : agParams.value;
+                        overHeader = isNil(rowIndex),
+                        value = overHeader ? headerTooltip : agParams.value,
+                        record = overHeader ? null : api.getDisplayedRowAtIndex(rowIndex).data
 
                     const record = api.getDisplayedRowAtIndex(rowIndex)?.data;
 
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -393,10 +393,9 @@ export class Column {
                         return div(colDef.headerTooltip);
                     }
 
-                    const value = agParams.value,
-                        record = api.getDisplayedRowAtIndex(rowIndex).data;
-                    // ag-Grid encodes the value, so we decode it before passing to the renderer
-                    return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
+                    const record = api.getDisplayedRowAtIndex(rowIndex).data,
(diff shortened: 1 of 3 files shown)


cmp/grid/impl/ColumnHeader.js
```diff
@@ -71,6 +71,18 @@ export const columnHeader = hoistCmp.factory({
             headerName = xhColumn.headerName({column: xhColumn, gridModel});
         }
 
+        const showTooltipWhenElided = isNil(impl.xhColumn.headerTooltip),
```
amcclain, line 74: I believe that `impl.xhColumn` can be null - note that line 69 above checks for this - due to direct usage of agGrid features around auto grouping columns or something similar.  Would need to make this nullsafe as well.

cnrudd: Thank you!  Fixed + cleaned up line 69.

cmp/grid/impl/ColumnHeader.js
```diff
@@ -66,11 +66,23 @@ export const columnHeader = hoistCmp.factory({
         ];
 
         let headerName = props.displayName;
-        if (impl.xhColumn && isFunction(impl.xhColumn.headerName)) {
+        if (isFunction(impl.xhColumn?.headerName)) {
             const {xhColumn, gridModel} = impl;
             headerName = xhColumn.headerName({column: xhColumn, gridModel});
         }
 
+        const showTooltipWhenElided = isNil(impl.xhColumn?.headerTooltip),
```
amcclain, line 74: Good catch thanks was just about to suggest

cmp/grid/columns/Column.js
```diff
@@ -380,8 +380,11 @@ export class Column {
                 getReactContainerClasses() {return ['xh-grid-tooltip']}
                 render() {
                     const agParams = this.props,
-                        {value, rowIndex, api} = agParams,
-                        record = api.getDisplayedRowAtIndex(rowIndex).data;
+                        {rowIndex, api, colDef} = agParams,
+                        {headerTooltip} = colDef,
+                        value = isNil(rowIndex) ? headerTooltip : agParams.value;
```
amcclain, line 385: This might be more self-documenting here - assuming we are sure the `isNil(rowIndex) == true` only when the mouse is over the header. Just checking that can't happen under any other conditions.
```suggestion
                        overHeader = isNil(rowIndex),
                        value = overHeader ? headerTooltip : agParams.value,
                        record = overHeader ? null : api.getDisplayedRowAtIndex(rowIndex).data
```

cmp/grid/columns/Column.js
```diff
@@ -380,8 +380,11 @@ export class Column {
...
+
+                    const record = api.getDisplayedRowAtIndex(rowIndex)?.data;
 
                     // ag-Grid encodes the value, so we decode it before passing to the renderer
                     return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
```
amcclain, line 390: I think this could be problematic in that these renderers would be written by the developer to expect the cell contents - e.g. a number to format - and now they would be unexpectedly getting the `headerTooltip` string.

I think we would need to short-circuit the call to `me.tooltipElement` in this case and return the `headerTooltip` directly - probably needs to be wrapped in some minimal element...  

Am I correct, would this be a potential real-world problem?

cmp/grid/impl/ColumnHeader.js
amcclain: Or rather, I did suggest - didn't realize I had submitted that comment immediately :)

cmp/grid/columns/Column.js
cnrudd: Right, I thought of this also, but the AGGrid docs say that a tooltipComponent needs to ready to handle both potentials.

If the tooltipComponent returns just the headerTooltip string, it gets rendered rather plainly, not in the nice AG-Grid tooltip box.  So our TooltipComponent would need to style the header tooltip itself.  Which defeats the purpose of a tooltipElement (which I guess  is to give the dev total control over look of the tooltip).  Maybe it needs to be made more clear in the docs that a tooltipElement also needs to be ready to style a tooltipHeader if a tooltipHeader is on the column - that is essentially what AG-Grid is saying.

The tooltipElement api could be made more specific to accept 
`me.tooltipElement({cellValue: decodeURIComponent(value), headerValue: headerTooltip}, {record,....`
That might make it more clear as well.

cmp/grid/columns/Column.js
cnrudd: Correct.  AGGrid says rowIndex == true only when mouse is over header:
"The tooltipComponent detects that it's being rendered by a header because the params object does not contain a rowIndex value." --https://www.ag-grid.com/javascript-grid-tooltip-component/#header-tooltip-with-custom-tooltip

cmp/grid/columns/Column.js
amcclain: My thinking there is that `Column.tooltipElement` is intended to show rich tooltips for the grid data itself, and the fact that it gets called when hovering over the header is honestly somewhat weird - we have the simpler `headerTooltip` string config for that. 

So if we handled the hover-over-header case to return a reasonable-looking (but determined by Hoist) wrapper around any configured `headerTooltip` text I don't think it would violate developer expectations. We could also clarify in our docs that the non-header tip configs are only applied to data.

I don't think that I would want to change the signature of `tooltipElementFn` because it lines up with all the other tip and renderer fns we support where a value is passed as the first arg (and would be a breaking change).

I dunno - would like someone else to weigh in on it, also we have #2006 outstanding and it seems like that's in the same part of the code.

(@haynesjm42 or @lbwexler if you have any time to look at this change I would appreciate your input.)

lbwexler:
definitely agree we should combine this with #2006, and I think I pointed
out what needs to be done there.  Will hand it to Colin now for inclusion
in this PR.

Like your idea about simplifying, but worried a little bit about the case
where somebody wants to document column contents in a rich tooltip.  I feel
like we have seen that before.
Will continue to look at this ticket.

cmp/grid/columns/Column.js
cnrudd: Ok I can see that argument, that `Column.tooltipElement` is a Hoist optimization for rich data tooltips.   So, if a developer really needs a rich header tooltip, he can not use `Column.tooltipElement` and instead use the escape hatch 'agOptions.tooltipComponentFramework'.   

I'll go ahead and create a nicely styled header tooltip in the Column.js' `tooltipComponentFramework`.

cnrudd:
@lbwexler I have resolved #2006 in this branch as per your suggestion.

cmp/grid/columns/Column.js
cnrudd: Hi,  I've reworked `column.tooltipComponentFramework` so that if it is called for a headerTooltip it returns a div that will be styled by agGrid tooltip classes.   

I also found a prop `location` that seems to be a better, clearer choice for detecting if overHeader or not.

cmp/grid/columns/Column.js
```diff
@@ -377,14 +378,24 @@ export class Column {
...
+                        ['ag-tooltip'] :
+                        ['xh-grid-tooltip'];
+                }
                 render() {
                     const agParams = this.props,
-                        {value, rowIndex, api} = agParams,
-                        record = api.getDisplayedRowAtIndex(rowIndex).data;
+                        {rowIndex, api, colDef, location} = agParams;
+
+                    if (location === 'header') {
```
amcclain, line 392: This `location` param looks like a very helpful find!

cmp/grid/columns/Column.js
```diff
@@ -377,14 +378,24 @@ export class Column {
...
+                        return div(colDef.headerTooltip);
+                    }
 
-                    // ag-Grid encodes the value, so we decode it before passing to the renderer
-                    return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
+                    const record = api.getDisplayedRowAtIndex(rowIndex).data,
+                        value = record.get(colDef.colId);
```
amcclain, line 397: Believe we should use `colDef.field` here instead of `colId` - they are going to be the same in almost all cases, but you can have cases with multiple cols pulling the same record field (e.g. to render the same primary value differently). In that case the dev has been required to set unique colIds which won't map onto record fields.

It's also true that `Column.field` is optional and might not be specified - say for a col where there is no one field you want to map to, but instead you are rendering based on the entire record. 

I thought that was gonna be fine - primary value passed to tooltip renderer can be null - but I see that `record.get()` will throw on an unknown field - part of the value-add with the `get()` method is that it checks that access. For this use-case I would pull from `record.data[colDef.field]` - hopefully that's the last little wrinkly... 🤞

cnrudd: Ok, thanks for the tips re record.data. I have applied the suggested change.

lbwexler:
looking now

cmp/grid/columns/Column.js
```diff
@@ -377,14 +378,24 @@ export class Column {
...
-                        record = api.getDisplayedRowAtIndex(rowIndex).data;
+                        {rowIndex, api, colDef, location} = agParams;
+
+                    if (location === 'header') {
+                        return div(colDef.headerTooltip);
+                    }
 
-                    // ag-Grid encodes the value, so we decode it before passing to the renderer
-                    return me.tooltipElement(decodeURIComponent(value), {record, column: me, gridModel, agParams});
+                    const record = api.getDisplayedRowAtIndex(rowIndex).data,
```
lbwexler, line 396: do we always get a valid rowIndex?  Do we always get a Record?  What about when the user mouses over a group row or other interstitial space?  I would feel more comfortable if this code was on the lookout for that, and rendered nothing rather than a DNP.

lbwexler: I also think we should be using the value getter on the column.  Its called out in constructor, and while it defaults to what you have here, it can be overridden by users:

 * @param {Column~getValueFn} [c.getValueFn] - function for getting the column value

amcclain:
Excellent! 🎉

Issue #2012, opened 2020-07-22
Implement find functionality for CodeInput


amcclain:
The codemirror library uses a virtual scroller / windowed rendering model to avoid rendering markup for all lines of a long code file to the DOM. This is efficient and works well, but it means that the browser-supplied "find in page" feature won't work for any lines that are below the visible area, since that text hasn't been rendered into markup.

There is a `viewportMargin` setting that can be customized to allow for a larger buffer of rendered rows, or set to Infinity to disable the smart rendering.  For smaller code files the performacne impact is not noticeable, but when you get to thousands of lines there is of course a slowdown.

It would be great for us to implement an optional UI for searching within an input - this could be a prop on our component that could e.g. display a toolbar and, when the user enters text, display the match count and let you navigate to the next/prev match, much like a simpler version of the editor find in InitelliJ.

Codemirror provides a `searchcursor` addon that could power this feature. Looks like they also ship a more bundled up `search` addon that would provide some kind of a dialog-drive UI for an out-of-the-box impl. but my guess is that we would want to go our own way there so the entire thing looks nice and consistent with Hoist standards.

https://codemirror.net/doc/manual.html#addons

amcclain:
This came up as a request from another client and made me think about it a bit more.

I would suggest we also introduce a new `showToolbar` prop that, if true, would render the buttons the comp already supports + the new find field in a toolbar.  If you have enabled the find field, showToolbar would be defaulted/required to be true - I do _not_ think we should attempt to do any special rendering or UI pattern with a floating find input.

Having support for showing a toolbar could clear the way for some additional functionality to be added down the line - another client request related to code folding - without us needing to stretch too far with the floating buttons.

That said, I think by default and for most use-cases where this is a simple / small input, we _do_ want to continue using the floating buttons and not show a toolbar.  The toolbar would be only for when this is used in a context where richer functionality such as find is required - I think it is reasonable to assume that if you need a find box you are giving the comp a good amount of space to show large code blocks and would be OK with allocating internal space to a toolbar.

PR #2034, opened 2020-08-06
CodeInput search feature

petradish:
This draft PR addresses https://github.com/xh/hoist-react/issues/2012

Contrary to Anselm's request, I did add another floaty button that expands on click and collapses when the search query is cleared. The width of the search bar is also fixed right now -- part of the desire to minimize the toolbar to a button is to avoid having to size the toolbar based on the size of the text area box. Open to changing the implementation!

Hoist P/R Checklist
-------------------

- [x ] Caught up with `develop` branch as of last change.
- [ Not yet ] Added CHANGELOG entry, or determined not required.
- [ Not yet ] Created Toolbox branch / PR, or determined not required.


desktop/cmp/input/CodeInput.js
```diff
@@ -179,28 +209,125 @@ export class CodeInput extends HoistInput {
...
-
             className: this.getClassName(),
             onBlur: this.onBlur,
             onFocus: this.onFocus,
-
             ...props
         });
     }
 
+    renderToolbar() {
```
amcclain, line 221: This is a case where it would be a very good optimization to actually factor this out into its own sub-component, since it looks like this will re-render often.

amcclain:
Had a chance to actually test drive this in context - eager to move forward on it! Some additional thoughts:

+ I think the find functionality might be pushing the floating, very minimal controls too far. 
  + Keeping it floating puts pressure on how big the find field can be - it really needs to stay very small. I think ideally we would have it be a bit larger so if you are searching for something like a longer name or GUID you can see a bit more of your input. 
  + The variable number of results also causes it to grow/shrink a bit in a somewhat distracting way. 
  + Right now you need to click twice - once on the icon to show the bar and then again to focus the input and start typing. We could hopefully fix this to avoid, but if the input was always visible when search was enabled it wouldn't be an issue.
  + Dismissing the find box and going back to the mode where there is only the magnifying glass icon is a bit inconsistent - I think you can only do it if you have entered some text and then clear the text using the input's clear button. If you click the icon and then realize you don't need to search for anything, it's unclear how to get it back how it was. 
  + It's relatively easy to click somewhere on the control (like the hit counts) that causes it to dismiss unexpectedly.
+ I also think there could be some value in committing to the idea of a proper panel toolbar that could be shown on this control regardless of whether or not search was enabled. We should absolutely keep the default behavior of the floating minimal buttons when find not enabled, but in use cases like our current client project where you know you are dealing with big codeInputs and can devote the space to a toolbar, it would be natural to have a toolbar mode, make the existing buttons more obvious / easier to find, and provide a dependable/larger space for the find input. 
  + I would like the option to enable toolbar w/o enabling find - again maybe you have space and it makes the existing button functionality more discoverable.
  + But I don't think it's a bad limitation to say that enabling search requires the toolbar to be shown (and will auto-enable it for you).

Unrelated to toolbars...
+ The first click of the down arrow does not appear to do anything - if I type a new term and press enter the hits light up and it will already scroll to ensure the first one is visible and gets the extra highlight, so it behaves as though the first hit has already been selected and the arrow would take you to the second hit
+ The arrows should disable if no hits
+ Sorry if you mentioned this already, but did you look at searching-as-you-type by reacting to changes (with a decent debounce) to the bindable input text?

desktop/cmp/input/CodeInput.js
```diff
@@ -96,12 +110,20 @@ export class CodeInput extends HoistInput {
...
 
         /** True (default) to display Fullscreen button at bottom-right of input. */
-        showFullscreenButton: PT.bool
+        showFullscreenButton: PT.bool,
+
+        /** True to display Search button at bottom-right of input. When clicked, it will expand
+         * to a toolbar component at the bottom-right of input. Will collapse when search input is
+         * cleared. Search is case-insensitive.
+         * */
+        enableSearch: PT.bool
```
amcclain, line 119: Two admittedly pretty OCD nits here:

+ I've been alphabetizing prop-types to help with scanning for / locating a known prop, and simply to provide a consistent standard across components. 
+ For the multi-line comments, pls add a newline after the opening `/**` so it's on its own line.

desktop/cmp/input/CodeInput.js
```diff
@@ -179,28 +209,125 @@ export class CodeInput extends HoistInput {
...
+                                formField({
+                                    field: 'query',
+                                    label: null,
+                                    item: textInput({
+                                        leftIcon: Icon.search(),
+                                        width: 150,
+                                        enableClear: true,
+                                        rightElement: hbox(
+                                            label({
+                                                className: 'xh-code-input--label',
```
amcclain, line 245: We are definitely not perfect at following the BEM CSS classname convention throughout the code - not by any means! - but where we can it's nice.  Here I would use `xh-code-input__label` since it's a nested block element, not a modifier (like e.g. `--selected` would be)

desktop/cmp/input/CodeInput.js
```diff
@@ -300,6 +430,86 @@ export class CodeInput extends HoistInput {
...
+        this.cursor = this.editor.getSearchCursor(this.query, 0, true);
+
+        const {cursor, editor, selectedMatches} = this;
+        while (cursor.findNext()) {
+            const anchor = cursor.from(),
+                head = cursor.to();
+            selectedMatches.push({
+                anchor,
+                head,
+                textMarker: editor.markText(anchor, head, {className: 'selection-highlight'})
```
amcclain, line 444: Here I would use the class `xh-code-input--highlight` or similar - key part being to put that prefix in.  We're not worried about breaking other custom app styles - since we nest our rules under `.xh-code-input` - but you could imagine an app defining a `selection-highlight` class for some other, completely unrelated reason and breaking _our_ styling here.

styles/vars.scss
```diff
@@ -200,11 +200,16 @@ body {
   --xh-bg: var(--bg, white);
   --xh-bg-alt: var(--bg-alt, var(--xh-blue-gray-light));
   --xh-bg-highlight: var(--bg-highlight, #{mc-trans('blue', '200')});
+  --xh-selected-highlight: var(--selected-highlight, #{mc-trans('indigo', '300', 0.3)});
+  --xh-match-highlight: var(--match-highlight, #{mc-trans('amber', '700')});
```
amcclain, line 204: We will want to use more specific code-input scoped var names here, or maybe just not bother defining vars for these as long as we have something that looks reasonable. (Apps can always target the CSS classes and style without using vars.)

The issue with these is that they sound very general - if I were looking through our vars I might well assume the selected one applies to grid rows, or all text - or I wouldn't be quite sure what it did. 😄

PR #2037, opened 2020-08-13
Remove FieldFilter.fieldType

TomTirapani:


lbwexler:
I think this is great, and we can merge it.  

I don't want to do any premature optimization, BUT it does seem like if we felt the test function was redoing too much work on each record (and there is some other extra work, beyond the parsing), we could add a method that would have a  Filter create a test function, given a store. FilterModel (which is mutable and has state) could create a cached aggregate function *by store* that it would use to do its filtering.

PR #2042, opened 2020-08-19
Support for Column.displayName + defaulting

amcclain:
+ New config `Column.displayName` as general-purpose entry-point for the user-facing name of a Column.
+ Sourced via GridModel pre-processing from the `displayName` property of a linked store `Field` or config, if available, allowing further DRYing up of field metadata.
+ Also default columns bound to numeric field types to right alignment if not otherwise specified.
+ Add utility `genDisplayName()` function and use consistently in place of multiple independent calls to `startCase`.
+ Doc tweaks and auto-formatting.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/impl/FilterChooserFieldSpec.js
+++ b/cmp/filter/impl/FilterChooserFieldSpec.js
@@ -74,13 +74,28 @@ export class FilterChooserFieldSpec {
     get isRangeType() { return this.filterType === 'range'}
     get isValueType() { return this.filterType === 'value'}
 
+    get isDateBasedFieldType() {
+        const {fieldType} = this;
+        return fieldType == FieldType.DATE || fieldType == FieldType.LOCAL_DATE;
+    }
+
+    get isNumericFieldType() {
+        const {fieldType} = this;
+        return fieldType == FieldType.INT || fieldType == FieldType.NUMBER;
+    }
+
+    get isBoolFieldType() {return this.fieldType == FieldType.BOOL}
+
     /**
      * @return {string} - a rendered example / representative data value to aid usability.
-     *      TODO - default return of upper fieldType is a bit rough (e.g. "INT")
      */
     get example() {
-        const {exampleValue, fieldType} = this;
-        return exampleValue ? this.renderValue(exampleValue) : fieldType.toUpperCase();
+        const {exampleValue} = this;
+        if (exampleValue) return this.renderValue(exampleValue);
+        if (this.isBoolFieldType) return 'true';
+        if (this.isDateBasedFieldType) return 'YYYYMMDD';
+        if (this.isNumericFieldType) return this.renderValue(1234);
+        return 'value';
     }
 
     /**
@@ -131,7 +146,7 @@ export class FilterChooserFieldSpec {
         let ret;
         if (isFunction(this.valueRenderer)) {
             ret = this.valueRenderer(value, op);
-        } else if (this.fieldType === FieldType.DATE || this.fieldType === FieldType.LOCAL_DATE) {
+        } else if (this.isDateBasedFieldType) {
             ret = fmtDate(value);
         } else {
             ret = value?.toString();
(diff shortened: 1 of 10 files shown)


cmp/filter/impl/FilterChooserFieldSpec.js
```diff
@@ -74,13 +74,28 @@ export class FilterChooserFieldSpec {
...
      * @return {string} - a rendered example / representative data value to aid usability.
-     *      TODO - default return of upper fieldType is a bit rough (e.g. "INT")
      */
     get example() {
-        const {exampleValue, fieldType} = this;
-        return exampleValue ? this.renderValue(exampleValue) : fieldType.toUpperCase();
+        const {exampleValue} = this;
+        if (exampleValue) return this.renderValue(exampleValue);
+        if (this.isBoolFieldType) return 'true';
+        if (this.isDateBasedFieldType) return 'YYYYMMDD';
```
lbwexler, line 96: 👍

cmp/grid/GridModel.js
```diff
@@ -938,38 +915,103 @@ export class GridModel {
         }
     }
 
-    parseStore(store) {
+    // Store fields and column configs have a tricky bi-directional relationship in GridModel,
+    // both for historical reasons and developer convenience.
```
lbwexler, line 919: great comment

cmp/grid/columns/Column.js
```diff
@@ -202,16 +213,21 @@ export class Column {
         this.colId = withDefault(colId, field);
         throwIf(!this.colId, 'Must specify colId or field for a Column.');
 
-        this.hidden = withDefault(hidden, false);
-        warnIf(rest.hide, `Column ${this.colId} configured with {hide: true} - use "hidden" instead.`);
+        this.isTreeColumn = withDefault(isTreeColumn, false);
+
+        // Note that parent GridModel might have already defaulted displayName from an associated
+        // `Store.field` when pre-processing Column configs - prior to calling this ctor.
+        this.displayName = withDefault(displayName, genDisplayName(this.colId));
```
lbwexler, line 220: now might be a good time to actually rewrite all of these withDefault() statements to the more concise `??`.
The only time withDefault() is warranted is if we want to allow the application to force a setting of null or undefined.
(don't think that is the case here for any of these)

data/Field.js
```diff
@@ -94,11 +95,19 @@ export const FieldType = Object.freeze({
     STRING: 'string'
 });
 
+/**
+ * @param {string} fieldName - short name / code for a field.
+ * @return {string} - fieldName transformed into user-facing / longer name for display.
+ */
+export function genDisplayName(fieldName) {
+    return fieldName == 'id' ? 'ID' : startCase(fieldName);
```
lbwexler, line 103: nit: favoring triple equals now

lbwexler:
only nits here -- looks good!

cmp/grid/columns/Column.js
amcclain: It's a good question and I thought about it some more.

* I changed the `displayName` statement to ensure the default off of colId is always assigned. I agree I don't think we ever intend or want an app to force a null there. However, here and in some other cases I think we want `||` not `??` as we don't want to allow `''` or another false value here.
* The headerName is different - there you might well not want a header (for a minimal icon column, e.g.).  So we need something more than a falsely check and withDefault works nicely.
* The bool flags that default to false could go to `!!flagConfig` but I think withDefault ends up be more immediately readable and then is consistent with the flags we wish to default to true. Dev could still pass and end up with null, but that should be fine as a false value anyway.
* Reviewed one or two others - see a9780b43400bcc64c5b31aa68a7ec18adc64f61f

data/Field.js
amcclain: I can jump on the `===` bandwagon, no problem.

amcclain: Fixed. We could get serious about this and add it back into the linter if you're ready - probably best to do a bombing there (with some care, `==` has its place) at some point.

PR #2044, opened 2020-08-23
LocalDate - additional getters/caching/docs

amcclain:
+ Add methods/getters for start/end of month.
+ Add internal map to cache getter/method returns that are immutable for a given instance.
+ JsDoc return types.

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/datetime/LocalDate.js
+++ b/utils/datetime/LocalDate.js
@@ -32,6 +32,7 @@ export class LocalDate {
     _isoString;
     _moment;
     _date;
+    _getterCache = new Map();
 
     //------------------------
     // Factories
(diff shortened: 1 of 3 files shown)


utils/datetime/LocalDate.js
```diff
@@ -100,35 +101,50 @@ export class LocalDate {
...
         return this._date.getTime();
     }
 
+    /** @return {string} */
     format(...args) {
         return this._moment.format(...args);
     }
 
+    /** @return {string} */
     dayOfWeek() {
```
lbwexler, line 130: This little pattern seems like a poster child for an annotation honestly -- would be called something like "@lazy" or "@computeOnce".  It could even store the property on the object itself as ._lazyPropName if that was considered more efficient or straightforward (not sure that it is)

lbwexler: And reading more carefully your comment, see that you mention that.

lbwexler: am curious to know what prompted the caching bit -- are we doing some of these operations in a renderer, or otherwise where it feels like we are doing it in bulk wastefully?

Certainly we could easily add a decorator for this -- the question is how to generalize, it and choose an implementation.  Could see it becoming more important in other classes, where perhaps it also takes into account the parameters to the function (see Lodash memoize) just want to get this right, but also not hold up the unrelated changes.  

Also we might as well just add endOfYear/startOfYear while we are at it.   Happy to make these changes.

lbwexler:
Let me know your thoughts -- I made the changes discussed above.

+ Could have called the decorator 'memoize' but though maybe to save that name for a more general decorator that would actually cache based on arguments.  Seemed worth having the special case implemented for now.

+ Also note that using the decorator requires not freezing the object -- but I think that's appropriate.  Honestly, you would have to be especially malicious to damage this object, which already hides all state with `_` variables.

amcclain:
Hi - thank you for taking the time to code up the decorator - that was exactly the kind of thing I was thinking about.

I wasn't motivated by any observed performance issues or demands - was just adding more of these getters and transforms to LocalDate and it seemed a shamed to be doing what was clearly repeated work (even if hopefully pretty trivial) over and over again on the same instance throughout the life of the app. 

I wouldn't want to start using this everywhere, but for LocalDate it seemed appropriate since you can certainly have thousands of references to an instance and call these getters many thousands of times if loading up larger collections of objects with a few dates each.

I'm reviewing this now on the Github iPad app (for novelty) and am a bit too conservative to merge from here, but will merge tomorrow from my old fashioned computer, or please do so anytime. Thanks too for adding the end/start of year conveniences.

PR #2047, opened 2020-08-27
Refactor filterChooser

lbwexler:
+ Add validation of favorites, value
+ Clarify structure of options
+ Breakout Query processing in to its own file.
+ Tweak to option sorting logic to support future changes

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/FilterChooserFieldSpec.js
+++ b/cmp/filter/FilterChooserFieldSpec.js
@@ -88,9 +88,10 @@ export class FilterChooserFieldSpec {
     }) {
         this.field = field;
         this.store = store;
-        const storeField = this.storeField = store?.getField(field);
 
-        this.fieldType = this.fieldType ?? storeField?.type ?? FieldType.AUTO;
+        const storeField = store?.getField(field);
+        this.storeField = storeField;
+        this.fieldType = fieldType ?? storeField?.type ?? FieldType.AUTO;
 
         this.displayName = displayName ?? storeField?.displayName ?? genDisplayName(field);
         this.ops = this.parseOperators(ops);
--- a/cmp/filter/FilterChooserFieldSpec.js
+++ b/cmp/filter/FilterChooserFieldSpec.js
@@ -13,9 +13,8 @@ import {stripTags, throwIf} from '@xh/hoist/utils/js';
 import {isFunction, isNil} from 'lodash';
 
 /**
- * Defines the filters options available for a given field and provides useful metadata
- * for including these options in UI affordances, including available data values for suggestion
- * if applicable / so configured.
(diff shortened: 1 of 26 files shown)


cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -51,7 +48,64 @@ export class FilterChooserFieldSpec {
...
+        valueRenderer,
+        valueParser,
+        exampleValue,
+        store
+    }) {
+        this.field = field;
+        this.store = store;
+        const storeField = this.storeField = store?.getField(field);
+
+        this.fieldType = this.fieldType ?? storeField?.type ?? FieldType.AUTO;
```
TomTirapani, line 93: Shouldn't this be `this.fieldType = fieldType`?

cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -51,7 +50,66 @@ export class FilterChooserFieldSpec {
...
+     *      If unspecified, values will be extracted and updated from Store data if applicable.
+     * @param {FilterOptionValueRendererCb} [valueRenderer] - function to produce a suitably
+     *      formatted string for display to the user for any given field value.
+     * @param {FilterOptionValueParserCb} [valueParser] - function to parse user's input from a
+     *      filter chooser control into a typed data value for use in filtering comparisons.
+     * @param {*} [exampleValue] - sample / representative value displayed by `FilterChooser`
+     *      components to aid usability.
+     * @param {Store} [store] - set from controlling `FilterChooserModel.sourceStore` config, used
+     *      to source matching data `Field` and extract values if configured.
+     */
```
amcclain, line 80: I had specifically documented these via a typedef so we could then use it in the documentation for `FilterChooserModel.fieldSpecs` - the actual entry point for the developer. I had done this for `FieldConfig` as well (in Field.js).

IMO pushing it back to the ctor here just encourages a dev to construct this object, which we don't want them to do, and doesn't provide them any inline help when specifying fieldSpecs, which is exactly what we _do_ need them to do.

I left this in its new location on the ctor and updated the tags to `@param` consistently but would like to discuss as this is a pattern we have in a few places where we have really nice docs, but they aren't easily discoverable and don't provide the dev-time assistance they could because they are on a ctor that devs never call (`Column` is another prime example).

lbwexler:
Happy to discuss this of course  I have a huge number of additional changes
in my sandbox from last night.  Will look to get in for review ASAP
tommorrow.

I think we do need

lbwexler:
Have reviewed this extensively with Tom and John.  It's by no means perfect yet, but this moves the ball substantially forward from what we have, and the new structure has much more ability to take further enhancements.  

Am going to merge it it now, so that John can begin testing and we can show Dima.

Issue #2050, opened 2020-08-28
Admin monitor should make it clear if no monitors defined / enabled


amcclain:
If no monitors are defined or none enabled, we should mask the monitor reporting tab with a message that makes that clear. Currently the panel is weirdly empty and the "run now" button remains enabled despite the fact that we know it will do nothing.

PR #2055, opened 2020-08-31
Monitor status tab is masked when monitors are undefined or disabled

petradish:
This PR and [Hoist-Core PR](https://github.com/xh/hoist-core/pull/163) addresses Issue #2050 to provide a mask+message for the monitor status tab when monitors are either undefined or disabled.

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/monitor/MonitorResultsModel.js
+++ b/admin/tabs/monitor/MonitorResultsModel.js
@@ -5,7 +5,6 @@
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
 import {HoistModel, LoadSupport, managed, XH} from '@xh/hoist/core';
-import {mask} from '@xh/hoist/desktop/cmp/mask';
 import {action, computed, observable} from '@xh/hoist/mobx';
 import {Timer} from '@xh/hoist/utils/async';
 import {SECONDS} from '@xh/hoist/utils/datetime';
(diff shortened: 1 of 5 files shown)


admin/tabs/monitor/MonitorResultsModel.js
```diff
@@ -35,6 +36,21 @@ export class MonitorResultsModel {
         return this.results.filter(monitor => monitor.status === 'FAIL').length;
     }
 
+    get monitorMask() {
```
lbwexler, line 39: A couple of minor comments here:  

1) I think its totally appropriate for this to just be a function in the Component file-- this is really view logic. (I would make it a function,that takes a model).

2) I think you could add a getter on the model of the form 'active', that could be symmetrical  with the other getters.
I also think we should treat 'undefined' as 'active' for now, so we can release this without being dependent on hoist core.

3) Do you need to actually do the bit with the loadModel?  Can't you just always return 'onLoad' on line 50 and have the Panel handle it for you?

petradish: Thanks, Lee! Good points -- just tested again, and even with the new active flag, the monitors need to be run in order for the monitorResult to register the change of its monitor... not sure what the best way to handle this would be. Perhaps a label in the toolbar to display the active monitor count?

Issue #2058, opened 2020-09-03
Returning null from a column's tooltip func will break the tooltip on subsequent loads


febbraiod:
Example branch in toolbox:

https://github.com/xh/toolbox/tree/nullTooltipBreaksNextLoad

To repro:

load sample grid, all citys will be set to New York, which will produce a null tooltip.
Refresh the tab to reload the grid, the city data is set back to normal, but the tooltip will no longer work.

If I'm not mistaken the tooltip function runs after the second load and does return the proper string, but we do not see the tooltip render it.

lbwexler:
So this is actually the same as the following ag-Grid bug, which has a pending fix and PR.

https://github.com/ag-grid/ag-grid/issues/4039

If we don't wish to wait for this, we could always use the tooltipComponentFramework API internally.  Just need to come up with some sensible defaults, as simply a tooltipElement function that returns a string does not render well and needs styling

Arguably we need to improve those defaults anyway

lbwexler:
So our friend Gil at ag-Grid reporting this is fixed and merged -- not sure when it will get released, but something for us to keep an eye on.

PR #2059, opened 2020-09-07
Grid tooltip issues

TomTirapani:
Column.tooltipValueGetter must *always* returns a value - if it returns a nil value, the tooltip won't be shown, even if there is a value later. See issue https://github.com/xh/hoist-react/issues/2058 for more details. To work around this, a placeholder is returned instead of a nil value.

Column always uses tooltipComponentFramework. This allows us to handle the placeholder values set by tooltipValueGetter.

+ Provide minimal default styling for tooltip component.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -193,6 +193,17 @@
 
 .xh-grid-tooltip {
   position: absolute;
+
+  &:empty {
+    display: none;
+  }
+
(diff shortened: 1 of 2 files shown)


lbwexler:
I think this is a great fix.  I think the consistency that all hoist tooltips are rendered by the same component will also be a huge benefit for styling.

Issue #2062, opened 2020-09-09
Remove use of inline pre-flight browser check script


amcclain:
As per CSP (content security policy) we want to avoid inline scripts such as the one we embed in:

https://github.com/xh/hoist-react/blob/98a703a3b3bd76075c79551c48f98935b4682436/static/index.html

Having this in place requires us to use the `'unsafe-inline'` option within CSP header `script-src` directives. 

We should package this as a standalone file within static and ensure that dev-utils handles appropriately.

TomTirapani:
I've moved the inline script into a standalone file, which is imported into each app bundle by hoist-dev-utils:
https://github.com/xh/hoist-react/pull/2063
https://github.com/xh/hoist-dev-utils/pull/27

PR #2063, opened 2020-09-10
Move inline scripts into standalone file.

TomTirapani:
See issue #2062.
See corresponding hoist-dev-utils PR here: https://github.com/xh/hoist-dev-utils/pull/27

Hoist P/R Checklist
-------------------

### DIFF
--- a/static/bootstrap.js
+++ b/static/bootstrap.js
@@ -1,0 +1,19 @@
+window._xhLoadTimestamp = Date.now();
+
+(function(W) {
+    var D = document,
+        // Mobx 5+ has a hard requirement for Proxy (test w/string access for max safety).
+        hasProxy = W['Proxy'];
+
+    if (!hasProxy) {
+        D.body.innerHTML = (
+            '<div style="margin:80px auto; font-family: sans-serif; width:450px; text-align:center; padding: 20px; border: 1px solid #ccc;">' +
+            'This application can not be loaded in your current browser<br/>' +
+            'due to missing support for modern web technologies.<br/><br/>' +
+            'Try again with an updated browser such as Chrome.' +
+            '</div>'
+        );
+        try {W.stop()} catch (e) {}
+        try {D.execCommand('Stop')} catch (e) {}
+    }
+})(window);
--- a/static/index.html
+++ b/static/index.html
@@ -7,27 +7,6 @@
 </head>
 <body class="xh-app">
 
-    <script type="text/javascript">
-      window._xhLoadTimestamp = Date.now();
-      (function(W) {
-          var D = document,
-              // Mobx 5+ has a hard requirement for Proxy (test w/string access for max safety).
-              hasProxy = W['Proxy'];
-
-          if (!hasProxy) {
-              D.body.innerHTML = (
-                  '<div style="margin:80px auto; font-family: sans-serif; width:450px; text-align:center; padding: 20px; border: 1px solid #ccc;">' +
-                  'This application can not be loaded in your current browser<br/>' +
-                  'due to missing support for modern web technologies.<br/><br/>' +
-                  'Try again with an updated browser such as Chrome.' +
-                  '</div>'
-              );
-              try {W.stop()} catch (e) {}
-              try {D.execCommand('Stop')} catch (e) {}
-          }
-      })(window);
-  </script>
-
   <noscript>
     You need to enable JavaScript to run this app.
   </noscript>


amcclain:
Hi Tom - thanks for looking at this!

Here is my concern (not yet verified) with this approach... The goal of this script is to cleanly identify older browsers that can't run the code at all and to display a basic inline error message. With this approach, my understanding is that the little outdated-browser detection script will end up bundled in the same file as the rest fo the codebase - or at least some common chunk of it. Is it possible then that an older browser would load and attempt to parse that first chunk and fail to even get to the execution step due to unknown keywords or syntax being present in the file?

I had been thinking that we would need a process, distinct from the main bundles, that would bring this file into the HTML source as its own standalone script tag, to be loaded/parsed/executed before any bundles.

But if this works I don't want to over-complicate things. Were you able to test, or does the concern above make sense in terms of requiring some testing to see how it behaves?

lbwexler:
Guys -- I suspect this might work -- but to answer the question we really just need to test this on IE11, which is the best example of a browser we are trying to defeat with it.  

I am hoping that the main problem is just that mobx 5 calls `new Proxy()` when it actually executes, but this script will execute first and prevents that.

lbwexler:
@haynesjm42 -- can you try this PR out on your windows machine with IE 11 and let us know if cleanly yields the error message we are trying to produce?

haynesjm42:
This is what I got in IE11 - looks good!

image

lbwexler:
Thanks John!

lbwexler:
Am going to merge this, based on atm's comments,  the successful test, and our desire not to get too caught up on support for unknown legacy browsers....

amcclain:
Thanks for testing/merging - just catching up here apologies am a bit behind. 

- @haynesjm42 just checking, did you use a locally linked version of dev-utils to test? 
- I will update Toolbox to use the dev-utils snapshot so we can ensure we're testing this change there - right now it's still on the prior released version so a bit of a mismatch.
- To that point, I think we need a changelog entry highlighting that the current hoist-react release should be paired with a dev-utils release which we need to do.
- But before that I wanted to revisit small question of possible confusion over `bootstrap.js` name - @lbwexler I think you had raised that.  Should we rename the file in `/static/` to something like `preflight-checks.js` (or any other suggestions - it also sets the starting timestamp)?
- We should also add a comment to `/static/README.md` as otherwise the new file is a bit of a mystery.  Again wanted to confirm the name first.

Sorry for being behind here - these dev-utils related/coordinated changes sometimes have a few wrinkles so I didn't want to miss anything.

amcclain:
I updated toolbox-dev to put it back on snapshot for hoist-react and and hoist-dev-utils - 

if you look at the source and the bundles I don't know / think that the new preflight code is the first JS run on the page. Let me know if I am missing something here?

haynesjm42:
When I tested I had copied the entirety of webpack.config.js from the branch on hoist-dev-util and ran locally, then opened in IE.

Just tried opening toolbox-dev in IE11 and am not getting the unsupported browser message - so it appears that I did not have something correct when testing or that testing approach was flawed :(

amcclain:
Potentially related to how the bundles/chunks are built and ordered in production bundling mode vs. dev-time mode - or something similar.  Thanks for testing again with the deployed version.

We can revisit some more early next week w.r.t. different options for getting the bootstrap file into the source but maybe not including in the core bundles, if that is indeed the issue...

amcclain:
I think I have a working approach. Just-released hoist-react 36.1.0 includes both inline and no-inline versions of the index.html template.  I also renamed the `static/bootstrap.js` to `preflight.js` as a slightly more descriptive / non-overlapping name.

https://github.com/xh/hoist-react/commit/bfa46360f57f9f4246c8bd46a2fa720d8050167e

This way hoist-react should have the resources to work with either the existing dev-utils versions or a new one.  For the new one, I used the `CopyWebpackPlugin` we already had to copy the preflight script into `/public/` within the build output. Then used new `HtmlWebpackTagsPlugin` to inject a script tag w/cache busting hash into the generated HTML output, before all other bundle script tags:

https://github.com/xh/hoist-dev-utils/commit/3b9b2b61879bd314cf9f140a3fc267808353d042

In my test Win10 VM this now shows the error message correctly in IE11 - John if you can test again that would be great. And I was able to tighten up the CSP header for Toolbox to restrict any inline script tags, which was the ultimate goal. 

Hopefully all set with this for the time being!

PR #2066, opened 2020-09-11
ExceptionHandler supports Error.requireReload

amcclain:
+ Allow errors to be thrown with requireReload:true to be picked up as such by standard exception handling.

Hoist P/R Checklist
-------------------


amcclain:
Does this look reasonable to you, if you want to require a reload when throwing an exception that you expect to be caught by `XH.handleException`, without having to check the exception and set the flag in the catch+call to handle?

lbwexler:
The addition of requireReload on the exception seems fine to me, but I still think in all cases the option bundle (first clause) needs to be determinate, if defined.  So I think the '??' is correct and the '||' is not in this case.  Let me know if I am missing something?

PR #2075, opened 2020-09-16
Add bulk delete action to rest grid, apply to Config and Preference panels.

febbraiod:
Hoist P/R Checklist
-------------------


desktop/cmp/rest/RestGridModel.js
```diff
@@ -55,6 +55,13 @@ export const deleteAction = {
     actionFn: ({record, gridModel}) => gridModel.restGridModel.confirmDeleteRecord(record)
 };
 
+export const bulkDeleteAction = {
+    text: 'Delete',
+    icon: Icon.delete(),
+    intent: 'danger',
+    recordsRequired: true,
+    actionFn: ({gridModel}) => gridModel.restGridModel.confirmBulkDeleteRecords()
+};
```
febbraiod, line 64: Note from CR in prev branch: 
Could use this everywhere we want delete...rather than keeping old single del around. Then bulk is driven by multi select enabled or not. Is this too dangerous?

lbwexler: I think we should just make the delete action *be* bulk delete.  Simpler. If there is one record in the selection it can use the regular rest deletion endpoint and  otherwise use the bulk deletion endpoint, with

desktop/cmp/rest/RestGridModel.js
```diff
@@ -102,7 +109,8 @@ export class RestGridModel {
      * @param {Object[]|RecordAction[]} [toolbarActions] - actions to display in the toolbar. Defaults to add, edit, delete.
      * @param {Object[]|RecordAction[]} [menuActions] - actions to display in the grid context menu. Defaults to add, edit, delete.
      * @param {Object[]|RecordAction[]} [formActions] - actions to display in the form toolbar. Defaults to delete.
-     * @param {Object} [actionWarning] - map of action (e.g. 'add'/'edit'/'delete') to string.  See default prop.
+     * @param {Object} [actionWarning] - map of action (e.g. 'add'/'edit'/'del') to string or a fn to create one. See default prop.
+     *      Function will be passed selected record(s) to be acted upon.
```
lbwexler, line 113: probably simplest to always pass an *array* of records for *all* actions, and specify that you are doing so here.  This will mean the developer doesn't have to guess based on the action whether it receives one or all records.  Edit, for axample, could be multi-record enabled at some point in the future.

desktop/cmp/rest/impl/RestFormModel.js
```diff
@@ -83,11 +84,13 @@ export class RestFormModel {
         }
 
         if (warning) {
-            const confirmed = await XH.confirm({
-                message: warning,
-                title: 'Warning',
-                icon: Icon.warning({size: 'lg'})
-            });
+            const message = isFunction(warning) ? warning(this.currentRecord) : warning,
```
lbwexler, line 87: see note above about just always passing in an array of records, even if it is just a single record (or zero records) for some actions

desktop/cmp/rest/RestGridModel.js
lbwexler: if you want a version of the delete action that can *only* be used for single deletion, an application could spread in recordsRequired: 1 to the standard action.  But I think the simplest default is true

desktop/cmp/rest/RestGridModel.js
```diff
@@ -174,60 +172,52 @@ export class RestGridModel {
...
     }
 
-    @action
-    editSelection() {
-        const record = this.selectedRecord;
-        if (record) this.editRecord(record);
+    async bulkDeleteRecordsAsync(records) {
+        throwIf(this.readonly, 'Records not deleted: this grid is read-only');
+        const resp = await this.store.bulkDeleteRecordsAsync(records).catchDefault(),
+            intent = resp.fail > 0 ? 'warning' : 'success',
```
lbwexler, line 185: don't think this does what you want if the server throws.  Gonna still toast some nonesense, even if exception is thrownt.

desktop/cmp/rest/RestGridModel.js
```diff
@@ -174,60 +172,52 @@ export class RestGridModel {
...
+            delFn();
+        } else {
+            const message = isFunction(warning) ? warning(records) : warning;
             XH.confirm({
-                message: warning,
+                message,
                 title: 'Warning',
                 icon: Icon.warning({size: 'lg'}),
-                onConfirm: () => this.deleteRecord(record)
+                onConfirm: () => delFn()
```
lbwexler, line 212: don't need the extra fat arrow function on 212, can just set to onConfirm: delFn

desktop/cmp/rest/RestGridModel.js
febbraiod: Thanks again. Need to learn to think more about the unhappy path.

lbwexler: sorry to be a pain, but for symmetry's sake if deleteRecord is catching the exception, bulkDeleteRecordsAsync should be catching the exception.  Also, let's make the method `deleteRecordAsync` to follow our conventions and for symmetry.

PR #2077, opened 2020-09-17
Bulk regroup for Admin config and preference panels

febbraiod:
Hoist P/R Checklist
-------------------


febbraiod:
Requires: xh/hoist-core#161

admin/regroup/RegroupDialog.js
```diff
@@ -0,0 +1,57 @@
...
+            filler(),
+            button({
+                text: 'Cancel',
+                onClick: () => model.close()
+            }),
+            button({
+                text: 'Save',
+                icon: Icon.check(),
+                intent: 'success',
+                disabled: model.groupName == null,
```
lbwexler, line 52: why can't I bulk change all group names to null?

admin/regroup/RegroupDialogModel.js
```diff
@@ -0,0 +1,61 @@
...
+        this._parent = parent;
+    }
+
+    async saveAsync() {
+        const {_parent, groupName} = this,
+            {selection, store} = _parent.gridModel,
+            ids = selection.map(it => it.id),
+            resp = await store.bulkUpdateRecordsAsync(ids, {groupName}),
+            intent = resp.fail > 0 ? 'warning' : 'success',
+            message = `Change group complete with ${resp.fail} failures.`;
```
lbwexler, line 41: Can we update this to "Successfully updated x records.", if there were any successes.  If there were any failures, just append "Failed to update 'y' records".  Want to be a bit more friendly on the expected happy path.

lbwexler:
Hi Don -- It occurs to me that this could easily be adapted to do a bulk edit on *any* single field value as specified by the user at run-time.  We could add a little two-stage menu to the context menu that allowed you to choose the field you wanted to edit, before bringing up a form with the single control that would be used to select the new value.  Would use the same control and validation that the full edit form would have used.  We could allow RestGridModel to suggest which fields (if any) were eligible for this treatment.  

Let's ask around the group and see how useful people would find it.  I bet people would find it very useful for, say, disabling rows in bulk

admin/tabs/preferences/PreferenceModel.js
```diff
@@ -76,6 +86,17 @@ export class PreferenceModel {
         actionWarning: {
             del: 'Are you sure you want to delete? Deleting preferences can break running apps.'
         },
+        toolbarActions: [
```
lbwexler, line 89: toolbarActions is the default -- should be able to remove

admin/tabs/preferences/UserPreferencePanel.js
```diff
@@ -4,12 +4,16 @@
  *
```
lbwexler, line 4: this entire file no longer changing, can be reverted

admin/regroup/RegroupDialog.js
febbraiod: Preference's and Config's groupName property is non-nullable via Domain Object constraints. I guess I could see that not always being the case if we decide to use this elsewhere, but do we really want to allow the request to throw?

lbwexler: thanks for the correction -- my bad

lbwexler:
Thanks for working through this -- think this landed in a great place, and can be the starting point for more general functionality of bulk editing built in to the framework.

PR #2083, opened 2020-09-21
GridModel checks isReady before reading selection

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -628,8 +628,7 @@ export class GridModel {
     noteAgSelectionStateChanged() {
         const {selModel, agGridModel, isReady} = this;
 
-        // Debounce allows for cases where grid has been unmounted and is no longer ready when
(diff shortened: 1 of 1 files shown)


amcclain:
This is a non-critical update to avoid throwing from within `GridModel.noteAgSelectionStateChanged()` during an edge-case where the grid changes its selection and is immediately unmounted / destroyed. I had a client app pattern that triggered this and noticed an error thrown to the console for a failed call to `agGridModel.getSelectedRowNodeIds()`. 

You will see that we deliberately throw if not ready there, making this a reasonable change I think to avoid:

https://github.com/xh/hoist-react/blob/fcfb1a756bac7320e47056e0d42a54163d9ce94a/cmp/ag-grid/AgGridModel.js#L380

Only question in my mind was if it would be better to clear the selection vs. leaving it as-is, but it seemed slightly more conservative to do nothing.

PR #2085, opened 2020-09-21
fixes #2068 :  Dragging grid column with element based header display…

### LINKED ISSUE #2068: Dragging grid column with element based header displays [object Object] in draggable
Not critical by any means but would be interested if we can avoid this by supplying an alternate string (e.g. column/field displayName w/fallback to colId) to show in the dragger here:

Screen Shot 2020-09-11 at 7 50 40 AM

cnrudd:
**Resolution:** 
A grid column header that has a function for headerName argument or element/object for resolved headerName
  now falls back to column.chooserName for the value to display when dragging the column header
  or in the column header tooltip.

Hoist P/R Checklist
-------------------

- [X] Added CHANGELOG entry.
- [X] Reviewed for breaking changes: no breaking changes.
- [X] Updated doc comments: not required.
- [X] Reviewed and tested on Mobile: not required.
- [X] Created Toolbox branch: https://github.com/xh/toolbox/compare/gridHeaderNameOnDrag?expand=1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,7 @@
 * Resetting Grid columns to their default state (e.g. via the Column Chooser) retains enhancements
   applied from matching Store fields.
 * Desktop `DateInput` now handles out-of-bounds dates without throwing exception during rendering.
-* A grid column header that has a function for headerName argument or element/object for resolved headerName
-  now falls back to column.chooserName for the value to display when dragging the column header
-  or in the column header tooltip.
+* Dragging a grid column with element based header no longer displays [object Object] in draggable.
 
 ### 📚 Libraries
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
 * Resetting Grid columns to their default state (e.g. via the Column Chooser) retains enhancements
   applied from matching Store fields.
 * Desktop `DateInput` now handles out-of-bounds dates without throwing exception during rendering.
+* A grid column header that has a function for headerName argument or element/object for resolved headerName
+  now falls back to column.chooserName for the value to display when dragging the column header
+  or in the column header tooltip.
 
 ### 📚 Libraries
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -29,6 +29,9 @@
 * Resetting Grid columns to their default state (e.g. via the Column Chooser) retains enhancements
   applied from matching Store fields.
 * Desktop `DateInput` now handles out-of-bounds dates without throwing exception during rendering.
+* A grid column header that has a function for headerName argument or element/object for resolved headerName
```
lbwexler, line 32: too much detail about the technical details of how this workaround happened -- let's just briefly explain what bug fixed

cmp/grid/columns/Column.js
```diff
@@ -317,7 +317,9 @@ export class Column {
                 colId: this.colId,
                 headerValueGetter: (agParams) => {
                     return agParams.location === 'header' ?
-                        isFunction(headerName) ? headerName({column: this, gridModel, agParams}) : headerName :
+                        isFunction(headerName) ?
```
lbwexler, line 320: why not checking to see if this is not a string -- couldn't be an element, or a function as well?

Also, this should default to 'displayName', not chooserName -- that seems tailor made for that!

cmp/grid/impl/ColumnHeader.js
```diff
@@ -77,7 +77,8 @@ export const columnHeader = hoistCmp.factory({
         if (isDesktop && isUndefined(xhColumn?.headerTooltip)) {
             onMouseEnter = ({target: el}) => {
                 if (el.offsetWidth < el.scrollWidth) {
-                    el.setAttribute('title', headerName);
+                    const title = isObject(headerName) ? xhColumn?.chooserName : headerName;
```
lbwexler, line 80: Seems like this should also be isString(headerName)?  headerName : props.displayName

A lot of this code would be clearer if we referred to the local variable as `headerElem` as that is the most general description of iut.

Issue #2084, opened 2020-09-21
Add commitOnChange functionality to ColChooser


petradish:
ColChooserButton's popover commits changes and omits the 'Save' button that exists in the ColChooserDialog implementation. We should give the popover an option to not commit on change and render the 'Save' button.

haynesjm42:
While we are enhancing the col chooser could we also make a couple of usability improvements?
* Allow the size of the col chooser to be configurable (helpful when dealing with many columns)
* Provide a means of adding an entire chooser group in 1 click (perhaps just a button on the group row?)

petradish:
I like those ideas! Maybe for the group, we can select all rows by clicking the group row?

PR #2086, opened 2020-09-21
Enhance ColChooserModel config and expose it via GridModel

petradish:
This PR addresses https://github.com/xh/hoist-react/issues/2084.

- Desktop ColChooserModel supports new config options: commitOnChange, showRestoreDefaults, width, and height, to specify ColChooser implementations in popover and dialog.
- ColChooserModel config can be specified via GridModel config: colChooserModel. It also accepts true or false to show or hide default components -- upgraders must replace enableColChooser with this new flag.
  - Toolbox PR: https://github.com/xh/toolbox/pull/427
- Added new 'restoreDefaults' action to default StoreContextMenu, since 'commitOnSave' will remove the 'Reset' button. (Also makes this action more clearly aligned with all grid defaults (not just columns)).

Hoist P/R Checklist
-------------------


amcclain:
Petra and I discussed and I had a few ideas for your input when you have a chance, @lbwexler -

- There's an issue currently where the dev has no way to configure/set options on the `ColChooserModel`, at least when it's launched from the default context menu. This PR has a workaround to add a new mode prop to the button, which will then set the mode on the model when the button is rendered, but Petra and I discussed how that's a bit unexpected.
  - I'm thinking it might be worth supporting an optional `GridModel.colChooserModel` config, which could be a model instance (rare) or a config/partial config for one. This would allow us to configure any aspect of the chooser API at the model level.
  - We could still support `enableColChooser: true|false` for convenience and backwards compat. But a dev could leave `enableColChooser` unspecified - passing any object to `colChooserModel` would enable the chooser.
- I found the mode concept to be a bit tricky, as modes often are, especially since we are looking at only two cases at the moment (and I couldn't think of any others). 
  - I'm favoring `commitOnChange` as a boolean - Petra mentioned she had that, so I wanted to discuss before we changed anything to avoid flipping around. To me it's a _little_ bit weird in the context of the col chooser, but not really - and once you think about it, it really seems to fit.
  - I would then also favor an explicit option such as `showRestoreDefaults` so apps could set this independently of the commitOnChange/mode. Petra explained the inconsistency with allowing restore of defaults while in batch/save mode, but I think on balance it ends up being more confusing to sometimes see the button and sometimes not - so I would default this to true. (Although I understand that for our new client app project we have other technical reasons for always disabling - hence supporting the flag.)
  - I would also suggest labelling the button "Restore Defaults" to match (almost) the context menu option. We could take the step of prompting the user if `commitOnChange:false` (or always) to make it clear that _all_ grid customizations will be _immediately_ reset.

Happy to discuss more - it's a busy day so I agreed with Petra I would write up this comment for discussion when we have a few minutes the three of us.

desktop/cmp/button/ColChooserButton.js
```diff
@@ -46,6 +47,8 @@ export const [ColChooserButton, colChooserButton] = hoistCmp.withFactory({
             return cloneElement(displayButton, {disabled: true});
         }
 
+        colChooserModel.setMode(mode);
```
lbwexler, line 50: don't want to set model state in render

lbwexler:
I agree entirely with your second major item, regarding a single 'commitOnChange'  and the "showRestoreDefaults" flag.

lbwexler:
I like the first suggestion as well.  Its analogous to the selection model.  Its also better for it to behave the same way whether it comes up via context menu or button, unlike what we have now!

lbwexler:
could deprecate enableColumnChooserModel, and ask upgraders to just set colChooserModel: true instead for the default behavior

CHANGELOG.md
```diff
@@ -2,6 +2,26 @@
 
 ## v37.0.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes
+* New `GridModel` config `colChooserModel` replaces `enableColChooser` to allow for more flexible
```
lbwexler, line 6: love this documentation -- a little more detailed than usual for changelog, but really clear. Should we move it to a jsdoc  for colChooserModel

cmp/grid/GridModel.js
```diff
@@ -1085,9 +1090,26 @@ export class GridModel {
...
     }
 
+    parseChooserModel(chooserModel) {
+        if (XH.isMobileApp) {
+            return chooserModel ? this.markManaged(new MobileColChooserModel(this)) : null;
+        }
 
-    createChooserModel() {
-        return XH.isMobileApp ? new MobileColChooserModel(this) : new DesktopColChooserModel(this);
+        if (chooserModel instanceof ColChooserModel) {
```
lbwexler, line 1098: 1099 is the case where we *don't* want to mark it as managed --

desktop/cmp/grid/impl/ColChooserModel.js
```diff
@@ -26,11 +27,26 @@ export class ColChooserModel {
...
+     * @param {Object} c - ColChooserModel config
+     * @param {GridModel} c.gridModel - model for the grid to be managed.
+     * @param {boolean} [c.commitOnChange] - Immediately render changed columns on grid.
+     *      Set to false to show Save button to commit changes on save. Default true.
+     * @param {boolean} [c.showRestoreDefaults] - show Restore Defaults button. Default true.
+     * @param {number} [c.width] - chooser width for popover and dialog
+     * @param {number} [c.height] - chooser height for popover and dialog
      */
-    constructor(gridModel) {
+    constructor({gridModel, commitOnChange, showRestoreDefaults, width, height}) {
```
lbwexler, line 42: easiest and also most self documenting to default all four of these in the method signature

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -137,6 +140,12 @@ export class StoreContextMenu {
                     hidden: !gridModel?.autosizeEnabled,
                     actionFn: () => gridModel.autosizeAsync()
                 });
+            case 'restoreDefaults':
+                return new RecordAction({
+                    text: 'Restore Grid Defaults',
+                    icon: Icon.reset(),
+                    actionFn: () => gridModel.restoreDefaults()
```
lbwexler, line 147: Should the action have a built in confirm?

lbwexler: I think so -- its a pretty rare and drastic action, and can also clarify that it is going to be *immediately* happening, which might help in the case that the button is clicked when commitOnChange is true

cmp/grid/GridModel.js
```diff
@@ -241,7 +248,6 @@ export class GridModel {
         compact,
 
         enableColumnPinning = true,
-        enableColChooser = false,
```
amcclain, line 244: I'm on the fence re. not transforming this for developers to make for a smoother transition, but after some thought I think it's the right choice.

However this is a great use for our `apiDeprecated` util to warn if developers are still specifying `enableColChooser`. Otherwise it would be very easy for a dev who doesn't catch the release note to disable column customization in all grids and not necessarily notice it.

I wonder if it would be a better to enable it by default as a core/normal part of Grid. My sense is that we enable in the large majority of grids, and disable only when we know all columns are shown or know we are programatically managing column selection in some other special way.

If we did that we could note it as an additional change and then devs could simply remove the `enableColChooser: true` configs that are likely sprinkled throughout their code.  

Thoughts? Does it seem risky in any way to allow column selection by default? (Especially considering that we *don't* wire up any default persistence).

cmp/grid/GridModel.js
```diff
@@ -310,12 +316,24 @@ export class GridModel {
...
 
-        this.colChooserModel = enableColChooser ? this.createChooserModel() : null;
+        this.colChooserModel = this.parseChooserModel(colChooserModel);
         this.selModel = this.parseSelModel(selModel);
         this.persistenceModel = persistWith ? new GridPersistenceModel(this, persistWith) : null;
         this.experimental = this.parseExperimental(experimental);
     }
 
+    /**
+     * Confirm before restoring defaults.
```
amcclain, line 326: Sorry to nitpick but here is an example of a comment we can zap. 😃

cmp/grid/GridModel.js
```diff
@@ -1085,9 +1103,26 @@ export class GridModel {
         };
     }
 
+    parseChooserModel(chooserModel) {
+        if (XH.isMobileApp) {
+            return chooserModel ? this.markManaged(new MobileColChooserModel(this)) : null;
+        }
```
amcclain, line 1109: 👍 this seems pragmatic and correct to me.

cmp/grid/GridModel.js
```diff
@@ -1085,9 +1103,26 @@ export class GridModel {
...
+
+        if (chooserModel instanceof ColChooserModel) {
+            return new DesktopColChooserModel(chooserModel);
+        }
+
+        if (isPlainObject(chooserModel)) {
+            return this.markManaged(new DesktopColChooserModel(defaults(chooserModel,
+                {gridModel: this})));
```
amcclain, line 1117: Would you mind fixing up the whitespace/line nesting here?  This is a bit of an odd line break.

cmp/grid/GridModel.js
```diff
@@ -1085,9 +1103,26 @@ export class GridModel {
...
+        }
 
-    createChooserModel() {
-        return XH.isMobileApp ? new MobileColChooserModel(this) : new DesktopColChooserModel(this);
+        if (isBoolean(chooserModel)) {
+            return chooserModel ? this.markManaged(new DesktopColChooserModel({gridModel: this})) : null;
+        }
+
+        console.warn(`colChooserModel config in GridModel not supported: ${chooserModel}`);
```
amcclain, line 1124: I can go either way on the final step to validate that it's a boolean - would also be fine just using truthiness at this point - i.e. after the check for plainObject simply `return chooserModel ? this.markManaged(new DesktopColChooserModel({gridModel: this})) : null;`

But if we are going to check for the boolean we should throw instead of console.warning.  That's a dev-time error and there's no need to be subtle about it.

desktop/cmp/button/ColChooserButton.js
```diff
@@ -20,14 +20,15 @@ import {button, Button} from './Button';
...
  *
- * Requires the `GridModel.enableColChooser` config option to be true.
+ * Requires the `GridModel.colChooserModel` config option. Set to true for default implementation.
  */
 export const [ColChooserButton, colChooserButton] = hoistCmp.withFactory({
     displayName: 'ColChooserButton',
     model: false,
 
-    render({icon, title, gridModel, popoverPosition, chooserWidth, chooserHeight, ...rest}) {
+    render({icon, title, gridModel, popoverPosition, ...rest}) {
```
amcclain, line 29: That ends up being a nice cleanup!

desktop/cmp/button/ColChooserButton.js
```diff
@@ -42,7 +43,7 @@ export const [ColChooserButton, colChooserButton] = hoistCmp.withFactory({
         }
 
         if (!colChooserModel) {
-            console.error('No ColChooserModel available on bound GridModel - ensure enableColChooser config is set to true.');
+            console.error('No ColChooserModel available on bound GridModel - ensure colChooserModel config is set to true.');
```
amcclain, line 46: Suggest ` - enable via GridModel.colChooserModel config`

desktop/cmp/contextmenu/StoreContextMenu.js
```diff
@@ -37,7 +37,10 @@ export class StoreContextMenu {
      *          `exportExcel` - same as above.
      *          `exportCsv` - export grid data to CSV via Hoist's server-side export capabilities.
      *          `exportLocal` - export grid data to Excel via ag-Grid's built-in client side export.
-     *          `'autosizeColumns` - autosize columns to fit their contents.
+     *          `autosizeColumns` - autosize columns to fit their contents.
+     *          `restoreDefaults` - restore grid defaults.
```
amcclain, line 41: Let's clarify a bit, briefly, as to what state is restored - column selection and ordering, grouping, and sort - something like that

cmp/grid/GridModel.js
petradish: ok -- will take your first suggestion and remove the console warning

cmp/grid/GridModel.js
petradish: I'm not sure about other apps and their use of col selection, but I can think of a couple places where it would be a bad breaking change - also harder to search the code for the config when it wasn't there to begin with to add `colChooserModel: false`. I like the idea of using apiDeprecated for now to make the transition easier and show exactly where in the code the config needs to be changed.

cmp/grid/GridModel.js
```diff
@@ -162,6 +166,10 @@ export class GridModel {
      *      `store.SummaryRecord` to be populated. Valid values are true/'top', 'bottom', or false.
      * @param {(StoreSelectionModel|Object|String)} [c.selModel] - StoreSelectionModel, or a
      *      config or string `mode` with which to create one.
+     * @param {(ColChooserModel|Object|boolean)} [c.colChooserModel] - ColChooserModel, or a
+     *      config with which to create one, or boolean `true` to enable default.
+     *      @see ColChooserModel desktop impl for valid config options.
```
lbwexler, line 171: We should have a jsdoc type for what ColChooserModel can be, and do *not* need to support users actually providing an instance.   The instance is currently in the 'impl' category which suggests we are not looking to make it part of the API in its current form.

cmp/grid/GridModel.js
lbwexler: yes, but we mean 'apiRemoved', I think.

cmp/grid/GridModel.js
```diff
@@ -255,8 +261,10 @@ export class GridModel {
         useVirtualColumns = false,
         autosizeOptions = {},
         experimental,
+        enableColChooser,
         ...rest
     }) {
+        apiDeprecated(enableColChooser, 'enableColChooser', "Use 'colChooserModel' instead");
```
lbwexler, line 267: think we want this *not* to appear as an argument, and call apiRemoved(rest.enableColChooser, )

cmp/grid/GridModel.js
```diff
@@ -1085,9 +1102,20 @@ export class GridModel {
         };
     }
 
+    parseChooserModel(chooserModel) {
+        if (XH.isMobileApp) {
+            return chooserModel ? this.markManaged(new MobileColChooserModel(this)) : null;
+        }
+
+        if (chooserModel instanceof ColChooserModel) {
```
lbwexler, line 1110: think we should remove this section here, as discussed above -- but if we were to leave it, this code incorrectly recreating a new model on line 1111

cmp/grid/GridModel.js
```diff
@@ -310,12 +318,21 @@ export class GridModel {
...
         });
 
-        this.colChooserModel = enableColChooser ? this.createChooserModel() : null;
+        this.colChooserModel = this.parseChooserModel(colChooserModel);
         this.selModel = this.parseSelModel(selModel);
         this.persistenceModel = persistWith ? new GridPersistenceModel(this, persistWith) : null;
         this.experimental = this.parseExperimental(experimental);
     }
 
+    confirmRestoreDefaults() {
```
lbwexler, line 327: wonder if a better pattern here would be to just add an additional flag to the existing restoreDefaults method of the form `requireConfirm= false`?

the name `confirmRestoreDefaults` seems a bit ambiguous to me

desktop/cmp/grid/impl/ColChooserDialog.js
```diff
@@ -14,17 +14,19 @@ export const colChooserDialog = hoistCmp.factory({
...
 
         return dialog({
             icon: Icon.gridPanel(),
             title: 'Choose Columns',
             isOpen: true,
             onClose: () => model.close(),
             item: colChooser({model}),
-            className
+            className,
+            style: {width}
```
lbwexler, line 29: where does height get set?

lbwexler (commented):
Just one comment about the height, other than that am fine with this

desktop/cmp/grid/impl/ColChooserDialog.js
petradish: The dialog will flex to the height of the colChooser (but not its width)

lbwexler:
Discussed with Petra. - She is doing a bit more testing, and also one more change to make the reset alert easily configuable with a new model property

PR #2097, opened 2020-09-29
Add remote diff functionality to JsonBlob admin

TomTirapani:
See hoist-core PR: https://github.com/xh/hoist-core/pull/170

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/differ/DifferModel.js
+++ b/admin/differ/DifferModel.js
@@ -26,6 +26,9 @@ export class DifferModel  {
 
     parentGridModel;
     entityName;
+    displayName;
+    columns;
+    matchFields;
     url;
     clipboardContent;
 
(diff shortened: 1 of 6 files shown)


admin/tabs/preferences/JsonBlobModel.js
```diff
@@ -0,0 +1,125 @@
...
+            {field: 'description', formField: {item: textArea()}},
+            {field: 'value'},
+            {field: 'dateCreated'},
+            {field: 'lastUpdated'},
+            {field: 'lastUpdatedBy'}
+        ]
+    });
+
+    @managed
+    differModel = new DifferModel(this.gridModel, 'jsonBlob');
```
haynesjm42, line 119: It appears the DifferModel will match local and remote records by just comparing the name, which works fine for configs and prefs, but for json blobs I think we also need to take the owner and type into account to match the records. Perhaps we should pass the matcher fn into the DifferModel constructor?

TomTirapani:
Thanks for taking a look John!

I've updated the DifferModel to be able to handle matching across different fields, and also to display different fields. Unfortunately, its difficult to test this properly without the hoist-core side being deployed at one of the remote Toolbox instances :/

admin/differ/DifferModel.js
```diff
@@ -26,6 +26,9 @@ export class DifferModel  {
 
     parentGridModel;
     entityName;
+    displayName;
+    columns;
```
haynesjm42, line 30: Should this be `columnFields`?

admin/differ/DifferModel.js
```diff
@@ -77,12 +90,12 @@ export class DifferModel  {
...
                 {field: 'status', hidden: true},
-                {field: 'name', width: 200},
-                {
-                    field: 'type',
-                    width: 80,
-                    renderer: this.valueTypeRenderer
-                },
+                ...this.columnFields.map(field => {
+                    const ret = {field, renderer: this.fieldRenderer};
+                    if (field === 'valueType') ret.headerName = 'Type';
```
haynesjm42, line 95: I know this differ stuff is only used internally right now so this is probably fine, but having these checks for specific fields which come from specific types of records seems wrong...

haynesjm42: Maybe we allow providing column defs instead of just fields? Can keep the support for building the def from a field if they pass a string to make the api easy to use

TomTirapani: Makes sense - will push an update shortly

admin/differ/DifferModel.js
```diff
@@ -270,25 +306,22 @@ export class DifferModel  {
 
     valueRenderer(v) {
```
haynesjm42, line 307: I think this should be moved out of here and be part of the ConfigModel column def

Issue #2096, opened 2020-09-26
ExceptionDialog closes instantly when it is open after hitting "Enter" key on TextInput that throws an error


cnrudd:
Noticed in a client app that when this sequence happens:
1 Hit "Enter" on a formless TextInput to commit a value
2 A reaction to the value change makes an ajax call that returns with an error
3 And opens an ExceptionDialog via .catchDefault()

The ExceptionDialog opens with the "Close" button autofocused, and the "Enter" key event is bubbling up/propagating
and activates the Close button so that the dialog closes instantly.

To the user/developer, this may look like the dialog:
1.  never opens
2. or flashes open and closes in an instant

I looked into solutions involving ev.preventDefault(), ev.stopPropagation(), ev.nativeEvent.preventDefault(), ev.nativeEvent.stopPropagation(), ev.nativeEvent.stopImmediatePropagation() on onKeyDown and onKeyPress on the TextInput component - in hoist-react  and in the app.    Nothing stopped the event bubbling in time to let the dialog render AFTER the bubbling had been stopped.

I ended up fixing this by delaying the ajax call with wait(500).  Could optimize that fix further by delaying the promise resolution only if an error is returned.   

But, of course this is something that would be better addressed with event management on textInput (doesn't seem possible)  or maybe delaying the autofocus on the dialog close button.

cnrudd:
I think another possible fix that I prefer over the others is to add a check in the ExceptionDialog's close btn onClick handler to see if the ev.target value is the close button.  If it is not, ignore the event and do not close the dialog.   That seems like a clean solution if it works.

cnrudd:
The ev.target idea did not work.  By the time the event triggers the close button click, the target is always the close button, whether the event started at the textInput, or at the close button.

lbwexler:
Potential good news!  It sounds to me like this might be fixed by react-17

https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation

lbwexler:
Before I forget, want to say that I saw this error in a client app with a hotkey that brings up a dialog, being input right in to that dialog's default text.  A slight delay fixed the issue, but it remains very surprising and ugly for app developers to have to work around in this way

cnrudd:
[...]

PR #2100, opened 2020-09-29
Delay focus on exceptionDialog dismiss Button. (#2096)

cnrudd:
Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry: added.
- [x] Reviewed for breaking changes.  No breaking changes.
- [x] Reviewed and tested on Mobile: The bug does not exist on Mobile, because the exceptionDialog close button is not focused on Mobile.  This fix is null safe for Mobile.
- [x] Created Toolbox branch: [This branch](https://github.com/xh/toolbox/tree/exceptionDialogFocusDelay) reproduces the bug in the [desktop OHLC chart](http://localhost:3000/app/charts) and in this [mobile grid](http://localhost:3000/mobile/grids) demonstrates that the bug does not exist in mobile.

### DIFF
--- a/desktop/appcontainer/ExceptionDialog.js
+++ b/desktop/appcontainer/ExceptionDialog.js
@@ -4,6 +4,8 @@
  *
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
+import {FocusStyleManager} from '@blueprintjs/core';
+
 import {ExceptionDialogModel} from '@xh/hoist/appcontainer/ExceptionDialogModel';
 import {filler, fragment} from '@xh/hoist/cmp/layout';
 import {hoistCmp, uses, XH} from '@xh/hoist/core';
@@ -27,7 +29,12 @@ export const exceptionDialog = hoistCmp.factory({
     render({model}) {
         const {exception, options} = model;
 
-        if (!exception) return null;
+        if (!exception) {
+            FocusStyleManager.onlyShowFocusOnTabs();
+            return null;
+        }
+
+        FocusStyleManager.alwaysShowFocus();
 
         const onClose = !options.requireReload ? () => model.close() : null;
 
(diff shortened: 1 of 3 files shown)


desktop/appcontainer/ExceptionDialog.js
```diff
@@ -27,7 +29,12 @@ export const exceptionDialog = hoistCmp.factory({
     render({model}) {
         const {exception, options} = model;
 
-        if (!exception) return null;
+        if (!exception) {
+            FocusStyleManager.onlyShowFocusOnTabs();
```
cnrudd, line 33: This is necessary only when programmatically setting focus on an element.   It wasn't necessary when autoFocus:true.

desktop/appcontainer/ExceptionDialog.js
```diff
@@ -27,7 +29,12 @@ export const exceptionDialog = hoistCmp.factory({
...
+            return null;
+        }
+
+        FocusStyleManager.alwaysShowFocus();
```
cnrudd, line 37: This `FocusStyleManager. alwaysShowFocus();` is necessary only when programmatically setting focus on an element.   It wasn't necessary when autoFocus:true.

lbwexler:
So I have been looking at this one in depth and discussed at length with Colin.  Not a huge fan of the solution in this PR, as it seems a bit brittle, and does not get to the core of the problem.

It looks like we are unable to successfully consume the keyDown event, and it indeed immediately generates a pseudo "click" event on the focused button. 

That pseudo event bit, is actually built-in to html, as far as I can tell.  The thing that seems wrong is that we are unable to consume the event with either preventDefault(), or stopPropagation().  See the following and related for hints about the source of this problem: 

https://medium.com/@ericclemmons/react-event-preventdefault-78c28c950e4


Note that if I change to keyUp for committing the text input's value, everything works fine -- the built in keyDown or keyPress handlers have already been fired. That behavior is actually fine for me, and its a simple workaround, but still does not seem incredibly satisfying. Also,  I wonder if it will have any material effect on existing code, and if we would logically extend it to to anywhere in hoist we listen for an enter key.

PR #2101, opened 2020-09-29
GridModel supports treeGridStyleMode property - Fixes #1250

### LINKED ISSUE #1250: Distinct styling for tree grid parent row
Would like to explore (optional) distinct styling for tree-grid parent rows.  Something in between the full-on dark background shading of the full-width row groups and what we have now, where parents and children are only identified by their indent level.

One option we have used in the past successfully is a lighter background shading + ledger-style top border on group rows.

Would need to consider how this styling applies to nested (multi-level) groups. Should it appear on top-level groups only? Be the same for all groups? Have some gradation based on color or intensity between the levels?

febbraiod:
Hoist P/R Checklist
-------------------


cmp/grid/GridModel.js
```diff
@@ -130,6 +130,8 @@ export class GridModel {
     @observable showSummary = false;
     /** @member {string} */
     @observable emptyText;
+    /** @member {string} */
+    @observable treeGridStyleMode; // Not sure this can't be immutable, but for our example I need observable
```
febbraiod, line 134: Maybe safer to leave observable for as yet unimagined use cases?

lbwexler: no, I don't think so.  There are many immutable properties on the GridModel.  Also, can we call it 'treeStyle'?

febbraiod: Sounds good. And yeah that name is bulky, although last I spoke to Anselm he said "let us leave some room to grow by using a config with “mode” in the name". Maybe we go with treeStyleMode as grid is really the obv throw away

cmp/grid/GridModel.js
```diff
@@ -253,7 +258,8 @@ export class GridModel {
         showHover = false,
         rowBorders = false,
         cellBorders = false,
-        stripeRows = true,
+        treeGridStyleMode = 'none',
+        stripeRows = treeGridStyleMode == 'none' ? true : false,
```
lbwexler, line 262: triple equals.  

Also, If we are going to support 'none' (which I support) why also support null?  Let's go ahead and use a psuedo-enumeration like RenderMode.

lbwexler: I think having the enumeration in place will give us room to grow, and also obviate the need for 'mode' in the name.  Mode is confusing in this context because we already have the main flag with the name 'treeMode'

amcclain: Yes I didn't mean to insist on that word

febbraiod: Got it. Didn't mean to imply insistence, just wanted to put all the info on the table

lbwexler:
We need a toolbox toggle for this in the 'options' sidebar.  Saw that the
associated branch has this manipulable in the portfolio, example -- that
really just makes sense for dev purposes, we won't want this for merging.

I am going to go back on what I said earlier -- for symmetry lets make this
visual property behave just like the myriad of other visual properties on
GridModel, that is , not observable, but settable -- e.g.:

get rowBorders() { return this.agGridModel.rowBorders }
    setRowBorders(rowBorders) { this.agGridModel.setRowBorders(rowBorders) }

cmp/grid/Grid.scss
```diff
@@ -289,4 +289,97 @@
...
+//------------------------
+// Tree grid styling modes
+//------------------------
+
+.xh-grid--highlight-groups {
+
+  .ag-theme-balham.xh-ag-grid, .ag-theme-balham.xh-ag-grid--no-stripe-rows {
+    div[class*=" ag-row-group"], div[class^="ag-row-group"] {
+      &:not(.ag-row-selected), &.ag-row.ag-row-odd:not(.ag-full-width-row):not(.ag-row-selected) {
+        background-color: hsl(207, 90, 94);
```
amcclain, line 304: We should declare new CSS vars for this group background color and the border color.

Then it looks like we should be able to simply set them to different default values for the light/dark theme variations in `vars.scss` and not need to duplicate all these rules here.  That's the approach we should be using consistently for the grid and many other places.

amcclain (commented):
As mentioned, very excited to get this improvement into Hoist.

I was wondering if there would be any value in having the new `treeStyle` config take an object, which would currently support `{highlightGroups, topLevelGroupBorders}` as two boolean options you can set independently of each other. 

We could export an enum that you could use to reference the expected common combinations of these settings so you didn't need to pass an object. Could still even support the string form which would look for that property on the enum.

Would that be worth the slight extra complexity to make room for this API better supporting more independent options around gradients, supporting shading on extra levels, etc. in the future?

amcclain (approved):
I think this is looking great - I would be ready to merge and get feedback from app developers.

PR #2103, opened 2020-09-30
Select can now show a check mark by the selected items in multiMode & hideSelectedOptions: false

cnrudd:
This change allows the checkmark on selected options to be shown when the select has this config:
```
enableMulti: true,
rsOptions: {
   closeMenuOnSelect: false,
   hideSelectedOptions: false
}
```

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry..
- [x]  Reviewed and tested on Mobile: not required.
- [x] Created Toolbox branch / PR.   Did not create new toolbox branch, but you can test this by running the develop toolbox branch and testing the last select on http://localhost:3000/admin/tests/select.

### DIFF
--- a/desktop/cmp/input/Select.js
+++ b/desktop/cmp/input/Select.js
@@ -517,14 +517,7 @@ export class Select extends HoistInput {
     };
 
     get suppressCheck() {
-        const {props} = this;
-        let dflt = false;
-
-        if (props.enableMulti) {
-            dflt = this.hideSelectedOptions;
-        }
-
-        return withDefault(props.hideSelectedOptionCheck, dflt);
+        return withDefault(this.props.hideSelectedOptionCheck, this.hideSelectedOptions);
     }
 
     // Match react-select defaulting.
--- a/desktop/cmp/input/Select.js
+++ b/desktop/cmp/input/Select.js
@@ -509,7 +509,13 @@ export class Select extends HoistInput {
 
     get suppressCheck() {
         const {props} = this;
-        return withDefault(props.hideSelectedOptionCheck, props.enableMulti && !props.rsOptions?.hideSelectedOptions);
+        let dflt = false;
+
+        if (props.enableMulti) {
+            dflt = props.rsOptions?.hideSelectedOptions;
+        }
+
+        return withDefault(props.hideSelectedOptionCheck, dflt);
     }
 
     //------------------------
--- a/desktop/cmp/input/Select.js
+++ b/desktop/cmp/input/Select.js
@@ -176,11 +176,15 @@ export class Select extends HoistInput {
     get creatableMode() {return !!this.props.enableCreate}
     get windowedMode() {return !!this.props.enableWindowed}
     get multiMode() {return !!this.props.enableMulti}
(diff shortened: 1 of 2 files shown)


cnrudd:
The formatOptionLabel and optionRenderer methods have 1 test each that run for every option that could be pushed up to a higher level, making the rendering of a large number of options a little faster.
Here is an "optimized" branch that includes this PR change and the optimization of these tests, if interested:  https://github.com/xh/hoist-react/compare/selectOptionLabelFormatterOptimized

amcclain:
We should consider promoting the two options you mention to props on the component (instead of rsOptions) if we are finding them useful to recommend in certain cases. That would make this change a bit less obscure as well, avoiding the need to reach into a potentially null `props.rsOptions` object.  They both seem like pretty straightforward props - in general we didn't want to simply go and mirror the entire props API (hence rsOptions), but we shouldn't hesitate to do so if we need it.

Thoughts?

cnrudd:
@amcclain I think the usage of `select` in the way it is proposed in this fix ((enable multi, user wants to see the selected items in the menu, doesn't want to have to re-open the menu after each selection)) is a pretty good and standard usage, that we should support in a way that makes it as easy as possible for a developer to setup.   So I agree.  I'll promote them to props.

cnrudd:
@amcclain Might even venture that `closeMenuOnSelect` should default to false if enableMulti = true.

cnrudd:
> @amcclain Might even venture that `closeMenuOnSelect` should default to `false` if `enableMulti` = `true`.

I this idea and it feel right.

desktop/cmp/input/Select.js
```diff
@@ -509,7 +518,18 @@ export class Select extends HoistInput {
...
+        let dflt = false;
+
+        if (props.enableMulti) {
+            dflt = this.hideSelectedOptions;
+        }
+
+        return withDefault(props.hideSelectedOptionCheck, dflt);
+    }
+
+    // Match react-select defaulting.
```
cnrudd, line 530: See https://github.com/JedWatson/react-select/blob/397e95337f54074cf0705cb5b06ce8951be72f2a/packages/react-select/src/Select.js#L1152

desktop/cmp/input/Select.js
```diff
@@ -170,11 +176,15 @@ export class Select extends HoistInput {
...
-    get filterMode() {return withDefault(this.props.enableFilter, true)}
+    get filterMode() {return this.props.enableFilter ?? true}
     get selectOnFocus() {
-        return withDefault(this.props.selectOnFocus,
-            !this.multiMode && (this.filterMode || this.creatableMode));
+        return this.props.selectOnFocus ??
+            (!this.multiMode && (this.filterMode || this.creatableMode));
     }
+    get suppressCheck() {return this.props.hideSelectedOptionCheck ?? this.hideSelectedOptions}
+    // Match react-select defaulting for hideSelectedOptions.
```
cnrudd, line 185: See https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L1152

PR #2113, opened 2020-10-08
Dynamic Tabs for TabContainer

lbwexler:
See also: https://github.com/xh/toolbox/pull/430

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/tab/TabContainerModel.js
+++ b/cmp/tab/TabContainerModel.js
@@ -140,31 +140,40 @@ export class TabContainerModel {
         ensureUniqueBy(tabs, 'id', 'Multiple tabs have the same id.');
 
         tabs = tabs.filter(p => !p.omit);
-        this.activeTabId = this.initialActiveTabId(tabs);
-        const newTabs = tabs.map(t => (
-            t instanceof TabModel ? t : new TabModel({...t, containerModel: this})
-        ));
 
-        this.tabs = newTabs;
+        // Adjust state -- intentionally setting activeTab *before* instantiating new tabs.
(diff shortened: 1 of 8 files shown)


cmp/tab/TabContainerModel.js
```diff
@@ -134,6 +122,57 @@ export class TabContainerModel {
...
+    addTab(tab, index = this.tabs.length) {
+        const {tabs} = this;
+        this.setTabs([...tabs.slice(0, index), tab, ...tabs.slice(index)]);
+    }
+
+    /**
+     * Remove a Tab for display.
+     * @param {Object|String} tab - TabModel or id of TabModel to be removed.
+     */
+    removeTab(tab) {
```
petradish, line 168: Hi! I can't leave a comment on the non-diff, but in line 206, we should not activate a tab if it's not found. (if you were to add an onClick 'x' handler to the tab title, it would try to activate what's been removed)

lbwexler: Fixed!

petradish (approved):

lbwexler:
I will create a seperate branch to add the x affordance as part of the toolkit.  Hard to do it generally without a fixed size tab and/or better delineation between tabs.

lbwexler:
Remaining issues:

+ See tabCloseAffordance branch for an example of how we could build the affordance in to the tab.  Not sure if this is strictly neccessary.  Seems like doing it properly requires a bit better seperation of tabs, and constant size tabs, a la Chrome.

Issue #2118, opened 2020-10-12
Chart's isVisible check is not correct all the time.


cnrudd:
In a client app I am upgrading from Hoist 29 to Hoist 36.x,
I am finding that this visibility test, released into Hoist-React v30,
https://github.com/xh/hoist-react/blob/e80b8eefa8520086d85cb57936c96ed2451d8963/cmp/chart/Chart.js#L157

can incorrectly cause a chart not to be rendered.

The error happens when a chart component has no width or height specs, and is put into a flex container that has no width or height specs, and the chart is expected to fill all available space, and expand its container to fill all its available space.

This [S.O. issue has a good discussion](https://stackoverflow.com/questions/19669786/check-if-element-is-visible-in-dom) of the strategies that could be used to test DOM node visibility.  It mentions [jquery's solution](https://github.com/jquery/jquery/blob/master/src/css/hiddenVisibleSelectors.js):
```
return !!( elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length );
```

and it looks like we stopped just short using jquery's entire solution in our test:
```
if (parentDims.width === 0 || parentDims.height === 0) return;
```

Does anybody remember why we left out getClientRects in our test?

In my testing in my client app, adding the getClientRects().length test solves the bug.

I will put in a PR for an isElVisisble function that uses the 3 checks, and am then happy to discuss more.

cnrudd:
Closing; workaround found in client app: define chart container's height=400, the highcharts default: https://api.highcharts.com/highcharts/chart.height

Issue #1703, opened 2020-02-24
Chart resize and hidden tabs


TomTirapani:
We noticed an issue with our chart resize detection in a client project with slow chart requests.

If a chart is in a tab, it is possible to trigger a refresh on mount but complete the refresh when the tab is hidden. In this situation, Highcharts is unable to determine its container size when it first renders. Furthermore, `useOnResize` will not trigger when it becomes visible because the size hasn't changed since it was first mounted.

The end result is a chart drawn with default dimensions. You have to manually trigger another resize to get it to display correctly:

<img width="813" alt="Screenshot 2020-02-24 at 12 22 59" src="https://user-images.githubusercontent.com/3017757/75154693-3642f700-5706-11ea-8f3b-70e8373bb59a.png">

(Recreated in Toolbox by adding `await wait(5000)` to `PortfolioService.getLineChartSeriesAsync()`)

**To replicate:**

1) Mount a slow-loading chart in a tab, then immediately change tab. `useOnResize` will fire when mounted, tracking `prevWidth` & `prevHeight`.

2) Immediately change tab while the chart is still loading, and allow the chart to finish loading while the tab is hidden. When the chart finishes loading, it will call `renderHighChart()`, where it will attempt to derive its dimensions from its parent:

```
dims = this.getChartDims({
	width: parentEl.offsetWidth,
	height: parentEl.offsetHeight
});
```

Since the parentEl is currently hidden, these will be `0` and Highcharts will ignore them.

3) Return to the tab. Since the dimensions haven't changed since from the `prevWidth` & `prevHeight` set when it was mounted, `useOnResize` will not fire.

PR #2120, opened 2020-10-12
More thorough test of DOM node visibility.

cnrudd:
This proposes to resolve issue: https://github.com/xh/hoist-react/issues/2118

- [x] Reviewed for breaking changes: no breaking changes
- [ ] Reviewed and tested on Mobile: Ack! There are no Chart demos on mobile Toolbox!.
- [x] Created Toolbox branch: 
https://github.com/xh/toolbox/tree/isElVisible  This new branch reproduces the test that Tom described here: https://github.com/xh/hoist-react/issues/1703 back in Feb 2020 and confirms that the change proposed in this PR still solves the original problem:  prevents the chart from rendering if it is not visible.  You can test it when running this branch by loading this URL: http://localhost:3000/app/charts/line
This branch also demonstrates this bug if you run it against hoist-react's `develop` branch: load http://localhost:3000/app/charts/ohlc. to see 4 charts that do NOT render.
Switch hoist-react to the new bugfix branch `isElVisible` to see the charts render.

The OHLC page has been recoded to put the panel that contains the 4 charts in a div that has no height, and is not a flexbox.
So, the div does not expand vertically, and the panel inside it does not, so the current visible test fails the height === 0 test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### 🐞 Bug Fixes
 
-* Improved test of is Dom node visible.
+* Improved test of is DOM node visible.
 
 [Commit Log](https://github.com/xh/hoist-react/compare/v36.4.0...develop)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v37.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Improved test of is Dom node visible.
+
 [Commit Log](https://github.com/xh/hoist-react/compare/v36.4.0...develop)
 
 ## v36.4.0 - 2020-10-09
--- a/cmp/chart/Chart.js
+++ b/cmp/chart/Chart.js
@@ -16,7 +16,7 @@ import {
     useOnResize,
     useOnVisibleChange
 } from '@xh/hoist/utils/react';
-import {isElVisible} from '@xh/hoist/utils/js/';
+import {isNodeVisible} from '@xh/hoist/utils/js/';
 import {assign, castArray, clone, isEqual, merge, omit} from 'lodash';
 import PT from 'prop-types';
 import {ChartModel} from './ChartModel';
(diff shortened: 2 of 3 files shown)


lbwexler:
My comments:

Love the idea about a helper function here, but the burden of making it generally correct goes up if we do that. 

Also a bit concerned about the changing semantics of this, and just want to confirm its what we intend:

 + For both the charts and the ResizeObserver, we were saying they were invisible if *either* dimension was zero.  Now we are saying that they are invisible only if *both* are zero.  Is that intended?  Seems wrong to me.

+ What is the crux of the fix here?  The change of the boolean semantics for height/width or the addition of the extra check on clientRects

cnrudd:
@lbwexler Thanks for the good right questions.
1: The crux of the fix indeed was the "The change of the boolean semantics for height/width", not "the addition of the extra check on clientRects"

The fix works just as well if this function:
```
export function isNodeVisible(node) {
    return !!(node.offsetWidth || node.offsetHeight || node.getClientRects().length);
}
```
is
export function isNodeVisible(node) {
    return !!(node.offsetWidth || node.offsetHeight);
} 



Now that I realize the above, this brings up these questions:
I am not able to reproduce the bug raised here: https://github.com/xh/hoist-react/issues/1703
even after adding the wait(.. ) as suggested in #1703.  The chart renders fine when the tab is hidden and parentEl width & height are both 0.

So, maybe the proper fix is to remove this check from chart, and leave resizeObserver logic alone?
```
            if (parentDims.width === 0 || parentDims.height === 0) return;	
```
if someone can confirm that this #1703 is no longer an issue with that check removed, I think that would be the way to go.

cnrudd:
Here's a screen recording of my inability to reproduce #1703
Screen-Recording-2020-10-14-at-1

cnrudd:
I discussed this issue with @TomTirapani, and we confirmed that the bug described in #1703 should be happening in my local instance when the width & height check, and the onVisibleChange hook are commented out and the wait(..) is put on the data call and the "user flow to recreate bug" is followed  -- but the bug IS NOT happening.

However, Tom explained that the check is there also to prevent un-needed re-renders when the chart has an ancestor set to display:none or some other setting making it not visible, and is reluctant to make any changes to it.

So, I have recoded the client app to use defined heights on containers of charts that previously expanded their container heights.

Closing this PR and issue #2118.

PR #2125, opened 2020-10-13
Improve hover and active background-color for header tool buttons in …

### LINKED ISSUE #2123: In Panel>headerItems, button styling on hover should be more transparent.
This issue exist on my local instance of toolbox in Light Theme:
http://localhost:3000/app/panels/sizing

and in deployed develop in Light Theme:
https://toolbox-dev.xh.io/app/panels/sizing

Here is a screenshot of the button in my local instance when hovered over:
image

The issue does NOT exist in Dark Theme.

I'm seeing the same bug in a client app running the latest HR commits on my dev machine.

cnrudd:
…light theme. (fixes #2123)

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.
- [x] Updated doc comments:  not required.
- [x] Reviewed and tested on Mobile:  not required.

### DIFF
--- a/desktop/cmp/button/Button.scss
+++ b/desktop/cmp/button/Button.scss
@@ -370,3 +370,22 @@
     }
   }
 }
+
+//------------------------
+// Header "tool" buttons in panels and docks
+// Adjustment needed only in light theme.
+//------------------------
+*:not(.xh-dark),
+*:not(.bp3-dark) {
+  .xh-dock-view__header,
+  .xh-panel-header__items {
+    .xh-button.bp3-button.xh-button--minimal {
+      &.xh-button--intent-none {
+        &.xh-button--enabled {
+          &:hover {background-color: var( --xh-intent-neutral-trans1)}
+          &:active {background-color: var( --xh-intent-neutral-trans2)}
+        }
+      }
+    }
+  }
+}
--- a/desktop/cmp/button/Button.scss
+++ b/desktop/cmp/button/Button.scss
@@ -375,8 +375,7 @@
 // Header "tool" buttons in panels and docks
 // Adjustment needed only in light theme.
 //------------------------
-*:not(.xh-dark),
-*:not(.bp3-dark) {
+*:not(.xh-dark) {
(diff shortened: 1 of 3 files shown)


desktop/cmp/button/Button.scss
```diff
@@ -370,3 +370,22 @@
...
+//------------------------
+// Header "tool" buttons in panels and docks
+// Adjustment needed only in light theme.
+//------------------------
+*:not(.xh-dark),
+*:not(.bp3-dark) {
+  .xh-dock-view__header,
+  .xh-panel-header__items {
+    .xh-button.bp3-button.xh-button--minimal {
+      &.xh-button--intent-none {
```
cnrudd, line 383: I followed the suggestion from PanelHeader.scss, where we don't worry about other intents or styles other than minimal when header buttons are concerned.

amcclain (commented):
This seems very closely related to #2127 which I filed only yesterday - can you take a look at folding in a fix for that here as well?

desktop/cmp/button/Button.scss
```diff
@@ -370,3 +370,22 @@
     }
   }
 }
+
+//------------------------
+// Header "tool" buttons in panels and docks
+// Adjustment needed only in light theme.
+//------------------------
+*:not(.xh-dark),
+*:not(.bp3-dark) {
```
amcclain, line 379: Do you know if we need the `:not(.bp3-dark)` here?  Is there a case where that would be applied but `.xh-dark` is not?

cnrudd:
> This seems very closely related to #2127 which I filed only yesterday - can you take a look at folding in a fix for that here as well?

@amcclain #2127 fixed in latest commit.

desktop/cmp/button/Button.scss
cnrudd: @amcclain I've removed `:not(.bp3-dark)`.  I had just thrown it in because it seemed to be a pattern up higher.

amcclain (approved):
Looks good, thanks for addressing these issues.

PR #2128, opened 2020-10-14
Always wrap grid cell contents to support flexbox alignment and overflow

TomTirapani:
Grid cells were not eliding properly if they didn't use renderers. This is because non-renderer cells outputted a string (text node) as the direct contents of the flex container.

> `text-overflow: ellipsis` won't work on flex container (`display: flex`). The main reason is that the text node becomes an anonymous flex child and needs `min-width: 0` to behave (or else it won't shrink beyond its content size), but as one can't target a text node with CSS, we need to wrap it, here done with a span.
>
> https://stackoverflow.com/questions/48545178/use-text-overflow-ellipsis-and-flexbox-align-items-center-in-combination

The proposed change here always adds a renderer to our columns, which ensures the contents are wrapped in a span if no renderer provided. Then we can reliably target the containers child for text-overflow.

Are we concerned about performance implications with this? It's interesting that the statement above claims the browser will treat the text node as an anonymous flex child, which to me implies there shouldn't be much impact in making the flex child explicit. We certainly have grids in client apps where *not* using a renderer is the anomaly.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/ag-grid/AgGrid.scss
+++ b/cmp/ag-grid/AgGrid.scss
@@ -180,10 +180,11 @@
     display: flex;
     align-items: center;
 
-    & > span {
+    & > * {
       flex: 1;
       overflow: hidden;
       text-overflow: ellipsis;
+      min-width: 0;
     }
   }
 
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -475,6 +475,8 @@ export class Column {
 
                 refresh() {return false}
             });
+        } else {
+            setRenderer((agParams) => agParams.value);
         }
 
         const sortCfg = find(gridModel.sortBy, {colId: ret.colId});


cmp/grid/columns/Column.js
```diff
@@ -475,6 +475,8 @@ export class Column {
 
                 refresh() {return false}
             });
+        } else {
+            setRenderer((agParams) => agParams.value);
```
amcclain, line 479: Could we please add a comment on this line as to why we are installing a no-op renderer? I can easily imagine us forgetting that there is a difference in the generated markup between this passthrough renderer and no renderer at all (or remembering that there is _some_ reason why we have this but no longer being sure why, or if it applies in some future ag-grid version).

cmp/ag-grid/AgGrid.scss
```diff
@@ -180,10 +180,11 @@
     display: flex;
     align-items: center;
 
-    & > span {
+    & > * {
       flex: 1;
       overflow: hidden;
       text-overflow: ellipsis;
+      min-width: 0;
```
amcclain, line 187: It would also be good to comment this `.ag-cell-value` rule generally to lay out the goals of these styles - I have found myself coming back into stylesheets and not being sure why things are how they are, often I think because we tend to comment less but also because there seem to be fewer opportunities on the CSS side to make the code itself expressive.

In general I _think_ our goals here are to:
* Use `display: flex` on `.ag-cell-value` to support vertical centering of cell contents.
* Set `flex: 1` on the next nested element to ensure it takes up the entire cell width.
* Set the next three rules to ensure it clips and elides strings with ....  (Is `min-width` a trick to ensure that actually happens - something about putting the entire element into a "sized" mode where eliding could occur, even though we're clearly saying as little as possible about sizing?)

Issue #2126, opened 2020-10-13
Dashboard Layout not rounding its values


lbwexler:
Its outputting a silly amount of precision for its pixel values

TomTirapani:
PR here: https://github.com/xh/hoist-react/pull/2130

amcclain:
Fixed via PR above.

PR #2130, opened 2020-10-15
Round relative sizes to 2dp in DashContainer state.

TomTirapani:
See issue https://github.com/xh/hoist-react/issues/2126

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dash/impl/DashContainerUtils.js
+++ b/desktop/cmp/dash/impl/DashContainerUtils.js
@@ -6,7 +6,7 @@
  */
 import {serializeIcon} from '@xh/hoist/icon';
 import {throwIf} from '@xh/hoist/utils/js';
-import {isArray, isEmpty, isFinite, isNil, isPlainObject, isString} from 'lodash';
+import {isArray, isEmpty, isFinite, isNil, isPlainObject, isString, round} from 'lodash';
 
 /**
  * Lookup the DashViewModel id of a rendered view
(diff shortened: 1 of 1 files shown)


amcclain:
I wasn't sure where in the process this rounding was happening. 

Any sense that we should ensure all of our rounded relative sizes still add to 100%  (i.e. by adjusting the last item), or is this all so close enough that it doesn't really matter? Are the width and height values that are being rounded percentage-based at this point, or are they pixel widths?  

Asking more for my own education than anything else. I think I've spent a bit too much time talking about trading allocations recently where questions around "extra pennies" keep popping up. 😄

TomTirapani:
They're percentage sizes at this point - pixel sizes are an extra layer we added on top of GoldenLayout, which deals exclusively in relative sizes.

I *think* since we're just losing precision and not doing any math operations, they'll always add to 100? My understanding is that floating point issues come up when doing math operations, but rounding is pretty safe (but I may be wrong)

lbwexler:
I think that we are not rounding to integers -- still storing 2 dps of precision, so things are going to come out pretty damn near 100%, and almost certainly something that is going to round to 100 when rounded to an integer.
  
So probably not much different than what we have now (which probably also will add to something close to 100 and then round to 100) except just truncating the excess precision.

If this works well in practice, think we should merge it.

amcclain:
Great - I am all for it, and have no interest in holding it up!

My only (purely theoretical) concern was something along the lines of the sum rounding up to 101% of total and ending up triggering horizontal scrollbars or something like that with a tiny overflow. But my guess is that's not really how things work, and again if we ever see such a thing we can address then.

Thanks for the fix - I know we've had similar cases in the past and it's nice to have those specs not carry crazy precision around with them for no reason.

Issue #2148, opened 2020-10-21
Mobile NavigatorModel should support a `track` property


lbwexler:
Its the mobile equivalent to TabContainer, so would be nice to have this convenience as well.

lbwexler:
And we can definitely use it in toolbox

lbwexler:
Thinking about this some more, thing we should go ahead and make TabContainerModel and NavigatorModel just maintain a cache and only track the first time you visit a tab for Page.

Realize that is not what you always want, but think that is more likely to be what you want.

Thoughts?

amcclain:
Yes, I agree - a user swapping between two tabs within a browsing session would ideally not trigger a stream of back-and-forth tracking logs.  Main goal here was to track if a given tab was being visited/used _in general_ - I think that tracking once-per-session would accomplish this and be a net win.

PR #2150, opened 2020-10-21
Mobile NavigatorModel supports page tracking

petradish:
This PR addresses https://github.com/xh/hoist-react/issues/2148
Linked Toolbox PR: https://github.com/xh/toolbox/pull/437

Hoist P/R Checklist
-------------------


lbwexler:
Thanks for jumping on this Petra.  One subtlety here, I think is that we would like to track when you put something 'new' on the navigation stack for the first time, but not when you are just going back.

Its true we actually do this on TabContainer as well, and I now wonder if we can do better than that -- almost want to just track it when it becomes active for the first time.

We can think about it a bit more, but I think the noise from going "back" in the mobile example seems especially egregious to me.

amcclain:
Added linked ticket above for the change to track-once - I think this is orthogonal enough to tackle in a separate pass.

amcclain:
Eh - thought about it again, and actually think we should hold off merging this until we do #2399 (which seems like a reasonably small lift).

cmp/tab/TabContainerModel.js
```diff
@@ -128,7 +129,8 @@ export class TabContainerModel extends HoistModel {
                         category: 'Navigation',
                         message: `Viewed ${isString(title) ? title : id} tab`,
                         // If using routing, data field specifies route for non-top-level tabs.
-                        data: route && route !== 'default' ? {route: route} : null
+                        data: route && route !== 'default' ? {route: route} : null,
+                        oncePerSession: true
```
petradish, line 133: Could we make the 'oncePerSession' flag configurable on TCM and NavigatorModel? I think it could be useful information to track a user's entire navigation history, for traffic data on a particular tab/page or for understanding workflow that require navigation?

mobile/cmp/navigator/NavigatorModel.js
```diff
@@ -48,13 +51,16 @@ export class NavigatorModel extends HoistModel {
     /**
      * @param {Object[]} pages - configs for PageModels, representing all supported
      *      pages within this Navigator/App.
+     * @param {boolean} [track] - True to enable activity tracking of page views (default false).
```
petradish, line 54: Re: comment above: could be a boolean or config, and if set to true, will take the default {oncePerSession: true} flag

lbwexler:
good idea and I think we can definitely expand `track` in the future to
take a partial, if we need it.  But there are a few questions about how we
would do so -- would like to keep this minimal for now.

lbwexler:
Petra/Lee Co-dev

petradish:
Looks good then! Noted about expanding TCM/mobile NM taking a track partial config in the future

PR #2152, opened 2020-10-21
Add about item to app menus

haynesjm42:
Wasn't 100% sure where this item should go in the menu, ended up putting it just before the Logout item, so should be the last item in SSO environments which seemed the most appropriate for where you usually expect an about item to be.

Hoist P/R Checklist
-------------------


mobile/cmp/header/AppMenuModel.js
```diff
@@ -64,6 +66,12 @@ export class AppMenuModel extends MenuModel {
                 action: () => XH.showImpersonationBar(),
                 prepareFn: (item) => item.hidden = !XH.identityService.canImpersonate
             },
+            {
+                icon: Icon.info(),
+                text: `About ${XH.appName}`,
```
amcclain, line 71: Minor point - should we use `XH.clientAppName` here?  I think that might be preferable in some cases where the `appName` is the overall project and might not be familiar to end-users.

PR #2153, opened 2020-10-21
Fix action col width calculation

haynesjm42:
Am not sure why the sass import broke, but it really seemed like an unnecessary complication in this case so opted to just hardcode the padding value in the js code, and added a comment so we remember to keep the 2 places we have that value in sync.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/columns/Actions.js
+++ b/desktop/cmp/grid/columns/Actions.js
@@ -100,7 +100,11 @@ export const actionCol = {
  *      Default small minimal buttons with an icon will be 24px
  * @returns {number} - the width in pixels
  */
-export function calcActionColWidth(count, cellPadding = 5, buttonWidth = 24) {
+export function calcActionColWidth(
+    count,
+    cellPadding = 5,  // This value needs to be kept in sync with $action-col-pad-px
+    buttonWidth = 24
+) {
     // add 1 to cellPadding to account for 1px transparent border in default theme
     return (count * buttonWidth) + ((cellPadding + 1) * 2);
 }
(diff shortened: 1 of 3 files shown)


lbwexler:
I tweaked with an additional comment in the js file as well.  Will leave it to @TomTirapani  or @amcclainfor merge.

TomTirapani:
Do we even need to maintain the sass var? It would be great if we could drop that sass var and the need to sync entirely.

I can't see what purpose it serves - now that grid cells use flexbox for alignment, setting zero padding on `.xh-action-col-cell.ag-cell` still centers the buttons the cell. Am I missing something?

lbwexler:
please feel free to whack it and merge then!

thanks,

Lee

desktop/cmp/grid/columns/Actions.js
```diff
@@ -99,7 +100,11 @@ export const actionCol = {
  *      Default small minimal buttons with an icon will be 24px
  * @returns {number} - the width in pixels
  */
-export function calcActionColWidth(count, cellPadding = Number(actionColPad), buttonWidth = 24) {
+export function calcActionColWidth(
+    count,
+    cellPadding = 5,  // This value needs to be kept in sync with $action-col-pad-px
```
amcclain, line 105: Does the variable mentioned in this comment exist?  I couldn't find it - concerned we risk driving future versions of ourselves a bit crazy... 😃

lbwexler: good catch!  It was there!

Issue #2144, opened 2020-10-21
Tab Switcher Enhancements


lbwexler:
+ Handle overflow gracefully
+ Allow specification of size constraints on tabs (minSize, maxSize) to allow them to compress/expand within limits, or alternatively be a constant size across the container.

Realize the last item it a little under-specified.  Not exactly sure what is possible here and we need to explore.

lbwexler:
Also -- please note that regarding overflow, we should consider the behavior that the intellij IDE uses, where all of the partially offscreen tabs appear as a little dropdown list that you select from and have the tab container scroll to that.  Its actually really nice!

TomTirapani:
Opened a WIP PR for the overflow behaviour here: https://github.com/xh/hoist-react/pull/2162

Could use some polish and testing before merge - at this point, I'm just looking to see if we are happy with the overall user experience. I also intend to fold the tab size constraints into the same PR.

petradish:
Hi Tom! Spoke to Lee about this in the context of our new app-- could you also add support for a bindable tooltip on Tab model? Let me know if I should create a separate ticket for it :)

TomTirapani:
Hey Petra - sure thing, sounds like a great idea and should be straightforward enough to fold in :)

TomTirapani:
I've updated the PRs to include the following changes:
+ TabContainerModel can be passed `switcherProps`, which it passes along to its TabSwitcher.
+ TabSwitcher supports `tabSize`, `tabMinSize` and `tabMaxSize` props.
+ Added bindable TabModel.tooltip and display in the TabSwitcher.

lbwexler:
👍

PR #2162, opened 2020-10-27
TabSwitcher Enhancements

TomTirapani:
A proposed solution based on Intellij IDE behaviour.

See issue: https://github.com/xh/hoist-react/issues/2144
See toolbox PR: https://github.com/xh/toolbox/pull/440

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -64,7 +64,7 @@ export class TabModel {
     }) {
         throwIf(showRemoveAction && XH.isMobileApp, 'Removable Tabs not supported in Mobile toolkit.');
 
-        this.id = id;
+        this.id = id.toString();
         this.containerModel = containerModel;
         this.title = title;
         this.icon = icon;
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -57,7 +57,7 @@ export class TabModel {
         containerModel,
         title = startCase(id),
         icon = null,
-        tooltip= null,
+        tooltip = null,
         disabled = false,
         excludeFromSwitcher = false,
         showRemoveAction = false,
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -6,7 +6,7 @@
  */
 import {XH, HoistModel, managed, ManagedRefreshContextModel} from '@xh/hoist/core';
 import {action, bindable, computed, observable} from '@xh/hoist/mobx';
-import {throwIf} from '@xh/hoist/utils/js';
+import {throwIf, withDefault} from '@xh/hoist/utils/js';
 import {startCase} from 'lodash';
 
 /**
@@ -22,6 +22,7 @@ export class TabModel {
     id;
     @bindable.ref title;
     @bindable.ref icon;
+    @bindable tooltip;
     @observable disabled;
     excludeFromSwitcher;
     showRemoveAction;
@@ -38,6 +39,7 @@ export class TabModel {
      *      container when constructing these models - no need to specify manually.
      * @param {Element} [c.title] - display title for the Tab in the container's TabSwitcher.
      * @param {Element} [c.icon] - display icon for the Tab in the container's TabSwitcher.
+     * @param {string} [c.tooltip] - tooltip for the Tab in the container's TabSwitcher. Defaults to title.
      * @param {string} [c.disabled] - true to disable this tab in the TabSwitcher and block routing.
      * @param {string} [c.excludeFromSwitcher] - true to hide this Tab in the TabSwitcher,
      *      but still be able to activate the tab manually or via routing.
@@ -55,6 +57,7 @@ export class TabModel {
         containerModel,
         title = startCase(id),
         icon = null,
+        tooltip,
         disabled = false,
         excludeFromSwitcher = false,
         showRemoveAction = false,
@@ -68,6 +71,7 @@ export class TabModel {
         this.containerModel = containerModel;
         this.title = title;
         this.icon = icon;
+        this.tooltip = withDefault(tooltip, title);
         this.disabled = !!disabled;
         this.excludeFromSwitcher = excludeFromSwitcher;
         this.showRemoveAction = showRemoveAction;
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -6,7 +6,7 @@
  */
 import {XH, HoistModel, managed, ManagedRefreshContextModel} from '@xh/hoist/core';
 import {action, bindable, computed, observable} from '@xh/hoist/mobx';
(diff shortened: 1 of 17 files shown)


utils/react/Hooks.js
```diff
@@ -66,4 +66,16 @@ export function useOnVisibleChange(fn) {
...
     }, []);
+}
+
+/**
+ * Hook to run a function when component scrolls.
+ *
+ * @param {function} fn
+ * @returns {function} - callback ref to be placed on target component
+ */
+export function useOnScroll(fn) {
```
lbwexler, line 77: so had to go back to hook university a bit to remember how this all works. 

-  Should we be cleaning up after ourselves the way we do with onVisibleChange() and onResize().  That is, keep a ref to the listener function added, and remove it when the node is replaced, or we are unmounted.

- Should we set {passive: true} on this ?  Do we expect fn(e) {} to ever call e.preventDefault()?  I don't think so.

lbwexler: would also be great to document for all 3 of these hooks what arguments get presented to fn()

TomTirapani: > Should we be cleaning up after ourselves the way we do with onVisibleChange() and onResize(). That is, keep a ref to the listener function added, and remove it when the node is replaced, or we are unmounted.

I don't think this is necessary - the event listener is on the element itself, and my understanding is that it will be cleaned up automatically once the element is removed from the DOM. `useOnResize()` is different as it creates a `ResizeObserver` instance that needs to call disconnect().

> Should we set {passive: true} on this ? Do we expect fn(e) {} to ever call e.preventDefault()? I don't think so.

`scroll` is passive by default, so no need to explicitly set it to true. But yes, I agree that we don't expect the callback to call preventDefault, and thus shouldn't support setting `{passive: false}`. `e.preventDefault()` actually has no effect for `scroll` events anyway, because it triggers after the scroll has already happened.

lbwexler: Great!  Thanks for clarifying this!

cmp/tab/TabModel.js
```diff
@@ -64,10 +67,11 @@ export class TabModel {
     }) {
         throwIf(showRemoveAction && XH.isMobileApp, 'Removable Tabs not supported in Mobile toolkit.');
 
-        this.id = id;
+        this.id = id.toString();
```
lbwexler, line 70: how did this come up?

desktop/cmp/tab/TabSwitcher.js
```diff
@@ -28,51 +34,249 @@ export const [TabSwitcher, tabSwitcher] = hoistCmp.withFactory({
...
+// Implementation
+//-----------------
+const overflowMenu = hoistCmp.factory({
+    render({model, tabs, vertical}) {
+        if (isEmpty(tabs)) return null;
+
+        const items = tabs.map(tab => {
+            const {id, title: text, icon, disabled} = tab;
+            return menuItem({
+                icon,
```
lbwexler, line 155: + Should this menu get a width of tabSize/maxTabSize/minTabSize for consistency?  Just thinking about very long tab names blowing this menu up, and getting a consistent.
+ Should these menu items get the tab tooltops?

desktop/cmp/tab/TabSwitcher.js
```diff
@@ -28,51 +34,249 @@ export const [TabSwitcher, tabSwitcher] = hoistCmp.withFactory({
...
+                impl.switcherRef,
+                useOnResize(() => impl.updateOverflowTabs()),
+                useOnVisibleChange(() => impl.updateOverflowTabs()),
+                useOnScroll(() => impl.updateOverflowTabs())
+            );
+        }
+
+        // Create tabs
+        const tabStyle = {};
+        if (isFinite(tabSize)) tabStyle[vertical ? 'height' : 'width'] = tabSize + 'px';
```
lbwexler, line 66: should this always be width, even for vertical tabs? Seems like with the horizontal text we use for tabs, we always want to be manipulating the width.  will that work with the flexing in vertical orientations?

If we specify the new prop as xxxWidth, that also always leaves us the option for specifying the height in the future if we ever decide that makes sense.

lbwexler:
Some additional comments:

1) I think naively just clipping the tab title/icon combo is fine.  BUT we do need to always show the x affordance, even for a tiny tab, at least when the tab is visible.
(Chrome overlays it when the tab is active, otherwise clips it -- we could do something simpler, which is to always show it.)
Realize that this makes things more tricky, but its critical if we are ever going to clip tabs, that we be able to close them

2) Tooltip should take a string, or a function, and return an element.  If a function, function can be passed the tab itself, and the TabContainerModel.  this will be important and allow us to create rich tooltips (like Chrome).

3) Seems like {
            enableOverflow: false,
            minSize: 20
} should cause all the tabs in a horizontal configuation to compress, as in chrome.  Do you agree?

cmp/tab/TabModel.js
TomTirapani: In the dynamic tabs toolbox example, tab ids were ints. When we extract the tab id from the tab element's `data-tabId` attribute it's always a string, so we couldn't match it. TabModel id is documented as a string, and I don't think numeric ids are a realistic use case, so I thought it fine to cast to string here.

desktop/cmp/tab/TabSwitcher.js
TomTirapani: Very good point!

Although in vertical layout, setting the width of the tabs doesn't really make sense - they fill the width of the switcher container.

TomTirapani:
> I think naively just clipping the tab title/icon combo is fine. BUT we do need to always show the x affordance, even for a tiny tab, at least when the tab is visible.

100% agree, and I thought I had this in place, but looking again I can see that its not. Will fix.

> Tooltip should take a string, or a function, and return an element. If a function, function can be passed the tab itself, and the TabContainerModel. this will be important and allow us to create rich tooltips (like Chrome).

Okay - do we have a mechanism for creating rich tooltips? I know we do it for grids / charts, but that's leveraging ag-grid / highcharts - I don't think we have any infrastructure for creating rich tooltips for arbitrary components? Will take a look.

> Seems like {
   enableOverflow: false,
   minSize: 20
} should cause all the tabs in a horizontal configuation to compress, as in chrome. Do you agree?

So this is a potentially tricky change. In order for things to 'compress', we need to use flexbox, but we've only been able to get scrolling and overflow working by eschewing flexbox. Alternatively, we may be able to do it by taking control of sizing with js (e.g. available space / tabs). Let me play around with it.

lbwexler:
+ When the switcher  in overflow mode, and we don't yet have horizontal room for the tooltip (because we only have a few tabs) the rich tooltips will flicker uncontrollably.

+ So still not seeing the x affordance when the tab is clipped.  Are we punting on that one?  I think its pretty important in my use case.

PR #2179, opened 2020-11-15
Support for enhanced option filtering on select

lbwexler:
Hoist P/R Checklist
-------------------


desktop/cmp/input/Select.js
```diff
@@ -145,12 +145,25 @@ export class Select extends HoistInput {
...
+
+        /**
+         * Function to be used for filtering of values for a given query string  input.
+         * Used for filtering of options provided by `options` prop when `enableFilter` is true.
+         * Not to be confused with `queryFn` prop, used in asynchronous mode.
+         *
+         * Provided function should take an option and a query value and return a boolean.
+         * Defaults to a case-insensitive match on word starts.
+         */
+        filterOption: PT.func,
```
TomTirapani, line 157: I know this is the name used by React-Select, but I find the name of this prop unintuitive. Could we use `filterFn` to be consistent with `queryFn`?

lbwexler: Makes sense, since we long ago chose `queryFn` rather than the underlying `loadOptions` property.

cnrudd:
Well, it's a real shame that react-select's createFilter function config options' matchFrom confid only allows 'any' & 'start'.
https://react-select.com/advanced#custom-filter-logic
https://github.com/JedWatson/react-select/blob/397e95337f54074cf0705cb5b06ce8951be72f2a/packages/react-select/src/filters.js#L8

A 'startWord' option would might satisfy the needs that created this PR.

One thing that I see is lost - or broken - by this code change is that the default behavior of matching against label and value is taken away.

https://github.com/JedWatson/react-select/blob/397e95337f54074cf0705cb5b06ce8951be72f2a/packages/react-select/src/filters.js#L14

This default behavior worked well in my toolbox global nav search branch.
If it's not something you want to bring back in this new filterFn, would it be possible to have a way to easily allow the default react-select filtering?

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
 * `TabModel` now supports a bindable `tooltip`, which can be used to render strings or elements
   while hovering over tabs.
 * New `Placeholder` component provides a thin wrapper around `Box` with standardized, muted styling.
+* `Select` now implements more intelligent filtering of options based on keyboard input.  Filtering
```
cnrudd, line 14: "more intelligent" is perhaps too subjective for a document that strives to be factual?

lbwexler:
Thanks for the comments about matching the value.  Definitely something to
think about -- although I am not sure I would expect that to happen by
default, especially if the label is always the visible string.

I agree that "more intelligent" in the documentation leaves us open for
extra criticism. Although in some changes it might be warranted, in this
case "Improved" is probably a more appropriate claim.

lbwexler:
Colin -- if you want the default rs behavior back , the simplest way to get it back would be to do:

`filterFn: createFilter`.    (where createFilter is imported from 'react-select'.)

I  think for something as specialized as the global nav branch filtering, would be more transparent to just go ahead and specify a custom function that does exactly what you want.  I don't think createFilter is likely to be that thing.  For example, the 'start' option doesn't actually work well with the searching in 'value'.

amcclain:
Reviewed/thought about this, and I agree it makes the most sense to match on the label by default. Slightly less surprising if we think about the value as a "behind the scenes" string/identifier that might / might not be what the user expects.  

Can of course imagine cases where matching both is convenient, but then supplying a custom function is pretty straightforward. (If we find ourselves doing this a lot in practice we could export/provide one or two preset implementations.)

I'm happy to tweak the changelog and merge so we can put it to use.

amcclain:
Do we want to make the same change on mobile?  

Less likely to be used there as we [actively discourage](https://github.com/xh/hoist-react/blob/4e34ddeca4ced9c7a7165e591a38b364b0d6fb48/mobile/cmp/input/Select.js#L48) enabling filtering on mobile (due to the keyboard appearing) but worth it still for consistency?

lbwexler:
Yes should be consistent with mobile -- I even looked at mobile, but
somehow convinced myself we didn't support the filtering.  Uggh.

I'll make the change

cnrudd:
Thanks Lee,

this does indeed work to restore the default:
```
import {createFilter} from 'react-select';
...
...
 select({
filterFn: createFilter
})
```

In the case of searchFromNav branch the default `matchFrom` option 'any' was what was being used and continues to feel like the best choice there.

This `searchFromNav` branch is the only place I can recall taking advantage of the "match on value" feature.  It is useful there because the select has grouped options where the value is a string that incorporates the group/parent name, so this lets you search for a group/topic and its children, even if the child labels do not have the group name in them.

Issue #2178, opened 2020-11-06
Cube.info metadata should be observable


amcclain:
I don't see any reason why the optional collection of metadata you can attach to a cube via it's info property is not directly observable. (We are taking extra steps at the app level to sync this to observable fields on a model.)

TomTirapani:
PR here: https://github.com/xh/hoist-react/pull/2182

PR #2182, opened 2020-11-17
Make Cube.info metadata directly observable.

TomTirapani:
See issue #2178

Hoist P/R Checklist
-------------------


amcclain (approved):
This looks good to me - and great that we were able to clean up the trampolining of info to `View`. Although that did make me a bit concerned that we had done that in the first place - i.e. that there is some subtle reason why that was preferable to what we now have here.

@lbwexler do you have any thoughts or recollections?

lbwexler:
yes, this was quite intentional, as the deleted comment indicates.  Idea being that a disconnected view is a snapshot of data at a given time, and you would like to similarly snapshot the info (e.g. timestamps) at that time as well.

Just as a view captures a snapshot of the cube data in 'result' it should capture the immutable state of the info at that time.

TomTirapani:
Makes sense - apologies for missing that subtlety.

I've reverted the change to CubeView, so its info is still a snapshot of the info at the time it was loaded.

lbwexler:
Thanks Tom!

Issue #2142, opened 2020-10-20
Allow specifying a minimum version of Hoist Core


lbwexler:
Would like to have a minimum version of Hoist Core be something that can be specified in Hoist-React.  

App should throw if environment service returns less than that.

PR #2183, opened 2020-11-17
Allow specifying a minimum version of Hoist Core.

TomTirapani:
See issue: #2142

Hoist P/R Checklist
-------------------


amcclain:
Hi - thanks for taking this on - I like the idea and think we should incorporate it.

* My main question is - should we set a default minimum version based on what we know hoist-react to expect/require? I can see how it's non-trivial given that some services (JsonBlobService) might require a given HC version but generally be considered optional in the context of hoist-react, so maintaining the ability to set / override at the app level also seems important.  But providing a default could help keep things sane without requiring any work on the part of app developers.
* We could provide a general method such as `requireHoistCoreVersion(minVersion)` that could be checked in app code as well as Hoist services such as jsonBlobService to cleanly prevent them from initializing, without throwing the entire app.

package.json
```diff
@@ -66,6 +66,7 @@
     "resize-observer-polyfill": "~1.5.1",
     "router5": "~7.0.2",
     "router5-plugin-browser": "~7.0.2",
+    "semver": "^7.3.2",
```
amcclain, line 69: Could you pls. tweak this to use the `~` qualifier, like the others?  That means we opt-in to minor version updates, which I think has been reasonable so far.

lbwexler:
I was actually thinking of this as something that we would use internally within hoist-react, when we added features that required core-support.  Apps would not need to know about it.

It could be as simple as a single line of code in XH.initAsync() that throw after a semver check.  Min version could be just a constant in XH.

I also like the idea of a new service like BlobService, quietly checking the version, and then just refusing to initialize if it doesn't have what it needs.  I think that's a pattern that could be enabled by the easily availability of semver in hoist-react, and good examples of using it to test a minimum version .  But not sure anything else to do in this ticket.

TomTirapani:
Thanks for taking a look guys.

I've updated it to check as part of the XH.initAsync() process, with the min version being a constant in XH. You can no longer set it in the AppSpec.

At the same time, I've also wrapped the semver stuff into util functions, which should make it easier for us to check version at the service / app level if we want to.

lbwexler:
Tom -- I am really sorry for just putting 2+2 together here now, but I think we already have your utils functionality imported and are using it for ag-Grid and Highcharts:

See 
https://github.com/xh/hoist-react/blob/develop/kit/ag-grid/index.js and 
https://github.com/xh/hoist-react/blob/develop/kit/highcharts/index.js

Can we just use a consistent style check in XH?

lbwexler:
Tom and I decided that the *max* concerns here (e.g. new hoist-core breaks hoist-react) are best handled as a *min* check on the hoist core side.  (which we can add at some point in the future)

The calculus is a little different for our internal stuff than the checks against ag-grid, highcharts, etc.

PR #2184, opened 2020-11-17
Add `skipBlank` flag to RestEditor fields.

TomTirapani:
See issue #2134

Hoist P/R Checklist
-------------------

**Pull request authors:** Review and check off the below. Items that do not apply can also be
checked off to indicate they have been considered. If unclear if a step is relevant, please leave
unchecked and note in comments.

- [x] Caught up with `develop` branch as of last change.
- [x] Added CHANGELOG entry, or determined not required.
- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.
- [x] Updated doc comments / prop-types, or determined not required.
- [x] Reviewed and tested on Mobile, or determined not required.
- [x] Created Toolbox branch / PR, or determined not required.

**If your change is still a WIP**, please use the "Create draft pull request" option in the split
button below to indicate it is not ready yet for a final review.

> **Pull request reviewers:** when merging this P/R, please consider using a *squash commit* to
> collapse multiple intermediate commits into a single commit representing the overall feature
> change. This helps keep the commit log clean and easy to scan across releases. PRs containing a
> single commit should be *rebased* when possible.

PR #2186, opened 2020-11-17
Skip empty readonly RestEditor fields when adding

### LINKED ISSUE #2134: RestEditors should be more flexible in hiding metadata fields
We have some baked in handling to remove `lastUpdated` and `lastUpdatedBy` in RestForms when they are empty, but there is no way to extend this behaviour. In the JsonBlobs admin, for example, I'd also like to hide `token` and `dateCreated`.

The simplest solution would probably be to support a `skipBlank` flag on the editor config, and check this in RestFormField:
https://github.com/xh/hoist-react/blob/develop/desktop/cmp/rest/impl/RestFormField.js#L28

We could also consider a more fleshed out solution where editors can support a mode (`all|create|edit`) or similar, to offer more fine grained control over skipping editors at create time.

amcclain:
+ Add assorted doc comments and member type hints on related classes.
+ Fixes #2134
+ Alternate approach to PR #2184

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/rest/RestGridModel.js
+++ b/desktop/cmp/rest/RestGridModel.js
@@ -65,9 +65,12 @@ export class RestGridModel {
     //----------------
     // Properties
     //----------------
+    /** @member {boolean} */
     readonly;
 
+    /** @member {RestGridEditor[]} */
     editors;
+
     toolbarActions;
     menuActions;
     formActions;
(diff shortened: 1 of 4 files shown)


TomTirapani:
I think this makes a lot of sense, and is a clear improvement over https://github.com/xh/hoist-react/pull/2184 - no need to set an explicit flag when the correct behaviour can be derived from `readonly`!

Thanks also for cleaning up the documentation in the classes. 👍

PR #2187, opened 2020-11-17
Fix Summary record handling in tooltip

lbwexler:
Hoist P/R Checklist
-------------------


lbwexler:
This one is a little strange, but think we should truly own the way we do these tooltips and its relationship to our records.

petradish (approved):

TomTirapani (approved):
This makes sense - happy to see it doesn't affect the value passed to tooltipSpec.

Am I right in understanding we never really expect the tooltipValueGetter to be used outside of our tooltip render function? (since people should be using the provided `tooltip`|`tooltipElement` props?)

amcclain:
I believe this was a pre-existing issue and _not_ something introduced by this fix, but FYI I just pushed a tweak/fix in 62daf210d92ea81444615e86a3bfd4d6a1d93ea8 to defend against records with an undefined value throwing (badly!) on hover / tip show.

lbwexler:
great fix!

PR #2197, opened 2020-11-21
Mobx6

lbwexler:
See:  https://github.com/xh/toolbox/pull/445

Note that this branch is off of the existing react17 branch/PR, and incorporates those changes.  I think we should probably upgrade these together for testing purposes.  **I would be fine to just merge this and abandon the react17 branch.**

Some notes, beyond what is in the changelog:   

* I only did *part* of the change to inheritance in our core model/service classes.  I have a plan to bundle up ReactiveSupport/XhIdSupport/ManagedSupport/LoadSupport in to a new base class (e.g. "HoistBase" (?)).  I want to wait on that until we have gotten rid of @HoistComponent and class based components, which still require the mixin approach.
(I can explain why).

* There is a bit of risk in this upgrade that applications might either forget to call makeObservable() in a model constructor when needed, or forget to modify a destroy method().  The latter is the most serious concern, and can cause parts of your app to simple become unreactive without any other warning!  I don't have a great way to mitigate either of these concerns.  One good trick in code review to make sure that whenever we import from '@xh/hoist/mobx' (for the annotations) we also always import makeObservable (which we will also need to call).

Hoist P/R Checklist
-------------------


core/AppSpec.js
```diff
@@ -28,7 +28,7 @@ export class AppSpec {
      *      'MyApp Mobile').
      * @param {boolean} c.isMobileApp - true if the app should use the Hoist mobile toolkit.
      * @param {Class} c.modelClass - root Model class for App, decorated with `@HoistAppModel`.
```
TomTirapani, line 30: This `@HoistAppModel` comment is outdated

core/HoistModel.js
```diff
@@ -4,39 +4,41 @@
...
+        this.initLoadSupport();
+    }
 
+    matchesSelector(selector) {
+        if (isFunction(selector)) return selector.isHoistModel ? this instanceof selector : selector(this);
+        if (isString(selector)) return !!this['is' + selector];
+        return false;
+    }
+}
+HoistModel.isHoistModel = true;
```
TomTirapani, line 44: I'm confused by this line - is this necessary when we the getter on line 32?

TomTirapani: Ah, nevermind - static vs instance lookups

core/mixins/LoadSupport.js
```diff
@@ -45,19 +41,50 @@ export function LoadSupport(C) {
...
+                    });
+            },
+
+            /**
+             * Does this object implement loading?
+             * True if an implementation of doLoadAsync been provided?
+             */
+            implementsLoading: {
+                get() {
+                    return !!this.doLoadAsync;
```
TomTirapani, line 76: Should we use `isFunction()` here?

cmp/form/FormModel.js
```diff
@@ -221,5 +222,6 @@ export class FormModel {
 
     destroy() {
         XH.safeDestroy(this.fieldList);
```
haynesjm42, line 224: Does `@managed` not work on getters? If not, should we provide some kind of `markManagedProp` similar to `markManaged` for dynamically adding a prop to be managed?

haynesjm42 (commented):
I think we need to move away from `destroy()` as a chainable method and add functionality in `ManagedSupport` for tracking custom disposer methods along with the managed properties and instances - which I think will cover any other use-cases where models need to implement a destroy method. Thoughts?

haynesjm42:
We can't just put a call to `makeObservable(this)` in the `HoistModel` ctor? Does that not work because of class initialization order or something? If it doesn't work, could we put that call in a `setTimeout` or is there some reason it absolutely needs to be in the constructor?

Just trying to brainstorm some ways to avoid requiring all models to add a ctor and remember to make that call ... especially since it seems like it will not be necessary forever if they would finally approve a decorators proposal for js

lbwexler:
John -- Believe me, I tried to call makeObservable in Super. That was
originally the dream.  :-)

You are correct -- calling does not work to class initialization order
(which I now understand better than ever, at least for the time being. :-))

I think using a timer would be very dangerous, and a non-starter given that
we often start reactions within constructors.

lbwexler:
Agree with your point about enhancing managed, and writing destroy methods
only when necessary.

Issue #2177, opened 2020-11-06
GroupingChooser (DimensionChooser 2.0)


lbwexler:
DimensionChooser showing its age, especially next to some of our recent work on FilterChooser.  Thinking about:

+ More of a omni-box based approach, like FilterChooser?
+ Drag and Drop,  like our recent client work on column choosing?
+ Use Favorites, rather than History, to be like FilterChooser?

Also thinking to make a new control rather than breaking the API of the existing control.  Not sure about this one.

TomTirapani:
Opened up draft PRs for review and feedback:
https://github.com/xh/hoist-react/pull/2200
https://github.com/xh/toolbox/pull/446

TomTirapani:
<img width="427" alt="Screenshot 2020-11-24 at 13 28 24" src="https://user-images.githubusercontent.com/3017757/100100651-07a2aa00-2e59-11eb-9e28-9b9988357bbb.png">

lbwexler:
This image is enough for me.  Its love at first sight.

PR #2200, opened 2020-11-24
Add GroupingChooser component

TomTirapani:
Opening this up for review and feedback.

See issue: https://github.com/xh/hoist-react/issues/2177
See Toolbox PR: https://github.com/xh/toolbox/pull/446

Hoist P/R Checklist
-------------------


desktop/cmp/grouping/GroupingChooser.js
```diff
@@ -0,0 +1,363 @@
...
+        styleButtonAsInput = true,
+        emptyText = 'Ungrouped',
+        popoverWidth = 250,
+        popoverTitle = 'Group By',
+        popoverPosition = 'bottom'
+    }) {
+        const {editorIsOpen, favoritesIsOpen, value} = model,
+            isOpen = editorIsOpen || favoritesIsOpen;
+
+        // Todo: Tidy this up?
```
lbwexler, line 46: This comment appropriate for most of our code, always. :-)

desktop/cmp/grouping/GroupingChooser.js
```diff
@@ -0,0 +1,363 @@
...
+});
+
+const addDimensionControl = hoistCmp.factory({
+    render({model}) {
+        const options = getDimOptions(model.availableDims, model);
+        if (isEmpty(options) || model.atMaxDepth) return null;
+        return div({
+            className: 'xh-grouping-chooser__add-control',
+            item: select({
+                // By changing the key each time the value changes, we can
```
lbwexler, line 261: good workaround, comment should be `each time the options change`

TomTirapani:
I've updated this branch following a review with Lee last week and opened it up for review

petradish:
I think the GroupingChooser looks great! I like the indentation when the rows are static, but it feels a little too jumpy/snap-in-place-like on drag end, when the left margins recalculate.

lbwexler:
Snap-in effect growing on me.    But Tom had the same concern that you do.

-Lee

lbwexler:
Tom incorporated lots of my feedback, and I am very jazzed about where this landed!  I think it's a good combination of being clear/compact/clean, consistent with filter chooser, and not requiring too many extra clicks to operate.

We discussed the indentation which I was originally a proponent of -- if people felt strongly that it helped we could restore it, but I like the way everything lines up now, and its nice not to lose the valuable horizontal space.

Two things to consider before merging:

+ We will want to eventually support this on mobile.  Should we take the TabContainer approach of having the single cross-platform API.

+ I think that we should support the option to have it set to `null`, with potentially the option to *not* allow that.  It's seems like there are some application uses where that would make sense.  It could even be paired with empty text of 'Ungrouped' to provide an option of no grouping.

desktop/cmp/grouping/GroupingChooser.js
```diff
@@ -0,0 +1,399 @@
...
+                    }
+                }
+            })
+        });
+    }
+});
+
+GroupingChooser.propTypes = {
+    /** Icon for target button. */
+    buttonIcon: PT.element,
```
amcclain, line 91: What about `icon` or `leftIcon` here - the latter would match `FilterChooser`

desktop/cmp/grouping/GroupingChooser.js
```diff
@@ -0,0 +1,399 @@
...
+
+    /** Static text for target button, or null (default) to display current dimensions. */
+    buttonText: PT.node,
```
amcclain, line 94: Do we think we need this? I don't mind it, but wondering if we have any use cases where we don't want the button displaying its value.

desktop/cmp/grouping/GroupingChooser.js
```diff
@@ -0,0 +1,399 @@
...
+
+    /**
+     * Prefix for button text - applied when value not-empty and static text not specified.
+     *      E.g. "Group by" to render "Group by Fund > Trader".
+     */
+    buttonValueTextPrefix: PT.node,
+
+    /** Width in pixels of the target button. */
+    buttonWidth: PT.number,
```
amcclain, line 103: Here I would also vote for simplifying / standardizing to `width` - we can keep the distinct `popoverWidth` prop so this (by process of elimination) would pretty clearly apply to the button.  

⭐  Would also like to ensure we support setting a `flex` prop

cmp/grouping/GroupingChooserModel.js
```diff
@@ -0,0 +1,285 @@
...
+        if (isEmpty(this.availableDims)) return 'All dimensions added';
+        if (this.atMaxDepth) return 'Grouping at max depth';
+        return null;
+    }
+
+    /**
+     * @param c - GroupingChooserModel configuration.
+     * @param {(DimensionSpec[]|CubeField[]|string[])} c.dimensions - dimensions available for
+     *     selection. When using GroupingChooser to create Cube queries, it is recommended to pass
+     *     the `dimensions` from the related cube (or a filtered subset thereof).
```
amcclain, line 68: Thoughts on renaming this `fields` or `fieldSpecs`?  I am *not* sure about this, but wanted to raise the idea in case  people were sympathetic to it and we wanted to take this opportunity to standardize a bit more.

amcclain (commented):
Hi - I still want to look over a bit more carefully, but it's looking great and I only have these few comments - and there the only one I'd like to ensure we get in is support for `flex` and the suggested change to use `width` instead of `buttonWidth`

PR #2206, opened 2020-11-25
Remove remaining class-based components from internal Hoist code

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@
   * Use `switcher: false` to not include a TabSwitcher. (previously `switcherPosition: 'none'`)
   * Use `switcher: {...}` to provide customisation props for the `TabSwitcher`. See `TabSwitcher`
     documentation for more information.
+* The `HoistInput` base class has been removed. Application components extending `HoistInput` should
+use the `useHoistInputModel` hook instead.  This change marks the completion of our efforts to
+remove all internal uses of React class-based Components in Hoist.
 
 ### 🐞 Bug Fixes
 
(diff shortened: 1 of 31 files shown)


lbwexler:
Seperated this out from some of the API disruptions related to mobx6.  Would like to go forward with this.

lbwexler:
Tom and I reviewed this, and it reminded me that we should highlight the change that this relaxes the requirement that any  model for a functional component be a `HoistModel`.

The class-based component did not require that, and I don't think the functional component needs to either.  Not so uncommon for users to want to bind their inputs to simpler observable "models"

cnrudd:
@lbwexler I think something in this merge commit has affected the way refs are handled.

For example I notice that logic like this:
      ```return this.selectRef.current?.reactSelectRef?.current;```

no longer returns the the ReactSelect component.  It just returns the DOM element at this.selectRef.current

If i check out the prior commit: 06554433c0230a543d57354ef93c28bd80522554, the line above works.

lbwexler:
Will fix.  Thanks for catching!

lbwexler:
So this is a great catch.  Am hoping that this is isolated to reactSelectRef, but please read below.

It's true -- the 'refs' for all the inputs were changed -- they now refer to the top level dom object of the component, rather than the component instance itself.  This is just the 'default' for refs for all functional components.  The idea is that there is no longer really a component object available that external instances should be manipulating directly.  It should all be done with props.

Are there other 'properties/methods' on our class based inputs that we were using from the  outside?  Do we poke at their internal methods to get them to close  Did I miss something here? If that is the case, I could restore the existing behavior on refs using `useImperativeHandle` to return the underlying model as the 'ref'.  But I would rather not.

If its just reactSelectRef, would like to go ahead and just make that a new first level prop on the functional component that we pass through.

Might have a similar issue on DateInput, however that apparently has  
     inputRef = createObservableRef();
    buttonRef = createObservableRef();
    popoverRef = createObservableRef();

Agree there is an issue here we have to address ASAP

PR #2208, opened 2020-11-26
Core Refactor + Mobx6

lbwexler:
See https://github.com/xh/toolbox/pull/450
One note, beyond what is in the changelog:

There is a bit of risk in this upgrade that applications might either forget to call makeObservable() in a model constructor when needed, or forget to modify a destroy method(). The latter is the most serious concern, and can cause parts of your app to simple become unreactive without any other warning! I don't have a great way to mitigate either of these concerns. One good trick in code review to make sure that whenever we import from '@xh/hoist/mobx' (for the annotations) we also always import makeObservable (which we will also need to call).

Hoist P/R Checklist
-------------------


core/HoistBase.js
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    //-----------------------
+    // Load Support
+    //-----------------------
+    /**
+     * Does this object implement loading?
+     * True if an implementation of doLoadAsync been provided.
+     */
+    get implementsLoading() {
```
amcclain, line 78: Didn't want to stick with `hasLoadSupport`?  I still like that notion of "support" in that it encompasses a few features that get enabled when `doLoadAsync` has been implemented. (I see we have it still for the section block comment above.)

I will see if I can come up with any alternate suggestions, but I confess I find `implementsLoading` a bit awkward.

core/HoistBase.js
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+
+    /**
+     * Clean up resources associated with this object
+     */
+    destroy() {
+        this._xhManagedProperties?.forEach(p => XH.safeDestroy(this[p]));
+        this._xhManagedInstances?.forEach(i => XH.safeDestroy(i));
+        XH.safeDestroy(this._loadModel);
```
amcclain, line 233: Any preference for marking this as managed above, where it's constructed, vs. explicitly destroying it here? Do we want to encourage one over the other as a general best practice?

core/HoistBaseDecorators.js
```diff
@@ -0,0 +1,93 @@
...
+import {PersistenceProvider} from '@xh/hoist/core';
+
+import {cloneDeep, isUndefined} from 'lodash';
+import {start} from '@xh/hoist/promise';
+
+/**
+ * Decorator to make a property "managed". Managed properties are designed to hold objects that
+ * are created by the referencing object and that implement a `destroy()` method.
+ *
+ * See also (@see HoistBase.markManaged}.
```
amcclain, line 16:
```suggestion
 * @see HoistBase.markManaged
```

core/HoistBaseDecorators.js
```diff
@@ -0,0 +1,93 @@
+/*
```
amcclain, line 1: Should/could we validate (or at least comment) that these decorators can _only_ be used on classes which extend `HoistBase`?  The filename makes that pretty clear, but might be worth stressing in comments and/or having these throw.

core/HoistModel.js
```diff
@@ -4,39 +4,25 @@
...
-            }
-        }
-    });
-}
+export class HoistModel extends HoistBase {
 
+    get isHoistModel() {return true}
 
+}
+HoistModel.isHoistModel = true;
```
amcclain, line 28: ~~This is interesting - what's the purpose of installing this static on the class?~~

For use in the model context selector below, is that right? Worth a comment?

amcclain (commented):
I read as far through as `core/Reactive` tonight - will pick it up tomorrow and eager to do so, but didn't want the comments I had entered so far to be trapped in limbo.

Great stuff, thanks!

CHANGELOG.md
```diff
@@ -2,14 +2,49 @@
...
 ### 🐞 Bug Fixes
-
 * Fix issue where a `View` would create a root row even if there were no leaf rows.
 * Fixed regression in `LeftRightChooser` not displaying description callout.
 
+### 💥 Breaking Changes
+* The `Ref` class, deprecated in v26, has now been removed.  Use `createObservableRef` instead.
+* `@HoistComponent` has been deprecated and moved to `@xh\hoist\deprecated`
+* All `HoistModel` and `HoistService` classes will have to be adjusted as described above.
+
```
amcclain, line 41:
```suggestion
* The `uses()` function (called within `hoistComponent()` factory configs for model context lookups) no longer accepts class names as strings. Pass the class itself (or superclass) of the model you wish to select for your component. This util will now throw if given any string other than "*", making the need for any updates clear.
```

core/modelspec/creates.js
```diff
@@ -17,7 +17,7 @@ import {ModelPublishMode, ModelSpec} from './ModelSpec';
  * for access by all sub-components.
  *
  * The model created will be considered to be 'owned' by the receiving component. If it implements
- * `@LoadSupport` it will be loaded on component mount, and it will always be destroyed on
+ * `LoadSupport` it will be loaded on component mount, and it will always be destroyed on
```
amcclain, line 20: Here's another reference to `LoadSupport` - I'm still leaning towards `hasLoadSupport` as the getter name. (Same in `uses()` and `RefreshContextModel` docs.)

core/modelspec/uses.js
```diff
@@ -57,10 +56,19 @@ export class UsesSpec extends ModelSpec  {
 
     constructor(selector, fromContext, publishMode, createFromConfig, createDefault) {
         super(fromContext, publishMode);
-        throwIf(!selector, 'Must specify selector in uses().');
+        throwIf(
+            !isFunction(selector) && selector !== '*',
+            'A valid Class, function, or "*" is required as a selector in uses().'
```
amcclain, line 61: 👍

core/HoistModel.js
amcclain: I see we have this pattern consistently in place - don't think it needs a comment so much now that I understand it better and have read further down.

Any concerns having a static class property and the instance getter of the same name?

core/refresh/RefreshContextView.js
```diff
@@ -18,7 +19,7 @@ import {useEffect} from 'react';
  */
 export const [RefreshContextView, refreshContextView] = hoistCmp.withFactory({
     displayName: 'RefreshContextView',
-    model: uses('RefreshContextModel', {publishMode: ModelPublishMode.LIMITED}),
+    model: uses(RefreshContextModel, {publishMode: ModelPublishMode.LIMITED}),
```
amcclain, line 22: This ends up being a nice pickup to be able to leave the strings behind and pass the imported class. 👍

data/Store.js
```diff
@@ -28,9 +28,7 @@ import {Record} from './Record';
 /**
  * A managed and observable set of local, in-memory Records.
  */
-@ReactiveSupport
-@ManagedSupport
-export class Store {
+export class Store extends HoistModel {
```
amcclain, line 31: Thoughts on extending `HoistBase` here instead? Any reason to take the extra step of declaring Stores to be Models? 

(No functional difference between the two at this point, right? Unless some other code would be evaluating `isHoistModel` on a store and expecting it to be true in order for some necessary wiring to happen.)

data/cube/Cube.js
```diff
@@ -54,6 +54,7 @@ export class Cube {
         info = {},
         lockFn
     }) {
+        makeObservable(this);
```
amcclain, line 57: So here is a case where we have a class that doesn't extend `Reactive` using the mobx decorators and activating them with `makeObservable()`. That works a-OK - standard mobx right? - but any sense that we should still extend Reactive as a best practice?

data/UrlStore.js
```diff
@@ -5,14 +5,13 @@
...
 
-import {LoadSupport, XH} from '@xh/hoist/core';
+import {XH} from '@xh/hoist/core';
 
 import {Store} from './Store';
 
 /**
  * A store with built-in support for loading data from a URL.
  */
-@LoadSupport
```
amcclain, line 15: This ends up being quite nice, that you can simply implement `doLoadAsync()` to enable all the LoadSupport functionality.

desktop/cmp/panel/impl/dragger/DraggerModel.js
```diff
@@ -24,6 +23,7 @@ export class DraggerModel {
     maxSize = null;
 
     constructor(panelModel) {
+        super();
```
amcclain, line 26: Does the use of `createObservableRef()` require a call to `makeObservable() here`, or is that all encapsulated within the object returned by this function? I.e. that object is itself observable, this class does not need to be...

I think that must be the case, but checking to be sure as the implementation of `createObservableRef` is a bit 🤯

desktop/cmp/rest/RestGridModel.js
```diff
@@ -157,7 +156,7 @@ export class RestGridModel {
     }
 
     /** Load the underlying store. */
-    doLoadAsync(loadSpec) {
+    async doLoadAsync(loadSpec) {
```
amcclain, line 159: Good catch! This looks like something for which we could write a custom linting function, doesn't it?

mobile/cmp/button/RefreshButton.js
```diff
@@ -22,10 +22,10 @@ export const [RefreshButton, refreshButton] = hoistCmp.withFactory({
     model: false,  // For consistency with all other buttons -- the model prop here could be replaced by 'target'
 
     render({model, icon = Icon.sync(), onClick, ...props}) {
-        const refreshContextModel = useContextModel('RefreshContextModel');
+        const refreshContextModel = useContextModel(RefreshContextModel);
 
         if (!onClick) {
-            errorIf(model && !model.isLoadSupport, 'Provided model to RefreshButton must be decorated with LoadSupport.');
+            errorIf(model && !model.implementsLoading, 'Provided model to RefreshButton must be decorated with LoadSupport.');
```
amcclain, line 28: Error message needs updating here

mobx/decorators.js
```diff
@@ -18,10 +18,10 @@ import {action, observable} from 'mobx';
...
     const name = 'set' + upperFirst(property);
     if (!target.hasOwnProperty(name)) {
-        const fn = action.bound(target, name, {
-            value: function(v) {this[property] = v}
-        });
-        Object.defineProperty(target, name, fn);
+        const value = function(v) {
+            runInAction(() => {this[property] = v});
+        };
+        Object.defineProperty(target, name, {value});
```
amcclain, line 24: The new code here seems more clear to me - curious if this was required by mobx changes or just to learn more about when we discuss.

utils/async/PendingTaskModel.js
```diff
@@ -19,15 +19,16 @@ import {isUndefined} from 'lodash';
  * @see Promise#linkTo
  * @see Panel.mask
  */
-@HoistModel
 export class PendingTaskModel {
```
amcclain, line 22: Should we extend `HoistModel` here?

amcclain (commented):
OK read through the rest - looking forward to discussing on Monday!

core/HoistBase.js
lbwexler: Will revert to hasLoadSupport?  Agree its the best name, and on sum, the benefits to recycling the name outweigh any confusion.   (had been a bit concerned about developers that were familiar with it as a mixin flag getting confused)

core/HoistBase.js
lbwexler: yes -- will mark it as managed.  In an earlier iteration, I think it was not possible, but now it makes total sense.  At the very least, it provides a good example of how HoistBase and @managed is supposed to work.

core/HoistModel.js
lbwexler: no concerns with having the two flags be the same -- its a feature -- if you don't really care (or understand) the difference between whether you have an instance or the class, you will still get the right question to the answer you are asking.

It's what we did with the marker flags that were supplied by the mixin handling.

mobx/decorators.js
lbwexler: was required due to a change in how mobx decorators work, specifically `action`

utils/async/PendingTaskModel.js
lbwexler: We cannot.  Util is a lower level package, even lower level than  core.  The PendingTaskModel is actually built in to AppBase, so you would run in to a circular reference problems.

In retrospect the name PendingTaskModel was probably a mistake.  Should have just been 'PendingTask' and there would be no confusion.

Note that PendingTaskModel actually still serves as a `model` for the mask()!  It's a point that we don't document, but there was *never* actually a requirement that the model behind any component is actually a HoistModel, and this PR continues with that flexibility and actually extends it, allowing any 'model' to be looked up/placed in context as well.

 It's a useful feature -- we have one or two examples in toolbox of components that use XH as their model.   I would be surprised if applications do this as well, and I think its an important bit of flexibility -- but something we need to discuss.  HoistModel itself is just an (important) convention in the toolkit right now for easily getting an object that has the support of HoistBase. Let's discuss if we are ok with this, and how to best document it.

desktop/cmp/panel/impl/dragger/DraggerModel.js
lbwexler: definitely not -- the reference to the object returned is not the thing that typically gets observed -- its typically static set once on any object.  

The thing that is observable is ObservableRef's  "current" property....

admin/AppModel.js
```diff
@@ -14,11 +14,10 @@ import {loggingTab} from './tabs/logging/LoggingTab';
 import {monitorTab} from './tabs/monitor/MonitorTab';
 import {userDataTab} from './tabs/userData/UserDataTab';
 
-@HoistAppModel
-export class AppModel {
+export class AppModel extends HoistAppModel {
```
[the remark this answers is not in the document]
lbwexler: glad you brought this up.  Wanted to discuss.  So (weirdly) mobx *requires* that there be annotations on your objects if you are going to call makeObservable(this) with no second argument.  Otherwise it throws.  

We could put a dummy observable that always appears on HoistBase -- or actually just stop making the loadSupport related observables dynamic.  Think the latter is probably the right thing to do. 

also need to make triple-sure that its ok in mobx to be re-registering these observables twice, and calling makeObservable multiple times in the constructor chain

core/HoistBase.js
lbwexler: Let's discuss the naming -- got a bit more complicated with the latest refactor.

core/HoistBaseDecorators.js
lbwexler: Great idea -- did this, and actually caught a case where we were using the decorator and it was doing nothing!  

👍

cmp/form/field/BaseFieldModel.js
```diff
@@ -100,6 +101,8 @@ export class BaseFieldModel {
         readonly = false,
         rules = []
     }) {
+        super();
+        makeObservable(this);
```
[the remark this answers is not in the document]
lbwexler: yes, that was the idea.  Each class responsible for making any observables it adds get registered.

utils/async/PendingTaskModel.js
amcclain: Got it, thanks.  I recall the early discussions around what can qualify as a "model," and I still agree that it's a feature to have that be wide open.  Just as you could have a component that's not a product of the hoistComponent factory, you might have something called a model that doesn't extend `HoistModel`.

cmp/pinpad/PinPadModel.js
```diff
@@ -71,7 +72,7 @@ export class PinPadModel {
     get completedPin() {
         const {_enteredDigits, _pinLength} = this;
         return _enteredDigits.length === _pinLength ?
-            _enteredDigits.toJS().join('') :
+            _enteredDigits.join('') :
```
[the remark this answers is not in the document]
lbwexler: this was necessary for earliest versions of observable arrays in mobx, which were wrapped in a more destructive way.  Think it was obsolete, deprecated, and removed in this release.  (we do not actually observe arrays much in our code -- so this was a rare case)

lbwexler:
Made an additional round of changes that actually separates the LoadSupport nicely from the other utilities.  Think its actually substantially clearer.  Curious to get your opinion on the 'Loadable'/'implementsLoading'.  I know you aren't in favor of the latter, but am now looking for a combination of terms that make sense for the class and the property pair.

Would rather not use the xxxSupport convention on the class. -- better to leave it as a convention for mixins.
And with the class called `Loadable`, seemed weird than that instances would have` .isLoadable=true, hasLoadSupport=false `
Also various fixes related to CR and some documentation updates.

data/UrlStore.js
```diff
@@ -5,19 +5,21 @@
...
 export class UrlStore extends Store {
 
     url;
     dataRoot;
 
+    @managed
+    loadSupport = new LoadSupport(this);
```
amcclain, line 21: Should we add
```
async loadAsync(loadSpec) { this.loadSupport.loadAsync(loadSpec) }
```
to get the public API in there?  I saw in `LogViewerModel` we were calling `store.loadSupport.loadAsync(loadSpec)`. Any reason not to do so that I'm missing?

desktop/cmp/rest/RestGridModel.js
```diff
@@ -157,8 +156,8 @@ export class RestGridModel {
     }
 
     /** Load the underlying store. */
-    doLoadAsync(loadSpec) {
-        return this.store.loadAsync(loadSpec);
+    async doLoadAsync(loadSpec) {
+        return this.store.loadSupport.loadAsync(loadSpec);
```
amcclain, line 160: Another reference we could update if we wished to declare `loadAsync()` on `UrlStore`

utils/js/Decorators.js
```diff
@@ -46,16 +41,11 @@ export function computeOnce(target, key, descriptor) {
...
+        baseFn = isMethod ? value : get;
     return {
         ...descriptor,
         [baseFnName]: function() {
-            let val = this[cacheKey];
-            if (isUndefined(val)) {
-                val = this[cacheKey] = baseFn.call(this);
-            }
-            return val;
+            return getOrCreate(this, '_xh_' + key, () => baseFn.call(this));
```
amcclain, line 48: Noting for the record here that I haven't yet taken the time to really look at / grok these changes. Making this comment for myself more than anything, as they look like interesting uses of this new `getOrCreate` util!

amcclain (approved):
Let's go for it! 

Big refactor here but I think the choices are very solid and I hope we can get even more mileage out of them as we go forward.

PR #2212, opened 2020-11-30
Admin console tab to display MemoryMonitorService snapshots

amcclain:
+ Server-side functionality available in Hoist Core v8.7.
+ See https://github.com/xh/hoist-core/pull/183 for required Hoist Core PR.

Hoist P/R Checklist
-------------------


admin/tabs/monitor/MemoryMonitorModel.js
```diff
@@ -0,0 +1,144 @@
...
+    /** @member {GridModel} */
+    gridModel;
+    /** @member {ChartModel} */
+    chartModel;
+
+    get minVersionWarning() {
+        const minVersion = '8.7.0',
+            currVersion = XH.environmentService.get('hoistCoreVersion');
+        return checkMinVersion(currVersion, minVersion) ? null : `This feature requires Hoist Core v${minVersion} or greater.`;
+    }
```
amcclain, line 22: I know this is a bit elaborate, but I thought it was worth expressing that this is an optional admin feature that could be cleanly disabled if an app has an older version of HC - instead of incrementing the overall required version.

amcclain:
Screen Shot 2020-11-30 at 2 02 34 PM

admin/tabs/monitor/MemoryMonitorModel.js
lbwexler: a lighterweight/declarative way to do this would be just to add to the tab declaration in AppModel.js the following one-liner:

omit: !checkMinVersion(XH.environmentService.get('hoistCoreVersion') 8.7.0).

Just want to avoid cruft in the code from something like this...

PR #2214, opened 2020-12-01
Convert mobile ButtonGroup to functional component

TomTirapani:
When testing the recent input changes, I was surprised to notice that the mobile ButtonGroup (not ButtonGroupInput) was the last non-functional component. This PR addresses that, bringing it inline with the rest of the toolkit.

Hoist P/R Checklist
-------------------

- [x]  Updated doc comments / prop-types, or determined not required.

### DIFF
--- a/mobile/cmp/button/ButtonGroup.js
+++ b/mobile/cmp/button/ButtonGroup.js
@@ -4,38 +4,33 @@
  *
  * Copyright © 2020 Extremely Heavy Industries Inc.
  */
+import {hoistCmp} from '@xh/hoist/core';
 import {hbox} from '@xh/hoist/cmp/layout';
-import {elemFactory, HoistComponent, LayoutSupport} from '@xh/hoist/core';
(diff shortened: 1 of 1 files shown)


mobile/cmp/button/ButtonGroup.js
```diff
@@ -4,38 +4,33 @@
...
+            items,
+            className,
             ...rest,
-            ...this.getLayoutProps(),
-            className: this.getClassName()
+            ref
         });
     }
-}
-
```
lbwexler, line 40: Do we need to mark this with LayoutSupport?  That's been a requirement for FormField, which uses that marking to size its children?  

Agree this is kind of a clunky mechanism -- would be very happy to revisit.

TomTirapani: I don't think so - this is `ButtonGroup`, not `ButtonGroupInput`, which has `ButtonGroupInput.hasLayoutSupport = true`

PR #2221, opened 2020-12-05
Focus Management added to FormModel/FieldModel

cnrudd:
+ This PR adds some basic focus management to Form/FieldModel
+ This PR add a `boundInput` property to FieldModel for imperative access
+ This PR provides standard implementations of focus/blur/select to HoistInputs.

See toolbox PR, for enhanced forms that allow testing of the new Focus Management methods.

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.
- [x] Reviewed for breaking changes: no breaking changes
- [x] Reviewed and tested on Mobile: implemented equivalent changes on mobile HoistInput components.
- [x] Created Toolbox branch: https://github.com/xh/toolbox/compare/focusBlur?expand=1


desktop/cmp/input/Select.js
```diff
@@ -613,7 +634,7 @@ const cmp = hoistCmp.factory(
                 onFocus: model.onFocus,
                 filterOption: model.filterOption,
 
-                ref: model.reactSelectRef
+                ref: composeRefs(model.inputRef, props.inputRef, model.reactSelectRef)
```
lbwexler, line 637: why does the model need two identical refs? why are we introducing a new prop ref here?  this seems to me to be cruft.

lbwexler:
This looks great!

+ So we will want to add documented methods signatures (with empty implementations) for all of these on HoistInputModel itself.

+ Also would like to provide mobile implementations

+ interesting question as to whether the imperative focus() should (or even easily can) display the focus in the way that happens when tabbing in to a control.

desktop/cmp/input/Select.js
cnrudd: Fixed.

cnrudd:
> This looks great!
> 
> * So we will want to add documented methods signatures (with empty implementations) for all of these on HoistInputModel itself. - Done
> * Also would like to provide mobile implementations -Done
> * interesting question as to whether the imperative focus() should (or even easily can) display the focus in the way that happens when tabbing in to a control.  
- I looked into the focus question.  The focus on the slider, switch and checkbox is prevented by this line:
https://github.com/xh/hoist-react/blob/46ac834704dafb67635041d087dd8455ce941acc/kit/blueprint/index.js#L14
Comment it out, and those components show focus.
I think the solution is to give those components css styles similar to the other inputs that bypass blueprints focus styling overrides.  Let me know if you want me to look into that.

cnrudd:
@lbwexler I got the focus outlines working slider handles, checkbox and switch inputs.

cnrudd:
I've added focus & blur to Radio Groups & ButtonGroupInput on Desktop.
I've added focus & blur to Checkbox, Switch, and ButtonGroupInput on Mobile.  (Mobile doesn't have a RadioGroup)
2 Issues on Mobile:  

1. There doesn't seem to be a way to to select the content of a select input, so maybe don't implement the select() method.
2. There doesn't seem to be a way to put onFocus, onBlur event handlers on Onsen Buttons, so I had to put the handlers on the buttonGroup hBox.  These handlers are necessary to get the input model's hasFocus property to change accordingly.

cnrudd:
@lbwexler As discussed, I've reverted the css/style changes, and moved those to this branch:  https://github.com/xh/hoist-react/tree/focusBlur

desktop/cmp/input/ButtonGroupInput.js
```diff
@@ -49,6 +50,33 @@ ButtonGroupInput.hasLayoutSupport = true;
...
+            this.noteBlurred();
+        }
+    };
+
+    focus() {
+        this.enabledButtons[0]?.focus();
+    }
+
+    get enabledButtons() {
+        const btns = ReactDOM.findDOMNode(this.domRef.current)
```
lbwexler, line 73: Yikes!  This seems like a mistake.  This should be built-in.  Thought domRef was already returning a ref to the dom node1

lbwexler: Let's go over these implementations.  Not sure I fully understand this, but I see that it is complex.

desktop/cmp/input/RadioInput.js
```diff
@@ -95,7 +120,15 @@ const cmp = hoistCmp.factory(
...
                 value: opt.value,
-                className: 'xh-radio-input-option'
+                className: 'xh-radio-input-option',
+                onFocus: (e) => {
+                    model.onFocus();
+                    opt.onFocus ? opt.onFocus(e) : null;
+                },
+                onBlur: (e) => {
+                    model.onBlur(e);
+                    opt.onFocus ? opt.onBlur(e) : null;
```
lbwexler, line 130: Are these call backs intended to be a new API on the options (and buttons?).  Also, looks like a copy/paste error here.

desktop/cmp/input/SwitchInput.js
```diff
@@ -42,6 +44,23 @@ SwitchInput.propTypes = {
 //-----------------------
 // Implementation
 //-----------------------
+class Model extends HoistInputModel {
+
+    inputRef = createObservableRef();
```
lbwexler, line 49: Think we need to build this inputRef into the Base class at this point.  Its going to dry up a ton of code.

desktop/cmp/input/RadioInput.js
cnrudd: Fixed the copy/paste error.
The callbacks are not intended to be "new" on the button elements.  They've always been there, haven't they?
They're just necessary to make sure the button group model is notified that it has focus if one of its buttons is focused, same re blur, but a little more complicated with blur, since blurring one button does not mean the buttongroup is blurred, if another button is focused.

lbwexler:
cool -- sorry if I misunderstood there

cnrudd:
@lbwexler very nice refactor!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
   `filterFn` prop.
 * New Admin Console Monitor > Memory tab added to view snapshots of JVM memory usage. (Requires
   Hoist Core v8.7 or greater.)
+* Focus, blur, and select methods now available on the model returned by the `ref` prop on most HoistInputs.
```
amcclain, line 25:
```suggestion
* Focus, blur, and select methods now available on the model returned by the `ref` prop on most HoistInputs.
* `FormModel` and `FieldModel` also gain new focus-related APIs.
```

cmp/input/HoistInputModel.js
```diff
@@ -255,10 +298,10 @@ export class HoistInputModel {
     }
 
     containsElement(elem) {
-        const thisElem = this.domRef.current;
-        if (thisElem) {
+        const {domElem} = this;
```
amcclain, line 301: Looks like this should be `domEl` ?

mobile/cmp/input/Select.js
```diff
@@ -130,6 +130,24 @@ class Model extends HoistInputModel {
...
+    focus() {
+        this.reactSelect?.focus();
+    }
+
+    // not working when enableCreate: true or enableFilter: true.
+    // react-select not putting created content into input.
+    // maybe not necessary, anyways, on mobile?
+    select() {
+        this.selectText();
+    }
```
amcclain, line 149: I agree that we could leave this clearly un-implemented on mobile

amcclain (commented):
Hi - looks great, thank you.  I gave it a careful read here on Github - would suggest a CHANGELOG line re. the Form/FIeldModel updates + I think we have one typo - otherwise I am all for merging.

PR #2222, opened 2020-12-05
Call DOMPurify within parseFieldValue

amcclain:
+ Sanitize incoming String content to defend against possible XSS attacks or other issues with malformed HTML or scripting content before it is rendered into grids or other components

Hoist P/R Checklist
-------------------


amcclain:
Happy to discuss this, of course, as I realize it's a low-level change. But I think it's worth taking - can show you a POC exploit that this will defend against. Even if we're rarely operating with any untrusted/unsanitized inputs, I believe this might be a prudent additional layer of protection and general sanity to apply - in terms of what kind of string content we never expect or want to see in our apps.

PR #2229, opened 2020-12-11
Focus blur styles

cnrudd:
This is related to this PR:  https://github.com/xh/hoist-react/pull/2221

This PR is just the CSS changes to make an input control's focused outline look the same when the input is focused programmatically or via tab.  The resulting look is the more subtle, more compact XH styling.

This PR fixes the issues noticed on Desktop:  Radio, Checkbox, Switch, and Slider Handle.  

It does not yet fix focus issues on Desktop ButtonGroupInput, or Mobile Checkbox, Switch, or ButtonGroupInput.

Hoist P/R Checklist
-------------------


styles/XH.scss
```diff
@@ -51,7 +51,7 @@ body.xh-app {
     margin: var(--xh-pad-px) 0;
   }
 
-  :focus:not(.bp3-input),
+  :focus:not(.bp3-input, .bp3-slider-handle),
```
cnrudd, line 54: Seems like this block could be moved to kit/blueprint/styles.scss since it deals specifically with blueprint overrides.

styles/XH.scss
```diff
@@ -51,7 +51,7 @@ body.xh-app {
     margin: var(--xh-pad-px) 0;
   }
```
cnrudd, line 53: Seems like this block could be moved to kit/blueprint/styles.scss since it deals specifically with blueprint overrides.

cnrudd, line 53: Seems like this block below could be moved to kit/blueprint/styles.scss since it deals specifically with blueprint overrides.

styles/XH.scss
```diff
@@ -51,7 +51,7 @@ body.xh-app {
...
 
-  :focus:not(.bp3-input),
+  :focus:not(.bp3-input, .bp3-slider-handle),
   .bp3-control input:focus ~ .bp3-control-indicator {
```
cnrudd, line 55: Not sure why this class sequence does not override the blue print outline for radios, checkboxes, and switches.
The same class sequence with outline:0 in the new kit/blueprint/styles.scss change does.

kit/blueprint/styles.scss
```diff
@@ -260,6 +260,22 @@ textarea.bp3-input,
   box-shadow: var(--xh-form-field-box-shadow);
 }
 
+body.xh-app {
+  .bp3-control,
+  .bp3-dark .bp3-control,
+  .bp3-dark .bp3-control.bp3-switch {
```
TomTirapani, line 266: This looks like it only targets `.bp3-switch` in dark theme? Is that what we want?

styles/XH.scss
TomTirapani: Yes, I agree - would be best to consolidate these styles in one place

styles/XH.scss
TomTirapani: Were you testing in dark theme? It could be the inclusion of `.bp3-dark` in kit/blueprint/styles.scss

kit/blueprint/styles.scss
```diff
@@ -260,6 +260,22 @@ textarea.bp3-input,
   box-shadow: var(--xh-form-field-box-shadow);
 }
```
TomTirapani, line 262: I'm sure you've tried, but it would be great if we can get the below styles working *without* the high specificity added by the `body.xh-app` wrappers. We have a similar setup with our button styling, and it leads to a kind of arms-race of highly specific rules whenever you want to override the core style in a component or app.

styles/XH.scss
cnrudd: Moved to kit/blueprint/styles.scss

kit/blueprint/styles.scss
cnrudd: I'm pretty sure it's targeting a bp3-control-indicator that is adjacent to a focused input,
when that bp3-control-indicator is contained by any of these:
`.bp3-control`   (control in light mode)
`.bp3-dark .bp3-control`  (control in dark mode)
`.bp3-dark .bp3-control.bp3-switch`.  (switch control in dark mode)

That's what I'm seeing happen.

kit/blueprint/styles.scss
cnrudd: Seems to work today with body.xh-app removed.   Not sure how got lucky today or unlucky before.

kit/blueprint/styles.scss
TomTirapani: Yes... I guess my question was how come we're not targeting switch control in light mode?

cnrudd:
I've looked into putting a focus outline or border on mobile switchInput and checkbox.   Not very easy to get there.  Not sure it is worth the effort.  

The focus outline on the button group container is coming from the user-agent.  Not very pretty, but also not sure worth the effort of cleaning up.   Might be easier to clean up than the mobile switchInput and checkbox, though.

kit/blueprint/styles.scss
cnrudd: Oh, ok.  I've re-tested and find that .bp3-switch specificity is not needed now.  I've removed it.

styles/XH.scss
cnrudd: Now removed entirely.  Does not appear to be needed.

TomTirapani:
Think this landed in a great spot 👍

lbwexler (approved):

Issue #2233, opened 2020-12-17
Component to display an exception


amcclain:
It would be useful to have a component that took a JS `Error` as a prop and rendered its message in a reasonable and consistent way. This could be used to swap out data-driven components as a placeholder when they cannot be rendered due to an error loading or otherwise preparing their content.

- Could take base styling cues from `Placeholder` - centered text in callout box, in this case with some coloring to indicate "error" without being too alarming.
- Could support optional additional props such as `actionText` and `actionFn` to display a link or button below the exception message to run an action - e.g. `Retry` in the case of a timeout. This would be at the discretion of the app/developer.
- Related to #1825 but potentially without the mask/LoadSupport integration - instead this could start with a simpler, standalone component.

I am not sure about the name... `ExceptionMessage`, `ExceptionRenderer`, `ExceptionDisplay` come to mind - could replace `Exception` with `Error` in any of those suggestions.

PR #2237, opened 2020-12-18
Add ErrorMessage Component 

TomTirapani:
Can be used to display an exception or user-friendly error message, along with an optional action.

See issue https://github.com/xh/hoist-react/issues/2233

Hoist P/R Checklist
-------------------


amcclain:
Following up on our Slack chat from today - I pulled this and also created an `errorMessage` Toolbox branch to demo the desktop version and think a bit more about the API - https://github.com/xh/toolbox/commit/308687c4ef39523c4803b7feb96be70b4f6568d6 

A few thoughts:

* Suggest renaming the `message` prop to `error` - could still accept a string but would be less weird when taking an actual `Error` instance, and might often be more convenient when the dev has code like `const {error} = model` and is going to pass that local `error` var into the comp.
* Wondering if we want something like an optional `title` prop that could be used for a leading, app-specific message such as "Failure loading order blotter." The exception message might be a bit arcane or minimal - this would ensure you always provided the user with some useful context, while still allowing the exception to explain _what_ specifically went wrong (depending on how good of a message it carries). If provided, this text could be rendered on a line above the error message, perhaps in bold. (Not sure `title` is the best option as I am _not_ thinking this would look like a panel title, but it was my best suggestion.)
* Suggest replacing the actionIcon/Text props with `actionButtonProps` to allow customizing other aspects of the action button - we took the same approach with our message model and it works well. E.g. minimal, outlined, maybe sizing.
* I agree could be a nice enhancement to show the stacktrace as an optional disclosure, but I don't think we should block on it.  I would rather merge this and get it in, then loop back the first time we found ourselves actually wanting that in a particular app. Not at all opposed to the idea, but I do think it's a bit secondary. For many usages, I expect the message (and maybe optional title/intro text) will be plenty - e.g. a timeout or network failure.

These are all only suggestions, so let's agree and confirm any changes first, then get them in and merge it.

lbwexler:
I generally agree with these changes -- thanks for taking the time to write
it all out! What you have in mind with `error` and `title` is pretty much
what I was thinking.

Not sure about the "action button", and how that would interact with a
built-in affordance to show the exception dialog.  Maybe they are two
separate things, and agree we don't have to block on the exception dialog
affordance.

Note that we already have a `lastLoadException` observable on HoistModel --
that would be a good
candidate for any toolbox example, I think.

Think its the right thing to call it ErrorMessage/error -- even though we
also refer to "Exceptions" more generally in the Hoist API.


-Lee

amcclain:
I was thinking about the action button concept as an additional, optional feature for the app to let the user "do something" right from the errorMessage component to attempt to clear the error and continue. Would expect this to almost always be a retry of a failed load, but could imagine an app resetting back to a default query first or something like that.

I can see the additional "show technical details (stacktrace)" option being useful, although in general I think we're planning to use this component in contexts where a timeout or data unavailable or other common issue is something we've planned for - as opposed to an unhandled runtime exception with our current pop-up dialog and error reporting. Hence my thinking that we could wait a tick on that to see if / when we want to have it available for an app use-case.

lbwexler:
👍

TomTirapani:
Thanks for taking a detailed look and for all the great suggestions.

I've updated the PRs to include Anselm's suggestions above. I'm inclined to agree that we should hold off on the stacktrace until we need it, so haven't included that yet.

Issue #2201, opened 2020-11-24
TreeStyle.BORDERS does not apply to top-level rows without children


haynesjm42:
Looks like the css will only apply if the row has children, which looks a bit off in grids where the root parent rows may not all have children. I think we should just make the borders apply to all of the top-level rows in tree grids (except the first row)

amcclain:
I _think_ the same should be true of the background shading, if enabled. In another app use-case, I found it a bit jarring that this top-level group row is styled more like a child row (because it does not itself have any children):

<img width="305" alt="Screen Shot 2020-11-24 at 12 32 12 PM" src="https://user-images.githubusercontent.com/225806/100148172-39b00e00-2e51-11eb-865a-e3272aade271.png">

febbraiod:
Seems logical that I should have a look at this after having worked on the tree styles originally and in light of recent research on https://github.com/xh/hoist-react/issues/2191.

amcclain:
Thanks, Don!

PR #2249, opened 2021-01-14
Ensure tree styled group rows get hover styles, Fixes: #2201

febbraiod:
+Add new var to accommodate light theme hover color, previously too close to the tree style color

Hoist P/R Checklist
-------------------


styles/vars.scss
```diff
@@ -360,6 +360,7 @@ body {
   --xh-grid-group-text-color: var(--grid-group-text-color, white);
   --xh-grid-total-row-bg: var(--grid-total-row-bg, hsl(207, 87%, 95%));
   --xh-grid-tree-group-bg: var(--grid-tree-group-bg, hsl(207, 87%, 95%));
+  --xh-grid-tree-group-bg-hover: var(--grid-bg-hover, #{mc-trans('blue', '200', .2)});
```
febbraiod, line 363: Think this is the correct approach but not confident in the color itself.

TomTirapani:
This works, but I'm not sure it's the best approach - this puts us in the unfortunate position of having 2 hover colors in light theme and 1 in dark theme.

Couldn't we solve the problem by tweaking the existing light theme colors so that the hover and tree group colors are distinct enough from each other? That would be consistent with the current dark theme behaviour, solve the issue and wouldn't require a new var.

TomTirapani:
I would just like to add that if it proves too difficult to have visually distinct enough highlight across the 3 states (hover, select and tree group), the PR above works great and would be happy to merge it.

Just think we should aim for consistent behaviour across the themes if we can.

febbraiod:
Agree with Tom in principle. Consistent behavior across themes is best. Trouble is we need three distinct but related versions of blue here. So far it seems we always end up with one that is too light or too dark, otherwise two of the three become indistinguishable. Will bang on this a bit more.

febbraiod:
Think slightly darkening tree group highlights does the trick but forces us to also darken the selected row color, then we can leave the highlight color as light as it already is. Dark mode colors remain the same:

Screen Shot 2021-01-19 at 3 09 45 PM

TomTirapani:
This is great! I tweaked the colors a bit so that the order goes (darkest > lightest):

Selection > Hover > Tree Groups

This provides a less jarring experience in side-by-side comparisons to current production toolbox. The color changes are subtle enough to not draw attention to themselves, but distinct enough to solve the original issue.

TomTirapani (approved):

amcclain:
I'm sorry guys as I can see there has been a good amount of back-and-forth on this, but I am finding the new highlight color to be too intense in light mode. It's really bright blue!

* I would prefer to not give up on having `--xh-grid-bg-highlight` default to `--xh-bg-highlight` - that seemed logical to me, with the latter var being the general "here's how we highlight active things" and the former grid var being the primary example of that in action. 
  * I would be happy though to set `--xh-bg-highlight: var(--bg-highlight, #{mc-trans('blue', '200', 0.4)})` - making it one click more opaque / bright than the current `0.5` opacity value and still providing some visible contrast between the three classes of row shading. I know it's subtle but `0.4` looks OK to my eyes, whereas `0.3` pushes it just over the edge...
* Why is `--xh-grid-total-row-bg` set to a one-off RGB value now? Previously it and `--xh-grid-tree-group-bg` had the exact same value. Looking at the results, they still appear to be very very close, indistinguishable to my eyes.
  * Can we set `--xh-grid-total-row-bg: var(--xh-grid-tree-group-bg)` if we intend for them to be the same? I think that makes a certain amount of sense, with the total row as a special kind of "tree" row.
* What role is the transparency playing on all / any of these values? Do we really want/need them to be semi-transparent, or were we just using that to adjust the intensity of the color? (That's part of why I had started to use HSL values for some similar cases like buttons where we were looking to setup these hierarchy of shades/intensities all in the same color family.)

TomTirapani:
Thanks for taking a detailed look. I've pushed some updates to respond to your suggestions.

+ `--xh-grid-bg-highlight` once again defaults to `--xh-bg-highlight` for both themes. Thanks for catching that regression - was something I missed on my review

+ `--xh-grid-total-row-bg` defaults to `--xh-grid-tree-group-bg` for both themes.

+ I replace the transparent colors with a set of HSL values for both themes. You're correct in your assessment that we were using transparency to adjust the intensity, and HSL values are clearly the superior approach. The previous use of transparency is what prompted the one-off rgb value - without it, the layering of semi-transparent colors caused a third color in the total row (see screenshot). Replacing with HSL colors fixes that.

Screenshot 2021-01-22 at 10 45 44

TomTirapani:
Note that the above change to use HSL colors now affects the colors in dark theme too. To my eyes, I can't see much discernible difference when compared side-by-side with current prod toolbox - but we all know how subjective color perception can be :)

amcclain:
Looks great, thanks!  I pushed one tweak (please review) to avoid getting a partial hover highlight on the pinned total row. Great to have the new variable for hovered bg and the three-level distinction between group/hovered/selected

PR #2255, opened 2021-01-18
Support for agGrid 25.0

lbwexler:
+ Use functional components in ag-Grid
+ Remove obsolete experimental flags
+ immutableData = true, for safety, to avoid expensive comparisons.
+ Only forceRefresh() for complexRenderer on rows that have actually changed.
+ New workaround to defeat new ag-Grid sort optimization that was breaking absolute sorting
+ Remove obsolete workaround for flex columns

Hoist P/R Checklist
-------------------


cmp/grid/Grid.js
```diff
@@ -193,13 +194,12 @@ class LocalModel {
...
-        const {model} = this,
-            {useDeltaSort, useTransactions} = model.experimental;
+        const {model} = this;
 
         let ret = {
             model: model.agGridModel,
-            deltaSort: useDeltaSort && !model.treeMode,
-            immutableData: !useTransactions,
+            immutableData: true,
+            rowDataChangeDetectionStrategy: 'IdentityCheck',
```
haynesjm42, line 202: Looks like this is the default value when `immutableData: true` - are we concerned that ag-grid won't always respect that defaulting? Might be better to let ag-grid use whatever it thinks is best with immutable data, especially since this is grid property is pretty special and not really exposed in the standard grid options

cmp/grid/Grid.js
```diff
@@ -387,28 +387,32 @@ class LocalModel {
...
 
                     this.updatePinnedSummaryRowData();
 
-                    const refreshCols = model.columns.filter(c => !c.hidden && c.rendererIsComplex);
-                    if (!isEmpty(refreshCols)) {
-                        agApi.refreshCells({columns: refreshCols.map(c => c.colId), force: true});
+                    if (!isEmpty(transaction?.update)) {
+                        const refreshCols = model.columns.filter(c => !c.hidden && c.rendererIsComplex);
+                        if (refreshCols) {
+                            const rowNodes = compact(transaction.update.map(r => agApi.getRowNode(r.id))),
```
haynesjm42, line 412: Nice optimization here! Is a place I had noted as something we could be more efficient with when I was looking into inline editing performance, though there is a caveat here - if you have cell renderers which depend on other records in the store they will no longer be refreshed when the records they may depend on are updated. Something we might want to note in the changelog if we think there may be apps which depend on that kind of behavior.

cmp/grid/Grid.js
lbwexler: Yea, was trying to just get all the advanced options that might be useful for us in our code, so we could pull all the levers independently.  I will put a comment there, or remove it.

Note we are changing immutableData to *true* but I think that is the right thing to do.  From what I can see,  the *only* thing that immutableData does right now is set the changeDetectionStrategy for when rowData is changed in bulk (which we never do now anyway because we use transactions.).  

We could have considered removing our use of own transactions but to me that still seems the most prudent -- its a supported part of the API -- we can create it efficiently, and we also use part of it for our isRendererComplex bit.

cmp/grid/Grid.js
lbwexler: I think the documented semantics are clear that `rendererIsComplex` supported dependencies on something else on the *row*.  Do you think that is not enough power?   Do we need a `rendererIsReallyComplex` property  ?   


=============================
true if this renderer relies on more than
     *      just the value of the field associated with this column.  Set to true to ensure that
     *      the cells for this column are updated any time the record is changed.  Setting to true
     *      may have performance implications. Default false.

haynesjm42: Ok yeah I think those docs are pretty clear that it is only refreshed when the record itself changes - but if an app was depending on other records it would have worked previously, but now it wont - so maybe just something for us to be aware of in case some renderers stop working?

Something like a `refreshMode` for Columns might be worth considering for more advanced use cases

cmp/grid/Grid.js
haynesjm42: 👍

PR #2258, opened 2021-01-22
View Refactor

lbwexler:
+ Dry up the Row classes with inheritance.
+ Remove the weird inversion of the meta objects
+ Add support for omitFn

Hoist P/R Checklist
-------------------


data/cube/impl/BaseRow.js
```diff
@@ -0,0 +1,150 @@
...
+        this.children?.forEach(it => it.noteBucketed(bucketSpec, bucketVal));
+    }
+
+    // Determine what should be exposed as the actual children in the
+    // row data.  This where we lock, skip degenerate rows, etc.
+    applyVisibleChildren() {
+        let {children, view, data} = this,
+            {lockFn, omitFn} = view.cube;
+
+        if (!children) return null;
```
haynesjm42, line 47: I think this should just be a return

data/cube/impl/BaseRow.js
```diff
@@ -0,0 +1,150 @@
...
+
+        // Remove all children from the data if the query is not configure to include leaves and
+        // this row has leaves as children
+        if (!view.query.includeLeaves && children[0]?.isLeaf) {
```
haynesjm42, line 51: I wonder if we have a potential bug here. Say we are bucketing leaf rows but `includeLeaves` is `false` in the query - and the first child here is a bucket row containing only leaves. In that case we would not remove any of those leaves, right? Should this be a `chilldren.some(it => it.isLeaf)` ?

data/cube/impl/BaseRow.js
```diff
@@ -0,0 +1,151 @@
...
+                return;
+            }
+        }
+
+        // ...otherwise wire up to your own children's data
+        data.children = children.map(it => it.data);
+    }
+
+    isRedundantChild(row) {
+        // TODO:  put this test in application code omitFn instead?
```
haynesjm42, line 82: I could see this being an option on the Query - `cullRedundantChildren` or something

lbwexler:
John and I reviewed this carefully.  Its a little late in the release, but we think its an important cleanup, and gives us the all-important omitFn

data/cube/impl/BaseRow.js
lbwexler: we resolved -- we don't create buckets of leaves

data/cube/impl/BaseRow.js
lbwexler: yep -- let's see how things shake out

PR #2261, opened 2021-01-28
Tweak RelativeTimestamp formatting

amcclain:
+ Moment outputs e.g. "a minute" instead of "1 minute". This creates some awkwardness when the leading number comes and goes - "<1 minute" -> "a minute" -> "2 minutes".

Hoist P/R Checklist
-------------------


amcclain:
I think this should be reliable, given that we already have explicit handling to catch and replace what I believe is the one output format from moment - "a few seconds..." - that would trip this up.

I was finding this annoying in the context of our client dashboard app, where timestamps in short mode where aging through the awkward "1 min ago" -> "a min ago" -> "2 mins ago"

PR #2270, opened 2021-02-01
Use approximate rendered value for StoreFilterField

lbwexler:
UPDATE:  Using rendered value only for dates for now, for performance reasons.

This isn't perfect, and in some cases may not be exactly what you want -- but on balance I think it handles things better --
WYSYG

+ Most Importantly searching for Date strings now works as your would expect.
+ Ledger format numbers work better -- no need to use '-' when you see '('.
+ Performance sees reasonable - tested on 50k, 7 column grid.

One potential downside is that searching for numbers now requires you to type the comma.    This is also probably the most significant behavioral difference.  If we think that is a bug, we could handle numeric fields differently, but I would be inclined to have the search work consistently on the rendered string for all fields.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/store/impl/StoreFilterFieldImplModel.js
+++ b/cmp/store/impl/StoreFilterFieldImplModel.js
@@ -131,12 +131,11 @@ export class StoreFilterFieldImplModel extends HoistModel {
 
         let newFilter = null;
         if (filterText && !isEmpty(activeFields)) {
-            // stripping commas improves number matching
-            const regex = this.getRegex(stripNumberCommas(filterText));
-            const valGetters = flatMap(activeFields, (fieldPath) => this.getValGetters(fieldPath));
-            newFilter = (rec) => valGetters.some(fn => {
-                return regex.test(stripNumberCommas(fn(rec)));
-            });
+            const regex = this.getRegex(filterText),
+                valGetters = flatMap(activeFields, (fieldPath) => this.getValGetters(fieldPath));
+            newFilter = (rec) => valGetters.some(fn => (
+                regex.test(fn(rec))
+            ));
         }
 
         if (filter === newFilter) return;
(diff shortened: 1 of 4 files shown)


cmp/store/impl/StoreFilterFieldImplModel.js
```diff
@@ -203,4 +202,27 @@ export class StoreFilterFieldImplModel extends HoistModel {
...
+        if (!cols) {
+            // lodash get() slower - use only when needed to support dot-separated paths.
+            return field.includes('.') ? (rec) => get(rec.data, field) : (rec) => rec.data[field];
+        }
+
+        // For each visible column, return closest approximation of rendered value.
+        return cols.map(column => {
+            const {renderer, getValueFn} = column;
+            return (record) => {
+                let ret = getValueFn({record, field, column, gridModel});
```
amcclain, line 221: We should call with the `store` as well - that's part of our API:
https://github.com/xh/hoist-react/blob/c13147124f2fbcaf05f39bb0e7680a9bfe06304b/cmp/grid/columns/Column.js#L714

amcclain (commented):
One small comment below - thinking about the question of numbers and commas...

This quickly starts to get into deeper waters re. search engine things like tokenization, stemming, etc. So I am very wary of going down that road at all. But still wondering if we should make some basic substitutions like stripping commas from both query and rendered match candidates....  Maybe that's the only one that would be reasonable, trying to think of how it could go wrong.

I guess I do (annoyingly I am sure 😄) think that it will be considered sub-optimal if we force the user to type the commas. I know that this component's main job is not to search numeric data, but still.  

What do you think about taking commas off the table all around?

cmp/store/impl/StoreFilterFieldImplModel.js
lbwexler: I suppose the weirdest thing that would happen with stripping commas is that you would get *more* matches than you wanted to on text when you were looking for a specific comma (e.g. in a company name) -- but that seems pretty unlikely.  Lets do that.

amcclain (approved):
Great - a few little tweaks and an hour or so later and ready to merge! 😃

Issue #2266, opened 2021-01-29
Mobile Popover Component


TomTirapani:
It would be useful for the mobile toolkit to have an equivalent of Blueprint's popover component.

It would be able to show an arbitrary component in a floating popover, anchored to another 'target' component on screen. It should handle show / hide (including hiding when tapping away), and smart positioning that accounts for viewport overflow.

See mobile's MenuButton, where we are currently using Popper.js (the same lib Blueprint uses) to handle popover menus. This would be a more generalised version of that, with more bells and whistles and a consistent API with desktop

TomTirapani:
See PRs:
https://github.com/xh/hoist-react/pull/2271
https://github.com/xh/toolbox/pull/463

PR #2271, opened 2021-02-01
Add Popover to mobile toolkit.

TomTirapani:
+ Use Popover to implement mobile MenuButton.

See issue: https://github.com/xh/hoist-react/issues/2266
See Toolbox PR: https://github.com/xh/toolbox/pull/463

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -44,7 +44,7 @@ export class TabModel extends HoistModel {
      *      but still be able to activate the tab manually or via routing.
      * @param {boolean} [c.showRemoveAction] - display an affordance to allow the user to remove
      *      this tab from its container.
-     * @param {(Element|Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
+     * @param {(ReactElement|Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
      *      Tab; or function returning react element to be rendered by this Tab.
      * @param {RenderMode} [c.renderMode] - strategy for rendering this tab. If null, will
      *      default to its container's mode. See enum for description of supported modes.
--- a/cmp/tab/TabModel.js
+++ b/cmp/tab/TabModel.js
@@ -44,7 +44,7 @@ export class TabModel extends HoistModel {
      *      but still be able to activate the tab manually or via routing.
      * @param {boolean} [c.showRemoveAction] - display an affordance to allow the user to remove
      *      this tab from its container.
-     * @param {(Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
+     * @param {(Element|Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
      *      Tab; or function returning react element to be rendered by this Tab.
      * @param {RenderMode} [c.renderMode] - strategy for rendering this tab. If null, will
      *      default to its container's mode. See enum for description of supported modes.
--- a/mobile/cmp/menu/MenuButton.js
+++ b/mobile/cmp/menu/MenuButton.js
@@ -5,7 +5,7 @@
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
 import {HoistModel, hoistCmp, useLocalModel} from '@xh/hoist/core';
-import {observable, action, makeObservable} from '@xh/hoist/mobx';
+import {bindable, makeObservable} from '@xh/hoist/mobx';
 import {popover} from '@xh/hoist/mobile/cmp/popover';
 import {Icon} from '@xh/hoist/icon';
 import {button, Button} from '@xh/hoist/mobile/cmp/button';
@@ -62,15 +62,10 @@ MenuButton.propTypes = {
 };
 
 class LocalModel extends HoistModel {
-    @observable isOpen = false;
+    @bindable isOpen = false;
 
     constructor() {
         super();
         makeObservable(this);
-    }
-
-    @action
-    setIsOpen(isOpen) {
-        this.isOpen = isOpen;
     }
 }
--- a/mobile/cmp/menu/MenuButton.js
+++ b/mobile/cmp/menu/MenuButton.js
@@ -36,7 +36,7 @@ export const [MenuButton, menuButton] = hoistCmp.withFactory({
             target: button({icon, ...props}),
             content: menu({menuItems, onDismiss: () => impl.setIsOpen(false)}),
             onInteraction: (nextOpenState) => impl.setIsOpen(nextOpenState),
-            withMask: true,
+            backdrop: true,
             ...popoverProps
         });
     }
--- a/mobile/cmp/menu/MenuButton.js
+++ b/mobile/cmp/menu/MenuButton.js
@@ -6,13 +6,10 @@
  */
 import {HoistModel, hoistCmp, useLocalModel} from '@xh/hoist/core';
 import {observable, action, makeObservable} from '@xh/hoist/mobx';
-import {fragment} from '@xh/hoist/cmp/layout';
+import {popover} from '@xh/hoist/mobile/cmp/popover';
 import {Icon} from '@xh/hoist/icon';
 import {button, Button} from '@xh/hoist/mobile/cmp/button';
-import {createObservableRef} from '@xh/hoist/utils/react';
-import {usePopper} from 'react-popper';
 import PT from 'prop-types';
-import ReactDom from 'react-dom';
 
 import {MenuItem} from './MenuItem';
 import {menu} from './impl/Menu';
@@ -27,35 +24,21 @@ export const [MenuButton, menuButton] = hoistCmp.withFactory({
     render({
         menuItems,
         menuPosition = 'auto',
-        popperOptions,
+        popoverProps,
         icon = Icon.bars(),
         ...props
     }) {
-        const impl = useLocalModel(LocalModel),
-            popper = usePopper(impl.targetEl, impl.menuEl, {
-                placement: menuPositionToPlacement(menuPosition),
-                strategy: 'fixed',
-                modifiers: [
-                    {name: 'preventOverflow', options: {padding: 10}}
-                ],
-                ...popperOptions
-            });
+        const impl = useLocalModel(LocalModel);
 
-        return fragment(
-            button({
-                icon,
-                ...props,
-                onClick: () => impl.open(),
-                ref: impl.buttonRef
-            }),
-            menu({
-                omit: !impl.isOpen,
-                menuItems,
-                style: popper?.styles?.popper,
-                onDismiss: () => impl.close(),
-                ref: impl.menuRef
-            })
-        );
+        return popover({
+            isOpen: impl.isOpen,
+            position: menuPosition,
+            target: button({icon, ...props}),
+            content: menu({menuItems, onDismiss: () => impl.setIsOpen(false)}),
+            onInteraction: (nextOpenState) => impl.setIsOpen(nextOpenState),
+            withMask: true,
+            ...popoverProps
+        });
     }
 });
 
@@ -74,14 +57,11 @@ MenuButton.propTypes = {
         'auto'
     ]),
 
-    /** Escape hatch to provide additional options to the PopperJS implementation */
-    popperOptions: PT.object
+    /** Props passed to the internal popover */
+    popoverProps: PT.object
 };
 
 class LocalModel extends HoistModel {
-
-    buttonRef = createObservableRef();
-    menuRef = createObservableRef();
     @observable isOpen = false;
 
     constructor() {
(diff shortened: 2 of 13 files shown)


TomTirapani:
I've implemented the mobile popover component and added an example page to Toolbox.

The API is based on the commonly used properties of Blueprint's popover API, and supports both controlled and uncontrolled usage.

I'm a bit on the fence about whether or not its worth implementing the little arrow affordance that is seen on desktop. It seems like it would be tricky to implement, and I'm not how much value it adds on the smaller screen.

mobile/cmp/menu/MenuButton.js
```diff
@@ -27,35 +24,21 @@ export const [MenuButton, menuButton] = hoistCmp.withFactory({
...
+        const impl = useLocalModel(LocalModel);
+
+        return popover({
+            isOpen: impl.isOpen,
+            position: menuPosition,
+            target: button({icon, ...props}),
+            content: menu({menuItems, onDismiss: () => impl.setIsOpen(false)}),
+            onInteraction: (nextOpenState) => impl.setIsOpen(nextOpenState),
+            withMask: true,
+            ...popoverProps
```
lbwexler, line 40: nice!

mobile/cmp/menu/MenuButton.js
```diff
@@ -90,49 +70,7 @@ class LocalModel extends HoistModel {
...
-    }
-
-    get targetEl() {
-        return ReactDom.findDOMNode(this.buttonRef.current);
-    }
-
-    get menuEl() {
-        return this.menuRef.current;
+    setIsOpen(isOpen) {
+        this.isOpen = isOpen;
```
lbwexler, line 74: for this private class, would be fine to use bindable.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,196 @@
...
+        className,
+        target,
+        content,
+        isOpen,
+        onInteraction,
+        withMask = false,
+        position = 'auto',
+        popperOptions
+    }) {
+        throwIf(!isValidElement(target), 'Popover requires a `target` component');
```
lbwexler, line 44: does this allow a string?

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,196 @@
...
+    isOpen: PT.bool,
+
+    /**
+     * Callback invoked in controlled mode when the popover open state _would_ change due to user interaction.
+     * Receives (nextOpenState: boolean)
+     */
+    onInteraction: PT.func,
+
+    /** Whether to display a mask behind the popover */
+    withMask: PT.bool,
```
lbwexler, line 108: would it be useful to call the property `mask` and have users be able to specify the mask itself, if they wanted to (or a boolean) we do this elsewheere in the API.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,196 @@
...
+                    onClick: () => impl.setIsOpen(false)
+                })
+            ]
+        });
+    }
+});
+
+Popover.propTypes = {
+    /** Component to display inside the popover */
+    content: PT.element,
```
lbwexler, line 93: Wondering if this should be something that could be parsed as  `elementFromContent`.   We use that in a number of places in the API.  Could use the same for target.

We could also expand `elementFromContent` to take an element itself as well as a Component/function.  Not sure why we never supported that.  Would be useful in other contexts as well.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,196 @@
...
+export const [Popover, popover] = hoistCmp.withFactory({
+    displayName: 'Popover',
+    model: false,
+    className: 'xh-popover',
+
+    render({
+        className,
+        target,
+        content,
+        isOpen,
```
lbwexler, line 38: Seems like the big question is whether we like this style of using controlled mode, vs. exposing a model.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,196 @@
...
+        this._controlledMode = true;
+        if (isFunction(onInteraction)) {
+            this._onInteraction = onInteraction;
+        }
+    }
+
+    @action
+    setIsOpen(isOpen) {
+        if (this._controlledMode) {
+            this._onInteraction?.(isOpen);
```
lbwexler, line 157: this is cool that this use on line 157 works!  I don't think I have ever seen this style

mobile/cmp/popover/Popover.js
TomTirapani: I think being able to specify a mask itself will confuse things... you don't want to show a spinner or tie this to a load model, etc.

The overlay is always there, it's necessary to be able to dismiss the popover. This prop is really just whether or not to color the overlay like a mask.

amcclain: Blueprint uses `hasBackdrop` for what sounds like the analogous setting on their `Drawer` component - that might be a good option here, to distinguish from our other usages of mask.  (I would probably go with `withBackdrop` or simply `backdrop`.)

I don't feel strongly about it - the prop as its named now avoids introducing a new term and is pretty clear - I don't think it's that likely that anyone would interpret it as putting a mask _over_ the popover, although maybe...

mobile/cmp/popover/Popover.js
amcclain: Yes, I like it! Why not, right?

mobile/cmp/popover/Popover.js
amcclain: Agree strongly (and see Tom has already made the change) that `content` should accept the same type of things across its usages in our container components.

And the fact that this is the _component_, as opposed to the tab or navigator models, makes me also agree that our util should support an element itself, so you can use it as you would `items`.  Thanks for making that change, too.

The one concern is that we deliberately had `content` taking a function that returns an element in the model case to support lazy rendering, as you are often configuring/constructing your model well prior to any content actually being rendered (if it ever is).  It seems that we should emphasize that usage so people don't start configuring models with elements and defeat the built-in laziness.  (I hope I am understanding / describing this correctly.)

amcclain: i.e. emphasize (as succinctly as possible) the reasons for wrapping in a closure - when and why you would do that vs. when you would provide an element directly.

Did we think about just going with `items` for the content?  Or is the relationship between `target` and `content` such that neither has a clearer claim to being the child than the other?

cmp/tab/TabModel.js
```diff
@@ -44,7 +44,7 @@ export class TabModel extends HoistModel {
      *      but still be able to activate the tab manually or via routing.
      * @param {boolean} [c.showRemoveAction] - display an affordance to allow the user to remove
      *      this tab from its container.
-     * @param {(Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
+     * @param {(Element|Object|function)} c.content - Hoist Component (class or functional) to be rendered by this
```
amcclain, line 47: We could leave `Element` this out of the signature, even if supported generally by the util, we don't need to declare it supported (in the sense of recommended in a documented way) here.

I also think that `Element` resolves to the DOM object, not the react one - we should confirm - might need to bomb the code with `{ReactElement}` instead, but I'd want to look into it before doing so to ensure I've got that right.

mobile/cmp/menu/impl/Menu.js
```diff
@@ -41,24 +40,18 @@ export const [Menu, menu] = hoistCmp.withFactory({
...
-                item: vbox({
-                    className: 'xh-menu__list',
-                    items
-                })
+        return vbox({
+            ref,
+            className,
+            style: {
+                ...style,
+                ...layoutProps
```
amcclain, line 48: Small thing, but any reason to not let the outer `vbox` accept these props directly at the top level and do this itself?  I think you could then destructure off only `menuItem` and `onDismiss` and pass the rest through.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,191 @@
...
+
+import './Popover.scss';
+
+/**
+ * Popovers display floating content next to a target element.
+ *
+ * The API is based on a stripped-down version of Blueprint's Popover component
+ * that is used on Desktop. Popover is built on top of the Popper.js library.
+ *
+ * @see https://popper.js.org/
```
amcclain, line 25: Sorry to mention this late in the game, but did you catch that Blueprint is starting to phase in a new implementation, based on the latest popper library?  

They are currently bundling it as a separate package - https://blueprintjs.com/docs/#popover2-package.  I just saw this the other day and was thinking that we should be sure to review this - and likely us it - for a desktop dialog component.  But it might well apply here too in terms of taking the latest popper and seeing how BP has changed their implementation / approach.

mobile/cmp/popover/Popover.js
```diff
@@ -0,0 +1,191 @@
...
+
+    _onInteraction
+    _controlledMode = false;
+
+    get targetEl() {
+        return ReactDom.findDOMNode(this.targetRef.current);
+    }
+
+    get contentEl() {
+        return this.contentRef.current;
```
amcclain, line 138: These two `xxxEl` getters are different in their implementation - why is that?

mobile/cmp/popover/Popover.scss
```diff
@@ -0,0 +1,27 @@
+.xh-popover {
+  &__content-wrapper {
+    position: fixed;
+    z-index: 21;
+    background-color: var(--xh-bg);
+    box-shadow: 0 0 20px rgba(0, 0, 0, 0.3);
```
amcclain, line 6: Candidate for a CSS var - specifically, does this work equally well in light and dark modes?

mobile/cmp/popover/Popover.scss
```diff
@@ -0,0 +1,27 @@
...
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    z-index: 10;
+    background-color: transparent;
+
+    &--mask {
+      background-color: var(--xh-mask-bg);
```
amcclain, line 24: If we did update the prop to `withBackdrop` or similar, we could add a distinct `--xh-backdrop-bg` var.

For a client project where we just added a usage of the Blueprint drawer, I found that I wanted distinct shading for the backdrop, which I set to be darker so as to better mask and then set off the modal content in front of it, vs the mask, which I wanted to be lighter so it was less jarring when rendered over e.g. an existing grid full of data that was simply refreshing.

I can show you if helpful in practice - I hadn't appreciated the distinction until just the other day

utils/react/ReactUtils.js
```diff
@@ -22,20 +23,19 @@ export function getReactElementName(obj) {
...
     // Note: Would be more general to look for a *react* component.
     if (content.isHoistComponent) {
         return createElement(content, addProps);
     }
 
-    const ret = content();
+    const ret = isFunction(content) ? content() : content;
     if (ret === null) return null;
     throwIf(!isValidElement(ret),
         'Must specify either a HoistComponent or a function that returns a React Element.'
```
amcclain, line 41: This message needs updating now

mobile/cmp/popover/Popover.js
TomTirapani: I like renaming it to `backdrop` to avoid any potential ambiguity - will make that change now

mobile/cmp/popover/Popover.scss
```diff
@@ -0,0 +1,27 @@
+.xh-popover {
+  &__content-wrapper {
+    position: fixed;
+    z-index: 21;
```
amcclain, line 4: Comment as to why 21?

mobile/cmp/popover/Popover.js
amcclain: (to be clear - I don't think we need to open this all up and swap over - but we should look at some point soon, and I wanted to mention it in case we are struggling at all with popper on mobile and there is a chance the new version is significantly improved)

mobile/cmp/popover/Popover.js
TomTirapani: Maybe I'm misunderstanding, but I thought that even if you provide an element as content for e.g. TabContainer, it won't get rendered until it becomes the active tab? Or, to put it another way, its not rendered until it actually gets added to the DOM in accordance with its RenderMode?

mobile/cmp/popover/Popover.js
TomTirapani: That was cruft from the previous implementation, where target was an onsen button and the ref would return the Component rather than the element. Now that targetEl is the wrapper div, its no longer needed - will clean up. Thanks for catching! 👍

mobile/cmp/popover/Popover.js
lbwexler: Tom -- w.r.t. to your question, I think you have it right --  the element generation should be pretty lightweight.  

According to my understanding, no actual Component rendering, render methods etc. are going to happen until React actually tries to mount that element.  So things would still be quite lazy.   It's really just a matter of if you are ready to execute the code itself to specify the element (vs. doing that a bit later).

So really 
      
`content: () => panel({})`
 `content: panel({})`.  

are not that different.  It seems like it would be hard to get yourself in too much trouble specifying an actual element declaratively.

mobile/cmp/popover/Popover.js
lbwexler: we did look at the new popover for sure, and should move to it.  did not seem like a major change

mobile/cmp/popover/Popover.js
lbwexler: regarding ATM's question about `items` vs. `target` -- it seems like we are trying to mirror the BP desktop API, for better or worse.   If we weren't doing that, I think we might go in a number of different directions.

Interesting to our discussion about returning elements that the new blueprint api actually *requires* a `renderTarget` (vs. target)   Weird that the always rendered portion is so declaratively lazy, and not the content (which may never be displayed).

PR #2273, opened 2021-02-03
GroupingChooser tweaks for usability

amcclain:
+ Auto-expand add dim select upon show.
+ Add text to toolbar buttons to make them easier click targets.
+ If add dim button clicked and only one dim remains to add, add it immediately instead of showing select.
+ Show remove dim button on row hover only, add danger intent for general consistency.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ decorators, in favor of a simpler inheritance-based approach to defining models 
   element. Its API is based on the Blueprint `Popover` component used on desktop.
 * `StoreFilterField` now matches the rendered string values for `date` and `localDate` fields when
   linked to a properly configured `GridModel`.
+* `GroupingChooser` gets several minor usability improvements + clearer support for an empty /
+  ungrouped state, when so enabled.
 
 ### 💥 Breaking Changes
 
--- a/cmp/grouping/GroupingChooserModel.js
+++ b/cmp/grouping/GroupingChooserModel.js
@@ -164,7 +164,11 @@ export class GroupingChooserModel extends HoistModel {
 
     @action
     showAddControl() {
-        this.addControlShown = true;
+        if (this.availableDims.length === 1) {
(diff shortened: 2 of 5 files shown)


amcclain:
After a few weeks of using our new grouping chooser, I would like to propose the included (minor) changes in the name of greater usability. 

I know that we discussed the auto-expand of the "add dim" select when we first released this component and decided to go without it, to avoid any surprising or over-eager interactions. However I must say that with regular use, that extra click to expand the select bugs me every time, and I almost never want to type an option.

I also find that I have a difficult time quickly hitting the Add button - I'd really like to lengthen that text to make it a bigger target, and I thought it made sense to add text to the "Apply" button as well (there you can click the underlying mask to apply, but that's not too obvious).  I don't think we need to be so minimal in these UI elements.  (I did try OK and Save as alternate text options for that button, but Apply still strikes me as the most appropriate.)

amcclain:
Here's what this rev looks like, for convenience.  Also, I didn't apply these changes to mobile as I wanted to get feedback on desktop first, but I think it would make sense to keep consistent and I can make those updates whenever we agree on taking/leaving changes. (Hopefully taking 😄 )

Screen Shot 2021-02-03 at 6 45 56 AM

lbwexler:
I am all for these changes

TomTirapani:
Just took a look at this; I think it's a huge improvement :)

Thanks for making these changes, I'm all for taking them. Only thing I can think of is maybe changing the `Add Level` label to just `Add`, the `Level` there feels a bit redundant.

lbwexler:
I am making the change I mentioned to handle the empty case.

lbwexler:
I applied the changes that made sense on mobile, added the empty display to the popover, and made a few additional tweaks.

Not sure about the new text on the mobile buttons -- seems a little bit clunkier in that context, but will let you be the judge ATM.

Thanks!

PR #2278, opened 2021-02-09
Null out references

lbwexler:
This got a bit more complicated.  Feeling much less sure about it with the needed action wrapper

Difficult to recognize props that are managed and observable (they are rare).  Could do some more research to create the actions selectively, or skip nulling out these.

Hoist P/R Checklist
-------------------


amcclain:
I would agree that the need for the run in action wrapper makes the entire thing less appealing - happy to leave it alone for the moment and simply keep it in mind whenever we're looking at memory usage to see if we can find an instance of retained properties in the wild where this would help.  It was a theoretical concern at the moment, for sure....

Issue #2196, opened 2020-11-21
Support for variable height rows in Grid


cnrudd:
If a dataset has some data in a column that requires multiple lines, agGrid provides an affordance for that:
`column.autoHeight: true`.

In a HR grid column, this would be:
```
agOptions: {
  autoHeight: true
}
```

This column.autoHeight option is ignored by agGrid when a getRowHeight function is also provided to the grid.

In a HR grid, a getRowHeight function is ALWAYS provided to the grid because it is used to set the rowheight according to sizing mode.

Possible solutions:

1. Promote autoHeight to a top level column config property, and do not provide any getRowHeight function if a column has an autoHeight:true setting.  agGrid will calculate rowheights based on the content of that multi-line column.
2. Maintain current behavior, but when agOptions.autoHeight:true is attempted, provide a console warning explaining that Hoist-React grids are really intended for single line data. Developer should consider a DataView component instead of a grid.

AgGrid is doubling down on its `autoHeight` feature.  In v24 it introduces `textWrap:true` which it says should be used in conjunction with `autoHeight:true`.  

This issue was discovered while upgrading a client app that attempts to display multiline data in a column.
I can think of one other client app that already displays comments in a grid column (currently limited to one line and elided).
Developers might expect to be able to take advantage of the textWrap: true, autoHeight:true feature for these comments when they become available in v24.

In v23, `autoHeight:true` all by itself works best when the column's width is preset, not flexed.  On flexed columns the height is incorrectly calculated based on the minimum starting width of the column, not the final flexed width, so empty data needs to be preloaded before real data is called for and loaded, to get ag-Grid to correctly expand the target flex column.
This issue isn't obvious on this demo: https://plnkr.co/plunk/yCrRkXSuZphaA6gm, but can be seen in the client app.

But, in v24, `autoHeight:true` & `wordwrap:true` work very well on a flex column.  See: https://plnkr.co/plunk/N6C2bTo4OoHBjoDQ
and also in the client app, without any need for the above described empty preload workaround.

cnrudd:
Here are branches in HR and Toolbox that demo a solution and its drawback:
https://github.com/xh/hoist-react/compare/agGridAutoHeight?expand=1
https://github.com/xh/toolbox/compare/agGridAutoHeight?expand=1

Drawback being that sizingMode applied to rowHeight must be ignored across whole grid when a column specifies `autoHeight: true`.   This is is not an issue if all rows have multiline content, but could be annoying if a minority of rows do.

[...]

PR #2283, opened 2021-02-11
Allow variable height in grid rows

cnrudd:
This is a WIP for this issue: #2196

Hoist P/R Checklist
-------------------


TomTirapani:
This branch works fine, but as discussed there are a couple of issues that unfortunately I haven't been able to work around.

### Sizing Modes.
I've scoured the agGrid docs, and there doesn't seem to be any way to get `Column.agOptions.autoHeight: true` to work with sizing modes.

If your grid has a `getRowHeight()` function **defined** then autoHeight is ignored, and our grids rely on the default `getRowHeight()` method to apply the sizing modes. Your grid becomes locked into 'standard' sizing mode if any column uses autoHeight.

I've looked to see if there is an API hook we could use to return the height from *within* the `getRowHeight()` function (and then `min()` with the sizing mode height), but no such hook seems to exists. The `AutoHeightCalculator` util that agGrid uses in its source is not exported.

The only approach I could think of would be to extend our autosize service to be able to determine height. It would render a placeholder component, and determine its height given a specified width. We could then use this within the `getRowHeight()` function to return the minimum height for the row. At that point we are effectively taking agGrid out of the equation. That is likely to be a considerable amount of work, so wanted to check if we think it is worth the investment.


### Autosize
Less of an issue than sizing modes, but column autosize will stretch out the column width so that its content is all in one line. While I think this behaviour is **correct**, it may be surprising and lead to _very_ wide columns if `autosizeMaxWidth` is not used. We could consider disabling autosize for columns with `Column.agOptions.wrapText: true`, similar to how we disable it for flex columns.


### Conclusion
At this point, I don't feel comfortable supporting auto row height as a "first-class" member of the grid API unless we can resolve the sizingMode conflict above. If this is required for a client app, this PR allows us to support it as a "hidden" option via `Column.agOptions` as long as the developer is willing to accept the caveats above.

TomTirapani:
Okay -- I managed to get this working while respecting sizing modes.

There is a hook on the `agApi` we can use to calculate the row height, even when the getRowHeight function is implemented. This allows us to compare the auto height against the sizing mode height and choose the largest.

Unfortunately, I'm having difficult getting it to work with DataViews (or rather, with elementRenderers in general), will continue to investigate...

cnrudd:
@TomTirapani,

Awesome that you found the solution to the autoheight on grids.
I've reached the same point/issue on dataview.  AgGrid is Ignoring the elementRender output when calculating autoheight.
Maybe a brief chat tomorrow?

cnrudd:
It looks like the cause of the autoHeight fail for cellRenderers that are React Components is that the agGrid team has not yet developed a solution for psuedo rendering reactjs cell renderer components to get their height.

Here are two variations of the same ag-grid plunker demo, in which I modified the cellRenderer component for the currency column to repeat the currency 3 times in divs.  This should produce a tall cell that expands the rowHeight.
It works in Angular, but does not in React.
Angular: https://plnkr.co/plunk/TPGICc4PNVV2cAUm
ReactJS: https://plnkr.co/plunk/lVwzyPGgDvChQPo0

When I step through the code in the Chrome debugger when running the React version, I see in this file:
webpack:///./node_modules/@ag-grid-community/core/dist/es6/rendering/row/autoHeightCalculator.js?a87b
image
some code that tries to render an Angular component, but does not try to render a React component.

When filtering for autoHeight in the changelog (https://www.ag-grid.com/ag-grid-changelog/),
the entries that come up suggest they have been working through some autoHeight issues for a few versions now.
image

cnrudd:
I found this bug discussed in this issue: https://github.com/ag-grid/ag-grid/issues/4138
TLDR:  it is being tracked in their [pipeline](https://www.ag-grid.com/ag-grid-pipeline/) as AG-4724.
It is not yet fixed, it is marked as "backlog".

TomTirapani:
Thanks for the research Colin.

Disappointing to learn that this is an issue on AgGrid's side, and that its been ongoing for a few versions now with no fix in sight.

I think for the `Grid` and `AgGrid` use cases, this PR should help as long as we make it clear that it doesn't work for `elementRenderers` (which we discourage anyway). We can open this PR and merge if somebody else (@lbwexler?) wants to sanity check it.

For the `DataView` case, unfortunately it doesn't look like variable row heights are going to be an option with AgGrid. That leaves us with a few options:

+ Allow `DataViewModel` to accept a `renderer` in addition to an `elementRenderer`, and render html in my client app. We could make this a requirement of using auto height. Devs will have to decide between using an `elementRenderer` with a fixed item height (good for most cases), or a `renderer` for variable row height.

+ Allow `DataViewModel.itemHeight` to be a closure that receives the record. Developers *may* be able to use this some heuristic to return a different height based on the record's content. Not as good as actually sizing it, but may be sufficient for some cases.

+ If all else fails, I can change my client app to display a vanilla scrollable list of components rather than a `DataView`. I'd lose the buffered rendering and would have to manually implement the grouping, but in this particular use-case there shouldn't be too many rows for performance to suffer.

cnrudd:
@TomTirapani Hi Tom,  I took a run on this idea:
"Allow DataViewModel.itemHeight to be a closure "
and got the dataview to work in these new branches:
https://github.com/xh/hoist-react/tree/dataViewAutoHeight
https://github.com/xh/toolbox/tree/dataViewAutoHeight

It's not totally clean.  Probably can be improved.  Happy to discuss tomorrow.

cnrudd:
@TomTirapani Hi Tom,
Here are new branches that demo an approach that adds the dataviewItem to the DOM to get the height:
https://github.com/xh/toolbox/tree/dataViewReactDomRender
https://github.com/xh/hoist-react/tree/dataViewReactDomRender

This preserves the ag-Grid virttualization feature.

cnrudd:
Todos from CR with Lee:

1. make autoHeight:true a HoistGridColumn prop that sets agOptions{autoHeight:true, wrapText:true}
2. add warning if a column has autoHeight:true and an elementRenderer component (won't work)
3. move comment test to a test grid in Admin

PR #2287, opened 2021-02-12
Add popoverClassName prop to mobile Popover

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/mobile/cmp/popover/Popover.js
+++ b/mobile/cmp/popover/Popover.js
@@ -37,6 +37,7 @@ export const [Popover, popover] = hoistCmp.withFactory({
         onInteraction,
         backdrop = false,
         position = 'auto',
+        popoverClassName,
         popperOptions
     }) {
         const impl = useLocalModel(() => new LocalModel()),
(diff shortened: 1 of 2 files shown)


mobile/cmp/popover/Popover.js
```diff
@@ -71,7 +72,7 @@ export const [Popover, popover] = hoistCmp.withFactory({
                                 ref: impl.contentRef,
                                 style: popper?.styles?.popper,
                                 className: 'xh-popover__content-wrapper',
-                                items: elementFromContent(content)
+                                items: elementFromContent(content, {className: popoverClassName})
```
lbwexler, line 75: This is going to overwrite any className specified on the content, no?  

I suppose this is not the worst thing in the world, but the fact that it will do it even if it is undefined seems harsh. :-)

TomTirapani: Thanks for catching!

For some reason, I thought that `elementFromContent()` would merge classNames rather than overwrite them, like when you have a className in the HoistComponent.

I've changed it to apply the class name to the outer wrapper instead - should be much safer and doesn't modify the content in unexpected ways.

lbwexler:
Thanks Tom!  Nice!

Issue #2289, opened 2021-02-12
TextInput does not pass DOM props onto the HTML input element


cnrudd:
This can cause inconveniences such as needing to wrap the TextInput component in a DIV to size it correctly or add event handlers like onPaste.

lbwexler:
Interesting idea, but am wondering why we would do this for just TextInput.  Seems like something we would to do consistently for all inputs (or at least most inputs)

For example, wouldn't the user using this feature expect NumberInput to work as well?

PR #2290, opened 2021-02-12
Fix: TextInput does not pass DOM props onto the HTML input element

cnrudd:
This fixes issue: https://github.com/xh/hoist-react/issues/2289
Which was raised most recently by @petradish when coding an onPaste event handler.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/TextInput.js
+++ b/desktop/cmp/input/TextInput.js
@@ -119,7 +119,7 @@ const cmp = hoistCmp.factory(
     ({model, className, ...props}, ref) => {
         const {width, flex, ...layoutProps} = getLayoutProps(props),
             // remove layoutProps, props not recognized as DOM elements, and props not accepted on BP's inputGroup component
-            {commitOnChange, enableClear, onCommit, selectOnFocus, textAlign, ...rest} = getNonLayoutProps(props);
+            {bind, commitOnChange, enableClear, onCommit, selectOnFocus, textAlign, ...rest} = getNonLayoutProps(props);
 
         const isClearable = !isEmpty(model.internalValue);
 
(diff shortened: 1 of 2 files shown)


desktop/cmp/input/TextInput.js
```diff
@@ -117,32 +117,34 @@ class Model extends HoistInputModel {
...
+                // These styles get passed onto the HTML INPUT element, 
+                // they are not placed on the DIV created by Blueprint's InputGroup component.
                 style: {
-                    ...props.style,
+                    ...rest.style,
                     ...layoutProps,
-                    textAlign: withDefault(props.textAlign, 'left')
+                    width: withDefault(width, 200),
+                    flex: withDefault(flex, null),
+                    textAlign: withDefault(textAlign, 'left')
```
amcclain, line 147: I am confused by some of these being passed here in the `inputGroup.style` block and then also in the style block below for the outer wrapper div we are installing.  (We have spent a lot of time on these inputs and layout issues in the past, so I'm on the lookout for any possible issues or regressions.)

- Does it make sense to specify flex on the input? Is it in a flex parent container?
- OK to have the width of the input and the width of the outer wrapper be exactly the same? Could the wrapper have padding or borders or anything else?
  - Note also that there is an `InputGroup.fill` prop - if we are deciding the wrap in a div and manage the width there, perhaps this should always be set to true for the inner input group?

These longstanding sizing/layout questions have always been a pain point for me, and I hate to have to bring them into this PR.  If we can use this to clarify them and get a better handle on the layouts, that would be excellent. If we are introducing more uncertainty, however, perhaps there is a way to restructure this change to get the other props passed through that we want without needing to rock the boat re. layout....

lbwexler:
Closing this based on coversation above.  Significant risk for little gain here.

PR #2294, opened 2021-02-16
Increase specificity of group leaf indent override

### LINKED ISSUE #2293: Add more specificity for tree grid leaf indentation
I can't reproduce this in toolbox, but in one of our client apps, the group row leaf indentation specified by ag-theme-balham.css file takes precedence over Hoist's overrides in Grid.scss. I'm not sure exactly why this is happening in our client app vs toolbox, but I did notice that we are leaving out the .ag-row-level-{i} class selector. PR for possible fix forthcoming.

petradish:
Resolves #2293 
+ Dry up .ag-row-group-indent selector
+ $i loops from 1-20 (only those indices are supported by ag-grid)

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -109,49 +109,17 @@
 .xh-grid--hierarchical:not(.xh-data-view) {
   // Generate indentations for tree grids with hierarchical data.
   .ag-ltr {
-    .ag-row-group-indent-0 {
-      padding-left: calc(0 * var(--xh-grid-tree-indent));
-    }
+    @for $i from 1 to 20 {
+      .ag-row-group-indent-#{$i} {
+        padding-left: calc(#{$i} * var(--xh-grid-tree-indent));
+      }
 
-    .ag-row-group-indent-1 {
-      padding-left: calc(1 * var(--xh-grid-tree-indent));
+      .ag-row-level-#{$i} {
+        .ag-row-group-leaf-indent {
+          margin-left: var(--xh-grid-tree-indent);
+        }
+      }
     }
-
-    .ag-row-group-indent-2 {
-      padding-left: calc(2 * var(--xh-grid-tree-indent));
-    }
-
-    .ag-row-group-indent-3 {
-      padding-left: calc(3 * var(--xh-grid-tree-indent));
-    }
-
-    .ag-row-group-indent-4 {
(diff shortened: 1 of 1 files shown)


TomTirapani:
This looks and works great. Did you test it in your client app to see if it resolves the issue? Curious to understand why its manifesting in the client app but not in toolbox 🤔 

DRYing up the rule generation with the for loop is great. Interesting that we used to do that way, and then we changed to listing them out in this commit: https://github.com/xh/hoist-react/commit/258966ea2b3c70c06a94c4bb9ae8cf9a4dd80b5d . @amcclain - can you remember the rationale behind that change?

petradish:
Hi Tom! Yup, tested in the client-app and works well :) I wonder if it is the way webpack loads in the css files? I also removed the extra className on the grid thinking that might be changing the specificity, but that did not change the way the browser was interpreting the css.

amcclain:
Hi - thanks for taking another run at this pesky little issue - getting these grid levels reliably indented has always been a bit of a pain point for us.

@petradish I think that you're right about the difference in bundling resulting in the different ordering of these rules within the bundled stylesheet - at the same specificity, whichever one comes last wins - what a tricky bug to detect!

Re. why we moved away from the SASS approach - I am not positive, but I think it's because we wanted to use a CSS var to control the indent level, and maybe it never occured to me that we could use SASS to dynamically create the rules _and_ embed the var reference within them. That's great.

I pushed one change to setup rules for idx 0 - without that, you would get uneven indentation on any top-level rows that might not have children, e.g. here - EU should line up with US:

Screen Shot 2021-02-17 at 3 29 34 PM

Holding off on merging tonight as I'd like to look a bit more and ensure we're good. I'm on the fence about the nested inner `.ag-row-group-leaf-indent` rule being generated for each level. We actually don't need that many rules - we're just looking for more specificity. Is this the best way to get that here, and in general what approach do we want to take to make it more likely our rules will "win" without themselves being really difficult to override.  E.g. we could wrap all the `.xh-grid` rules in an outer `body.xh-app` rule. Would that be more reliable and generalizable to other components where we need to override existing rules from the source library?

amcclain:
Good morning - I'm going to merge this now to get this fix in, as it's clearly an improvement. Would still like us to think about the approach to specificity, both in this specific case and a bit more generally.

PR #2297, opened 2021-02-17
Async Selection methods

lbwexler:
See https://github.com/xh/toolbox/pull/467

+ Improve reactions in StoreSelectionModel
+ Misc cleanups

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/activity/clienterrors/ClientErrorsModel.js
+++ b/admin/tabs/activity/clienterrors/ClientErrorsModel.js
@@ -188,10 +188,8 @@ export class ClientErrorsModel extends HoistModel {
             });
 
             gridModel.loadData(data);
+            await gridModel.selectFirstAsync({ifNoSelection: true});
 
-            if (!gridModel.hasSelection) {
-                await gridModel.selectFirstAsync();
-            }
         } catch (e) {
             gridModel.clear();
             XH.handleException(e);
--- a/admin/tabs/activity/clienterrors/ClientErrorsModel.js
+++ b/admin/tabs/activity/clienterrors/ClientErrorsModel.js
@@ -12,7 +12,6 @@ import {HoistModel, managed, XH} from '@xh/hoist/core';
 import {fmtDate, fmtSpan} from '@xh/hoist/format';
 import {Icon} from '@xh/hoist/icon';
 import {action, bindable, comparer, observable, makeObservable} from '@xh/hoist/mobx';
-import {wait} from '@xh/hoist/promise';
 import {LocalDate} from '@xh/hoist/utils/datetime';
 import moment from 'moment';
 
@@ -190,8 +189,9 @@ export class ClientErrorsModel extends HoistModel {
 
             gridModel.loadData(data);
 
-            await wait(1);
-            if (!gridModel.hasSelection) gridModel.selectFirst();
+            if (!gridModel.hasSelection) {
+                await gridModel.selectFirstAsync();
+            }
         } catch (e) {
             gridModel.clear();
             XH.handleException(e);
--- a/admin/tabs/activity/clienterrors/ClientErrorsModel.js
+++ b/admin/tabs/activity/clienterrors/ClientErrorsModel.js
@@ -188,7 +188,7 @@ export class ClientErrorsModel extends HoistModel {
             });
 
             gridModel.loadData(data);
-            await gridModel.selectFirstAsync({ifNoSelection: true});
+            await gridModel.preSelectFirstAsync();
 
         } catch (e) {
             gridModel.clear();
--- a/cmp/ag-grid/AgGridModel.js
+++ b/cmp/ag-grid/AgGridModel.js
@@ -5,7 +5,7 @@
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
 import {HoistModel} from '@xh/hoist/core';
-import {action, bindable, observable, makeObservable} from '@xh/hoist/mobx';
+import {action, bindable, observable, makeObservable, computed} from '@xh/hoist/mobx';
 import {throwIf, apiDeprecated} from '@xh/hoist/utils/js';
 import {
     cloneDeep,
@@ -106,8 +106,9 @@ export class AgGridModel extends HoistModel {
     /**
      * @returns {boolean} - true if the grid fully initialized and its state can be queried/mutated
      */
+    @computed
     get isReady() {
-        return !isNil(this.agApi);
+        return !isNil(this.agApi) && !isNil(this.agColumnApi);
     }
 
     /**
--- a/cmp/dataview/DataViewModel.js
+++ b/cmp/dataview/DataViewModel.js
@@ -115,7 +115,7 @@ export class DataViewModel extends HoistModel {
     }
 
     // Getters and methods trampolined from GridModel.
-    // Explicit trampolining to aid code-editor, future docs.
+    // Explicit trampolining to aid tooling and docs.
     get store()                 {return this.gridModel.store}
     get empty()                 {return this.gridModel.empty}
     get selModel()              {return this.gridModel.selModel}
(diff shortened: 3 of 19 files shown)


data/StoreSelectionModel.js
```diff
@@ -105,16 +102,11 @@ export class StoreSelectionModel extends HoistModel {
...
-            run: (records) => {
-                const storeIds = records.map(it => it.id),
-                    selection = this.ids,
-                    newSelection = intersection(storeIds, selection);
-                if (selection.length !== newSelection.length) {
-                    this.select(newSelection);
-                }
-            }
+            track: () => store.records,
+            run: () => remove(this.ids, id => !store.getById(id))
```
amcclain, line 109: By default `store.getById()` does not respect filtering - believe that makes this change problematic as it would no longer cull records excluded via filter from the selection.

Along those same lines, should the call to getById within the records getter should pass true for its `respectFilter` arg? 

Also was unsure if the new comment was as intended - "not mutating the ref" meaning not setting `this.ids` to an new array? Because now we _are_ mutating the array itself. I assume this relates to use of `observable.ref` on the property.

data/StoreSelectionModel.js
```diff
@@ -35,29 +35,26 @@ export class StoreSelectionModel extends HoistModel {
...
     get isEmpty() {
-        return this.ids.length === 0;
+        return this.count === 0;
     }
 
     /** @return {number} - count of currently selected records. */
-    @computed
     get count() {
-        return this.ids.length;
+        return this.records.length;
```
amcclain, line 57: Maybe silly, but would this be a good place to use `@computed` given that the records can change without the count changing? Would that avoid possible re-renders of things that only wanted to depend on `count` or `isEmpty` directly?

data/StoreSelectionModel.js
```diff
@@ -35,29 +35,26 @@ export class StoreSelectionModel extends HoistModel {
...
-    get singleRecord() {
-        const ids = this.ids;
-        return ids.length === 1 ? this.store.getById(ids[0]) : null;
-    }
-
     /** @return {Record[]} - currently selected Records. */
-    @computed
+    @computed({equals: comparer.shallow})
     get records() {
         return compact(this.ids.map(it => this.store.getById(it)));
```
amcclain, line 41: Checking to ensure I understand - will this re-compute if the recordset / records change, even if `this.ids` does not?

I am thinking it will because `store.getById()` is accessing the recordSet and therefore mobx knows this depends on it. And then the `comparer.shallow` will detect if any of the selected records themselves have actually changed (given that they might have been left unchanged and would therefore be reused across store updates).

amcclain (commented):
Hi - a few questions and discussion items below - happy to talk over whenever we have a chance, probably much easier than trying to type out replies!

cmp/grid/GridModel.js
```diff
@@ -443,22 +449,39 @@ export class GridModel extends HoistModel {
...
+     * Any selected records that are hidden because their parent rows are collapsed will first
+     * be revealed by expanding their parent rows.
+     *
+     * This method imposes a minimal delay to allow the underlying grid implementation to
+     * render all pending data changes.
      */
-    ensureSelectionVisible() {
-        const {records} = this.selModel,
-            {agApi} = this;
+    async ensureSelectionVisibleAsync() {
```
cnrudd, line 459: The changes to ensureSelectionVisibleAsync in this branch do not fix an issue tested 
in this toolbox branch: https://github.com/xh/toolbox/tree/gridEnsureSelectionVisible2
on this page http://localhost:3000/admin/tests/grid by clicking on the "Select and Go" button.

That button changes the ag-grid row selection via the agrid api and then calls ensureSelectionVisible.

In this branch the new ensureSelectionVisibleAsync does not wait for the store's selection count to catch up to the grid's selection count.  So, it finds no selections.

This "not checking Grid rows selected may be a deliberate choice.  Perhaps to try to guide the user to use Hoist functions to make selections.  But then maybe the method docs should point this out, or the name might be changed to indicate this, ie: ensureRecordSelectionVisibleAsync

cmp/grid/GridModel.js
```diff
@@ -443,22 +449,39 @@ export class GridModel extends HoistModel {
...
+            indices = [];
+
+        // 1) Expand any selected nodes that are collapsed
+        records.forEach(({id}) => {
+            let rowNode = agApi.getRowNode(id);
+            while (rowNode.parent && !rowNode.parent.expanded) {
+                agApi.setRowNodeExpanded(rowNode.parent, true);
+                rowNode = rowNode.parent;
+            }
+        });
```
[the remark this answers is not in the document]
lbwexler: ah hopefully that is it -- great catch of some additional complexity here

cmp/grid/GridModel.js
```diff
@@ -443,22 +449,39 @@ export class GridModel extends HoistModel {
...
+        records.forEach(({id}) => {
+            const rowNode = agApi.getRowNode(id);
+            if (!isNil(rowNode?.rowIndex)) indices.push(rowNode.rowIndex);
         });
 
         const indexCount = indices.length;
-        if (indexCount != records.length) {
-            console.warn('Grid row nodes not found for all selected records - grid data reaction/rendering likely in progress.');
+        if (indexCount !== records.length) {
+            console.warn('Grid row nodes not found for all selected records.');
```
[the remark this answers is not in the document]
lbwexler: no, we can probably remove

data/StoreSelectionModel.js
lbwexler: good catch here as well -- most important in the records getter -- not sure how I missed this.

data/StoreSelectionModel.js
lbwexler: Yes!  If we think there are transitions where records goes from one collection to another with the same size, that would indeed cut down on noise.

data/StoreSelectionModel.js
lbwexler: yep

lbwexler:
Thanks for the careful look colin!

admin/tabs/activity/clienterrors/ClientErrorsModel.js
```diff
@@ -190,8 +189,9 @@ export class ClientErrorsModel extends HoistModel {
 
             gridModel.loadData(data);
 
-            await wait(1);
-            if (!gridModel.hasSelection) gridModel.selectFirst();
+            if (!gridModel.hasSelection) {
+                await gridModel.selectFirstAsync();
```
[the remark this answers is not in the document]
lbwexler: thinking about this more -- 

gridModel.selectFirstAsync({ifNoSelection:true})
gridModel.selectFirstIfNeededAsync({})
gridModel.initSelectionAsync();
gridModel.ensureSelectionAsync();  (too much like ensureSelectionVisibleAsync(), for sure)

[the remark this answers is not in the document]
amcclain: I like `ifNoSelection` as an option to the one method

cmp/ag-grid/AgGridModel.js
```diff
@@ -106,8 +106,9 @@ export class AgGridModel extends HoistModel {
     /**
      * @returns {boolean} - true if the grid fully initialized and its state can be queried/mutated
      */
+    @computed
```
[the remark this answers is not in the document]
TomTirapani: My understanding is that without @computed, things observing `isReady` will trigger when either of the underlying observables change. With @computed, they will trigger when the **resulting** boolean change. So without computed, you may get more reactions than you want. But I may be wrong... @lbwexler...? :)

cmp/dataview/DataViewModel.js
```diff
@@ -125,13 +125,19 @@ export class DataViewModel extends HoistModel {
     get groupBy()               {return this.gridModel.groupBy}
     get sortBy()                {return this.gridModel.sortBy}
```
TomTirapani, line 126: ^ *Very* minor OCD thing, but we should probably tab these over so they align with the methods below

cmp/grid/Grid.js
```diff
@@ -495,9 +479,9 @@ class LocalModel extends HoistModel {
     }
 
     columnStateReaction() {
-        const {agGridModel} = this.model;
+        const {model} = this;
         return {
-            track: () => [agGridModel.agApi, agGridModel.agColumnApi, this.model.columnState],
+            track: () => [model.agApi, model.agColumnApi, model.columnState],
```
TomTirapani, line 484: Should this track `[model.isReady, model.columnState]`?

cmp/grid/impl/ColumnHeader.js
```diff
@@ -124,10 +123,10 @@ class LocalModel extends HoistModel {
     _lastTouchStart = null;
     _lastMouseDown = null;
 
-    constructor({gridLocalModel, xhColumn, column: agColumn}) {
+    constructor({gridModel, xhColumn, column: agColumn}) {
```
TomTirapani, line 126: Great that we were able to get rid of this weirdness 👍

PR #2308, opened 2021-02-22
New Class for LoadSpec, with additional properties

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/refresh/LoadSpec.js
+++ b/core/refresh/LoadSpec.js
@@ -23,6 +23,8 @@
  */
 export class LoadSpec {
 
+    get isLoadSpec()    {return true}
+
     /** @member {number} - index of the associated load on this object.  0 for the first load. */
     loadNumber;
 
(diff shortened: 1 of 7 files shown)


core/refresh/LoadSpec.js
```diff
@@ -0,0 +1,63 @@
...
+    owner;
+
+    get isLoadSpec()    {return true}
+
+    get isLatest()      {
+        return this === this.owner._lastRequested;
+    }
+
+    get isObsolete()    {
+        return this.owner._lastSucceeded.loadNumber > this.loadNumber;
```
haynesjm42, line 48: `_lastSucceeded` could be `null` on the owner here

core/refresh/LoadSpec.js
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     *  @private
+     *  Not for application use.  Used by LoadSupport.
+     */
+    constructor({loadNumber, isRefresh, isAutoRefresh, owner}) {
+        this.loadNumber = loadNumber;
```
haynesjm42, line 56: Where is `loadNumber` ever being passed in by calling code? Should we just set it in the ctor here based on `owner._lastRequested`?

core/refresh/LoadSpec.js
lbwexler: great catch

core/refresh/LoadSpec.js
lbwexler: fixed

amcclain (approved):
Looks good to me, thanks!  Please verify that the changes + one fix I made look correct and complete.

Any opportunities to check for stale loads in Hoist or Toolbox? Would be great to have an example to point to.  (Thinking Toolbox could support a dedicated tab on LoadSupport, given that it is both really important/powerful but also not all that obvious - but we certainly should not block on that.)

PR #2309, opened 2021-02-22
Unescape encoded characters after DOMPurify sanitization

petradish:
DomPurify.sanitize will escape characters '<', '>', '&' when it interprets '<' character as an opening angle bracket. 
This could be problematic for records that might contain this character in one of its fields during an update. Upon creation, characters are saved to the db correctly unescaped, but it will return to the client escaped -- and if left unchanged, its new escaped value will be written to the db.

This PR ensures that these characters are unescaped after the sanitize method is run.

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Field.js
+++ b/data/Field.js
@@ -9,7 +9,7 @@ import {XH} from '@xh/hoist/core';
 import {isLocalDate, LocalDate} from '@xh/hoist/utils/datetime';
 import {withDefault} from '@xh/hoist/utils/js';
 import equal from 'fast-deep-equal';
-import {isDate, isString, toNumber, isFinite, startCase} from 'lodash';
+import {isDate, isString, toNumber, isFinite, startCase, unescape} from 'lodash';
 import DOMPurify from 'dompurify';
 
 /**
@@ -66,7 +66,10 @@ export function parseFieldValue(val, type, defaultValue = null, disableXssProtec
     if (val === undefined || val === null) val = defaultValue;
     if (val === null) return val;
 
-    if (!disableXssProtection && isString(val)) val = DOMPurify.sanitize(val);
+    if (!disableXssProtection && isString(val)) {
+        val = DOMPurify.sanitize(val);
+        val = unescape(val); // Unescape characters (i.e. '<', '>', '&', ''','"') after sanitization
+    }
 
     const FT = FieldType;
     switch (type) {


amcclain:
Hi - can you please confirm if/how this does not undo or weaken any of the protection offered by the sanitization?

The main advantage in my mind of using dompurify is that we delegate the responsibility for identifying what is and is not a vector for an XSS or other attack to the security library, which we count on doing its job better than we could ourselves. 

I'm just looking at this PR in isolation and haven't researched, but my thinking is that dompurify is either removing or modifying additional content from the strings it sanitizes, in which case they could still be broken, or it is not, in which case I worry this could simply revert the protections it offers.

Thanks in advance for any more info or thoughts here.

petradish:
No worries - here are some examples from the DOMPurify [readme](https://github.com/cure53/DOMPurify):

```
DOMPurify.sanitize('<svg><g/onload=alert(2)//<p>'); // becomes <svg><g></g></svg>
DOMPurify.sanitize('<p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p>'); // becomes <p>abc</p>
DOMPurify.sanitize('<math><mi//xlink:href="data:x,<script>alert(4)</script>">'); // becomes <math><mi></mi></math>
DOMPurify.sanitize('<TABLE><tr><td>HELLO</tr></TABL>'); // becomes <table><tbody><tr><td>HELLO</td></tr></tbody></table>
```

The default `sanitize` will strip out malicious HTML (examples 1-3) or modify bad HTML tags that are not malicious (example 4). I think in its implementation, the opening angle bracket triggers a recursive call that escapes subsequent characters, and after the sanitization is complete, known HTML tags are converted back to their original form, leaving all other 'unmatched' brackets and escaped characters as is.

```
DOMPurify.sanitize('<p>2 & 3 > 1</p>'); // becomes <p>2 &amp; 3 &gt; 1</p>
DOMPurify.sanitize('1 < 2 & 3'); // becomes 1 &lt; 2 &amp; 3

\\ Whereas, because there is no opening < bracket:
DOMPurify.sanitize('2 & 3 > 1'); // becomes 2 & 3 > 1
```

I think if we trust `sanitize` to have done its job, any remaining escaped characters should be safe, but they need to be unescaped.

petradish:
Also, somewhat related, I'm not sure how to get around the sanitization for email addresses stored as strings of this format:  APP <APP@client.com> (in this case, the sanitize method will strip the entire email)

lbwexler:
Hi Petra -- I had the same question as ATM.

Did you look at the following customization docs?

https://github.com/cure53/DOMPurify#can-i-configure-dompurify

PR #2318, opened 2021-02-28
Store performance

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -13,7 +13,6 @@ import {
     castArray,
     defaultsDeep,
     differenceBy,
-    reduce,
     isArray,
     isEmpty,
     isNil,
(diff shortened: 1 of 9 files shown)


data/Store.js
```diff
@@ -786,21 +798,56 @@ export class Store extends HoistBase {
...
+            const {name} = field,
+                updateVal = update[name];
+            if (!isUndefined(updateVal)) {
+                const val = field.parseVal(updateVal);
+                if (!shareDefaults || val !== field.defaultValue) {
+                    ret[name] = val;
+                }
+            } else {
+                const existingVal = data[name];
+                if (!isUndefined(existingVal)) {
```
amcclain, line 827: Is there any value in also checking if `!sharedDefaults`? 

I believe that, in that (default) case we can count on no `undefined` values being present in existing `data` - i.e. `data` will be complete, with a field for all keys and no `undefined` values for those keys. 

That said, it still might not be a bad idea for symmetry and to ensure that `ret` always has the full set of fields when shared defaults not in use.

amcclain (commented):
Looks good - one minor note/question below, but happy to get this in to try it out in practice. 

If/when we promote from `experimental` we can look at adding `Record.getData()` or similar for the fully populated object and could take a stab at fully documenting the caveats around enumerable keys on `data`, but this looks like what we wanted to start trying it all out.

data/Store.js
lbwexler: great idea, done

PR #2331, opened 2021-03-06
errorMessage returns null if error not passed in.

cnrudd:
Hoist P/R Checklist
-------------------

- [x] Created [Toolbox branch / PR](https://github.com/xh/toolbox/pull/469)


lbwexler:
I think this is a great improvement -- especially useful in JSX scenarios that Colin is looking at where omit is not available.

But shouldn't the test be !(isNil(error) || error === false).

Seems like the empty string ought to still show, and any otherwise null item

cnrudd:
@lbwexler 

Is the test you are proposing:
`if (isNil(error) || error === false) return null;`


In any case, the outcome: "empty string ought to still show"  does allow for showing an empty styled error panel, which my proposed test did not allow for.  I guess that is a good option to provide.  One more option never hurts.   
But maybe we could add a prop to make that more explicit?  "omit" ?  "showEmpty" ? so that the error prop is not overloaded with functionality?

lbwexler:
Can you guys try to take a look at this?  Tried to synthesize some of the improvements discussed above in to something a *bit* more general/flexible.

desktop/cmp/error/ErrorMessage.js
```diff
@@ -18,31 +18,52 @@ import './ErrorMessage.scss';
  */
 export const [ErrorMessage, errorMessage] = hoistCmp.withFactory({
     className: 'xh-error-message',
-    render({className, error, title, actionFn, actionButtonProps}, ref) {
+    render({className, error, message, title, actionFn, actionButtonProps}, ref) {
+        if (!error) return null;
```
cnrudd, line 22: If `error` is '' (empty string)  this will return null.
This behaviour conflicts with the `error` proptype doc.

Same note for the mobile cmp.

desktop/cmp/error/ErrorMessage.js
```diff
@@ -18,31 +18,52 @@ import './ErrorMessage.scss';
...
-    /** If provided, will render an action button which triggers this function,
-     preconfigured with the text 'Retry' (see `actionButtonProps`) */
+    /**
+     *  Message to display for the error.
+     *  Defaults to the error, or any 'message' property contained within it.
+     */
+    message: PT.oneOfType([PT.element, PT.string]),
+
+    /**
+     * If provided, will render an action button which triggers this function,
```
TomTirapani, line 64: Comment should state that the `actionFn()` receives the `error` argument

desktop/cmp/error/ErrorMessage.js
```diff
@@ -57,22 +78,21 @@ const titleCmp = hoistCmp.factory(
...
 
 const actionButton = hoistCmp.factory(
-    ({actionFn, actionButtonProps}) => {
+    ({actionFn, actionButtonProps, error}) => {
         if (!isFunction(actionFn)) return null;
         return button({
             text: 'Retry',
             minimal: false,
-            onClick: () => actionFn(),
+            onClick: () => actionFn({error}),
```
TomTirapani, line 95: Should the argument here just be `error` (not an object)?

I can see the object being useful if we think we're going to extend this, but not sure what else we would pass. As it stands, I think an object with one property is more unexpected than just passing the error.

desktop/cmp/error/ErrorMessage.js
```diff
@@ -18,31 +18,52 @@ import './ErrorMessage.scss';
...
-    /** Error to display. Either an exception, an element or a string */
-    error: PT.oneOfType([PT.instanceOf(Error), PT.element, PT.string]).isRequired,
+    /**
+     *  Error to display. If null or undefined this component will not be displayed.
+     */
+    error: PT.oneOfType([PT.instanceOf(Error), PT.object,  PT.string]),
 
-    /** Optional title to display above the error. Either an element or a string */
+
+    /** Optional title to display above the label. */
```
TomTirapani, line 54: Comment should be `above the message`

lbwexler:
great points tom and colin.  will fix up now

PR #2342, opened 2021-03-12
Fix overeager scrolling in TreeMap/Grid combo.  Related TreeMap cleanups

lbwexler:
+ Related and Unrelated tweaks

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/treemap/TreeMap.js
+++ b/desktop/cmp/treemap/TreeMap.js
@@ -108,6 +108,7 @@ class LocalModel extends HoistModel {
         const chartDependencies = () => ([
             this.chartRef.current,
             model.highChartsConfig,
+            model.algorithm,
             model.data,
             XH.darkTheme
         ]);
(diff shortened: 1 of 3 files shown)


desktop/cmp/treemap/TreeMap.js
```diff
@@ -103,24 +104,23 @@ class LocalModel extends HoistModel {
...
 
-        // Render HighChart when chartElem container ready in DOM, a relevant observable on the
-        // model API changes, and/or the source data array itself changes.
+        // Render HighChart when chartElem container ready in DOM, or dependencies updated
+        const chartDependencies = () => ([
+            this.chartRef.current,
+            model.highChartsConfig,
+            model.data,
+            XH.darkTheme
+        ]);
```
TomTirapani, line 113: We're missing the dependency on the `model.algorithm` here. The chart needs to be redrawn when the layout algorithm changes

lbwexler:
I built it in to the dependency for  the model's `data` itself
intentionally.  Thought that was more straightforward, and that it is
unlikely to change.

TomTirapani:
But, it has nothing to do with how we process the data? It's essentially part of the highcharts config that we've lifted to be a bindable. Don't see how its more straightforward to rebuild the data in order to redraw the chart, when we can just redraw the chart?

desktop/cmp/treemap/TreeMap.js
lbwexler: restored

PR #2343, opened 2021-03-12
TreeMap & SplitTreeMap resizing

TomTirapani:
Mask TreeMap & SplitTreeMap when doing 'significant' resizing, to hide the rendered whitespace.

Use a heuristic to determine if the amount of whitespace is 'significant'. Whitespace is deemed to be significant if it extends beyond 50px in either direction.

Hoist P/R Checklist
-------------------


lbwexler:
Tom and I reviewed.  There is a regression to the label clipping/hiding that he will look at next week.
We will also change the name of the model flag to `isMasking`

lbwexler (commented):
Approved, pending fix to regression, rename

TomTirapani:
I spent a while looking into the label visibility regression.

Our previous masking approach included hiding the chart SVG with CSS. However, it seems that any labels that are reused after the masking are then unable to be sized - their dimensions are always zero, and I couldn't figure out a way to get the true dimensions.

I tried a different approach to the masking, where we instead show a mask *over* the chart rather than modifying the chart styles. This approach does not modify the chart component at all, and thus restores the label visibility functionality.

lbwexler:
👍

PR #2350, opened 2021-03-17
TilingContainer

TomTirapani:
+ Use TilingContainer to render status monitor tiles

Screenshot 2021-03-17 at 16 33 33

Hoist P/R Checklist
-------------------


cmp/tile/TilingContainerModel.js
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * Copyright © 2021 Extremely Heavy Industries Inc.
+ */
+import {HoistModel} from '@xh/hoist/core';
+import {observable, computed, action, makeObservable} from '@xh/hoist/mobx';
+import {isEmpty} from 'lodash';
+
+/**
+ * Model for a TilingContainer.
+ *
```
lbwexler, line 13: Thinking about how this relates to DataView, and whether we would want to bind it to a store.....

cmp/tile/TilingContainerModel.js
```diff
@@ -0,0 +1,219 @@
...
+
+        // Generate all possible tile layouts, from single column > single row.
+        for (let i = 1; i <= count; i++) {
+            const cols = i,
+                rows = Math.ceil(count / cols);
+
+            layouts.push(this.generateLayout(cols, rows, false));
+        }
+
+        // Choose layout with the best (i.e. lowest) score
```
lbwexler, line 131: looks like a job for lodash maxBy

cmp/tile/TilingContainerModel.js
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * Load a new and complete dataset.
+     *
+     * Data is an array of objects, models, etc. For each entry, a tile will be rendered using
+     * `content` and sized and placed according to the layout.
+     *
+     * @param {*[]} data - source data to load
+     */
+    @action
+    loadData(data) {
```
lbwexler, line 105: wonder if this is really just a simple setter, if it might be more straightforward to call it setData?   

On the point about a store again, you could imagine it would be real nice to support setting a store here, and having an internal computed actually dereference (and observe) the actual records.

cmp/tile/TilingContainer.js
```diff
@@ -0,0 +1,60 @@
...
+    className: 'xh-tiling-container',
+
+    render({model, className}, ref) {
+        ref = composeRefs(
+            ref,
+            useOnResize(dimensions => model.onResize(dimensions), {debounce: 100})
+        );
+
+        const {data, emptyText} = model;
+        if (isEmpty(data)) {
```
lbwexler, line 34: should still wire up the ref, even if its empty.

cmp/tile/TilingContainerModel.js
```diff
@@ -0,0 +1,219 @@
...
+    desiredRatio;
+    spacing;
+    minRatio;
+    maxRatio;
+    minWidth;
+    maxWidth;
+    minHeight;
+    maxHeight;
+
+    @computed.struct
```
lbwexler, line 47: might put `layout` and `onResize` in special block above implementation, indicating they are for calling by the component only, not applications.  Might change name of `onResize` to something like `noteDimensions()` (rename less important)

TomTirapani:
Thanks for taking a look Lee - will make some updates based on your suggestions.

I really like the idea of being able to bind in to store records. I wouldn't want to lose the ability to render an arbitrary collection of objects (I can imagine that being a common use-case), but could easily be persuaded. Can also imagine it being useful to be able to supply _any_ observable directly, allowing it to key off `MonitorResultsModel.results` directly.

Would love to go over the `data` part of the API with you today if you have time.

lbwexler:
One different option with this is to skip the dedicated model, and just have component take the configuration as props.  In some ways this makes sense, because its really layout only -- and in particular there is no selection to be manipulated and no real programmatic interface on the model as with say GridModel or TabModel.

In that world you could have a 'bind' prop that would just give you the ability to bind to a property name on the model that would be the (observable or not) collection or store. .  Observable data would never need to be copied locally.

lbwexler:
to me, the key decision at this point, is will we ever support selection with this thing.  Also, does it scroll?

TomTirapani:
It tries to avoid scrolling, but will scroll vertically if you have set a min height on components and it can't find a non-scrolling layout.

Regarding selection, I can think of a client usage for this that would include selection. But I'm not sure that *this* component needs to handle the selection - that can be left as an implementation detail for the app. I really like the idea of this being a component that just handles the layout and nothing else.

TomTirapani:
Updated this branch so that TilingContainer binds to an observable array. Also, removed the TilingContainerModel in favour of using a LocalModel

lbwexler:
Tom had idea to go one-step further in the simplification and just have it take the raw children, and focus totally on layout.

TomTirapani:
I've updated this branch so that the newly named TilingFrame acts on its raw children. Think we landed in a good spot

cmp/layout/TilingFrame.js
```diff
@@ -0,0 +1,264 @@
...
+            ref,
+            useOnResize(dimensions => localModel.setDimensions(dimensions), {debounce: 100})
+        );
+
+        return frame({
+            ref,
+            className,
+            ...layoutProps,
+            items: items.map((item, idx) => {
+                if (!localModel.layout) return null;
```
lbwexler, line 71: little surprising that you make the child an array of n  nulls, rather than a single null

TomTirapani: I had the same thought, but it seems we'd always be dealing with an array of nulls regardless: https://github.com/xh/hoist-react/blob/develop/core/elem.js#L45

Issue #2274, opened 2021-02-03
Mobile typeahead select control


TomTirapani:
The mobile Select, built on top of react-select, disables type-to-filter functionality by default. This is because it can be problematic to bring up the device keyboard when the Select is in the lower portion of the screen.

However, for large datasets, it can be quite painful to scroll through the list to find the entry you're looking for. A fully-supported type-to-filter control is necessary to provide a good user experience in mobile apps.

At this point, would like to brainstorm / research potential solutions.

TomTirapani:
I took a stab at implementing a "full-screen" mode for mobile selects (I'm not keen on the name), with the select at the top of the screen. This ensures there is always space to bring up the device keyboard.

https://github.com/xh/hoist-react/pull/2383
https://github.com/xh/toolbox/pull/473

The one wrinkle with the above approach is that its seemingly impossible to automatically bring up the device keyboard in fullscreen mode. Calling input.focus() *outside of a user action* doesn't bring up the keyboard. (https://stackoverflow.com/questions/6837543/show-virtual-keyboard-on-mobile-phones-in-javascript). As such, we can show the list on focus but can't bring up the keyboard.

We could take a different approach, where we make sure we use the same input and only move its position somehow. I tried this and hit a bit of a wall (we need to use a portal to reliably position at the top of the screen), but can reopen investigations if we think bringing up the keyboard is vital.

lbwexler:
could call it "modal"

TomTirapani:
Both PRs have been updated to latest from develop, and are ready for review.

These PRs build on top of react-select, rendering it in a full-screen portal on focus. IIRC, we had also discussed an alternative approach, wherein we would drop react-select in favour of a custom full-screen dialog (e.g. grid of options with a filter input) that would conform to the HoistInput API. Is this approach worth pursuing?

PR #2383, opened 2021-03-30
Mobile Select - "Fullscreen" mode

TomTirapani:
Allow showing mobile selects in "fullscreen" mode to facilitate space for the device keyboard.

See issue: https://github.com/xh/hoist-react/issues/2274
See toolbox PR: https://github.com/xh/toolbox/pull/473

Hoist P/R Checklist
-------------------


mobile/cmp/input/Select.js
```diff
@@ -105,6 +116,9 @@ Select.propTypes = {
     /** Text to display when control is empty. */
     placeholder: PT.string,
 
+    /** Text to display in header when in fullscreen mode. */
```
lbwexler, line 119: nit:  might just call this 'title' and mention that it is currently used in `fullScreen` mode

mobile/cmp/input/Select.js
```diff
@@ -385,14 +433,51 @@ const cmp = hoistCmp.factory(
...
+                    className: 'xh-select__fullscreen-toolbar',
+                    items: [
+                        button({
+                            icon: Icon.chevronLeft(),
+                            onClick: () => model.setFullscreen(false)
+                        }),
+                        span(title)
+                    ]
+                }),
+                children
```
lbwexler, line 479: realize there is only a single child, but this is confusing.  Might either spread with ...children, or use React.Children.only(children)

lbwexler (approved):
just some super-minor nits.  Looks great

TomTirapani:
Thanks for taking a look Lee - I've updated based on your suggestions.

What did you think about the concern raised on the issue re: inability to automatically bring up keyboard on device?
https://github.com/xh/hoist-react/issues/2274

mobile/cmp/input/Select.js
```diff
@@ -42,18 +47,25 @@ Select.propTypes = {
...
-     * showing the dropdown menu on click. Should be used with caution, as mobile keyboard can
-     * cause undesirable results.
+     * showing the dropdown menu on click. Usually used with enableFullscreen, to ensure access
+     * to the mobile device keyboard.
      */
     enableFilter: PT.bool,
 
+    /**
+     * True to render the select control in a full-screen modal dialog when focused.
+     * Usually applicable when enableCreate|enableFilter, as we can guarantee the control
```
amcclain, line 64: Suggest replacing "Usually..." in these comments with "Recommended for use with..."

mobile/cmp/input/Select.scss
```diff
@@ -81,3 +81,37 @@
...
+
+  .xh-select {
+    margin: var(--xh-pad-px);
+
+    &__fullscreen-wrapper {
+      position: absolute !important;
+      top: 0;
+      left: 0;
+      right: 0;
+      bottom: 0;
```
amcclain, line 99: Is it standard to specify right and bottom like this (in addition to top/left) for an absolutely positioned element?

mobile/cmp/input/Select.scss
```diff
@@ -81,3 +81,37 @@
...
+      background: var(--xh-bg);
+    }
+
+    &__fullscreen-toolbar {
+      background-color: var(--xh-appbar-bg);
+      color: var(--xh-appbar-title-color);
+      padding: var(--xh-pad-half-px) var(--xh-pad-px) var(--xh-pad-half-px) 0;
```
amcclain, line 106: Seeing a funny artifact when testing on my iPhone 12, where the back button is stretched against the left edge of the toolbar.  Weird, as I see we have left padding specified...

IMG_0615

amcclain (commented):
I tested on my iPhone and the keyboard is showing up automatically OK for me - is that what you guys were seeing? (This is in response to https://github.com/xh/hoist-react/issues/2274#issuecomment-810458552).

I had been thinking of the totally dedicated component with the text box and the grid, etc. but this seems to work well and has several very big benefits in terms of re-using `select` and all of its various API options, instead of re-inventing a more minimal component. 

I also like that the change is restricted to the mobile select component only, which gives me more confidence in merging, and I think it remains a usability win for working with longer select lists on the phone.

Let me know if the keyboard issue is still current / requires more investigation, or if there are any other causes for hesitation. Otherwise I'd be in favor of adding a changelog entry and merging.

mobile/cmp/input/Select.scss
TomTirapani: It's another way of making an absolute positioned element fill its container. We could also give it `top: 0; left: 0; width: 100%; height: 100%;`, but I think that might have some unexpected behaviour in some circumstances (i.e. container is floating).

https://rajatamil.medium.com/css-make-a-div-full-screen-402f9a935d15

We're using it in many other places throughout the code (CodeInput, TreeMap, Mask, etc)

mobile/cmp/input/Select.scss
TomTirapani: Actually, that button is supposed to be transparent (similar to the main app bar colors)... will update the styling. Thanks for catching!

TomTirapani:
Thanks for taking a look Anselm!

I'm happy to merge at this point - agree there are some benefits to building on top of `Select` rather than a new component. I've pushed some tweaks based on your feedback and added a CHANGELOG entry.

mobile/cmp/input/Select.scss
amcclain: Nice, thanks for the info - will (attempt to) remember it for when I need to do the same!

amcclain:
Great - glad this got merged in. 👍

PR #2384, opened 2021-03-30
Inline editing

haynesjm42:
Toolbox branch: https://github.com/xh/toolbox/tree/inlineEditing

Still need to add an inline select editor, will do that in the next couple days as part of this pr or a follow-up

Hoist P/R Checklist
-------------------


TomTirapani:
Been using this in its current state in a client project and looks and works great so far!

One thing to consider is how the inline editors impact our autosizing routine. Currently, if you autosize a column so that its display fits, the padding added by the editor makes it _just small enough_ that the input elides.

We could consider checking the editable flag (or the perhaps better the presence of an cellEditor) in the ColumnWidthCalculator and adding a buffer to account for this - making it so that autosizing an editable column ensures its wide enough to edit without eliding.

haynesjm42:
Thanks for trying it out Tom! I tweaked the css to avoid an extra border that was being added to the cell when inline editing which I think resolves that issue you mentioned with auto-sizing for text and number inputs.

I also changed the default date editor setup to be more in line with our non-inline-editor defaults (showing button for date picker popup) - which presents it's own challenge with auto-sizing since it will not have room for the button after an auto-size - not entirely sure the best approach here since some editors may need more space than others and that probably can't be integrated into the auto-size routine. Popup editors may be the way to go for some of these, though that requires a very different path for styling the editor components properly since they are no longer contained within the cell - will explore that for the date editor.

haynesjm42:
Another thing I need to test and polish is the styling of the editors in different sizing modes - hopefully no one will try to have a tiny sized grid be editable but we should be able to at least make a good effort at making sure it is usable.

amcclain:
Hi - also pulled this into a client project where I needed to quickly implement a simple inline editing story.  Just FYI - absolutely still considering this a WIP (in a good way, i.e. we can keep changing whatever we need) and personally looking forward to feeling better and discussing in more detail next week.

amcclain:
[> I also changed the default date editor setup to be more in line with our non-inline-editor defaults (showing button for date picker popup) 

For my use case I like the previous setup without the inline button - it is not aligning correctly with custom grid sizing we have in this app, but in general I am not sure what is the best default.

amcclain:
Added DateInput ticket #2415 as we think that might be an important step towards getting that particular editor working more reliably with inline.

TomTirapani:
Just pulled the latest into my project and updated. 👍 

I'm still seeing sizing issues on autosized columns, particularly at the smaller sizing modes. These images show what happens immediately after opening an editor on an autosized column:

Screenshot 2021-05-05 at 14 55 23
Screenshot 2021-05-05 at 14 55 32

I was playing around in the dev tools, and the issue seems to be the inner left / right padding of the input. Think we might be able to get some leverage out of applying the appropriate `--xh-grid-cell-lr-pad-px`, `--xh-grid-compact-cell-lr-pad-px` left / right padding to the input when used in a cell editor. From my tests, it looks like this will lead to the text in the input lining up nicely with the text in the cell when switching to edit mode:

Screenshot 2021-05-05 at 14 55 23
Screenshot 2021-05-05 at 15 00 53

I'll play around with it and push a branch if it looks like a potential solution.

TomTirapani:
Went ahead and created this small PR (into this branch) which tries to render the inputs with similar font, font-size and padding as the grid across the various sizing modes: https://github.com/xh/hoist-react/pull/2423

cmp/grid/columns/Column.js
```diff
@@ -140,6 +151,8 @@ export class Column {
      *      flashing the cell's background color. Note: incompatible with rendererIsComplex.
      * @param {boolean|Column~editableFn} [c.editable] - true to make cells in this column
      *     editable.
+     * @param {Column-editorElementFn} [c.editorElement] - function which returns a React
```
lbwexler, line 154: I think we could just call this `editor` for simplicity

desktop/cmp/grid/editors/HoistInlineEditorModel.js
```diff
@@ -0,0 +1,144 @@
+/*
```
lbwexler, line 1: This could also be called "CellEditorModel".  In any case, seems like filename has an extra ''Hoist"

desktop/cmp/grid/editors/HoistInlineEditorModel.js
```diff
@@ -0,0 +1,144 @@
...
+
+        inputEl.focus();
+        if (selectOnFocus) inputEl.select();
+    }
+
+    stopEditing() {
+        this.agParams?.stopEditing();
+    }
+
+    // Stub for subclasses to implement for special handling when value is committed for certain
```
lbwexler, line 88: document this template method with jsdoc

desktop/cmp/grid/editors/HoistInlineEditorModel.js
```diff
@@ -0,0 +1,144 @@
...
+        return this.agParams?.cellStartedEdit;
+    }
+
+    /** @returns {HoistInputModel} */
+    get inputModel() {
+        return this.ref.current;
+    }
+
+    get inputEl() {
+        return this.inputModel?.inputRef.current;
```
lbwexler, line 52: HoistInputModel has an `inputEl` getter that does a bit more work for you and can be delegated to.

desktop/cmp/grid/editors/InlineCheckboxEditor.js
```diff
@@ -0,0 +1,24 @@
...
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2021 Extremely Heavy Industries Inc.
+ */
+import {hoistCmp} from '@xh/hoist/core';
+import {checkbox} from '@xh/hoist/desktop/cmp/input';
+import {useHoistInlineEditorModel} from './HoistInlineEditorModel';
+import {HoistInlineEditorPropTypes} from './HoistInlineEditorProps';
+
+export const [InlineCheckboxEditor, inlineCheckboxEditor] = hoistCmp.withFactory({
```
lbwexler, line 12: maybe lose the word "inline".  Will be clear from package context.

desktop/cmp/grid/editors/InlineCheckboxEditor.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2021 Extremely Heavy Industries Inc.
+ */
+import {hoistCmp} from '@xh/hoist/core';
+import {checkbox} from '@xh/hoist/desktop/cmp/input';
+import {useHoistInlineEditorModel} from './HoistInlineEditorModel';
```
lbwexler, line 9: Replace "HoistInline" prefix with "Cell".  All will be clear from package context.

haynesjm42:
Nope no blockers or major concerns by all means take a look if you have the time! I got a select editor working decently in a client app - there are some complications from keyboard nav for the options list (and styling react-select can be challenging) - can point you to the client app example I have

PR #2386, opened 2021-04-01
Add shared AppBanner

TomTirapani:
Adds a shared AppBanner, and convenience methods for showing / hiding it.
Also updates the implementation of our version update banner to use it.

See Toolbox PR: https://github.com/xh/toolbox/pull/474
See issue: https://github.com/xh/hoist-react/issues/887

Hoist P/R Checklist
-------------------


core/XH.js
```diff
@@ -424,6 +424,30 @@ class XHClass extends HoistBase {
...
+     */
+    showBanner(config) {
+        return this.acm.appBannerModel.show(config);
+    }
+
+    hideBanner() {
+        return this.acm.appBannerModel.hide();
+    }
+
+    get bannerIsShowing() {
```
TomTirapani, line 447: Remove this

desktop/appcontainer/AppContainer.js
```diff
@@ -129,6 +130,17 @@ const appContainerView = hoistCmp.factory({
...
 });
 
+const bannerList = hoistCmp.factory({
+    render({model}) {
+        const {bannerModels} = model.bannerSourceModel;
+        return (isEmpty(bannerModels)) ?
+            null :
+            fragment({
+                items: bannerModels.map(model => banner({model, key: model.xhId}))
+            });
```
amcclain, line 140: I didn't think it was important by any means to avoid returning an empty fragment, but it seemed slightly cleaner (and hopefully not worse in any way).

desktop/appcontainer/Banner.js
```diff
@@ -0,0 +1,83 @@
...
+        });
+    }
+});
+
+const actionButton = hoistCmp.factory(
+    ({model}) => {
+        const {actionFn, actionButtonProps} = model;
+        if (!isFunction(actionFn)) return null;
+        return button({
+            text: 'Action',
```
amcclain, line 61: The lack of a good default text here makes me wonder if we should remove `actionFn` and only take `actionButtonProps` (where user would specify `onClick`). The user almost certainly needs to spec the latter if they are spec'ing the former, why have them split across two top-level configs?

The pattern we have now is from `ErrorMessage`, where we supply a somewhat-more-likely-to-be-useful default of "Retry" for the button text. Weighing in my mind these options:

- Leave everything as it is right now 😃 
- Modify new banner config to accept only `actionButtonProps`
- Modify new banner config + change recently-added `ErrorMessage` comp to also support only the props, removing the default. I doubt many apps have used the comp yet, but we could deprecate/migrate to avoid the breaking change.

Looking at this case here, I'm inclined to go for the last option, but could pretty easily be persuaded to any of them. Let's discuss quick then get on with the merge.

desktop/appcontainer/Banner.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  &__message {
+    margin-right: var(--xh-pad-px);
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+  }
+
+  &__dismiss-button {
+    color: white !important;
```
amcclain, line 22: I wasn't a super big fan of adding these `!important`s as I did here and below, but I really did want the button to stand out, and I didn't want to string together all the classnames otherwise required to override. Hope this + the `xh-intent-none` change seem reasonable.

desktop/appcontainer/Banner.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  & > .xh-icon {
+    margin-right: var(--xh-pad-half-px);
+  }
+
+  &__message {
+    margin-right: var(--xh-pad-px);
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
```
amcclain, line 18: I tested with a very long message (pretty easy to imagine use case) and the message div pushes the action and dismiss buttons off the edge - we should resolve that if you have a chance to look.

Might be easier to do if the action button also floats to the right - like on mobile - so the message div gets the flex but no more.  I was wondering if that would look better anyway, so the action button is in a more consistent place. (Realize existing styling comes from update bar.). 

I'm happy either way, but I do think we should ensure it clips so those buttons remain accessible at all screen widths.

amcclain (commented):
Looks great - thanks for putting this so neatly together.

I pushed some minor tweaks + added these hopefully quick to review comments. Happy to talk it over and get it merged without delay tomorrow.

desktop/appcontainer/AppContainer.js
TomTirapani: Makes sense! Unless there are any objections, I might change this to an early return rather than a ternary - find those easier to read

desktop/appcontainer/Banner.scss
TomTirapani: Floating the action button to the right is a great idea, had the same thought myself. Then the message will take the flex and elide on overflow.

desktop/appcontainer/Banner.scss
TomTirapani: 👍

desktop/appcontainer/Banner.js
TomTirapani: I'm happy to implement the last option - makes sense to collapse into one prop. For ErrorMessage, I can't imagine many users are using it, and should be relatively straightforward to deprecate cleanly like you say.

amcclain: Great - let's do it.  I think that makes the most sense, and `actionButtonProps` seems like a reasonable enough config value for the model to trigger the display of an action button.  (Could also go with `actionButton`, but I think the name we have makes it slightly more clear that you should not provide the button element itself.)

TomTirapani: Great - will make those changes shortly

PR #2388, opened 2021-04-06
New img-based Spinner component

amcclain:
+ Used for the platform-specific `Mask` and `LoadingIndicator` components.
+ Replaces previous SVG-based implementations to reduce overhead when rendering spinners over remote connections.

Hoist P/R Checklist
-------------------


amcclain:
This should be tested with real apps, especially client apps accessed over remote connections, by developers who are currently experiencing issues with the existing spinner implementations (I am not, although I have seen it in the past).

A hoist-react build is available off of this branch - `@xh/hoist@40.0.0-SNAPSHOT.1617728122879` - for use in such testing.

The image quality of the new spinner over the remote connection is not super great in my testing, but nor is it especially bad. On par with e.g. reduced quality of fonts and other rendering subtleties when not working with a full-res, high-DPI connection.

If this approach does show clear benefits in practice, I think it's also a win to remove the use of two more "kit" dependencies.  We could stop re-exporting wrappers for them as well.

lbwexler:
love the way it drops in so nicely and has such a minimal dependency.  Think we should definitely remove the wrappers quietly as well -- don't think we make any strong API guarantees there.  Eager to see the results of the testing and to get this merged.

lbwexler (approved):

amcclain:
I'll merge this now - we can see if it actually improves the situation vis-a-vis VMs and Citrix. Even if it does not, we can amend the comments/CHANGELOG to avoid claiming that it does, but still likely worth it to take the change and reduce our use of kit dependencies where we can supply a simpler replacement.

Issue #2394, opened 2021-04-09
Submenu added to appMenuButton opens automatically when parent menu opens


febbraiod:
This is a known blueprint issue, for which there is a workaround:

[https://github.com/palantir/blueprint/issues/2863](https://github.com/palantir/blueprint/issues/2863)

Should be address simultaneously with [https://github.com/xh/hoist-react/issues/2393](https://github.com/xh/hoist-react/issues/2393)

Issue #2393, opened 2021-04-09
AppMenuButton's buildMenuItems method should recurse over children configs


febbraiod:
Right now if you want to add a submenu you have to add the children as menuItem() or menuDivider(), rather that as configs to build them, as with top level items.

febbraiod:
Should be address simultaneously with #2394

PR #2397, opened 2021-04-12
Improve support for SubMenus in AppMenu

TomTirapani:
Brings our AppMenu API in line with our other menu implementations.

+ Recursively parse AppMenu items to build submenus from configs. #2393
+ Don't automatically open submenus. #2394

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/button/AppMenuButton.js
+++ b/desktop/cmp/button/AppMenuButton.js
@@ -44,10 +44,12 @@ AppMenuButton.propTypes = {
     className: PT.string,
 
     /**
-     * Array of app-specific menu items. Can contain `MenuItem` configs, React Elements, or the
-     * special string token '-' (to render a `MenuDivider`).
+     * Array of extra menu items. Can contain:
+     *  + `MenuItems` or configs to create them.
+     *  + `MenuDividers` or the special string token '-'.
+     *  + React Elements or strings, which will be interpreted as the `text` property for a MenuItem.
      */
-    extraItems: PT.array,
+    extraItems: PT.arrayOf(PT.oneOfType([PT.object, PT.string, PT.element])),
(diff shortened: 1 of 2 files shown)


desktop/cmp/button/AppMenuButton.js
```diff
@@ -155,15 +156,30 @@ function buildMenuItems({
...
         .map(it => {
             if (it === '-') return menuDivider();
-            return isValidElement(it) ? it : menuItem(it);
+            if (isValidElement(it)) {
+                return ['Blueprint3.MenuItem', 'Blueprint3.MenuDivider'].includes(it.type.displayName) ?
+                    it :
+                    menuItem({text: it});
+            }
+
+            const cfg = {...it};
```
amcclain, line 178: This second case is expecting `it` will be a plain object, suitable as a config for `menuItem`, is that correct?

Can we tweak the proptype doc on `extraItems` for a bit more clarity? I think the valid options are:
- A `menuItem` or `menuDivider` element, which will be used as-is.
- A `ReactElement` or string (`isValidElement(someString) == true` ??) which will be set as the text of a menu item.
- A plain object to be used as a `menuItem` config, with support for a nested `items` array which can contain this same list of things.

Not saying we need to use those exact / verbose words in the prop-type doc, but enumerating them here to ensure I understand what we're accepting. At some point soon we really need to provide a first-class, fully documented general menu component, without relying on importing BP kit comps directly.  Still don't think we have an issue for that, will add.

TomTirapani: Yes, that's correct, and those are the options supported. I've pushed an update to the docs and propTypes for extra items that should hopefully clarify this a bit.

PR #2408, opened 2021-04-21
Exception stringify

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/exception/Stringify.js
+++ b/exception/Stringify.js
@@ -5,44 +5,52 @@
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
 import {stripTags, trimToDepth} from '@xh/hoist/utils/js';
+import {omitBy, isNil} from 'lodash';
 
 /**
- * Stringify an error object (typically an exception) safely for submission to server.
+ * Stringify an error object safely for submission to server.
  * This method will avoid circular references and will trim the depth of the stack.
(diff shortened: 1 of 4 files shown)


exception/Stringify.js
```diff
@@ -5,44 +5,54 @@
...
-        // Delete noisy grails exception
-        const {serverDetails} = err;
+        // 1) Create basic structure.
+        // Raw Error does not have 'own' properties, so be explicit about core name/message/stack
+        // Order here intentional for serialization
+        let ret = {
+            name: error.name,
+            message: error.message
+        };
+        Object.assign(ret, error);
```
amcclain, line 27: To ensure I understand this - we might get a "Raw Error" with no "own" properties, so we explicitly extract name/message/stack - and in that case, does `Object.assign()` does nothing?

But we might also get an Error that's not "raw", in which case it would have additional properties installed - primary use case being exceptions created by our `Exception.js` methods.  In that case, any additional custom properties will get copied over with the assign.

Give a shout if that's *wrong* - but otherwise looks good, recording here only to confirm and have a note on it.

amcclain (approved):

exception/Stringify.js
lbwexler: great -- you got it

Issue #2396, opened 2021-04-12
Columns should have a sortValue prop


febbraiod:
Would be symmetrical with exportValue.

In a client app we had a column where the sorting was broken because the underlying value is an object with a value property.

Solved by adding a comparator to the column that will pull out the obj.value, and pass that to the defaultComparator, but providing a sortValue would be cleaner application code.

If both a sortValue and a custom comparator are provided. Ideally the resolved comparator just gets fed the values that arise from sortValue, but if that is difficult in practice, could be fine to just have comparator ‘win’.

PR #2420, opened 2021-05-04
Columns should have a sortValue prop, Fixes: #2396

febbraiod:
Hoist P/R Checklist
-------------------


febbraiod:
Can test with : https://github.com/xh/toolbox/tree/SortValue

cmp/grid/columns/Column.js
```diff
@@ -548,8 +562,20 @@ export class Column {
...
+            recordA = agNodeA?.data,
+            recordB = agNodeB?.data;
+
+        if (isString(sortValue)) {
+            v1 = recordA?.data[sortValue] ?? v1;
+            v2 = recordB?.data[sortValue] ?? v2;
+        } else if (isFunction(sortValue)) {
+            v1 = sortValue(v1, {recordA, column: this, gridModel});
+            v2 = sortValue(v2, {recordB, column: this, gridModel});
+        }
```
TomTirapani, line 577: This code block is very similar to the one above. Could we DRY these up with a `getSortValue(record)` method?

cmp/grid/columns/Column.js
```diff
@@ -532,6 +536,16 @@ export class Column {
                         agNodeB
                     };
 
+                if (isString(sortValue)) {
+                    // If sortValue points to a different field
+                    valueA = recordA?.data[sortValue] ?? valueA; // Group rows have no record.
```
lbwexler, line 541: this note about groupRows seems problematic?  So will this even work or a groupedGrid that is trying to use the new sortValue property?

febbraiod: Well the value for the grouped row was being passed to the comparator before regardless. It was an empty string. Which it remains here.

cmp/grid/columns/Column.js
febbraiod: yes good idea.

cmp/grid/columns/Column.js
febbraiod: Tested whilst paying attn to the group rows. The sort works, but the group rows now jump around with the when sorting by this column. Previously the groups would stay put. Will revisit.

lbwexler:
great -- eager to merge this when you work this out.  Was it sorting the
groups based on 'null', or some other sort value of the group, in that case?

febbraiod:
Was using an empty string for the group rows. And I guess since "" === "" the group rows stay put. Just want to look at together for a sanity check since it's gone this far with a bit of confusion.

Issue #1316, opened 2019-08-15
Improve coloring for SplitTreeMap/TreeMap


lbwexler:
Some issues to consider:

+ The default coloring scale should be settable by the applications according to domain knowledge of the data.  This will allow colors to be compared meaningfully from data set to data set

+ In any case, split tree map should never use different color ranges for its two sub maps!  The map is conceptually one thing.  The brightest profit should be of an equivalent relative heat as the brightest loss!

+ <s>Consider using gray for nodes that should not have color (e.g. no risk info available).  This will look better than white in both dark and light modes and is more clearly 'disabled' </s>

+ <s> Need a clearer boundary between the two maps, or at least the option for an app to set the width.  In certain degenerate cases it can be difficult to grok the whole notion of the split tree map without that.</s>

lbwexler:
@TomTirapani  -- is there anything in this ticket that we still need to address?

TomTirapani:
I don't think so. I think all of the above has since been addressed

lbwexler:
I actually think the first two are still important issues.  We spread any heat data we get between 0 and 1, _for each individual map_.  I may be misunderstanding this, but I don't think this allows the application to determine a consistent scale for multiple maps.

TomTirapani:
Okay, I'll take another look.

Just to make sure I understand the requirements, for:
1) We want apps to be able to pass min/max values which determine the scale, otherwise we determine the range from the data.
2) The range used for each map is derived from the min / max of both maps. e.g. A wide range of negative values and a narrow range of positive values would render the positive values with a narrower color range.

Is this correct?

lbwexler:
Sorry for the delay.  I think the key is that the app should typically pass the split treemap a single absolute `maxHeat` value that represents the heat value that will have the maximum brightness (positive or negative) within the entire HeatMap.  This allows the coloring to be co-ordinated and consistent, across multiple maps, and multiple loads.  

Right now we use the largest value in the dataset, which makes the coloring always nice and bright in each submap, but does not allow the user to attach any consistent meaning to that brightness.

lbwexler:
FIXED!  Thanks Tom!

PR #2431, opened 2021-05-10
Tree map colors

TomTirapani:
Add maxHeat config to TreeMap and SplitTreeMap.
See issue: https://github.com/xh/hoist-react/issues/1316
See Toolbox PR: https://github.com/xh/toolbox/pull/477

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/treemap/TreeMapModel.js
+++ b/desktop/cmp/treemap/TreeMapModel.js
@@ -355,7 +355,7 @@ export class TreeMapModel extends HoistModel {
                     } else {
                         it.colorValue = this.normalizeToRange(heatValue, minHeat, midHeat, 0.6, 0.8);
                     }
-                } else if (colorMode === 'linear' || heatValues.length === 2) {
+                } else if (colorMode === 'linear' || heatValues.length <= 2) {
                     it.colorValue = this.normalizeToRange(heatValue, minHeat, maxHeat, 0.6, 1);
                 }
             } else if (heatValue < 0) {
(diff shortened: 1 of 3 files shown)


desktop/cmp/treemap/TreeMapModel.js
```diff
@@ -324,28 +331,14 @@ export class TreeMapModel extends HoistModel {
...
-        // 2) Split heat values into positive and negative
-        let [posHeatValues, negHeatValues] = partition(heatValues, it => it > 0);
-        posHeatValues = sortBy(posHeatValues);
-        negHeatValues = sortBy(negHeatValues.map(Math.abs));
-
-        // 3) Calculate bounds and midpoints for each range
-        let minPosHeat = 0, midPosHeat = 0, maxPosHeat = 0, minNegHeat = 0, midNegHeat = 0, maxNegHeat = 0;
-        if (posHeatValues.length) {
-            minPosHeat = posHeatValues[0];
-            midPosHeat = posHeatValues[Math.floor(posHeatValues.length / 2)];
```
lbwexler, line 339: So I am still just now trying to Grok the Balanced mode.  It seems like it distorts the data to try to spread it out a bit more across the range.  But in any case, it seems like the use of the Median, and the "min" that we are using are now changed (min now zero)

Would like to discuss with ATM, but I think since it is not the default, we could potentially drop it, if it no longer made sense to us.

TomTirapani: Yeah, the intention is balanced mode is to make a nicer looking chart where you have outliers. It does not do an 'accurate' mapping of color to value like 'linear' does, but grants a wider proportion of the range to whichever half is more heavily weighted so you can more easily see the difference.

It's a bit of a philosophical question whether its better to have accurate colors or instead display colors that are easier to compare against each other. The move towards having a set maxHeat certainly seems like we're favoring accuracy over aesthetics, so would be happy to drop it.

lbwexler:
OK -- removed 'balanced' after confirming with ATM .  Tom -- as I mention in the changelog, I hope that maxHeat also eliminates the need for 'balanced' mode as well

PR #2453, opened 2021-05-18
Inline Editing Validation

TomTirapani:
First pass at validation for Grid inline editors.

+ Move validation / constraints to data package
+ Store field can be configured with a collections of rules constraints
+ Add methods to validate and check validation state at the Store, Record and record Field levels
+ Support for async validation
+ Use cellClassRules to style invalid cells
+ Render errors in a tooltip
+ Minor (sometimes unrelated) fixes / tweaks

I'm not particularly happy with the complex way I'm storing the validation errors, but couldn't think of an alternative given we can't store them on the record. Hopefully something we can improve in review.

See Toolbox PR: https://github.com/xh/toolbox/pull/480


TomTirapani:
Screenshot 2021-05-18 at 13 09 07

cmp/grid/Grid.js
```diff
@@ -412,7 +413,7 @@ class LocalModel extends HoistModel {
 
                     if (transaction?.update) {
                         const refreshCols = model.columns.filter(c => !c.hidden && c.rendererIsComplex);
-                        if (refreshCols) {
+                        if (!isEmpty(refreshCols)) {
```
lbwexler, line 416: good catch!!

cmp/grid/Grid.js
```diff
@@ -561,6 +561,23 @@ class LocalModel extends HoistModel {
         };
     }
 
+    validationDisplayReaction() {
+        const {model} = this,
+            {store} = model;
+
+        return {
+            track: () => [model.isReady, store.isValidationPending],
```
lbwexler, line 569: this relies on the validation briefly flickering the pending state -- just noting this -- seems brittle

data/Record.js
```diff
@@ -118,6 +120,24 @@ export class Record {
         return this.store.getAncestorsById(this.id, false);
     }
 
+    /** @return {boolean} */
+    @computed
```
lbwexler, line 124: think the use of computed breaks the immutability here, it implies a cached value that will change..

data/impl/StoreValidator.js
```diff
@@ -0,0 +1,187 @@
...
+        return this.getValidationState();
+    }
+
+    /** @return {boolean} - true if currently recomputing validation state. */
+    get isPending() {
+        return this._validationTask.isPending;
+    }
+
+    /** @return {string[]} */
+    @computed
```
lbwexler, line 50: think this *might* need computed.struct.  

Need to think about this -- there may be many cases where the error object will be the same (e.g. [] and []) across validation runs.  Or, for example, that the user has a single specific validation error, and causes re-validation to occur, and still has the same single error.

data/impl/StoreValidator.js
```diff
@@ -0,0 +1,187 @@
...
+
+    /** @member {Store} */
+    store;
+
+    @managed _validationTask = new PendingTaskModel({mode: 'all'});
+
+    // Stores the result of evaluating each record. Errors are store in a deep map structure
+    // to facilitate access at the record, field and rule levels:
+    // e.g. _errors[record.id][field.name][rule] = errors[]
+    @observable.ref _errors = {};
```
lbwexler, line 30: We could consider adding a RecordErrors impl class that might be the thing that is stored for each record.

data/impl/StoreValidator.js
```diff
@@ -0,0 +1,187 @@
...
+
+    /** @return {boolean} */
+    @computed
+    get isValid() {
```
lbwexler, line 34: simple convenience getter, does not require computed.

data/impl/StoreValidator.js
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    //---------------------------------------
+    // Validating Records
+    //---------------------------------------
+    /**
+     * Recompute validations for a given record and return true if the record is valid.
+     * @returns {Promise<boolean>}
+     */
+    async validateRecordAsync(record) {
```
lbwexler, line 105: Thinking a RecordValidator object, that gets created on the fly with a pointer back to the Record will clean this up a lot!  Breaks up this file (which is looping on three different levels) and also means you don't have to pass the record to all of the methods below.  Should not be too much object creation, now that we are only validating modified records.

data/impl/StoreValidator.js
```diff
@@ -0,0 +1,187 @@
...
+        await Promise.all(promises).linkTo(this._validationTask);
+        return record.isValid;
+    }
+
+    /** @return {ValidationState} - the current validation state for a given record. */
+    getRecordValidationState(record) {
+        const VS = ValidationState;
+        if (!record.isModified) return VS.Valid;
+
+        const states = map(this.store.fields, field => this.getRecordFieldValidationState(record, field));
```
lbwexler, line 117: why is UNKNOWN not a possibility for a newly created Record?

Issue #1931, opened 2020-05-27
Improve arguments to fmtNumber


febbraiod:
A simple enhancement would be a useCommaDelimiters flag. 

We had a use case where the magnitude and type of the data (thousands of milliseconds)looked funny with the commas. It turned out to be not as straightforward to turn on/off as we hoped. We needed to use the numbro formatConfig directly.

Using the numbro formatConfig directly has its own sort of inelegance that could be improved upon but perhaps not as easily. If you provide the formatConfig to our method, it will stomp on precision and zeroPad flags. Which requires you to use the numbro equivalents directly. This is because the helper that transforms these flags (buildFormatConfig) is never called in this case. It would probably be better if a dev could specify one or more pieces of the numbro config and still have the formatter accept Hoist flags by intelligently merging them in.

amcclain:
See my thoughts on this Slack thread - https://extremelyheavy.slack.com/archives/C098Y9CDP/p1590178369024300

febbraiod:
Anslem's comments for easier reference:

I went to add a call to fmtNumber() to add commas for the odd very slow call, and saw that same jarring 1,234ms value.

Wondering if we should build in an option to our fmtNumber() function to be smart about this. I was surprised to find we don't currently have any flag controlling commas (although you can provide your own format string).

One option would be withCommas: true|false|auto - where "auto" mirrors the auto option for precision and would skip the comma for four-digit numbers.  A bit weird, but one proposal for how we might get it into the API.

I'm not sure how surprising ppl would find it if that were the default (which I think it would need to be if we went for "auto" - too weird for auto to not be the default).  For currency amounts, seeing 1,234 does not seem that weird.  I am not sure.

The precision might also be a factor - somehow less weird to me to see 1,234.00 in any context

febbraiod:
My concern about 'auto' being the default and then having that mean "skip the comma for 4 digit numbers" is that it's a minor breaking change for any application using our formatters and renderers. A numeric column that once aligned its values would in some cases no longer nicely align. 

1,000,000
        1,000
 becomes
1,000,000
        1000


I would hope this change could be implemented and not misalign nearly every numeric column in nearly every app we have.

PR #2463, opened 2021-05-26
Add withCommas and omitFourDigitComma options to number formatters, Fixes: #1931

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/format/FormatNumber.js
+++ b/format/FormatNumber.js
@@ -35,6 +35,8 @@ const UP_TICK = '▴',
  *      align vertically with negative ledgers in columns.
  * @param {boolean} [opts.withPlusSign] - true to prepend positive numbers with a '+'.
  * @param {boolean} [opts.withSignGlyph] - true to prepend an up / down arrow.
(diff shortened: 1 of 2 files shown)


format/FormatNumber.js
```diff
@@ -35,6 +35,8 @@ const UP_TICK = '▴',
  *      align vertically with negative ledgers in columns.
  * @param {boolean} [opts.withPlusSign] - true to prepend positive numbers with a '+'.
  * @param {boolean} [opts.withSignGlyph] - true to prepend an up / down arrow.
+ * @param {boolean} [opts.withCommas] - whether to include comma delimiters, defaults to true.
+ * @param {boolean} [opts.omitFourDigitComma] - true to omit delimiter if value is less than 10,000
```
febbraiod, line 39: Super not in love with this name, but do like the two property approach

lbwexler: I agree, and think the name is as good as it gets.  I do think that Anselm is correct in the slack conversation that it should be used only in the case of exactly 4 printed digits, that is, not if there are decimal places.

PR #2475, opened 2021-06-03
TileFrame tweaks

lbwexler:
Tom -- for your review/discussion.  

Would like to understand a tiny bit more about the algorithm.  Are there any links to the source of the scoring heuristic?

+ Simplify docs
+ More vertical/horizontal symmetry
~+ Change to defaults suggested by ATM.~

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/layout/TileFrame.js
+++ b/cmp/layout/TileFrame.js
@@ -16,9 +16,9 @@ import {Children} from 'react';
 import './TileFrame.scss';
 
 /**
- * This component renders its children as space filling tiles of the equal size.
- * It will generate a tiling layout that attempts to compromise between limiting empty space,
- * and keeping the tile width / height ratio as close to the desired ratio as possible.
+ * This component renders its children as equally-sized tiles, resized and arranged to fill the
+ * available space within the container while maintaining even padding between tiles and keeping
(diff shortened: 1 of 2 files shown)


TomTirapani:
Thanks for these tweaks Lee!

I think these look great - particularly like the increased readability from splitting the scored vs fallback paths, and the clarification in the code of the different ways they handle the constraints.

I'm curious about the origin of the new min / max ratio defaults? I'm not averse to them, and they certainly make sense for the common use-case, but I'm wary of introducing default constraints since they reduce the universe of scored layouts and increase the likelihood of ending up with the fallback.

Regarding the scoring heuristic - no source besides my own brain, unfortunately :| It was arrived at through experimentation in the client app where it was originally developed, and isn't very scientific. It's fairly straightforward - both straying from the desired ratio and having empty spaces negatively affect a layout's score. Empty spaces are weighted, so it prefers filling the space even if the ratio doesn't match. The "ideal" score of `0` would be achieved by a layout with the exact desired ratio and no empty spaces.

At the risk of over-complicating this thing, we could support a small library of different scoring algorithms (and accept a closure form?) that may be more appropriate for different use cases. For example, one could imagine having a different heuristic that just cares about the ratio and doesn't care about empty spaces.

PR #2477, opened 2021-06-03
Improve behavior of FormModel.init()

lbwexler:
+ See also toolbox branch

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/form/FormModel.js
+++ b/cmp/form/FormModel.js
@@ -84,9 +84,10 @@ export class FormModel extends HoistModel {
         this.readonly = readonly;
         const models = {};
         fields.forEach(f => {
-            const model = f instanceof FieldModel ? f : (f.subforms ? new SubformsFieldModel(f) : new FieldModel(f));
-            throwIf(models[model.name], 'Form cannot contain multiple fields with same name: ' + model.name);
-            models[model.name] = model;
+            const model = f instanceof FieldModel ? f : (f.subforms ? new SubformsFieldModel(f) : new FieldModel(f)),
+                {name} = model;
+            throwIf(models[name], 'Form cannot contain multiple fields with same name: ' + name);
+            models[name] = model;
         });
         this.fields = models;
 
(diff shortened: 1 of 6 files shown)


cmp/form/field/BaseFieldModel.js
```diff
@@ -23,9 +23,6 @@ import {PendingTaskModel} from '@xh/hoist/utils/async';
  */
 export class BaseFieldModel extends HoistModel {
 
-    /** @member {FormModel} */
```
amcclain, line 26: FYI I find these annotations helpful to get typing within this containing class's code, and when looking for usages of any of the member class's properties during refactoring. (Noticed one or two of them went away in the TODO example app.)

I certainly don't think we need these on _every_ member - not by any means - but for Hoist models with non-trivial APIs, I often find they have value in a few ways while we are working on the Hoist codebase itself.

(I don't think that they overly imply the member to be public. We could even go further and add `@private` to the member jsdoc to make that more clear, vs. the leading underscore convention, which I've seen developers very happily ignore. But I have not been doing that - waiting for us to take the plunge to the new private fields syntax instead.... 😃)

cmp/form/FormModel.js
```diff
@@ -127,8 +132,9 @@ export class FormModel extends HoistModel {
     /**
      * Set the initial value of all fields and reset the form.
      *
-     * This is the main entry point for loading new data (or an empty new record) into the form for
-     * editing. If initial values are undefined for any field, existing initial values will be used.
+     * This is the main programmatic entry point for loading new data (or an empty new record)
+     * into the form for editing. If initial values are undefined for any field, the original
+     * initial values specified during model construction will be used.
```
amcclain, line 137: Great - this makes a lot of sense to me and seems like a smart way to get the behavior we expect!

amcclain (approved):
I think this looks great and is a clear improvement.  Taking off now (literally on the plane) but let's touch base tomorrow AM - assuming the CHANGELOG is the remaining item - how much of a breaking change is this, etc.

Issue #2476, opened 2021-06-03
Mobile: Highlight grid rows on tap


TomTirapani:
Investigate what support AgGrid has for highlighting a row on tap. Note this is different than selection, it should be transient feedback for the user.

lbwexler:
Waiting on ag-Grid, as discussed in Pull Request.

amcclain:
Should be unblocked when #2684 resolved

TomTirapani:
New PR using AgGrid v26: https://github.com/xh/hoist-react/pull/2764

PR #2489, opened 2021-06-21
Highlight grid row on tap / click.

TomTirapani:
See issue:  #2476

Unfortunately, ag-grid doesn't offer any built-in solution for highlight grid rows on tap. After trying (and failing) a CSS animation based solution, I was able to get something working by leveraging ag-grid's highlight cells api.

I found the hoist default yellow inappropriate for this use-case, so changed it to blue. This has the unfortunate side effect of also affecting Column.highlightOnChange - I couldn't think of a reliable way for hoist to distinguish between flashes caused by row taps and those caused by cell changes.

We can:
+ Accept this with the knowledge that mobile grids (where this is most useful) are also unlikely to receive streaming updates.
+ Stick with the yellow color for both.
+ Continue trying to figure out a different approach, that doesn't conflate the two distinct use-cases for row flashing.

Hoist P/R Checklist
-------------------


cmp/grid/GridModel.js
```diff
@@ -303,6 +309,7 @@ export class GridModel extends HoistModel {
         autosizeOptions = {},
         restoreDefaultsWarning = GridModel.DEFAULT_RESTORE_DEFAULTS_WARNING,
         fullRowEditing = false,
+        highlightRowOnClick = XH.isPhone || XH.isTablet,
```
amcclain, line 312: I'd suggest we default to `XH.isMobileApp` (and adjust the doc comment accordingly).

cmp/grid/GridModel.js
```diff
@@ -250,6 +252,10 @@ export class GridModel extends HoistModel {
      *      should not immediately respond to user or programmatic changes to the sortBy property,
      *      but will instead wait for the next load of data, which is assumed to be pre-sorted.
      *      Default false.
+     * @param {boolean} [c.highlightRowOnClick] - Set to true to highlight a row on click. Intended
+     *      to provide feedback to users in grids without selection. Note that setting this changes
+     *      the default color provided to Column.highlightOnChange. Default true for phones and
+     *      tablets, false for desktop.
```
amcclain, line 258: Maybe something like "Note this setting overrides the styling used by Column.highlightOnChange, and is not recommended for use alongside that feature."

amcclain (commented):
Thanks for looking at this, and highlighting (no pun intended) the tradeoffs involved.

It's not ideal, but in practice I also don't think that overriding the highlightOnChange color/style would be much of an issue. More than anything it's just a pain to have to explain.

I was hoping we could get the highlight happening as soon as the user touches the row - i.e. `touchstart`. That's really where you get the sense of "I'm interacting with _this_ row."  I tried using the `onCellMouseDown` event with the same approach, but unfortunately the end result when tested on my phone appears to be the same. I'm guessing that's due to ag-grid [translating touch events into mouse clicks](https://www.ag-grid.com/javascript-grid/touch/) but not getting this one quite right (IMO).

With the events we appear to have, I do see the flash right at the moment the app navigates to the detail panel. I still think it helps indicate to the user that they tapped on the intended row, although I wish it were a bit more convincing given the tradeoffs above...

What do you think - on balance is the change/feature pulling its weight?

cmp/grid/GridModel.js
TomTirapani: Curious about this suggestion - I think this is valuable due to the touchscreen common to both phones and tablets. Defaulting true for one touchscreen device but not the other seems a bit arbitrary.

amcclain: I was thinking that since we most commonly expect tablets to run the desktop app, they will typically have master/detail grids with stateful selection, as opposed to the phone apps which more often navigate on tap to a new detail screen.

It's that latter case that was motivating the overall feature in my mind. (If you have a grid with selection you don't need the extra highlight, as you continue to see the selected row w/its natural selection highlight.)

TomTirapani: Ah, makes sense! Will adjust

amcclain:
I filed a support ticket with AG to see if they have any suggestions here, specifically calling out the fact that I expected the documented `onCellMouseDown` grid event to fire on `touchstart` - not on `touchend` as it appears to do now. Will see if they have any other suggestions for this implementation, or thoughts on whether the current events are firing as intended/ideal.

amcclain:
Ag-Grid acknowledged and filed this as "AG-5512 When tapping a grid cell on a mobile device, cellMouseDown doesn't fire until the touch ends". In their backlog at the moment, we can track it via https://www.ag-grid.com/ag-grid-pipeline/

TomTirapani:
AG-5512 is listed as fixed in today's new v26.0.0 release:
https://github.com/ag-grid/ag-grid/releases/tag/v26.0.0

TomTirapani:
Please alternate PR, which leverages the fix in AgGrid v26: https://github.com/xh/hoist-react/pull/2764

lbwexler:
Closing in favor of new PR

PR #2497, opened 2021-06-25
Update to Highcharts 9.1.2

### LINKED ISSUE #2494: Update Highcharts to 9.x
As per changelog, looks like this might not require any API changes to our (minimal) Hoist wrappers.

We don't actually spec the HC dep. within HR (for licensing reasons), do this issue really involves testing the latest 9.x release in Toolbox and ensuring that any HC integrations or API usages we have there + in the supporting hoist-react classes continue to work OK on the latest.

As per changelog, will want to focus on tooltips and anywhere else where we are rendering HTML within the chart: https://www.highcharts.com/blog/changelog/

saba-mo:
Closes #2494 
See matching toolbox PR

Hoist P/R Checklist
-------------------


amcclain:
Hi Saba - pls touch base any time today if you want to go over the latest here.  Was there another set of changes that you intended to push with that last commit - "refactored turning off resetZoomButton" ?

amcclain:
Hi - I pushed some minor tweaks here and then additional cleanups to Toolbox (not required by the upgrade, but while I was looking I figured I would do a bit of housekeeping on some older oddities that had crept into those examples).

One thing I should have mentioned is that the treemap visualizations are also sourced from Highcharts. I took a quick look and they appear OK, with the minor (but still good to fix) exception of the treemap label text.  We were overriding its styling before to remove the default boldface on the label font - that's no longer working.  I think it does look better w/o the bold given how cramped those labels can get. Could you please take a look?

That's the only other item on my list - if we can fix it I would love to get this in.  Thanks much 📈 ⭐

saba-mo:
The treemap data labels are no longer responding to the CSS rule for text > tspan - it's unclear why this stopped working with the update, and I did not find an answer in the CL. The data label should take a class name per the API but it did not apply that rule either. I adjusted the font weight directly in the component, and left the CSS untouched to further investigate.

amcclain (approved):

PR #2499, opened 2021-06-28
Date input commit on change

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/editors/impl/InlineEditorModel.js
+++ b/desktop/cmp/grid/editors/impl/InlineEditorModel.js
@@ -5,7 +5,6 @@
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
 import {useImperativeHandle} from 'react';
-import composeRefs from '@seznam/compose-react-refs';
 import classNames from 'classnames';
 import {isNil} from 'lodash';
 import {HoistModel, useLocalModel} from '@xh/hoist/core';
@@ -47,7 +46,7 @@ export function useInlineEditorModel(component, props, ref, isPopup = false) {
         model: impl,
         bind: 'value',
         commitOnChange: true,
-        ref: composeRefs(ref, impl.ref),
+        ref: impl.ref,
         ...inputProps
     });
 }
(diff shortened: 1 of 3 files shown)


desktop/cmp/input/DateInput.js
```diff
@@ -239,9 +273,8 @@ class Model extends HoistInputModel {
     };
 
     onKeyDown = (ev) => {
-        if (ev.key === 'Enter') {
+        if (ev.key === 'Enter' || ev.key === 'Tab') {
             this.doCommit();
-            this.consumeEvent(ev);
```
amcclain, line 244: What issues could we see from dropping this?

desktop/cmp/input/DateInput.js
```diff
@@ -43,6 +43,9 @@ DateInput.propTypes = {
     ...HoistInputPropTypes,
     value: PT.oneOfType([PT.instanceOf(Date), PT.instanceOf(LocalDate)]),
 
+    /** True to commit on every change/keystroke, default false. */
```
amcclain, line 46: I think this doc string will need more elaboration re. any special behaviors in the context of the date input, where the text input might not be valid, etc.

amcclain (commented):
Hi John - thanks for looking at this.

I see you’ve added `commitOnChange` with the standard default of false, but then it doesn’t appear that we’re setting to true in `DateEditor` - what’s the story there?  Was that not really needed? 

I am a bit thrown off by the date picker appearing but without the picker icon.  I suppose it’s not a big deal - are we simply unable to get the picker icon to align OK in the grid context, or are there other issues?

(Note this also breaks support for `enableClear` prop - maybe not that important.)

I wish I were feeling better about this set of components - they have given us issues in the past, and I still don't personally feel that I have a good grasp or understanding of how they work and what the options we expose are doing.

desktop/cmp/grid/editors/impl/InlineEditorModel.js
```diff
@@ -47,7 +46,7 @@ export function useInlineEditorModel(component, props, ref, isPopup = false) {
         model: impl,
         bind: 'value',
         commitOnChange: true,
-        ref: composeRefs(ref, impl.ref),
+        ref: impl.ref,
```
lbwexler, line 49: this seems correct to me -- surprised it was not causing us even more problems before

lbwexler:
The `commitOnChange:true` setting happens on all the editors in the hook

lbwexler:
closing in favor of the alternative PR we have been working on

PR #2502, opened 2021-06-29
Date input fix

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/DateInput.js
+++ b/desktop/cmp/input/DateInput.js
@@ -18,7 +18,7 @@ import {isLocalDate, LocalDate} from '@xh/hoist/utils/datetime';
 import {warnIf, withDefault} from '@xh/hoist/utils/js';
 import {getLayoutProps} from '@xh/hoist/utils/react';
 import classNames from 'classnames';
-import {assign, clone, trim} from 'lodash';
+import {assign, castArray, clone, trim} from 'lodash';
 import moment from 'moment';
 import {createRef} from 'react';
 import PT from 'prop-types';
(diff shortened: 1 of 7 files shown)


lbwexler:
Sorry for the delay — this takes several of the key elements of John’s PR, but I think implements it in a way that is perhaps a bit more conservative/targeted.  

The main elements are:

+ supporting multiple parse strings, to allow more flexible parsing.  A few obvious ones are included by default.
+ when the text in the textInput merely changes, always accept a strictly parseable value immediately (either null or a date) .  Commit will more aggressively parse.
+  when the tab entered, be sure to commit the value.  Needed to hack this a bit — we could also consider making TextInput and other inputs generally do this.
+ Don’t consume the <Enter> key.  This brings DateInput into line with TextInput and NumberInput, and allows <Enter> on DateEditor to dismiss the  editor in the way it does for NumberInput and TextInput

desktop/cmp/input/DateInput.js
```diff
@@ -308,26 +328,35 @@ class Model extends HoistInputModel {
...
+
+        return formatString ?? this.addTimeFmt('YYYY-MM-DD');
+    }
+
+    getParseStrings() {
+        const {formatString, parseStrings} = this.props;
+
+        if (parseStrings) return parseStrings;
+
+        const ret = ['YYYY-MM-DD', 'YYYYMMDD', 'YYYY/MM/DD', 'YYYY-M-DD', 'YYYY/M/DD'].map(s => this.addTimeFmt(s));
```
amcclain, line 342: Would anyone use the slash-delimited variations, with the year coming first? "2020/09/17" - is that a thing?

Any sense we should also support `L` - local date format - for e.g. `9/17/2020` in the US? (Unclear to me if this can be used along with the timeFmts - not anything we need to do now, but mentioning it in case that was more what you were looking to accept)

desktop/cmp/input/DateInput.js
```diff
@@ -308,26 +328,35 @@ class Model extends HoistInputModel {
...
-        } else if (timePrecision === 'second') {
-            ret += ' HH:mm:ss';
-        }
         return ret;
     }
 
+    addTimeFmt(fmt) {
+        if (this.timePrecision === 'minute') return fmt + ' HH:mm';
+        if (this.timePrecision === 'second') return fmt + ' HH:mm:ss';
+        return fmt;
```
amcclain, line 351: This keeps us in 24-hour time for now, which is the status quo. Could imagine also supporting `hh:mm:ssa` variants.

amcclain (commented):
This is working well in the client project I'm using to test these inline editing changes. 👍 

Minor questions below.

PR #2504, opened 2021-06-30
Fix streaming update bug in Cube

lbwexler:
We need to Flow aggregation updates all the way to the root, even if there is no change at an intermediate level.

Colin and I discovered this problem in a client app where aggregates on certain dimensions always aggregated to *null* but dimensions above them could still compute aggregates.

Without the changes in this PR, the dimension with null aggregates prevent the update of all aggregates for dimensions above them in the tree.

Hoist P/R Checklist
-------------------

- [] Added CHANGELOG entry, or determined not required.

### DIFF
--- a/data/cube/row/BaseRow.js
+++ b/data/cube/row/BaseRow.js
@@ -124,12 +124,10 @@ export class BaseRow {
             if (canAggregate[name]) {
                 const oldValue = data[name],
                     newValue = field.aggregator.replace(children, oldValue, update);
-                if (oldValue !== newValue) {
-                    update.oldValue = oldValue;
-                    update.newValue = newValue;
-                    myUpdates.push(update);
-                    data[name] = newValue;
-                }
+                update.oldValue = oldValue;
+                update.newValue = newValue;
+                myUpdates.push(update);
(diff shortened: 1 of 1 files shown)


data/cube/row/BaseRow.js
```diff
@@ -124,12 +124,10 @@ export class BaseRow {
...
-                if (oldValue !== newValue) {
-                    update.oldValue = oldValue;
-                    update.newValue = newValue;
-                    myUpdates.push(update);
-                    data[name] = newValue;
-                }
+                update.oldValue = oldValue;
+                update.newValue = newValue;
+                myUpdates.push(update);
+                data[name] = newValue;
```
amcclain, line 130: I really don't have this all loaded in my head, but reading the description of the issue, I'm wondering if it is still a problem in any way that this still only happens within the `if (canAggregate[name]) {}` conditional.

Apologies, but I would need to be led through the whole thing again for me to properly say I've reviewed and understood this.

amcclain (approved):
If you are unsure about this and want to talk through it all with me please let me know. I can't honestly say I understand the change.

PR #2511, opened 2021-07-06
Add GridModel.getIsReadyAsync()

### LINKED ISSUE #2507: GridModel async selection APIs await forever if grid not mounted
Calling e.g. `await gridMode.preselectAsync()` will await forever if the grid component is not mounted, due to await on `isReady` in `ensureSelectionVisibleAsync()` 

https://github.com/xh/hoist-react/blob/98aba786934109f451595e30bd582e7372a55973/cmp/grid/GridModel.js#L514

For a "real world" example, add an await to the Toolbox Contacts app call to this API on:

https://github.com/xh/toolbox/blob/9463583d14109a07451c96b3b6ff4e4772fcfcd6/client-app/src/examples/contact/DirectoryPanelModel.js#L94

then change your displayMode to tiles (so the grid won't be mounted) and refresh the app. The mask will stay up forever.

I propose we add a timeout so this method bails if the grid doesn't get ready within some reasonable amount of time (a few seconds maybe?). That seems like a better tradeoff vs. this attempting to make a strong guarantee that it will await for however long it takes for the grid to show up.

amcclain:
+ Improves `GridModel` async selection methods to ensure they do not wait forever if grid does not mount.
+ Fixes #2507

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -471,7 +471,7 @@ export class GridModel extends HoistModel {
      */
     async selectFirstAsync({ensureVisible = true} = {}) {
         const {selModel} = this,
-            isReady = await this.getIsReadyAsync();
+            isReady = await this.whenReadyAsync();
 
         // No-op if grid failed to enter ready state.
         if (!isReady) return;
@@ -513,7 +513,7 @@ export class GridModel extends HoistModel {
      * render all pending data changes.
      */
     async ensureSelectionVisibleAsync() {
-        const isReady = await this.getIsReadyAsync();
+        const isReady = await this.whenReadyAsync();
 
         // No-op if grid failed to enter ready state.
         if (!isReady) return;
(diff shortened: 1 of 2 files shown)


cmp/grid/GridModel.js
```diff
@@ -1014,6 +1019,23 @@ export class GridModel extends HoistModel {
...
     }
 
+    /**
+     * Returns true as soon as the underlying agGridModel is ready, waiting a limited period
+     * of time if needed to allow the component to initialize. Returns false if grid not ready
+     * by end of timeout to ensure caller does not wait forever (if e.g. grid is not mounted).
+     * @param {number} [timeout] - timeout in ms
+     * @return {Promise<boolean>} - latest ready state of grid
+     */
+    async getIsReadyAsync(timeout = 3 * SECONDS) {
```
lbwexler, line 1029: wonder if this could be called `whenReadyAsync()` instead.  Seems to be slightly clearer -- current name seems like a conventional getter around the flag.

amcclain: Yes, I went back and forth on the name. It felt a bit weird to have `whenReadyAsync` return an `isReady` value that might be false, but not that weird. I'm a-OK with where it landed - I don't expect this to be a common external API in any case - was very much focusing on a clean way to resolve the edge-but-bad case in the ticket. Thanks for merging!

PR #2513, opened 2021-07-06
JsDoc updates in data package

amcclain:
+ Define convenience typedefs for:
  - `RecordId`
  - `RecordOrId`
  - `StoreErrorMap`
  - `RecordErrorMap`

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/GridModel.js
+++ b/cmp/grid/GridModel.js
@@ -441,7 +441,7 @@ export class GridModel extends HoistModel {
     /**
      * Select records in the grid.
      *
-     * @param {(Object[]|Object)} records - single record/ID or array of records/IDs to select.
+     * @param {(RecordOrId|RecordOrId[])} records - single record/ID or array of records/IDs to select.
      * @param {Object} [options]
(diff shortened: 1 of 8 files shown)


lbwexler:
Want to just discuss RecordOrId vs. (Record|RecordId)

amcclain:
Hi - would have been a-OK with `(Record|RecordId)` and dropping `RecordOrId` if it seemed too weird - although it's a common enough pattern to accept both as function args it seemed worth it. See also the very similar example on the [doc for the typedef tag](https://jsdoc.app/tags-typedef.html) - in terms of this being a valid use of typedef generally. That's part of what convinced me it was worth doing.

Much bigger issue really is that in my IDE, `Record` is often resolved to the [proposed new native JS object](https://github.com/tc39/proposal-record-tuple) - an unfortunate naming collision with our core Hoist class.  I was looking at how best to make the minimal change to alias in jsDoc-land so our doc strings reliably worked, but didn't get too far before running out of time.

PR #2516, opened 2021-07-07
Use private fields in #core

lbwexler:
Hoist P/R Checklist
-------------------


amcclain:
Exciting! Looking forward to checking this out very soon.

lbwexler:
Let me know if you still want to do this one?  It's a safe/minor usage, and really just dipping our toes in to these waters.

As I mentioned, the fact that destructuring does not work with these private fields throws a bit of a damper on the whole project.  At one point, it seemed like it was proposed that you would not need to write `this.#foo` in class methods, but rather just `#foo`.   This would have removed alot of the need for destructuring.  But, don't see any evidence for this syntax in the babel behavior, or current class fields proposal.

https://tc39.es/proposal-private-fields/
(obsolete, but allowed the `#foo` access without `this`)

https://github.com/tc39/proposal-class-fields
(current stage 3 proposal, no shortcut access)

PR #2518, opened 2021-07-07
Grid export enhancements

petradish:
+ Add `exportOptions.timeout` soft config
+ Fix: Export column order based on sort order of `exportOptions.columns`

See Hoist-Core PR: https://github.com/xh/hoist-core/pull/200

Hoist P/R Checklist
-------------------

- [N/A ] Reviewed and tested on Mobile, or determined not required.


TomTirapani:
Looks great! Awesome that we were able to get rid of the `streamingCellThreshold` concept by reusing those group background styles.

Overall looks great, but I'm not clear on the need for downgrading tree exports to leaf rows given the hoist-core change. What is the strain of producing tree grids now that we're caching the group styles? Is it simply a way to reduce the cell count?

petradish:
Hi Tom! It is really a lot longer to create a tree-grid as we are currently doing it:
2021-07-08 15:38:05 | GridExportImplService [DEBUG] | Generating Export | export | 50002 rows | 6 cols | 1 depth | petra@xh.io | completed | 27349ms
2021-07-08 15:40:11 | GridExportImplService [DEBUG] | Generating Export | export | 50002 rows | 6 cols | 2 depth | petra@xh.io | completed | 109783ms
I just enhanced our debug to show that for the exact same row and col count (using the Test Grid in admin), just an additional level of depth will take 4 times as long. I think as you add more depth it might take even longer... not 100% sure what is causing this though.

lbwexler:
+  we clearly want to remove the warning around `streamingCellThreshold` -- this has already been changed in Hoist Core
+ We also clearly want to take  column ordering fix.
+ Not sure about fixing the timeout.  Perhaps that should also be configurable in the soft config?
+ Not sure about degrading to just showing the leaves when above the new threshold.  Would an application generally want that?  What about cases where the leaves are not all at the same depth?  (we have grids like this.)

petradish:
Hi @lbwexler could you please review these changes we discussed yesterday? Unwinding the change in hoist-core now

lbwexler:
Petra and I discussed this -- want to think a bit more about the timeout.  

+ Should this be per call?  Default in the config?
+ Also -- if we are contemplating long-running exports, do we need to show some kind of affordance indicating the export is in progress?  Should we allow specifying modal exports?  Petra has noted that the current setup makes it very easy to generate multiple exports and use up all the request threads on a tab.

PR #2519, opened 2021-07-08
Badge component

### LINKED ISSUE #2385: CSS class and/or minimal component for a "badge" indicator
It's often useful to display a small, distinctly styled "badge" inline with other text or a title - typically showing a count or other small indicator that something is new or has content. A tab title is a prime candidate for this kind of thing, like so:

Screen Shot 2021-03-31 at 9 02 56 AM

We could provide a base CSS class in Hoist that got you the default background, sizing, border radius, etc. Ideally it would be relative to the font-size of the containing block, so you could insert a badge inline with other content of various sizes and it would render proportionally (e.g. 0.5em).

If it seemed helpful, we could ship a minimal `badge` component that simply applied this class - e.g. `badge(itemCount)` vs. `span({item: itemCount, className: 'xh-badge'})`.

Ideally the badge class / component should combine well with our standardized intent classes to allow a user to create badges with those standard colors applied and automatically looking good in dark and light themes.

saba-mo:
Closes #2385
See matching toolbox PR

Hoist P/R Checklist
-------------------


lbwexler:
Hi -- this is looking good API-wise, and generally.

Two thoughts:

- 
- - can we make the sizing of this thing relative to the size of what its contained in?
- - any reason not to have this be a single cross-platform component?
-

TomTirapani:
I had a chat with Saba yesterday about the first item, and I actually made the case for the fixed sizing. Before that, it was using `em` sizes to be relative to its container.

Perhaps I'm misunderstanding the intention here, but I don't understand the need for relative sizing. All similar cases I've seen "in the wild" have been a consistent fixed size. Don't feel too strongly either way, just wanted to make it clear that the fixed sizing was my fault, not Saba's :)

lbwexler:
sounds like you are on top of it -- I also don't feel strongly about it.

Whatever works best in practice.

saba-mo:
Thank you for the feedback. I moved the badge directory to hoist-react/cmp so that it is available on all platforms, and I added a mobile example. 

It is currently intended to size relative to its container, per the suggestion in the original ticket, though my CSS rules may need some adjustment. It's also possible we don't want to include the current position prop, but rather let components position the badge themselves (another good suggestion from Tom). While the prop currently allows you to position the badge nicely in the upper right of the tab text, I don't think it will always look good in other places. I also don't think we need a 'bottom' position, but didn't want to just add 'top'/'center' options.

mobile/cmp/header/AppMenuButton.js
```diff
@@ -96,7 +98,14 @@ function buildMenuItems({
             omit: hideThemeItem,
             actionFn: () => XH.toggleTheme(),
             prepareFn: (item) => {
-                item.text = XH.darkTheme ? 'Light Theme' : 'Dark Theme';
+                item.text = hbox({
```
lbwexler, line 101: We'll move this example to toolbox -- discussed with saba

lbwexler:
Merging this to facilitate usage on the snapshot -- we can tweak the look-and-feel as we get more experience with it.
The API seems simple and solid.

PR #2523, opened 2021-07-09
Recheck isReady in Grid validationDisplayReaction

amcclain:
+ I observed this reaction throwing with an NPE on `agApi`, despite the check on the `isReady` param passed into the function.
+ Re-reading `isReady` off the model within the fn body confirmed that it can mis-match the arg - I assume this has to do with the `@computed` behavior on the `isReady` getter.
+ Would be good to review and ensure we have a handle on it, as this pattern is used in a number of places here. (This change only modified this one reaction.)

Hoist P/R Checklist
-------------------


lbwexler:
I am fine with merging this, for sure.   

We definitely want to understand this better -- but do not want to delay merging this fix for too long.   

This particular `isReady` flag and `agApi` have been a vexing problem for us all along.

amcclain:
Updated two other reactions that keyed off and passed in `isReady` to also re-read from the model for a possible short-circuit. I think these changes are safe enough that I'm going to go ahead and merge, especially as @petradish just ran into this same bug it the client app she's developing.

PR #2525, opened 2021-07-10
Style first-child icon within placeholder

amcclain:
+ If first child of placeholder is an icon, render 4x size with reduced opacity.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@
 * New `Exception.timeout()` util to throw exceptions explicitly marked as timeouts, used by
   `Promise.timeout` extension.
 
+### ✨ Style
+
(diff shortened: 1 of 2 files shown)


amcclain:
This could be considered a mildly-breaking change, but I didn't think it was quite worth calling out there. Let me know if you think this either a) doesn't make sense as a default or b) would be more of a breaking change for apps than I think it will be.

See https://github.com/xh/toolbox/pull/503 for TB PR to add an example.

TomTirapani:
I think this looks great, and makes a lot of sense!

We'll have to double-check usages in client apps, but I don't imagine this will be too much of a breaking change. In the examples I can think of, we're doing something very similar in the client app styling (that could not potentially be unwound). Furthermore, it's not too difficult to override this styling if you want something different.

PR #2527, opened 2021-07-13
Column filters

TomTirapani:
See Toolbox PR: https://github.com/xh/toolbox/pull/505

Hoist P/R Checklist
-------------------


CHANGELOG.md
```diff
@@ -2,6 +2,40 @@
...
+      are sourced from. The default `GridFilterModel` sets this to the `GridModel`'s `Store`.
+    + `GridFilterModel.target` - controls which `Store` or cube `View` the generated filter
+      is applied to. The default `GridFilterModel` sets this to the `GridModel`'s `Store`.
+    + `GridFilterModel.initialFilter` - sets an initial filter in the `GridFilterModel`.
+  + New `Column.filterable` to enable a column-level filter affordance in the column header.
+    Note the that the `GridModel` must have a `filterModel`. The filter control offers two tabs - a
+    "Values" tab for enumerative, value-based filter, and a "Custom" tab to build complex filtering
+    queries with multiple clauses.
++ Cube `View` now accepts a `loadModel` property - pass to `Cube.createView()` to link view
+  updates to a `PendingTaskMask`. If `loadModel` is not provided, the `View` will create one.
```
lbwexler, line 29: *PendingTaskModel*

lbwexler: We probably don't need to go in to quite so much detail in the Changelog as to how the API works -- We want to keep it readable and a high level.  I think its fine to point people to the docs of the key classes.

cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -4,157 +4,97 @@
...
+     *      specified as the function to be used for the matching. (If true a default "word start"
      *      matching against the formatted value will be used.)
      * @param {boolean} [c.forceSelection] - true to require value entered to be an available value
-     *      for '=' and '!=' operators.  Defaults to false.
-     * @param {FilterOptionValueRendererCb} [c.valueRenderer] - function to produce a suitably
+     *      for '=' and '!=' operators. Defaults to false.
+     * @param {FilterChooserValueRendererCb} [c.valueRenderer] - function to produce a suitably
      *      formatted string for display to the user for any given field value.
-     * @param {FilterOptionValueParserCb} [c.valueParser] - function to parse user's input from a
+     * @param {FilterChooserFieldSpec} [c.valueParser] - function to parse user's input from a
```
lbwexler, line 52: This should be {FilterChooserValueParserCb} no?

cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -239,23 +158,14 @@ export class FilterChooserFieldSpec extends HoistBase {
...
-        return this.isValueType ? ['=', '!=', 'like'] : ['>', '>=', '<', '<=', '=', '!='];
-    }
-
-    loadValuesFromStore() {
-        const {field, store} = this,
+    loadValuesFromSource() {
+        const {field, source} = this,
             values = new Set();
 
         // Note use of unfiltered recordset here to source suggest values. This allows chooser to
```
lbwexler, line 165: interesting that this behavior is exactly opposite what we are doing with the column filters enumeration

cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -4,157 +4,97 @@
...
 
-    /** @member {?Array} - data values available for suggestion. */
-    values;
-
-    /** @member {FilterOptionValueRendererCb} */
+    /** @member {FieldSpecValueRendererCb} */
     valueRenderer;
 
-    /** @member {FilterOptionValueParserCb} */
+    /** @member {FieldSpecValueParserCb} */
```
lbwexler, line 34: These  jsdoc names are  wrong, no?  I think `FilterChooserValueParserCb` and FilterChooserRendererCb

cmp/filter/FilterChooserModel.js
```diff
@@ -36,15 +37,18 @@ export class FilterChooserModel extends HoistModel {
     /** @member {Filter[]} */
     @observable.ref favorites = [];
 
-    /** @member {Store} */
-    sourceStore;
+    /** @member {(Store|View)} */
+    valueSource;
```
lbwexler, line 41: want to make sure I understand why we need to allow binding to the View  -- the View itself could be bound to a store.

cmp/filter/FilterChooserModel.js
```diff
@@ -136,12 +151,17 @@ export class FilterChooserModel extends HoistModel {
         this.setValue(value);
         this.setFavorites(favorites);
 
-        if (targetStore) {
+        if (target) {
```
lbwexler, line 154: given the two-way binding here, should we be renaming this argument to `bind` rather than `target`?

TomTirapani: I like this idea - less misleading now that we're doing two-way binding, and we're already in the business of doing a breaking rename here. Will do the rename. 👍

CHANGELOG.md
TomTirapani: Thanks Lee - will streamline it a bit, and make sure all the relevant info can be found in the right class docs

cmp/filter/FilterChooserModel.js
TomTirapani: It was your requirement, Lee :) My understanding is that filtering the cube view may be preferable because it happens before the view calculates its aggregations.

amcclain (commented):
Submitting partial review comments- only got through a small fraction of it so far

admin/tabs/activity/tracking/detail/ActivityDetailModel.js
```diff
@@ -51,26 +55,27 @@ export class ActivityDetailModel extends HoistModel {
                     }
                 },
                 {field: 'id', headerName: 'Entry ID', width: 100, align: 'right', hidden: true},
-                {field: 'username', ...usernameCol},
+                {field: 'username', filterable: true, ...usernameCol},
```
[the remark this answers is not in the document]
TomTirapani: I chose `filterable` to match the other functionality flags on column, e.g. `editable`, `sortable`, `resizable` etc. Agree that filterable is a little awkward, but didn't want to make it an outlier in the API...

admin/tabs/activity/clienterrors/ClientErrorsModel.js
```diff
@@ -111,8 +111,8 @@ export class ClientErrorsModel extends HoistModel {
         });
 
         this.filterChooserModel = new FilterChooserModel({
-            sourceStore: this.gridModel.store,
-            targetStore: this.gridModel.store,
+            bind: this.gridModel.store,
+            valueSource: this.gridModel.store,
```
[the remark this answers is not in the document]
TomTirapani: This sounds like a good idea... also struggling to come up with a scenario where you would use bind but not valueSource. In the overwhelming majority of cases you'd want them to be the same, so think it's okay to default as described.

cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -4,157 +4,97 @@
...
-        const {fieldType} = this;
-        return fieldType == FieldType.DATE || fieldType == FieldType.LOCAL_DATE;
-    }
-
-    get isNumericFieldType() {
-        const {fieldType} = this;
-        return fieldType == FieldType.INT || fieldType == FieldType.NUMBER;
+    supportsSuggestions(op) {
+        return this.values &&
+            this.suggestValues &&
```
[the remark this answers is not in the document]
TomTirapani: Double-checked and it's non-falsey checks all the way, which is correct. 👍

admin/tabs/activity/tracking/detail/ActivityDetailModel.js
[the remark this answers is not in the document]
amcclain: I think the consistency is ultimately worth it - i.e. stick with `filterable`

data/cube/Cube.js
```diff
@@ -106,19 +108,20 @@ export class Cube extends HoistBase {
      * @param {Object} query - Config for query defining the shape of the view.
      * @returns {Object[]} - data containing the results of the query as a hierarchical set of rows.
```
TomTirapani, line 109: Returns a promise<Object[]>

data/cube/View.js
```diff
@@ -113,22 +146,46 @@ export class View extends HoistBase {
...
+    getField(name) {
+        return this.fields.find(it => it.name === name);
+    }
+
+    /**
+     * @param {(Store[]|Store)} stores - Stores to be loaded/reloaded with data from this view.
+     */
+    setStores(stores) {
+        this.stores = castArray(stores);
+        this.fullUpdateAsync().linkTo(this.loadModel);
```
TomTirapani, line 163: We don't need to full update here... can just update the stores.

data/cube/Cube.js
```diff
@@ -106,19 +108,20 @@ export class Cube extends HoistBase {
...
      */
-    executeQuery(query) {
+    async executeQueryAsync(query) {
         query = new Query({...query, cube: this});
-        const view = new View({query}),
-            rows = view.result.rows;
+        const view = new View({query});
+        await when(() => view.result);
```
TomTirapani, line 114: Replace with View.initAsync (that does the first update)

data/cube/View.js
```diff
@@ -94,8 +120,16 @@ export class View extends HoistBase {
...
+        const newQuery = this.query.clone(overrides);
+        if (this.query.equals(newQuery)) return;
+
+        // If anything other than just filter changing, blow away row cache
+        if (!isEqual(keys(overrides), ['filter'])) {
+            this._rowCache.clear();
+        }
+
+        this.query = newQuery;
+        this.fullUpdateAsync().linkTo(this.loadModel);
```
petradish, line 132: I think we need to await this and make updateQuery async - having trouble with my masking operations in the client app

lbwexler: am looking at this now.  Not sure why this method needs to block on the update being completed?  Seems like potentially a different problem.  Are you passing in a shared PendingTaskModel?  Is it set mode = 'all'?

petradish: ah! i think that's it! it's set to the 'last' default mode

Issue #2492, opened 2021-06-23
Add GridModel.setColumnState() API method


amcclain:
See #2314 - app code should expect a clean, public API for taking a previously saved copy of `gridModel.columnState` and applying it back to a GridModel in one call.

lbwexler:
don -- any progress on this?  Seemed like you had a branch going.  Looking to close out this one

febbraiod:
Yep slipped off my radar, I think I might have thought I already opened the PR:

https://github.com/xh/hoist-react/pull/2540/files

PR #2540, opened 2021-07-23
Add GridModel.setColumnState() API method. Fixes: #2492

febbraiod:
Hoist P/R Checklist
-------------------


cmp/grid/GridModel.js
```diff
@@ -728,6 +728,11 @@ export class GridModel extends HoistModel {
             .map(({colId, width, hidden, pinned}) => ({colId, width, hidden, pinned}));
     }
 
+    setColumnState(colState) {
+        colState = this.persistenceModel.cleanColumnState(colState);
```
lbwexler, line 732: this will throw if there is no persistenceModel.  But its totally valid (and common) for a grid not to have a persistence model.  Also, needs jsdoc

cmp/grid/impl/GridPersistenceModel.js
```diff
@@ -95,7 +94,7 @@ export class GridPersistenceModel extends HoistModel {
         const {gridModel, state} = this;
         if (!state.columns) return;
 
-        const colState = this.cleanColumnState(state.columns);
+        const colState = gridModel.cleanColumnState(state.columns);
```
lbwexler, line 97: just use setColumnState() here

cmp/grid/impl/GridPersistenceModel.js
```diff
@@ -86,7 +85,7 @@ export class GridPersistenceModel extends HoistModel {
         return {
             track: () => this.gridModel.columnState,
             run: (columnState) => {
-                this.patchState({columns: this.cleanColumnState(columnState)});
+                this.patchState({columns: this.gridModel.cleanColumnState(columnState)});
```
lbwexler, line 88: why does this columnState have to be cleaned again if it is coming directly out of gridModel.  Isn't that by definition clean?

cmp/grid/impl/GridPersistenceModel.js
febbraiod: Probably to remove widths? IE:

// Remove the width from any non-resizable column - we don't want to track those widths as
// they are set programmatically (e.g. fixed / action columns), and saved state should not
// conflict with any code-level updates to their widths.

febbraiod: 10-4

cmp/grid/impl/GridPersistenceModel.js
febbraiod: Probably to remove widths? IE:

// Remove the width from any non-resizable column - we don't want to track those widths as
// they are set programmatically (e.g. fixed / action columns), and saved state should not
// conflict with any code-level updates to their widths.

lbwexler: can we have the method on the persistenceModel renamed to removeWidths and only do that?  Confusing to have the duplication of the other code.

PR #2544, opened 2021-07-26
Apply identical styling to upper level nodes with/without children in tree grids

### LINKED ISSUE #2414: Improve rendering of GridModel.treeStyle when some top-level nodes have no children.
We have a number of usages in a client app where some root level nodes have children, and some don't.  
(think, e.g. Traders and Trader Groups).  

They should all have the identical tree styling to indicate that they are siblings, but the nodes without children are themselves styled as children, which ends up being quite confusing.

saba-mo:
Closes #2414

Hoist P/R Checklist
-------------------


amcclain:
I have some questions / concerns here - will reach out to review together momentarily, but please hold off on merging this until we get a chance to discuss - thanks!

cmp/grid/enums/TreeStyle.js
```diff
@@ -24,14 +24,14 @@ export const TreeStyle = Object.freeze({
 });
 
 
-export function getTreeStyleClasses(treeStyle) {
+export function getTreeStyleClasses(treeStyle, maxDepth) {
     switch (treeStyle) {
         case TreeStyle.HIGHLIGHTS:
-            return 'xh-grid--tree-style-highlights';
+            return `xh-grid--tree-style-highlights-level-${maxDepth}`;
```
amcclain, line 30: We would still want to have a base `xh-grid--tree-style-highlights` class applied to generally key off of. If we went this route in general, we would apply another class with maxDepth baked in, as an addition.

cmp/grid/Grid.scss
```diff
@@ -209,28 +209,28 @@
...
+          }
         }
       }
     }
   }
 
   &.xh-grid--tree-style-borders {
-    .ag-row-level-0[class*=" ag-row-group"],
-    .ag-row-level-0[class^="ag-row-group"], {
+    .ag-row-level-0 {
```
amcclain, line 233: Wondering if there was some other reason why we needed those extra selectors here - I think there must have been... Would want to check if any other rows we're not accounting for could start matching (e.g. summary rows?)

amcclain (commented):
Wrote up a long response here but then GitHub lost it - I want to keep thinking on this - concerns over performance w.r.t. that reaction to calc maxdepth for this purpose, re-render of grid if/when that changes. Also concerned re. generating the big block of complex/nested level-specific rules - want to ensure we don't end up creating any CSS performance issues, which can be something to consider.

cmp/grid/Grid.js
```diff
@@ -143,6 +143,16 @@ class GridLocalModel extends HoistModel {
...
 
+    @computed
+    get depthOfLowestLeaf() {
+        let maxDepth = 0;
+        if (this.isHierarchical) {
+            const {records} = this.model.store;
+            records.forEach(record => maxDepth = Math.max(record.treePath.length - 1, maxDepth));
+        }
+        return maxDepth;
+    }
```
amcclain, line 154: @lbwexler - this seemed like a potentially expensive reaction to be running against `records` here, for this purpose alone. 

It's possible there is some completely alternate approach that doesn't depend on getting this stat out of the store, but I have not come up with any such alternative.  If we do need this state, could it be calculated/produced more efficiently/centrally within the Store or RecordSet classes?

lbwexler: not as worried about the performance of this -- we do a more signifigant O(n) operation when the records change just to generate any transaction -- this seems like a minor addition to that work.  

Not obvious that it would be signifigantly faster computing in recordset itself -- maybe marginally so, in that it has an internal representation of children only that we could limit the check to . could add a getMaxDepth() method on recordset if we wanted to, for clarity and to take advantage of

cmp/grid/Grid.scss
saba-mo: Just tested and summary rows are not taking on any new styles.

Issue #2547, opened 2021-07-26
GridModel config to control number of clicks required to start inline editing


amcclain:
Ag-Grid provides a `singleClickEdit` config to enable editing on the first click, instead of the default double-click requirement. This is a relatively common option for editable grids to enable - it should be possible to enable via a Hoist config. Another ag-Grid config `suppressClickEdit` disables editing on click entirely.

We could:
* Add `GridModel.singleClickEdit:boolean` config to mirror the ag-Grid config.
* Add `GridModel.singleClickEditing:boolean` to better match our existing `fullRowEditing` config
* Add `GridModel.clicksToEdit:int` config - would take `2` (default), `1`,  or `-1/0` to disable. 
  * This would cover both of the AG configs above.
  * Disabling editing via click would be for use when there as some other predefined way to programatically start editing a particular row/cell.

PR #2554, opened 2021-07-27
GridModel.clicksToEdit implemented.

cnrudd:
This resolves issue #2547.

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch / PR: https://github.com/xh/toolbox/pull/507

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -224,8 +224,7 @@ class GridLocalModel extends HoistModel {
             autoSizePadding: 3, // tighten up cells for ag-Grid native autosizing.  Remove when Hoist autosizing no longer experimental,
             editType: model.fullRowEditing ? 'fullRow' : undefined,
             singleClickEdit: model.clicksToEdit === 1,
-            suppressClickEdit: model.clicksToEdit <= 0
-
+            suppressClickEdit: model.clicksToEdit !== 1 && model.clicksToEdit !== 2
         };
 
         // Platform specific defaults
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -222,7 +222,10 @@ class GridLocalModel extends HoistModel {
                 suppressSizeToFit: true // Without this the auto group col will get shrunk when we size to fit
             },
             autoSizePadding: 3, // tighten up cells for ag-Grid native autosizing.  Remove when Hoist autosizing no longer experimental,
-            editType: model.fullRowEditing ? 'fullRow' : undefined
+            editType: model.fullRowEditing ? 'fullRow' : undefined,
+            singleClickEdit: model.clicksToEdit === 1,
+            suppressClickEdit: model.clicksToEdit <= 0
+
         };
 
         // Platform specific defaults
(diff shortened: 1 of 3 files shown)


cmp/grid/GridModel.js
```diff
@@ -172,6 +172,8 @@ export class GridModel extends HoistModel {
 
     /**
      * @param {Object} c - GridModel configuration.
+     * @param {int} [c.clicksToEdit] -  Set how many clicks are needed to open an editable cell for edting.
```
amcclain, line 175: Type here would be `{number}`

amcclain (commented):
Thank you for putting this implementation together - much appreciated and eager to get it in.

I had suggested `clicksToEdit` in the linked issue - and I still think it might be the best choice for this API to bundle up both 1 vs. 2 clicks and disabling click handling.  It is a bit weird - i.e. you can't spec any arbitrary higher number - which sounds silly when I type it - but still...

I am leaning towards going with this proposal, but if any other ideas are out there - now would be the time.  @haynesjm42 / @lbwexler  any thoughts?

cmp/grid/GridModel.js
cnrudd: Ok.  I saw {int} in these 2 places, so thought it was accepted:
https://github.com/xh/hoist-react/blob/develop/utils/js/LangUtils.js#L210
https://github.com/xh/toolbox/blob/develop/client-app/src/core/svc/PortfolioService.js#L21

JSDoc wouldn't choke on "int", but are we trying to stick to real javascript types?

cmp/grid/Grid.js
```diff
@@ -222,7 +222,10 @@ class GridLocalModel extends HoistModel {
                 suppressSizeToFit: true // Without this the auto group col will get shrunk when we size to fit
             },
             autoSizePadding: 3, // tighten up cells for ag-Grid native autosizing.  Remove when Hoist autosizing no longer experimental,
-            editType: model.fullRowEditing ? 'fullRow' : undefined
+            editType: model.fullRowEditing ? 'fullRow' : undefined,
+            singleClickEdit: model.clicksToEdit === 1,
+            suppressClickEdit: model.clicksToEdit <= 0
```
lbwexler, line 227: why don't we make suppress be != 1 && != 2

lbwexler:
I am fine with this.  Seems easier to reason about than the agOptions

lbwexler:
colin and I discussed -- could we imagine wanting to set this on the column level, especially if we e.g. wanted to edit checkboxes on a single click, but others with more clicks.  Can wait to see if that is needed.

But in general users will expect consistency here, I think!

cmp/grid/GridModel.js
amcclain: Yes - jsDoc will let you put literally anything in there, but if it doesn't resolve to a known JS primitive or another "namepath" that jsdoc can resolve, I don't think it will have any (meaningful) effect.

I'll clean up those other two references, thanks for catching. 👍

cmp/grid/Grid.js
amcclain: Done - we had the doc comment saying as much, so seemed worth implementing that way

PR #2556, opened 2021-07-28
Programmatic dismissal of Toast/Improve Export messaging

lbwexler:
+ expose ToastModel
+ improve toasting and display of exceptions from GridExportService
+ misc cleanup

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/appcontainer/ToastSource.js
+++ b/desktop/appcontainer/ToastSource.js
@@ -41,13 +41,10 @@ class LocalModel extends HoistModel {
         });
     }
 
-    //------------------------
-    // Implementation
-    //------------------------
+    /** @param {ToastModel[]} models */
     displayPendingToasts(models) {
         models.forEach(model => {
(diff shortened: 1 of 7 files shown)


desktop/appcontainer/ToastSource.js
```diff
@@ -80,8 +83,6 @@ class LocalModel extends HoistModel {
             container = containerRef ? containerRef : document.body,
             className = `xh-toast-container ${containerRef ? 'xh-toast-container--anchored' : ''}`;
 
-        position = withDefault(position, 'top');
```
amcclain, line 83: I thought it made more sense to default the position with the `ToastModel` ctor, which I did. 

Given that this entire class is doc'd as `@private` I am assuming it's safe the change the behavior of this method.

amcclain (approved):
Looks great.  I had removed the `return` of the `ToastModel` from the `XH` methods precisely because it had been marked private, but I meant to flag and revisit that - thanks for doing so here, with the correct fix/approach.

Issue #1331, opened 2019-08-27
Classes applied via GridModel.rowClassFn / Column.cellClassFn are additive only


amcclain:
As per the [the ag-Grid docs](https://www.ag-grid.com/javascript-grid-row-styles/#refresh-of-styles) row CSS classes applied via `getRowClass()` will only accumulate - updates to the data will not remove previously applied classes.  This is the method we rely on for our `GridModel.rowClassFn`, so our contract is not working as expected.

Same applies for [cellClassFn](https://www.ag-grid.com/javascript-grid-cell-styles/#refresh-of-styles)

haynesjm42:
Taking this over as will need the add/remove control for #2353

PR #2559, opened 2021-07-28
Apply CSS class to inline-editable grid cells

### LINKED ISSUE #2553: Editable Cells in a grid should indicate that they are editable
This could be something left up to the app developer, but it might be nice to have in the framework.
If built into the framework, it could be a config, either at gridmodel level or column level, that turns on a css rule on a cell when it is not being edited that helps indicate that the cell is editable.

The CSS could:
1: switch to pointer cursor on cell, to indicate that the cell would "respond in some way to a click"
2: put some special border or inset shadow on the cell to make it special.

The CSS could of course be overridden at the app level.

This feature would be especially useful in grids where only a few of the N columns are editable.  With this feature, new users don't have to guess or read the app docs to find out which columns are editable.

amcclain:
+ New `.xh-cell--editable` applied, but with no default styling provided by Hoist.
+ Changed previous `.xh-invalid-cell` to `.xh-cell--invalid` for consistency.
+ Fixes #2553

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -438,7 +438,7 @@ export class Column {
         }
 
         // Tooltip Handling
-        const {tooltip, tooltipElement, editor} = this,
+        const {tooltip, tooltipElement, editable, editor} = this,
             tooltipSpec = tooltipElement ?? tooltip;
 
         if (tooltipSpec || editor) {
@@ -610,7 +610,16 @@ export class Column {
                 throw XH.exception('Column editor must be a HoistComponent or a render function');
             });
             ret.cellClassRules = {
-                'xh-invalid-cell': ({data: record}) => record && !isEmpty(record.errors[field])
+                'xh-cell--invalid': (agParams) => {
+                    const record = agParams.data;
+                    return record && !isEmpty(record.errors[field]);
+                },
+                'xh-cell--editable': (agParams) => {
+                    const record = agParams.data;
+                    return isFunction(editable) ?
+                        editable({record, store: record.store, gridModel, column: this, agParams}) :
+                        editable;
+                }
             };
         }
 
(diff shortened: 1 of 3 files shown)


amcclain:
Just pushed Toolbox update https://github.com/xh/toolbox/commit/744d1472eb628e0fe6551cf373ec6a0ac0320bd1 to `develop` to demo the use of an app-specific style to highlight editable cells - assuming we'll merge this PR or some variation on it shortly.

cmp/grid/columns/Column.js
```diff
@@ -610,7 +610,16 @@ export class Column {
                 throw XH.exception('Column editor must be a HoistComponent or a render function');
             });
             ret.cellClassRules = {
```
cnrudd, line 612: I think ret.cellClassRules and agOptions.cellClassRules need to specially merged so as to really be combined together.  As is now, an app's agOptions.cellClassRules will wipe out ret.cellClassRules won't it?

I have a client app with 12 instances of cellClassRules...

cnrudd (changes requested):

cmp/grid/columns/Column.js
amcclain: Yes, agree 100% and that's something that I want to handle very soon via longstanding ticket #1331.

This PR doesn't change the current (suboptimal) behavior, which was introduced with the addition of the validation display to the first inline editing release.

Let me see if I can bump up that other ticket and get a clean approach into a dedicated PR for this overall Hoist release. I don't think it should hold up this merge, but agree it's a priority.

cnrudd: ok, sounds good.

amcclain:
Reviewed with Lee

Issue #2551, opened 2021-07-27
Unreliable async isReady reporting within GridModel


amcclain:
Noticed another previously-unfamiliar exception related to grid ready state, thrown from within `selectFirstAsync()` when swapping some panels/grids around in the process of dismissing a complex form. 

The return from `await this.whenReadyAsync()` below can be true, while a direct re-read of the property can evaluate to false:

```
async selectFirstAsync({ensureVisible = true} = {}) {
        const {selModel} = this,
            isReady = await this.whenReadyAsync();

        // No-op if grid failed to enter ready state.
        if (!isReady) return;

        // Observed to log "xh-id-89 true false"
        console.log(this.xhId, isReady, this.isReady);

...method continues
```

lbwexler:
wonder if its possible that this is related to the issues we had in the client app in question where grids were re-mounting on other component's GridModel via context

amcclain:
> wonder if its possible that this is related to the issues we had in the client app in question where grids were re-mounting on other component's GridModel via context

I am not positive, but I think that this is likely enough that we should close this now - will keep an eye out!

amcclain:
I could reproduce this again reliably, logging the `xhId` of my `GridModel` and `AgGridModel` and comparing the async return from `whenReadyAsync()` with a direct sync read from `this.agGridModel.isReady` and seeing them report different results.

Re-opening and will push update to branch to re-read isReady after these calls.

PR #2561, opened 2021-07-29
Add public GridModel APIs to stop/start editing

### LINKED ISSUE #2508: Add GridModel API to programmatically start inline editing for a given record/row
Would be especially useful when adding a new record to a store - user would expect to land in the first editable field on the newly added row with input focused, if possible.

amcclain:
+ New public methods `beginEditAsync()` and `endEditAsync()`.
+ New observable `isEditing` property.
+ Fixes #2508

Hoist P/R Checklist
-------------------


amcclain:
For consideration with https://github.com/xh/toolbox/pull/508

cmp/grid/columns/Column.js
```diff
@@ -344,8 +347,16 @@ export class Column {
     }
 
     /**
-     * Produce a Column definition appropriate for AG Grid.
+     * @param {Record} record
+     * @param {Object} [agParams] - raw params from agGrid event, if available.
```
amcclain, line 351: I'm calling this method from `GridModel.beginEditAsync()`, where I want to eval for a record but don't have the `agParams` as it's not being triggered from a grid event.

I wasn't sure as to value of that last arg - I think we have it for consistency with other callbacks in our API that often are attached to agParams, but it might be worse to *sometimes* pass it to the fn than to never pass it at all. The fn is getting a lot of other info from the rec, store, col, etc.

cmp/grid/GridModel.js
```diff
@@ -1031,6 +1032,93 @@ export class GridModel extends HoistModel {
...
+     * @return {Promise<void>}
+     */
+    async endEditAsync(dropPendingChanges = false) {
+        const isReady = await this.whenReadyAsync();
+        if (!isReady) return;
+
+        this.agApi.stopEditing(dropPendingChanges);
+    }
+
+    /** @package */
```
lbwexler, line 1112: well, this is a new jsdoc pattern

cmp/grid/GridModel.js
```diff
@@ -1031,6 +1032,93 @@ export class GridModel extends HoistModel {
...
+    /**
+     * Begin an inline editing session.
+     * @param {RecordOrId} [recOrId] - Record/ID to edit. If unspecified, the first selected Record
+     *      will be used, if any, or the first overall Record in the grid.
+     * @param {string} [colId] - ID of column on which to start editing. If unspecified, the first
+     *      editable column will be used.
+     * @return {Promise<void>}
+     */
+    async beginEditAsync({record, colId} = {}) {
+        const isReady = await this.whenReadyAsync();
```
lbwexler, line 1044: should we have a variant of this that throws if the grid cannot become ready?  

e.g. await this.ensureReadyAsync();

cmp/grid/GridModel.js
lbwexler: can't these handlers be annotated with @action?

cmp/grid/columns/Column.js
lbwexler: agree we should take the lowest common denominator approach on what gets passed to editable -- record/column/gridModel should be all we need

cmp/grid/columns/Column.js
```diff
@@ -658,6 +661,11 @@ export class Column {
         return null;
     }
 
+    // Normalize provided function or a boolean into a function form.
+    parseEditable(editable) {
+        return isFunction(editable) ? editable : () => !!editable;
```
lbwexler, line 666: seems a bit weird to lose the ability for app to read the editable flag on the column later -- might be useful for troubleshooting or app code in the very common case that booleans are being used.

cnrudd:
I've run the demo in Toolbox.  The user experience feels good.

cmp/grid/GridModel.js
amcclain: Yes, will just need to wrap when applied in comp code to get the scope right - can do that if preferred

For clarity, `@package` is to declare "these are called outside of this class/file, but they aren't considered a part of the general API". We have only two usages in hoist already, but I think it's helpful enough to use where appropriate

cmp/grid/GridModel.js
amcclain: As discussed, agree that would be a good addition - something we could look at along with #2551

PR #2563, opened 2021-07-29
Support rowClassRules/cellClassRules configs

### LINKED ISSUE #1331: Classes applied via GridModel.rowClassFn / Column.cellClassFn are additive only

amcclain:
+ Add as top-level configs to GridModel and Column, respectively.
+ Throw if detected within agOptions to ensure migration.
+ Remove obsolete `GridModel.compact` config - deprecated in v29.
+ Fixes #1331 

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### 🎁 New Features
 
+* New `GridModel.rowClassRules` and `Column.cellClassRules` configs added. Previously apps needed to
+  use `agOptions` to dynamically apply and remove CSS classes using either of these options - now
+  they are fully supported by Hoist.
+  * ⚠ Note that, to avoid conflicts with internal usages of these configs, Hoist will check and
+    throw if either is passed via `agOptions`. Apps only need to move their configs to the new
+    location - the shape of the rules object does *not* need to change.
 * New `GridAutosizeOptions.includeCollapsedChildren` config controls whether values from collapsed
   (i.e. hidden) child records should be measured when computing column sizes. Default of `false`
   improves autosize performance for large tree grids and should generally match user expectations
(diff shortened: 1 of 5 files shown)


amcclain:
See also https://github.com/xh/toolbox/pull/509 for TB demo of `cellClassRules` - need to find a reasonable place to use `rowClassRules` also (didn't come up with anything I really liked within SampleGrid, which is getting very busy - can look elsewhere).

Main decision points I wanted to discuss before finalizing:
- Are we OK just going with the agParams and not trying to normalize the signatures of the rule functions? (I.e. by transforming to pass in record, store, gridModel, etc as we typically do)?
- OK to throw if either config provided via agOptions?
- OK to maintain support for pre-existing `rowClassFn` and `cellClass` configs in addition to new configs?

cnrudd (approved):
Does this need a changelog entry?

Otherwise all good.

amcclain:
I added a changelog entry to move this forward in its current configuration.  Always happy to discuss, but I think this strikes the best balance:

1. Keeps the shape of the objects the same as it was for direct agGrid usage - should make it easy for apps to migrate
2. Keeps the apiDeprecated check in place, which will throw on existing app usages within agOptions, but calls out this caveat in the changelog. Given (1) I think this is reasonable to keep as a "minor breaking change" without triggering a new major version - use of `agOptions` by apps always carries caveats.
3. We could consider removing the alternate configs in a future release, but I don't think we need to force the issue now, and ag-Grid supports both.  The pre-existing ones are easier to use if you aren't worried about data changing and requiring them to be applied and removed within a session.

Issue #1825, opened 2020-04-07
Panel support for rendering an ErrorMessage in place of items


haynesjm42:
Would be nice if we had some easy way to show a mask over a panel which indicated that an error occurred. This could be a new type of mask component or an extension of the current mask component.

Could see how this could work nicely with `Panel` and `mask: 'onLoad`, we keep track of the last request error in `LoadSupport` models now, could determine if an error occurred on the last request to show the mask and default the message to the exception message.

For instances where you want something more minimal could also support this in the panel loading indicator by having it render an error icon and showing error information in a tooltip on the indicator.

amcclain:
Would be interesting to revisit this now that we have `ErrorMessage`.

I was writing a small new component yesterday that loads some data via its `LoadSupport` enabled model and shows the data in a grid - like so:
```
render({model}) {
        const error = model.lastLoadException;
        return panel({
            item: error
                ? errorMessage({
                      error,
                      actionButtonProps: {onClick: () => model.loadAsync()}
                  })
                : grid(),
            mask: 'onLoad'
        });
    }
```

Not bad, but unfortunate how the error case dominates the render method.  I was thinking it would be helpful if `errorMessage` had a more compact way to enable a default refresh option, which is a different ticket, but I think the idea here is worth exploring more.  I think I would prefer having the component swapped out entirely with a pre-configured `errorMessage` vs. a mask, but happy to discuss.

amcclain:
Closing as not directly actionable as a single Hoist feature - please see comment in the PR where we tried this out - https://github.com/xh/hoist-react/pull/2568#issuecomment-892213430

PR #2568, opened 2021-08-02
Panel error feature

dsilberger:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * New `GridModel.clicksToEdit` config controls the number of clicks required to begin inline-editing
   a grid cell (default remains 2).
+* `Panel` now has an `error` property to specify display in the event of an error.
 
 ### 🐞 Bug Fixes
 
--- a/desktop/cmp/panel/Panel.js
+++ b/desktop/cmp/panel/Panel.js
@@ -103,14 +103,14 @@ export const [Panel, panel] = hoistCmp.withFactory({
         let innerContents = null,
             coreContents = null;
 
-        if (contextModel.lastLoadException) {
-            if (isValidElement(error)) {
-                innerContents = error;
-            } else if (error === 'lastLoadException') {
-                innerContents = errorMessage({error: contextModel.lastLoadException});
-            } else {
-                innerContents = Children.toArray(children);
-            }
+        if (error === 'lastLoadException') {
+            innerContents = contextModel.lastLoadException ?
+                errorMessage({error: contextModel.lastLoadException}) :
+                Children.toArray(children);
+        } else if (typeof error === 'string') {
+            innerContents = errorMessage({error});
+        } else if (isValidElement(error)) {
+            innerContents = error;
         } else {
             innerContents = Children.toArray(children);
         }
@@ -201,6 +201,7 @@ Panel.propTypes = {
      * Error behavior customizer. When unset, the panel will load using local state, even upon error.
      * To replace panel contents in an error state, set to:
      *   + a ReactElement for custom error display,
+     *   + a string, to be displayed as an errorMessage,
      *   + the string 'lastLoadException' for a default errorMesssage bound to the contextModel.
      */
     error: PT.oneOfType([PT.string, PT.element]),
--- a/desktop/cmp/panel/Panel.js
+++ b/desktop/cmp/panel/Panel.js
@@ -20,7 +20,7 @@ import {toolbar} from '@xh/hoist/desktop/cmp/toolbar';
 import {useContextMenu, useHotkeys} from '@xh/hoist/desktop/hooks';
 import {PendingTaskModel} from '@xh/hoist/utils/async';
 import {splitLayoutProps} from '@xh/hoist/utils/react';
-import {omitBy} from 'lodash';
+import {omitBy, isString} from 'lodash';
 import PT from 'prop-types';
 import {isValidElement, useRef, Children} from 'react';
 import {panelHeader} from './impl/PanelHeader';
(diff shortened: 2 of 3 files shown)


lbwexler (commented):
Dan and I went over these changes together and they look very good to me to be what was envisioned in the ticket.

Two thoughts 
1) Having some questions about the value of 'lastLoadException' token?  Seems like nice symmetry with `mask: 'onLoad'`, but in practice are we really going to use it a lot?  Seems like we will still often want to customize the ErrorComponent, and how hard is it to write `error: model.lastLoadException`.  When we introduced the special syntax for mask, we immediately cleaned up a lot of existing code -- not the case here.

2) Note that toolbars are still shown.  The current solution is flexible, in that they can still always be turned off by application, but just something to be aware of.

desktop/cmp/panel/Panel.js
```diff
@@ -147,6 +149,16 @@ export const [Panel, panel] = hoistCmp.withFactory({
 
 });
 
+function parseError(error, contextModel) {
+    if (error === 'lastLoadException') {
+        const {lastLoadException} = contextModel;
+        return lastLoadException ? errorMessage({error: lastLoadException}) : null;
+    }
+    if (isString(error))        return errorMessage({error});
+    if (isValidElement(error))  return error;
```
amcclain, line 158: We should be able to take an actual `Error` as well - I can imagine a model choosing to save a reference to an actual error, distinct from loadSupport (i.e. it might not be implementing `doLoadAsync()`).

desktop/cmp/panel/Panel.js
```diff
@@ -106,7 +108,7 @@ export const [Panel, panel] = hoistCmp.withFactory({
                 style: {display: collapsed ? 'none' : 'flex'},
                 items: [
                     parseToolbar(tbar),
-                    ...Children.toArray(children),
+                    parseError(error, contextModel) ?? Children.toArray(children),
```
amcclain, line 111: I expect we still need to spread the Children array in the normal (non-error) case - otherwise wouldn't React want to insist upon a key?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## v42.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+* `Panel` now has an `error` property to specify display in the event of an error.
```
amcclain, line 6:
```suggestion
* New `Panel.error` prop accepts an `Error` object, `string`, or `ReactElement`. If provided, the panel will render the error - wrapped in an `errorMessage` component as needed - _instead of_ its child `items`. 
```

amcclain (commented):
(Tagging as Fixes #1825)

Thank you for opening for review here, appreciate that! Some thoughts:

- I agree with the implication that the special 'lastLoadException' text token isn't really worth it. It's not the most concise or elegant string, as far as special tokens go, and I can't really argue for it over just referencing the property off of `model`. 
- The question of the toolbars is tricky!
  - Sometimes the error is going to relate to an option/query selected via the toolbar, so you will need it to remedy or retry.
  - But many other times the toolbars (or most of their items) aren't going to make any sense without the actual contents of the panel rendering. So you end up having to switch on rendering those "manually" so this panel error feature less of a benefit overall.
  - Same is true for the (less common) `headerTools`.
  - Since apps can handle this a variety of ways, it's a matter of "what do we think will be most common and convenient?" 

I'll think on the toolbar question a bit more - most likely what we have now is safest / most conservative, and we can see how it shakes out a bit in the real world. 

Also thinking if there's any better name for the prop 😄 - looking to better capture the effect of the prop in the name somehow - but I can't say I have any ideas at the moment....

haynesjm42:
I think it makes the most sense to leave the toolbars rendering by default, especially considering cases like using a panel for dialog contents and using `bbar` for the Cancel/Confirm buttons - would want to be able cancel out of the dialog at any time (same for `headerItems` and having a close button)

We could also consider a more toolbar-like or drawer form for this error display instead of taking up the entire panel contents - I've often shown error messages in toolbars when toasts don't feel quite right for this exact reason of not wanting to cover the contents but also wanting the error message to stay around until the user has resolved it - especially useful when a dialog encounters some error from a server api call. Something more minimal like our loading indicator also may work to be less obtrusive to the contents.

desktop/cmp/panel/Panel.js
lbwexler: I made this change with Dan as part of a refactor, but I have to admit it was not intentional.

It works fine as is, relying on internal React behavior to flatten child arrays.    Not sure what is better -- we certainly don't rely on that flattening behavior in our code right now but we could .

lbwexler:
I think the main argument for leaving the toolbars in (Dan pointed this out) is that they can easily be omitted if needed, based on the error state.  Harder to bring them back.

Like John's idea about a partial error display -- a similar idea would be Dan's idea of using some kind of mask over the contents to show the error, rather than replacing the contents.

Starting to get very complicated however -- wonder if we really should have an opinion on this in Hoist.

desktop/cmp/panel/Panel.js
lbwexler: agree -- think we could actually just pass everything to errorMessage -- that would be the most straightforward.  But the issue of customization is something we need to discuss (see below)

amcclain:
Discussed today with @dsilberger and @lbwexler. While it was a very worthwhile exercise, we don't quite think that there is enough of a generally-applicable feature here to warrant inclusion in the Panel API. The question of toolbars and potential other props that would require switching in a real-world case (headerTools, title) points to how an attempt to build a solution into Panel could end up being close-but-not-usable for many cases.

Better instead to avoid the extra complication and direct our efforts to encouraging use of existing ErrorMessage component and other patterns for displaying errors like toasts within the context of particular apps.

PR #2571, opened 2021-08-03
Mobile swipe navigation

TomTirapani:
Added ability to swipe right to navigate back in mobile. Timings and pixel distances used in the gesture can be further tweaked, at this stage just looking to see if we're happy with the overall approach.

Has special handling around swiping on horizontally scrolling elements (i.e. grids), where it will only interpret it as a navigation gesture if the element can not scroll further at the start of the gesture.

See Toolbox PR: https://github.com/xh/toolbox/pull/514

Hoist P/R Checklist
-------------------


mobile/cmp/navigator/NavigatorModel.js
```diff
@@ -48,11 +57,18 @@ export class NavigatorModel extends HoistModel {
...
+    get swipeComplete() {
+        return this.swipeProgress === 1;
+    }
+
     /**
      * @param {Object[]} pages - configs for PageModels, representing all supported
      *      pages within this Navigator/App.
      * @param {boolean} [track] - True to enable activity tracking of page views (default false).
      *      Viewing of each page will be tracked with the `oncePerSession` flag, to avoid duplication.
+     * @param {boolean} [enableSwipe] - True to enable 'swipe to go back' functionality.
```
amcclain, line 71: How about `swipeToGoBack` as a slightly more verbose but also more clear name for this config?

It's going to be something that developers type once per mobile app (I assume), so I don't think the extra length is much of a penalty in practice.

I'd also vote for defaulting it to true - unless we find it causing some unexpected issues, it's such a nicely realized feature that I don't see a reason _not_ to enable it.  (If some users found it annoying for some reason, I could see this being an app-level option we allow users to turn on and off.)

mobile/cmp/navigator/Navigator.scss
```diff
@@ -1,4 +1,39 @@
...
+    align-items: center;
+    justify-content: center;
+    width: $size;
+    height: $size;
+    color: var(--xh-appbar-color);
+    background-color: var(--xh-appbar-bg);
+    border: 1px solid var(--xh-appbar-border-color);
+    border-radius: $half-size;
+    box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
+    z-index: 21;
```
amcclain, line 25: How did you land on this z-index value?

amcclain (approved):
Very nicely minimal in terms of the amount of code and the end effect - minor suggestions for discussion below but I'm all for it. 👍

mobile/cmp/navigator/NavigatorModel.js
lbwexler: agree

mobile/cmp/navigator/NavigatorModel.js
```diff
@@ -258,4 +276,48 @@ export class NavigatorModel extends HoistModel {
         this._callback = null;
     }
 
+    @action
```
lbwexler, line 279: All handlers can be defined as fat-arrow functions properties on the model -- avoids need for extra wrapper in render method.  This something we have been doing more and more of.

mobile/cmp/navigator/NavigatorModel.js
```diff
@@ -28,12 +28,21 @@ export class NavigatorModel extends HoistModel {
...
+    /** @member {boolean} */
+    enableSwipe;
+
     /** @member {RenderMode} */
     renderMode;
 
     /** @member {RefreshMode} */
     refreshMode;
 
+    /** @member {boolean} */
```
lbwexler, line 40: @package is a new jsdoc annotation  ATM has been recommending -- could consider here.
Or no jsdoc, and gather them in a more clearly defined implementation block.

mobile/cmp/navigator/NavigatorModel.js
lbwexler: should we be consuming the event when we accept it (e.g. e.stopPropogation(); e.preventDefault()).  Would that potentially circumvent the browser grabbing the event?  In any case, probably a best practice.

lbwexler (approved):

mobile/cmp/navigator/Navigator.scss
TomTirapani: This is the z-index used throughout the mobile toolkit in popups, toasts, etc to mean "show above everything else, including masks". IIRC, the origin comes from showing above react-selects dropdowns

mobile/cmp/navigator/NavigatorModel.js
TomTirapani: Great, will rename and change the default. Note that this change means we can close the Toolbox PR without merging: https://github.com/xh/toolbox/pull/514

TomTirapani:
Thanks for feedback guys - I've made the suggested changes. Unfortunately, consuming the event doesn't help to circumvent the browser behaviour :|

PR #2581, opened 2021-08-09
Swipe To Refresh + Refactor of Back Swiping

lbwexler:
Hoist P/R Checklist
-------------------


mobile/cmp/navigator/impl/BackSwiper.js
```diff
@@ -0,0 +1,133 @@
...
+const swipeIcon = hoistCmp.factory(
+    ({model}) => {
+        const {swipeProgress, swipeStarted, swipeComplete} = model,
+            left = -40 + (swipeProgress * 60),
+            className = classNames(
+                'xh-navigator__swipe-indicator',
+                swipeStarted ? 'xh-navigator__swipe-indicator--started' : null,
+                swipeComplete ? 'xh-navigator__swipe-indicator--complete' : null
+            );
+        console.log(left);
```
TomTirapani, line 44: console.log

mobile/appcontainer/RefreshSwiper.js
```diff
@@ -0,0 +1,129 @@
...
+        this.swipeProgress = Math.clamp(deltaY / 150, 0, 1);
+        this.consumeEvent(e);
+    }
+
+    @action
+    onDragEnd = (e) => {
+        if (!this.swipeStarted) return;
+        if (this.swipeComplete) {
+            XH.refreshAppAsync().finally(() => {
+                this.swipeProgress = null;
```
TomTirapani, line 119: We always need to set `swipeProgress` to null, even if not completed. Otherwise the icon stays on the screen

mobile/appcontainer/RefreshSwiper.js
```diff
@@ -0,0 +1,129 @@
...
+        );
+    }
+});
+
+const swipeIcon = hoistCmp.factory(
+    ({model}) => {
+        if (!model.swipeStarted) return null;
+
+        const {swipeProgress, swipeComplete} = model,
+            top = -40 + (swipeProgress * 150),
```
TomTirapani, line 39: I feel like this comes down too far, and something like 100 would be better (i.e. just below the top navigation).

mobile/appcontainer/RefreshSwiper.scss
```diff
@@ -0,0 +1,39 @@
...
+    border: 1px solid var(--xh-appbar-border-color);
+    border-radius: $half-size;
+    box-shadow: 0 2px 5px rgba(0, 0, 0, 0.2);
+    z-index: 21;
+    transition:
+      color 200ms ease,
+      background-color 200ms ease;
+
+    &:not(&--started) {
+      transition: left 200ms ease;
```
TomTirapani, line 31: This should be `top`, not `left`

TomTirapani:
Left a couple of comments.

In general, there is a lot of duplicated code - could we not use the same model / component for both gestures and call it e.g. `swipeDetector`? Also, not sure it's safe to have two competing gesture detectors which consume events on top on each other.

Also, I find it strange that one is in the AppContainer and one is in the Navigator. If we move to a single model, it should probably live at the AppContainer

mobile/appcontainer/RefreshSwiper.js
TomTirapani: Also, line 119 needs to be wrapped in `runInAction`

Issue #1889, opened 2020-05-07
Build in support for `sizingMode` to Hoist


lbwexler:
We would like to extend the support for GridModel.sizingMode to Hoist itself.  This will save applications from needing to wire it up to individual grids, and will also open us up to tie limited additional non-grid properties to it as well  (e.g. toolbar sizing).  In particular, the proposal is to:

+ add an observable `sizingMode` property to XH.  Should probably be implements in the same way that darkTheme is implemented.

+ GridModel should default to using that sizingMode, unless specifically instructed to use another sizingMode in its constructor. 

+ We will probably want grid models to react to changes in the global sizing mode, and make a call to autosizeColumns() when sizing mode is changed.  May or may want to defer this for hidden grids.

+ Would like to do an initial sizing for any grid when its never been sized, or has not yet been sized for the current sizingMode in-force.

[I realize the word sizingMode is a little wonky, but it seems generic enough to extend beyond Grid in the way we need it to]

amcclain:
This would be a great thing to clean up so we can encourage wider use of grid sizing mode as an app-wide option - making it a very easy thing to turn on all around.  Right now it's quite cumbersome, with the need to setup an observable property and then (even more annoying) reactions to update all app-level GridModels.

Marking as good-first-issue - it's not necessarily simple by any means, but could serve as a nice / comprehensive intro to a number of really key concepts.

TomTirapani:
Opened a number of PRs:

https://github.com/xh/hoist-react/pull/2593
https://github.com/xh/hoist-core/pull/205
https://github.com/xh/toolbox/pull/518

Looking forward to discussing!

PR #2593, opened 2021-08-13
Build in support for sizingMode

TomTirapani:
+ Add observable XH.sizingMode
+ Persist this with built-in preference
+ Use as default sizing mode in GridModel
+ GridModel.sizingMode reacts to changes to the global sizing mode
+ GridModel autosizes upon changing the sizing mode

See issue: https://github.com/xh/hoist-react/issues/1889

See hoist-core PR: https://github.com/xh/hoist-core/pull/205
See toolbox PR: https://github.com/xh/toolbox/pull/518

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,10 @@
 ### 🎁 New Features
 
 * Added observable `XH.sizingMode` to govern app-wide `sizingMode`. `GridModel`s will bind to this
-  `sizingMode` by default.
-* Added `GridAutosizeMode.ON_SIZING_MODE`, which autosizes Grid columns whenever `GridModel.sizingMode`
-  changes, and made it the default `GridAutosizeOptions.mode`.
+  `sizingMode` by default. Apps that have already implemented custom solutions around a centralized
+  `sizingMode` should endeavor to unwind in favor of this.
+* Added `GridAutosizeMode.ON_SIZING_MODE_CHANGE`, which autosizes Grid columns whenever
+  `GridModel.sizingMode` changes, and made it the default `GridAutosizeOptions.mode`.
 * `Icon` factories now support `intent`.
 * `TreeMapModel` and `SplitTreeMapModel` now supports a `theme` config, accepting the strings
   'light' or 'dark'. Leave it undefined to use the global theme.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## v43.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+* Added observable `XH.sizingMode` to govern app-wide `sizingMode`. `GridModel`s will bind to this
+  `sizingMode` by default.
+* Added `GridAutosizeMode.ON_SIZING_MODE`, which autosizes Grid columns whenever `GridModel.sizingMode`
+  changes, and made it the default `GridAutosizeOptions.mode`.
+
 [Commit Log](https://github.com/xh/hoist-react/compare/v42.0.0...develop)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
   `sizingMode` should endeavor to unwind in favor of this.
 * Added `GridAutosizeMode.ON_SIZING_MODE_CHANGE`, which autosizes Grid columns whenever
   `GridModel.sizingMode` changes, and made it the default `GridAutosizeOptions.mode`.
+* Added a library of reusable `AppOption` control configs, including `ThemeAppOption`,
+  `SizingModeAppOption` and `AutoRefreshAppOptions`. Apps that have implemented custom `AppOption`
+  controls to manage these hoist-provided options should consider migrating to these defaults.
 * `Icon` factories now support `intent`.
 * `TreeMapModel` and `SplitTreeMapModel` now supports a `theme` config, accepting the strings
   'light' or 'dark'. Leave it undefined to use the global theme.
--- a/appcontainer/SizingModeModel.js
+++ b/appcontainer/SizingModeModel.js
@@ -28,6 +28,10 @@ export class SizingModeModel extends HoistModel {
     @action
     setSizingMode(sizingMode) {
         throwIf(!values(SizingMode).includes(sizingMode), `Sizing mode "${sizingMode}" not recognised.`);
+
+        const classList = document.body.classList;
+        values(SizingMode).forEach(it => classList.toggle(`xh-${it}`, it === sizingMode));
+
         this.sizingMode = sizingMode;
         XH.setPref('xhSizingMode', sizingMode);
     }
--- a/appcontainer/SizingModeModel.js
+++ b/appcontainer/SizingModeModel.js
@@ -4,9 +4,10 @@
  *
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
-import {HoistModel, XH} from '@xh/hoist/core';
+import {HoistModel, XH, SizingMode} from '@xh/hoist/core';
 import {action, observable, makeObservable} from '@xh/hoist/mobx';
 import {throwIf} from '@xh/hoist/utils/js';
+import {values} from 'lodash';
 
 /**
  * Manage Sizing Mode.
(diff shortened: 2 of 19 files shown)


appcontainer/SizingModeModel.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Manage Sizing Mode.
+ *
+ *  @private
+ */
+export class SizingModeModel extends HoistModel {
+
+    @observable sizingMode = null;
+
+    modes = [
```
lbwexler, line 20: am fine with this being the general `SizingMode` even if right now it only refers to Grids

cmp/grid/GridModel.js
```diff
@@ -212,7 +214,10 @@ export class GridModel extends HoistModel {
      * @param {(string|string[])} [c.groupBy] - Column ID(s) by which to do full-width row grouping.
      * @param {boolean} [c.showGroupRowCounts] - true (default) to show a count of group member
      *      rows within each full-width group row.
-     * @param {string} [c.sizingMode] - one of large, standard, compact, tiny
+     * @param {string} [c.sizingMode] - one of large, standard, compact, tiny. If undefined, will
+     *      default and bind to `XH.sizingMode`.
+     * @param {boolean} [c.autosizeOnSizingModeChange] - trigger autosize whenever the sizing mode
```
lbwexler, line 219: this seems quite specific -- could we think about working this into the existing `autosizeOptions` bundle?

lbwexler (approved):
Love this change -- think its very elegant.

Only concern is about the wonky `autosizeOnSizingModeChange`

TomTirapani:
Thanks for taking a look Lee!

I worked the `autosize on sizing mode change` behaviour into the existing `GridAutosizeMode` enum, and adjusted the default there. Think that's cleaner and leverages what we already have in a better way.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## v43.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+* Added observable `XH.sizingMode` to govern app-wide `sizingMode`. `GridModel`s will bind to this
+  `sizingMode` by default.
+* Added `GridAutosizeMode.ON_SIZING_MODE`, which autosizes Grid columns whenever `GridModel.sizingMode`
```
amcclain, line 8:
```suggestion
* Added `GridAutosizeMode.ON_SIZING_MODE`, which autosizes Grid columns whenever `XH.sizingMode`
```

appcontainer/SizingModeModel.js
amcclain: Agree - I think it's a feature, even in advance of any other Hoist components reading it or changing as a result. 

E.g. in one of our current client projects, we talked about other aspects of very data-driven (but not grid) widget internals respecting sizing mode changes.

cmp/grid/Grid.js
```diff
@@ -518,6 +519,17 @@ class GridLocalModel extends HoistModel {
...
+    sizingModeReaction() {
+        const {model} = this;
+        return {
+            track: () => model.sizingMode,
+            run: () => {
+                if (model.autosizeOptions.mode !== GridAutosizeMode.ON_SIZING_MODE) return;
+                model.autosizeAsync();
+            }
+        };
+    }
```
amcclain, line 531: Is it weird that the Grid component is setting up a reaction on a model property that calls only model methods? Why not have this setup on the model itself?

(Was also wondering if cleaner to only setup the reaction in the first place if we're tracking the overall mode, but I don't think there's really a material difference, so if this is more consistent I'm a-OK with it.)

cmp/grid/GridAutosizeMode.js
```diff
@@ -16,9 +16,14 @@ export const GridAutosizeMode = Object.freeze({
...
     /**
      * Grid will offer affordances such as context menus and gestures to allow users to manually
-     * trigger an autosize only.  Applications should trigger with autosizeAsync().
+     * trigger an autosize only. Applications should trigger with autosizeAsync().
      */
-    ON_DEMAND: 'onDemand'
+    ON_DEMAND: 'onDemand',
+
+    /**
+     * Grid will autosize columns when the GridModel's sizingMode changes.
```
amcclain, line 24: We could clarify that onDemand resizing also still supported.

Would `onSizingModeChange` be a bit more clear? I don't think devs will ever need to type it, given that it's the default, so little penalty for length.

cmp/grid/GridModel.js
```diff
@@ -212,7 +212,8 @@ export class GridModel extends HoistModel {
      * @param {(string|string[])} [c.groupBy] - Column ID(s) by which to do full-width row grouping.
      * @param {boolean} [c.showGroupRowCounts] - true (default) to show a count of group member
      *      rows within each full-width group row.
-     * @param {string} [c.sizingMode] - one of large, standard, compact, tiny
+     * @param {string} [c.sizingMode] - one of large, standard, compact, tiny. If undefined, will
+     *      default and bind to `XH.sizingMode`.
```
amcclain, line 216: I think this makes sense - was wondering if an explicit option should also be supported, e.g. `app` or `auto` - but neither of those two suggestions seem more compelling than `undefined` 😄

core/XH.js
```diff
@@ -327,6 +327,19 @@ class XHClass extends HoistBase {
         return this.acm.themeModel.darkTheme;
     }
 
+    //------------------------
+    // Sizing Mode Support
+    //------------------------
+    /** Set the global sizing mode. */
```
amcclain, line 333: We might as well jsdoc these all the way
```suggestion
    /** @param {string} sizingMode - new app-wide sizing mode to apply */
```

core/XH.js
```diff
@@ -327,6 +327,19 @@ class XHClass extends HoistBase {
...
+    setSizingMode(sizingMode) {
+        return this.acm.sizingModeModel.setSizingMode(sizingMode);
+    }
+
+    /** Get the global sizing mode. */
```
amcclain, line 338:
```suggestion
    /** @return {string} - current app-wide sizing mode, observable for `Grid` and other responsive components. */
```

appcontainer/SizingModeModel.js
```diff
@@ -0,0 +1,42 @@
...
+        'large',
+        'standard',
+        'compact',
+        'tiny'
+    ];
```
amcclain, line 25: I think it would be worth an enum here - see `FieldType` in `Field.js`.  Devs could still use simple string values, but it formalizes the concept little and type-heavy folks like myself might want to pull it off the enum.

(Also, I have more than once tried to key off of 'medium' instead of 'standard', so I don't think it's entirely self-evident what the values are. 😄)

amcclain (commented):
I am excited to get this in.  Please see some comments and questions below.

What about applying a class to the `body` element? I think that would be useful if you wanted to have your app respond based purely on CSS style variations.

CHANGELOG.md
amcclain: Actually scratch that - GridModel does publish that var, which is what's actually read.

amcclain:
CHANGELOG entry should actively encourage apps that have already implemented custom solutions around a centralized sizing mode should endeavor to unwind - I think we have a bunch of them.  Although I don't believe that this change would be _breaking_ for any such apps.

I'd also like to create a standard options control (e.g. `buttonGroupInput`) to set the sizing mode - one that could be dropped into apps easily and consistently.  That is something we could look at for theme as well - how can we make it easy for apps to provide consistent and good looking controls to set these Hoist-supplied opts, even if we're staying pretty loose about the overall options an app presents.

cmp/grid/Grid.js
lbwexler: atm and I discussed this is ok

appcontainer/SizingModeModel.js
lbwexler: we will do this and then merge

cmp/grid/GridAutosizeMode.js
lbwexler: Anselm and I agree we should make the slight name changer here as well.

TomTirapani:
Thanks, will make these changes!

Adding the class to the body element is a bit tricky since we currently assign those once in `XH.initAsync()` using `classList.add()`. Since it's observable, we'd need some way to remove to the old class and add the new one when it changes. Furthermore, the `PrefService` isn't available yet to set the initial class name.

We could certainly do it, just wanted to point out that it's not as simple as the other body class names we apply.

TomTirapani:
Thanks for taking a look and the great suggestions! I've updated the PRs to include the suggestions made

PR #2607, opened 2021-08-19
Simplify grouping chooser

lbwexler:
Based on a discussion with ATM, decided to take a crack at this simplification we have been talking about for a while.   

I was originally the one against always showing the "Add..." control, but I actually think this is much better and has far fewer UI elements/state.  Sorry!

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * `Icon` factories now support `intent`.
 * `TreeMapModel` and `SplitTreeMapModel` now supports a `theme` config, accepting the strings
   'light' or 'dark'. Leave it undefined to use the global theme.
+* Various usability improvements and simplifications to `GroupingChooser`.
 ### 🐞 Bug Fixes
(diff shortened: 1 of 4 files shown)


TomTirapani:
Overall I like the simplification!

Do we need to be concerned about the inability to cancel out of inflight changes to grouping? Think about potentially triggering expensive requests etc.

Also, not sure about losing the toolbar and controls on mobile. I understand the desire to match desktop, but in mobile we're using a dialog rather than a popover so I'm not sure the 'click away to apply' really translates as well. This is the only case on mobile where tapping the dialog mask means "apply" rather than "cancel".

lbwexler:
Thanks for the comments tom -- yes, we are giving something up on the ability to cancel -- we thought it was worth the simplification.  Your comment about expensive queries does give me pause -- I guess our feeling was that an application like that would need to have application level protections there (e.g. Go Button).

I guess mobile is different.  You point about it being more aggressive about changing than all other popovers (right now, pre-change) is interesting.


If we were really unsure about this, we could add a flag `showToolbar` on one or more platforms.

Issue #2621, opened 2021-08-26
Investigate if we can force onsenui to use core-js v3


amcclain:
Noticed this relatively extreme warning when running `yarn upgrade`:

> warning @xh/hoist > onsenui > core-js@2.6.12: core-js@<3.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Please, upgrade your dependencies to the actual version of core-js.

We have latest core-js 3.x as a dependency for the rest of our code - could we use [a `resolutions` block](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/) in our `package.json` to force the use of v3 and avoid v2 coming in as a transitive dep from onsen? Would onsen still work / care if the newer version was used?

amcclain:
@TomTirapani as per comment on #2629, can you pls try the same resolutions block within Toolbox and verify that a) we can get only core-js 3 in play and b) onsen still works

If so, we can leave the resolutions in hoist-react as a guide for what transitive deps should be resolved to a newer version, even if we have to copy that into each affected app for it to do anything

lbwexler:
Thanks Tom!  Closing, Take 2.

PR #2629, opened 2021-08-31
Force OnsenUI to use core-js v3.

TomTirapani:
Did a smoke test of the Toolbox mobile app and the newer version of core-js doesn't seem to be causing any issues with Onsen.

Issue: https://github.com/xh/hoist-react/issues/2621

Hoist P/R Checklist
-------------------

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
   },
   "resolutions": {
     "jimp": "^0.16.1",
-    "resize-img": "^2.0.0"
+    "resize-img": "^2.0.0",
+    "core-js": "^3.0.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -3204,10 +3204,10 @@ core-js-compat@^3.14.0, core-js-compat@^3.16.0:
     browserslist "^4.16.7"
     semver "7.0.0"
 
-core-js@^2.6.12:
-  version "2.6.12"
-  resolved "https://registry.yarnpkg.com/core-js/-/core-js-2.6.12.tgz#d9333dfa7b065e347cc5682219d6f690859cc2ec"
-  integrity sha512-Kb2wC0fvsWfQrgk8HU5lW6U/Lcs8+9aaYcy4ZFc6DDlo4nZ7n70dEgE5rtR0oG6ufKDUnrwfWL1mXR5ljDatrQ==
+core-js@^2.6.12, core-js@^3.0.0:
+  version "3.16.4"
+  resolved "https://registry.yarnpkg.com/core-js/-/core-js-3.16.4.tgz#0fb1029a554fc2688c0963d7c900e188188a78e0"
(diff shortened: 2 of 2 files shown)


amcclain:
Thanks for looking at this. Unfortunately - and I should have called this out in the issue but wasn't sure - I don't believe that the `resolutions` block within hoist-react's package.json ends up having any effect for the client apps (like Toolbox).

I updated TB locally to use the latest HR snap:
```
❯ yarn hoistVer
yarn run v1.22.10
$ cat node_modules/@xh/hoist/package.json | grep version
  "version": "43.0.0-SNAPSHOT.1630406815513",
```
and then asked about core-js:
```
❯ yarn why core-js
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "core-js"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "core-js@3.16.4"
info Has been hoisted to "core-js"
info Reasons this module exists
   - Hoisted from "@auth0#auth0-spa-js#core-js"
   - Hoisted from "@xh#hoist#core-js"
info Disk size without dependencies: "7.73MB"
info Disk size with unique dependencies: "7.73MB"
info Disk size with transitive dependencies: "7.73MB"
info Number of shared dependencies: 0
=> Found "onsenui#core-js@2.6.12"
info This module exists because "@xh#hoist#onsenui" depends on it.
info Disk size without dependencies: "7.68MB"
info Disk size with unique dependencies: "7.68MB"
info Disk size with transitive dependencies: "7.68MB"
info Number of shared dependencies: 0
```

I believe we'd need to add that same override to toolbox's package.json and then test again.  Would you be able to take another look, or let me know if you have any sense I'm getting this wrong?

(Unfortunately if correct I'm not too sure what if anything we could do in HR to enforce that apps do this, but if onsen is actually OK with core-js 3 we could ensure our primary mobile apps are setup this way.)

TomTirapani:
Adding the resolutions block override to toolbox's package.json seems to force it to only use v3.16.4, and a smoke test of toolbox mobile doesn't seem to present any issues:
```
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "core-js"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "core-js@3.16.4"
info Reasons this module exists
   - "@auth0#auth0-spa-js" depends on it
   - Hoisted from "@auth0#auth0-spa-js#core-js"
   - Hoisted from "@xh#hoist#core-js"
   - Hoisted from "@xh#hoist#onsenui#core-js"
info Disk size without dependencies: "7.73MB"
info Disk size with unique dependencies: "7.73MB"
info Disk size with transitive dependencies: "7.73MB"
info Number of shared dependencies: 0
✨  Done in 0.86s.
```

PR #2630, opened 2021-08-31
Auto-apply column filters from popover

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/filter/GridFilterModel.js
+++ b/cmp/grid/filter/GridFilterModel.js
@@ -22,6 +22,8 @@ export class GridFilterModel extends HoistModel {
     gridModel;
     /** @member {(Store|View)} */
     bind;
+    /** @member {boolean} */
+    autoApply;
     /** @member {GridFilterFieldSpec[]} */
     @managed fieldSpecs = [];
 
@@ -39,6 +41,8 @@ export class GridFilterModel extends HoistModel {
      * @param {(Store|View)} c.bind - Store or cube View that should actually be filtered
      *      as column filters are applied, and used to provide suggested data values in
      *      column filters (if configured).
+     * @param {boolean} [c.autoApply] - true (default) to automatically apply filters to the
+     *      bound Store / View.
      * @param {(string[]|Object[]} [c.fieldSpecs] - specifies the fields this model supports
      *      for filtering. Should be configs for a `GridFilterFieldSpec`. These may be specified
      *      as field names in bound Store/View or omitted entirely, indicating that all fields
@@ -49,6 +53,7 @@ export class GridFilterModel extends HoistModel {
     constructor({
         gridModel,
         bind,
+        autoApply = true,
         fieldSpecs,
         fieldSpecDefaults
     }) {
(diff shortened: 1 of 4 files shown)


cmp/grid/filter/GridFilterModel.js
```diff
@@ -22,6 +22,8 @@ export class GridFilterModel extends HoistModel {
     gridModel;
     /** @member {(Store|View)} */
     bind;
+    /** @member {boolean} */
+    autoApply;
```
lbwexler, line 26: should this be `immediateApply` it does still get `auto`matically applied to the bound store

lbwexler: or perhaps, `commitOnChange` ?

TomTirapani: Will change to `commitOnChange`... I notice that's what we use for the equivalent setting on the Column Chooser

desktop/cmp/grid/impl/filter/ColumnHeaderFilterModel.js
```diff
@@ -38,15 +38,20 @@ export class ColumnHeaderFilterModel extends HoistModel {
         return this.gridFilterModel.getColumnFilters(this.field);
     }
 
+    @computed.struct
+    get pendingFilter() {
+        const {activeTabId} = this.tabContainerModel;
```
lbwexler, line 43: does this yield an immediate filter change just from changing tabs?   That seems surprising.

TomTirapani: Good point! :|

Not sure what the best approach is here. I could tweak the logic to wait until you've **made a change** in the new tab to commit the filter?

Alternatively, this could be a good justification to make require the user to explicitly clear out their filter upon switching tabs, which we currently only do when switching to the Values Tab from the Custom Tab.

lbwexler: I think we could just  flip things around so that each UI tab puts on its own reaction, such that *when it is visible*, and its value changes, then it should do the commit action

lbwexler: So playing with this.  Its really nice when the data set is the right size, and its all snappy.  

One issue I have is that for the custom tab, its now updating on every keystroke.  This to me seems overly aggressive.  Its especially problematic when you are trying to type in, say, a number, and you get intermediate results based on the first digits that you type.

Think we could potentially make the text fields on the Custom tab commit on change: false, so you had to hit enter, or leave the field for the new value to flow through.  (f we did that, would also probably need to always enable the apply button as well).

I also see a flash in the FilterChooser (in our toolbox example tab) whenever I update the filter via the single keystrokes.  Not sure what that is, but its annoying.  We might address the underlying cause, but that will be less important if we change the behavior as described above.

lbwexler: Could be good to add a little switch at the bottom of the toolbox example just to turn this `commitOnChange` behavior on and off

TomTirapani:
Thanks for continuing to look at this Lee!

I've adjusted the commitOnChange behaviour of the inputs on the custom tab, and added the toggle to Toolbox example to help with testing.

Thanks also for spotting that FilterChooser issue. That was not directly related to this change, just more apparent due to the commitOnChange. I've opened as separate PR which fixes that: https://github.com/xh/hoist-react/pull/2638

PR #2642, opened 2021-09-02
Reduce default grid header heights

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/ag-grid/AgGrid.js
+++ b/cmp/ag-grid/AgGrid.js
@@ -99,7 +99,7 @@ AgGrid.getRowHeightForSizingMode = (mode) => (XH.isMobileApp ? AgGrid.ROW_HEIGHT
 /**
  * Header heights (in pixels)
  */
-AgGrid.HEADER_HEIGHTS = {large: 28, standard: 24, compact: 20, tiny: 18};
+AgGrid.HEADER_HEIGHTS = {large: 28, standard: 24, compact: 22, tiny: 20};
 AgGrid.HEADER_HEIGHTS_MOBILE = {large: 42, standard: 38, compact: 34, tiny: 30};
 AgGrid.getHeaderHeightForSizingMode = (mode) => (XH.isMobileApp ? AgGrid.HEADER_HEIGHTS_MOBILE : AgGrid.HEADER_HEIGHTS)[mode];
 
--- a/cmp/ag-grid/AgGrid.js
+++ b/cmp/ag-grid/AgGrid.js
@@ -98,7 +98,7 @@ AgGrid.getRowHeightForSizingMode = (mode) => (XH.isMobileApp ? AgGrid.ROW_HEIGHT
 /**
  * Header heights (in pixels)
  */
-AgGrid.HEADER_HEIGHTS = {large: 36, standard: 32, compact: 28, tiny: 22};
+AgGrid.HEADER_HEIGHTS = {large: 28, standard: 24, compact: 20, tiny: 18};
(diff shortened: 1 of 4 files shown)


amcclain:
We're using these values in one of our active client projects and they appear to be working well.

- I've only updated desktop for now, but would be good to review if mobile also makes sense.
- These might be a bit _too_ aggressive - I've looped in you all as reviewers as I know you've either worked directly on the grid headers or have active projects that might be doing fancy things with their headers.

cmp/ag-grid/AgGrid.js
```diff
@@ -98,7 +98,7 @@ AgGrid.getRowHeightForSizingMode = (mode) => (XH.isMobileApp ? AgGrid.ROW_HEIGHT
 /**
  * Header heights (in pixels)
  */
-AgGrid.HEADER_HEIGHTS = {large: 36, standard: 32, compact: 28, tiny: 22};
+AgGrid.HEADER_HEIGHTS = {large: 28, standard: 24, compact: 20, tiny: 18};
```
petradish, line 101: I like the look of large and standard, but compact and tiny feel too squished, especially if the column filter menu is enabled. I like how {compact: 22, and tiny: 20} look personally.
I also think the column filter menu icon top margin needs to be tweaked to 2px or 3px with this change: https://github.com/xh/hoist-react/blob/555a550a0b0534cb241770a87d90b07ee63c0b37/desktop/cmp/grid/impl/filter/ColumnHeaderFilter.scss#L12.

haynesjm42:
I have no concerns with reducing the header heights, though I do agree that it is a bit aggressive - did we consider just matching the header heights to the row heights? I think it looks slightly odd for the headers to be shorter than the rows. Also might consider a different color for the header to keep them visually distinct from the rows if they are the same/similar height, I like to use `--xh-text-color-muted` for the headers in my apps

lbwexler:
A couple of thoughts:

+ I agree this is perhaps a bit aggressive, I could see giving back 2 pixels at every size.  But I am sympathetic with the idea that it needs to be enough so it's noticeably different from the grid rows.  Also, easy for apps to adjust if needed using the constant.

+ Should/can we also reduce the size of group rows as well?  They seem to me to be more "chrome" like in most grids, and look out of place now -- see image.    Just a thought -- curious to know what others think

image

amcclain:
Thanks - will review and reconcile the comments here as best I can as soon as I have a chance 👍

amcclain:
- I am happy to make the reduction a bit less drastic - this was admittedly the "farthest out" proposal - I'm sure it's wise to be a bit more incremental.
- I personally like having the headers a bit smaller relative to the rows themselves - I think it helps distinguish them as "grid chrome" in favor of emphasizing the data itself. That said, one motivation I had for the really tight sizes was to make this distinction look deliberate - would want to watch out for anything that felt like an uncanny valley effect where close-but-not-the-same heights looked somehow messy or random.
- I agree with you @haynesjm42 that the muted-text coloring is nicer - we are using that in a few apps and I would be happy to have our existing `--xh-grid-header-text-color` var default changed to `--xh-text-color-muted`
- Finally, while we're considering these headers and some style changes - I'm finding myself changing `--xh-grid-header-bg` to `--xh-grid-bg` to avoid yet-another-strip-of-gray. They end up looking a bit more modern and cohesive IMO - interested if you all agree and would propose making that change too while we're at it. (Again, some flagship apps already using these settings)

amcclain:
Updated if anyone would like to take a look for possible inclusion in Friday's release.

Lee you are right - I am eager to get those group rows slimmed down now!  (#2668)

lbwexler:
Looking now.  Will aim to merge

Issue #1552, opened 2020-01-16
Expand/collapse grid nodes on dbl click (desktop) and single tap (mobile) by default


amcclain:
This is a very natural way to expand/collapse tree grid rows. We are not wiring up anything by default - I believe because we didn't want to eagerly "claim" the double-click action for expand/collapse when an app might wish to use it for something else (e.g. opening a detail modal).

It would be an enhancement however if we could enable this behavior with a simple config vs. needing to wire up a custom handler.

Note we are setting this property within `Column` that appears related, but is not commented (and is somewhat weirdly at the Column, meaning cell, level).  We should understand what this is doing and how it might relate:

https://github.com/xh/hoist-react/blob/e8d32b5bd1713c7386efe7ea325be5f7eff5fc8c/cmp/grid/columns/Column.js#L300

amcclain:
Note the implementation for this might be as simple as:

```
onRowDoubleClicked({node, api}) {
    if (!node?.allChildrenCount) return;
    api.setRowNodeExpanded(node, !node.expanded);
}
```

I still think it would be worth bundling into a config, but of course am happy to discuss.

lbwexler:
Let's go ahead and call this "expandOnDoubleClick".  

As anselm, mentions  let's favor toggling the native ag property, over adding a new explicit handler.

amcclain:
Great thanks - this will help me clean up a few handlers in my code that I've added to do this consistently across grids. Will keep an eye out for it.

petradish:
> This is a very natural way to expand/collapse tree grid rows. We are not wiring up anything by default - I believe because we didn't want to eagerly "claim" the double-click action for expand/collapse when an app might wish to use it for something else (e.g. opening a detail modal).
> 
> It would be an enhancement however if we could enable this behavior with a simple config vs. needing to wire up a custom handler.
> 
> Note we are setting this property within `Column` that appears related, but is not commented (and is somewhat weirdly at the Column, meaning cell, level). We should understand what this is doing and how it might relate:
> 
> https://github.com/xh/hoist-react/blob/e8d32b5bd1713c7386efe7ea325be5f7eff5fc8c/cmp/grid/columns/Column.js#L300

[...]

PR #2651, opened 2021-09-06
Expand on double click

TomTirapani:
Opening this for discussion. See issue: https://github.com/xh/hoist-react/issues/1552

The code to support the new `expandOnDoubleClick` is quite small and straightforward. This PR also does the necessary work to distinguish between single and double clicks, and **only fire one or the other**. Currently, every double-click event is proceeded by two single-click events, and this makes expand to double click difficult (if not impossible) to pair with single-click event handling. On mobile in particular, we really need the ability to fire an "expand" event and a "navigate" event in a mutually exclusive way.

Hoist P/R Checklist
-------------------


cmp/grid/GridModel.js
```diff
@@ -213,6 +215,11 @@ export class GridModel extends HoistModel {
      *      rows within each full-width group row.
      * @param {SizingMode} [c.sizingMode] - one of tiny, compact, standard, large. If undefined, will
      *      default and bind to `XH.sizingMode`.
+     * @param {boolean} [c.expandOnDoubleClick] - true to expand / collapse tree rows on
+     *      double-click. Defaults to true if `treeMode=true`.
+     * @param {number} [c.doubleClickDelay] - Maximum time (in ms) between clicks to trigger a
+     *      double-click, i.e. `onRowDoubleClicked`|`onCellDoubleClicked`. Defaults to 300ms for
+     *      desktop, and 500ms for phone/tablet.
```
amcclain, line 222: Do we need to expose this as a config on (already _very_ configurable) GridModel? Wondering if it might be overkill and we should just decide on a value.

A bit of googling seemed to confirm that the closest thing to a standard is a 500ms default used by Windows.  There is also an accessibility setting in iOS that's defaulted to 250ms, but I'm not positive that it's the same thing.

This leads me to think that 500ms might be a good (single) value that we use and adjust based on feedback, if any. But of course let me know if the current approach w/two values had anything super specific behind it.

---------

EDIT - eh, understanding it more now, I see the (meaningful) tradeoff in lengthening the delay here - how that enforces a minimum wait for any on(Single)Click handler. 

Do we have any sense at all as to how those handlers are being used in apps - or even a single app?

---------

EDIT 2 - I see that we're using `onRowClick` on mobile for navigation - our common pattern of having a mobile grid w/selection disabled and a click (tap) handler applied to route to a detail page when a row is tapped.  

That seems like something that we would be loathe to delay further in terms of perceived mobile responsiveness.

Makes me wonder if the entire routine around the delay should only be applied if double clicks need to be detected at all - i.e. if there is a dblClick handler or expandOnDbl is enabled && row is parent. Something like that.  

Coupled with that is my nagging sense that **single** tap-to-expand is the most natural experience on mobile.

❗ **Please let's discuss together before changing anything because I am not sure about these suggestions and would want to be very clear on any direction before introducing / proposing additional complexity.**

lbwexler: Anselm -- just saw your comments now, and I think they mirror my conversation with Tom -- we are also  unsure how to proceed and wary about introducing the arbitrary delay.  

Certainly if we did this your idea of only introducing the delay if there was no double click handling is appealing -- just had the same thought.  And yet that makes the whole thing even more complex.

 We could also take the approach of not trying to be too clever, and just enabling the basic functionality and trusting applications not to shoot themselves in the foot with conflicting handlers.  What if we were to just add a new property `clicksToExpand` that behaved similarly to `clicksToEdit`?  Application could set it to 1 on a mobile grid, 2 on a desktop grid.  Use at your own risk, just like `clicksToEdit`

TomTirapani: Hey guys,

Some great ideas in this thread - comments below:

**Only delaying if necessary:**
Think this is a great idea, and think the logic you specified above is sound. No point in delaying single-clicks unless absolutely necessary. However, if we default to double-click expand on desktop, this does kind of necessitate single-click delay on all parent rows. I think that's worth living with given how rarely we use click handlers in desktop, typically preferring selection-based drilldown.

FWIW, I don't think its worth adding this feature _unless_ we default to enabling it - think its the kind of the core interactivity that we really want to be consistent across apps.

**Clicks to expand / single-tap on mobile:**
Agree with single-tap on mobile. The paradigm used in Sencha mobile (which I still think is best) is:

+ Leaf rows - single-tap to navigate
+ Parent rows - single-tap to expand, double-click to navigate.

I suspect many apps won't actually allow drilldown from parent rows, so for those we'd be single click all the way. Apps that wanted to drilldown from parent rows would be able to do so like this:

```
onRowClicked: () => {
	if (parent) return;
	navigate();
}

onRowDoubleClicked: () => {
	if (!parent) return;
	navigate();
}
```

Not sure about `clicksToExpand`, depends on if we want to enable this by default. If it's enabled by default, I think the behaviour should be hard-coded to double on desktop and single on mobile and not rely on app devs to configure. But if we wanted disabled by default, then passing `clicksToExpand` is a great way to enable **and** get the different mobile / desktop behaviours.

**Conflicting handlers:**
I don't see a path forward without resolving the conflicting handlers issue, don't think we can rely on apps to not shoot themselves in the foot. On mobile in particular, where we don't use selection-driven master-detail views, we **need** to be able to trigger navigation without triggering expand / collapse. It doesn't matter if it's expand on single or double, unless we resolved it we'd end up with either:

+ navigate() -> navigate() -> expand();
+ expand() -> collapse() -> navigate();

The only way we could go forward without it would be if we could say we never need to navigate from parent rows, or by simply disabling the gesture in grids where we do. Don't think either option is a good story for us.

TomTirapani:
Closing this in favour of: https://github.com/xh/hoist-react/pull/2701

PR #2652, opened 2021-09-06
Redesign custom filter rows

TomTirapani:
+ Use multi-select for multi-value operators
+ Remove confusing implicit join message

Hoist P/R Checklist
-------------------


lbwexler:
Love the direction this is moving -- think the multi-select much clearer than the current strategy.

But having seen it, not 100% sure about the new 2 line layout here  (which I pushed us to try).  Especially for the number filtering, it seems a little sub-optimal and confusing.  Wonder what the growing multi-select would look like within the three column layout.

As an orthogonal matter, I would remove the clear icon within the inputs -- it seems confusing and redundant, given the ability to just zap the entire filter entry with the red minus.

lbwexler:
Actually for the text columns,  the current layout is growing on me.  I think if you used the words `equals` and `not equals` now that you have the space, it would work really well.

A little less sure about what to do with numbers -- we definitely lost the ability to have multiple values for = and != -- not sure we thought  about that.  Putting that aside, as long as numbers only allow single values, could imagine going back to the single line layout.

Sorry for the ramble....

TomTirapani:
Good point about losing the ability to have multiple `=` and `!=` values for numeric fields, the same would apply to date fields :|

In light of that, not sure if this multi-select approach is going to work. For those types of values we'd need to allow multiple rows, which would bring us back to square one in terms of AND|OR confusion. I suppose we could use multi-select for those types of values too (at the cost of number formatting / date picker), and make each entry on the custom tab similar to a FilterChooser (could even leverage the same `valueParser` / `valueRenderer` configs on the FieldSpec), but I don't think the loss of functionality is justified.

If we do stick with the multi-select approach, I would vote for the same UI (2 line or inline) across all the types. I think switching across types would be weird, but could be persuaded.

Should we sit down with Anselm and discuss the options we have and weight up their pros and cons?

lbwexler:
This all makes sense -- yes, let's sit with Anselm and make some decisions
here.

-Lee

lbwexler:
Looks prmising -- could see using a less wonky tag for the non displayable filter's in FilterChooser  -- e.g. "Custom Filter on TradeDate"

lbwexler:
Would like to review in detail in person

cmp/grid/filter/GridFilterModel.js
```diff
@@ -175,4 +183,18 @@ export class GridFilterModel extends HoistModel {
             });
         });
     }
+
+    getOuterCompoundFilter(filter, field) {
+        if (!filter?.isCompoundFilter) return null;
+
+        // This is the outer compound filter if all its children
+        // are FieldFilters on the matching field.
+        if (every(filter.filters, it => it.field === field)) {
```
lbwexler, line 192: second clause to every can just be [field]

lbwexler (approved):

lbwexler:
Happy to go forward with this -- its a minimal change for the user experience, a nice functional enhancement, and a nice cleanup of our underlying model.  @amcclain  -- let us know if you want to take a look

PR #2654, opened 2021-09-07
Fix: In-app changelog prompt does not appear if category-based filtering results in an empty version

### LINKED ISSUE #2619: In-app changelog prompt should appear if category-based filtering has left it empty
We have a feature in `ChangelogService` that supports a config key to exclude certain categories from the in-app display of the log.  (So we can add e.g. Technical changes in the log but not have business users see and be confused by them.)

https://github.com/xh/hoist-react/blob/7cf837e2adbe5bc763328d2b863c36173bc16cc3/svc/ChangelogService.js#L148

However, it can be the case that use of this feature will cause _all_ entries for a given version to be filtered out, resulting in the user being prompted with the 🎁  icon to open the log and seeing a broken-looking display with no bullet points.

The service should take the filter into account when deciding if a version effectively "exists" before showing the prompt or printing the version header / date in the log.

saba-mo:
Closes #2619 

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/appcontainer/ChangelogDialog.js
+++ b/desktop/appcontainer/ChangelogDialog.js
@@ -5,7 +5,7 @@
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
 import {ChangelogDialogModel} from '@xh/hoist/appcontainer/ChangelogDialogModel';
-import {div, filler, h2, h3, h4, li, ul} from '@xh/hoist/cmp/layout';
+import {div, filler, h2, h3, li, ul} from '@xh/hoist/cmp/layout';
 import {hoistCmp, uses, XH} from '@xh/hoist/core';
 import {button} from '@xh/hoist/desktop/cmp/button';
 import {panel} from '@xh/hoist/desktop/cmp/panel';
@@ -75,7 +75,10 @@ const version = hoistCmp.factory(
                 ] :
                 [
                     h2(`v${version.version}`),
-                    h4('No release notes for this version.')
+                    div({
+                        className: 'xh-changelog__version__no-category',
+                        item: h3('No release notes for this version.')
+                    })
                 ]
         });
     }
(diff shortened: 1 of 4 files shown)


lbwexler:
Saba and I went over this, including how to test.  This solution seems fine, and works as requested.

Its a little unfortunate that if you do bring it up now, and the latest changes did *not* include any of your categories, you see an old version number.


We should probably *preserve* the empty versions and:

 1) Show some default text like (Misc. Changes).
 2) Get the lastest non-empty version, and show the gift box icon, only if *that* is out of date. (see line 159)

desktop/appcontainer/ChangelogDialog.js
```diff
@@ -58,19 +58,28 @@ const version = hoistCmp.factory(
...
+                            items: [
+                                h3(cat.title),
+                                ul(...cat.items.map(item => li(item)))
+                            ]
+                        });
+                    })
+
+                ] :
+                [
+                    h2(`v${version.version}`),
```
lbwexler, line 77: Think the title ought to be the same here (line 77 and 63, no?)

lbwexler:
Saba -- please review my latest commit, with some refactoring and slight change in logic regarding when the icon is shown.

lbwexler:
cr: Saba

amcclain:
Just checking - will the  🎁 icon now show if there is a non-empty CL entry for a version which you have yet to look at, even if the version which triggers the prompt is no longer the current app version?

I don't have a strong opinion on it, just confirming if that change was made in case anyone notices / asks about that change. I can see it being reasonable for the feature to behave either which way, and I can see an advantage to still encouraging the user to e.g. read the latest minor (feature) release notes even if a newer patch release was required that did not have a note.

saba-mo:
Yes, the 🎁 icon will show if there is a non-empty version that you have yet to look at, even if it's no longer the current app version. It will only not show if you have already seen all non-empty versions.

amcclain:
Great thanks - makes sense to me! 

🎁 's all around 😄

PR #2657, opened 2021-09-08
Setup grid cell sizes for groupUseEntireRow = false

cnrudd:
I was enthusiastically adding the app-wide Grid Sizing options in the Navbar App Menu for a client app, when I discovered that the Tiny, Compact, and Large sizes did not produce good group row sizes when the grid is using:

```
                    agOptions: {
                        groupUseEntireRow: false,
                        groupSuppressAutoColumn: true
                    }
```

which this client app uses A LOT.

This issue is reproducible in Toolbox in this branch: https://github.com/xh/toolbox/tree/groupUseEntireRow_false on the "Grouped Columns Grid", where I have set this agOptions config.

Here is a screenshot of the issue in toolbox:
gridCssError

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.
- [x] Reviewed and tested on Mobile.
- [x] Created Toolbox branch: https://github.com/xh/toolbox/tree/groupUseEntireRow_false

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🐞 Bug Fixes
 
 * Better coverage of Fetch abort error.
+* Grid SizingMode styles fixed to support agGrid `groupUseEntireRow = false`
 
 [Commit Log](https://github.com/xh/hoist-react/compare/v42.4.0..develop)
 
--- a/cmp/ag-grid/AgGrid.scss
+++ b/cmp/ag-grid/AgGrid.scss
@@ -243,9 +243,15 @@
       font-size: var(--xh-grid-large-header-font-size-px);
     }
 
+    .ag-cell-wrapper > *:not(.ag-cell-value):not(.ag-group-value) {
+      height: var(--xh-grid-large-cell-height-px);
+    }
+
+    .ag-full-width-row .ag-cell-wrapper.ag-row-group,
     .ag-cell {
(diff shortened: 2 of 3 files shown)


lbwexler:
Hi Tom -- Can you take a look at this one? Thanks

TomTirapani:
This works, but I'm not sure it's the right approach. Using CSS to height rows is something we want to avoid, as the row height is actually quite variable.

We have two different sets of row heights for mobile and desktop: https://github.com/xh/hoist-react/blob/develop/cmp/ag-grid/AgGrid.js#L90-L104. Furthermore, we allow apps to override these, and there's also the `GridModel.groupRowHeight`, `Column.rowHeight` and `Column.autoHeight` configs to take into account. We can't really rely on any particular row height in CSS.

For normal cells, we took the approach of vertically centering the cells to handle this variability: https://github.com/xh/hoist-react/blob/develop/cmp/ag-grid/AgGrid.scss#L180-L184. Could we take a similar approach with the group cells when `groupUseEntireRow: false` ?

cnrudd:
@TomTirapani thanks for the pointers.  looking into vertical centering solution now.

cnrudd:
@TomTirapani I think this one line change solution in the latest commit might be the one.  Thanks!
Colin

TomTirapani:
Looks great to me!

amcclain:
Note that the CL entry here ended up bucketed under the last version - easy to happen with a PR that stays open across a release.  Will fix now.

PR #2659, opened 2021-09-09
Model check

lbwexler:
Hoist P/R Checklist
-------------------


amcclain (approved):
I think this looks good - was on the fence re. utility of a "Technical" CL note, but don't feel strong about it.

IIRC from our earlier investigation, we really should not hit the throw within Grid LocalModel unless something has gone badly wrong - just re-affirming that is the case / expectation.

Also was on the fence re. special exception for FormField - if it needed a comment - but I am assuming it would be clear why it exists if we were trying to understand in the future and simply removed that exception.

lbwexler:
+ yes, don't expect to ever hit the Grid LocalModel throw -- your expectation is correct.

+ was also not thrilled about the special exception for FormField -- if we saw ourselves adding more carve-outs like that we could add a proper flag on the model spec -- was not ready to that at this point.  Could not really see any other way around this

PR #2660, opened 2021-09-09
New Column.autosizeBufferPx config

petradish:
New Column.autosizeBufferPx config to support overriding GridModel.autosizeOptions.bufferPx applied on all columns.

Hoist P/R Checklist
-------------------


cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -34,28 +34,32 @@ export class ColumnWidthCalculator {
...
 
     calcHeaderWidth(gridModel, column, options) {
-        const {autosizeIncludeHeader, autosizeIncludeHeaderIcons} = column,
-            {bufferPx} = options;
+        const {autosizeOptions} = column,
+            {includeHeader, includeHeaderIcons} = autosizeOptions;
 
-        if (!autosizeIncludeHeader) return null;
+        if (!includeHeader) return null;
+        let {bufferPx} = options;
```
TomTirapani, line 54: I think this line might be clearer moved to the top as:
```
const {autosizeOptions} = column,
    {bufferPxOverride, includeHeader, includeHeaderIcons} = autosizeOptions,
    bufferPx = isFinite(bufferPxOverride) ? bufferPxOverride : options.bufferPx;
```

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -92,8 +96,11 @@ export class ColumnWidthCalculator {
     }
 
     calcLevelWidth(gridModel, records, column, options, indentationPx = 0) {
-        const {field, getValueFn, renderer} = column,
-            {bufferPx} = options;
+        const {field, getValueFn, renderer, autosizeOptions} = column;
+        let {bufferPx} = options;
```
TomTirapani, line 100: Same as above

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@
 * Provide applications with the ability to override default logic for "restore defaults".
 This allows complex and device-specific sub-apps to perform more targeted and complete clearing of
 user state.  See new overridable method `HoistAppModel.restoreDefaultsAsync` for more information.
+* New `ColumnAutosizeOptions` config bundles existing column-specific autosize properties along with
+additional `bufferPxOverride` property, which overrides `GridAutosizeOptions.bufferPx`
```
TomTirapani, line 11: Did we consider losing the 'override' suffix? It seems clear enough to me that setting a property on the Column's autosizeOptions would override the property on the GridModel, and would keep the name consistent.

TomTirapani:
I'm like this! Bundling these options into an object gives us a hook to extend in the future, and allows us to shorten the property names by removing redundant `autosize` prefix. And allowing the dev to set bufferPx on a per column basis is a clear win. 👍 

I'd still like to go over the issue that prompted this change though, to get a better understanding of why the autosizing wasn't working for you.

CHANGELOG.md
petradish: I did consider losing the override suffix, but in the 2 methods above, I found the code confusing to read the same variable `bufferPx` off of either options (from grid) or autosizeOptions (from column). I like your refactor above, but it seems easier to read with the `override`. Also, I thought it might be ambiguous whether the column's bufferPx were additive vs replacing.

lbwexler:
I asked Petra to look into the new override on bufferPX, so thanks for putting this together!.  

But unfortunately,  at this point I have some misgivings with this PR:

-   We now think we don't actually need column specific `bufferPx`.  Tom is working on an unrelated fix to the problem that originally made it seemed like we needed that new flag.
-  Columns already have very flat options for many things (especially export).  It actually 
-  We need to avoid API churn unless we have a clear feature, or bug we are fixing.   

Thoughts?

lbwexler:
Tom -- let me know what you think of my reasoning?  I see you were liking it...

TomTirapani:
Lee -- you make a good point about API churn. It seems overkill to bring breaking changes to the API if we're able to solve the underlying issue in the other PR.

That said, I do think being able to set `bufferPx` on a per column basis is still valuable. I think of these `Column.autosizeXXX` options as "escapes hatches" for the dev to make autosize more stable (similar to `autosizeMaxWidth`, etc). If you have one particular column that could benefit from increasing the `bufferPx`, it seems reasonable to be able to do so, especially since we have these other column-specific flags available. It certainly seems better than the current solution of having to increase the `bufferPx` of **every** column, and widening **every** column in the grid, to fix an issue with a specific column.

I would vote for not adding the `ColumnAutosizeOptions`, but keeping the ability to set column specific `bufferPx` as a flat `Column.autosizeBufferPx` option, to make this a purely additive non-breaking change.

Thoughts?

lbwexler:
Sounds like a good compromise, and nice pickup!

PR #2661, opened 2021-09-09
Disable user select on tablets, Fixes #2655

### LINKED ISSUE #2655: Disable user select on tablets
We have had to do this on two of our flagship client apps. When a user holds down their finger to bring up the context menu the text on the page gets selected and it's not a good look. Is there any reason NOT to do this at the Hoist level?

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/styles/XH.scss
+++ b/styles/XH.scss
@@ -194,3 +194,7 @@ body.xh-app {
     text-transform: uppercase;
   }
 }
+
+.xh-tablet * {
+  user-select: none !important;
+}
--- a/styles/XH.scss
+++ b/styles/XH.scss
@@ -63,6 +63,10 @@ body.xh-app {
     cursor: pointer;
   }
 
+  &.xh-tablet * {
+    user-select: none !important;
+  }
+
   // Colors
   .xh-black { color: var(--xh-black); }
   .xh-blue { color: var(--xh-blue); }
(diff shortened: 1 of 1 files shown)


febbraiod:
I suppose this _could_ be a breaking change for an app that relies on user select on tablets, but that seems very unlikely, so I decided that we didn't really need a CL entry. Correct me if wrong.

styles/XH.scss
```diff
@@ -194,3 +194,7 @@ body.xh-app {
     text-transform: uppercase;
   }
 }
+
+.xh-tablet * {
+  user-select: none !important;
+}
```
amcclain, line 200: I would nest (like the other rules within this file) underneath the `body.xh-app` selector. Not sure if that would make a functional difference but still seems correct and consistent.  (So, move up inside braces and add leading ampersand: `&.xh-tablet * {...}`

Is the `!important` required?

amcclain:
> I suppose this _could_ be a breaking change for an app that relies on user select on tablets, but that seems very unlikely, so I decided that we didn't really need a CL entry. Correct me if wrong.

I agree, I do not think we need a CL entry here.

styles/XH.scss
febbraiod: Will move to that block now. 

The !important comes from the definition in the client app from which this change originated. I'm not sure it's needed, but someone thought it was. Could test in MY dashboard related client app but would need to cut a release and hit the app with my tablet via blackberry access to be sure (which is fine but will take some time)

amcclain:
👍 thanks for getting this in

Issue #2667, opened 2021-09-13
Autosize ColumnWidthCalculator needs to take row class names into account.


TomTirapani:
GridModels can use the `rowClassFn` and `rowClassRules` to conditionally apply CSS class names to grid rows. These styles need to be applied to the dummy cell used by ColumnWidthCalculator in case they affect the width of the cell (e.g. using bold font-weight, adding padding, etc.)

PR #2671, opened 2021-09-14
Autosize aims to respect class names applied to the row and cell

TomTirapani:
CSS class names applied to grid cells can interfere with column autosizing. This was seen in a client app where row classes were used to bold the cells, making them wider.

This PR incorporates `GridModel.rowClassFn`, `GridModel.rowClassRules`, `Column.cellClassFn` and `Column.cellClassRules` into the autosizing calculations. Please note due to the (necessary) optimisation to only sample a subset of values, there are still some cases that may slip through the gaps (e.g. conditionally applying a cell class to make an otherwise narrow cell much wider). However, for simpler cases such as bold fonts and added padding this should prove sufficient.

See Toolbox branch which replicates the issue seen in the app: https://github.com/xh/toolbox/tree/autosizeClassNames

See issue: https://github.com/xh/hoist-react/issues/2667

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnWidthCalculator.js
+++ b/cmp/grid/impl/ColumnWidthCalculator.js
@@ -88,28 +88,24 @@ export class ColumnWidthCalculator {
 
     calcLevelWidth(gridModel, records, column, options, indentationPx = 0) {
         const {field, getValueFn, renderer} = column,
-            {sizingMode} = gridModel,
+            {store, sizingMode} = gridModel,
             bufferPx = column.autosizeBufferPx ?? options.bufferPx;
 
-        // 1) Get unique values
+        // 1) Get unique values, and the ids of records that contain them
         const values = new Set(),
-            valueClassMap = {};
+            valueRecordMap = {};
 
         records.forEach(record => {
             if (!record) return;
-            const ctx = {record, field, column, gridModel, store: gridModel.store},
+            const ctx = {record, field, column, gridModel, store},
                 rawValue = getValueFn(ctx),
-                value = renderer ? renderer(rawValue, ctx) : rawValue,
-                cellClass = this.getCellClass(gridModel, column, record, rawValue),
-                rowClass = this.getRowClass(gridModel, record);
+                value = renderer ? renderer(rawValue, ctx) : rawValue;
 
             values.add(value);
 
-            // 1b) Track combinations of cell and row classes applied to this value.
-            // Stringify the combination to test Set equality.
-            let classNames = valueClassMap[value];
-            if (!classNames) classNames = valueClassMap[value] = new Set();
-            classNames.add(JSON.stringify({cellClass, rowClass}));
+            let recordIds = valueRecordMap[value];
+            if (!recordIds) recordIds = valueRecordMap[value] = new Set();
+            recordIds.add(record.id);
         });
 
         // 2) Use a canvas to estimate and sort by the pixel width of the string value.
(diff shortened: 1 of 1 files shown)


cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -87,26 +84,42 @@ export class ColumnWidthCalculator {
...
                 rawValue = getValueFn(ctx),
-                value = renderer ? renderer(rawValue, ctx) : rawValue;
+                value = renderer ? renderer(rawValue, ctx) : rawValue,
+                cellClass = this.getCellClass(gridModel, column, record, rawValue),
+                rowClass = this.getRowClass(gridModel, record);
+
             values.add(value);
+
+            // 1b) Track combinations of cell and row classes applied to this value.
+            if (!valueClassMap[value]) {
```
lbwexler, line 111: might be more straightforward (and efficient) to look this up once in to a variable, rather than looking up three times.

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -117,8 +130,20 @@ export class ColumnWidthCalculator {
...
         // 4) Render to a hidden cell to calculate the max displayed width
         return reduce(longestValues, (currMax, value) => {
-            const width = this.getCellWidth(value, renderer) + indentationPx + bufferPx;
-            return Math.max(currMax, width);
+            const {cellClasses, rowClasses} = valueClassMap[value];
+
+            // 4b) Loop through all combinations of cell and row classes applied to this value,
+            // and return the largest.
+            let ret = 0;
+            rowClasses.forEach(rowClass => {
```
lbwexler, line 138: this seems potentially expensive, and may even be testing combinations that don't exist, no?

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -87,26 +84,38 @@ export class ColumnWidthCalculator {
         } catch (e) {
             console.warn(`Error calculating max data width for column "${column.colId}".`, e);
         } finally {
-            this.setCellElActive(false);
+            this.resetClassNames();
```
lbwexler, line 87: think we can remove this -- we are doing a reset already before every measurement.

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -87,26 +84,38 @@ export class ColumnWidthCalculator {
...
+                cellClass = this.getCellClass(gridModel, column, record, rawValue),
+                rowClass = this.getRowClass(gridModel, record);
+
             values.add(value);
+
+            // 1b) Track combinations of cell and row classes applied to this value.
+            // Stringify the combination to test Set equality.
+            let classNames = valueClassMap[value];
+            if (!classNames) classNames = valueClassMap[value] = new Set();
+            classNames.add(JSON.stringify({cellClass, rowClass}));
```
lbwexler, line 114: Tom and I discussed doing this work of looking up the cells and row classes only after we have identified the records with the longest values.

Also, why not uniquify these things as a simple string with a pipe and skip the jsonification?

cmp/grid/impl/ColumnWidthCalculator.js
TomTirapani: This removes the 'xh-grid-autosize-cell--active' class, which hides the cell. Think we always want to do that after an autosize, especially if it throws at any point.

cmp/grid/impl/ColumnWidthCalculator.js
TomTirapani: Good suggestion about uniquifiying with a string pipe, will make that change

PR #2673, opened 2021-09-16
Grid Header Tweaks

### LINKED ISSUE #1896: Visual Tweaks to Grid Headers
+ Sometimes it is nice to have a grid column header just be an icon, and often in that case you also want it centered, but we are always adding some left margin to svg elements in our grid headers (presumably for the sort/filter icons) so things tend to look a bit off,
Offending css here: hoist-react/cmp/grid/Grid.scss - Line 55 in 597531e

+ Would it make sense to decrease/remove the spacing between the sorting icon and the header text?  It seems that we are pretty generous with that and it becomes increasingly important in the case of reserving size for headers. 

+ Would we consider *not* eliding the text of a header, but just clipping it?  (This is especially useful for grids not reserving space for the sorter, or the entire header).  It seems like quite often the '...' take up critical space that would give you one or two letters of context to understand what the header was....

petradish:
Resolves #1896

Adjust grid header margin values
+ Remove extra margins from icons and svgs in header - their fixed width already creates a sufficient margin between it and the header text. This is a nice win for header names that are only Icons - they center properly now.
+ Remove margin-top from grid header (this now aligns the text and sort/menu icons much better)

Apply hover color on column header filter icon
+ In line with hover color applied to the tree expand/collapse icon
+ Chose not to hover on sort icon, since clicking into the main body of the header accomplishes sorting as well... open to feedback (could apply hover color on sort icon when hovering over entire text as well)

Left ellipsis on overflow text as is. Clipping text has the unfortunate effect of looking like typos for most header text, especially when short. Found some exciting experimental text-overflow values that we could hopefully implement in the future (specifically, the fade function) when supported in browsers: https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -33,13 +33,11 @@
     }
 
     svg {
-      margin-left: var(--xh-pad-half-px);
       color: var(--xh-grid-header-icon-color);
     }
   }
 
   .xh-grid-header {
-    margin-top: 1px;
     display: flex;
     flex-direction: row;
     align-items: center;
@@ -56,7 +54,6 @@
     .xh-grid-header-menu-icon,
     .xh-grid-header-expand-collapse-icon {
       color: var(--xh-grid-header-icon-color);
-      margin-left: var(--xh-pad-half-px);
       width: 1.5em;
     }
 
@@ -388,7 +385,6 @@
 
   .xh-grid-header-sort-icon,
   .xh-grid-header-menu-icon {
-    margin-left: var(--xh-pad-half-px);
     width: 1.5em;
     display: none;
   }
--- a/desktop/cmp/grid/impl/filter/ColumnHeaderFilter.scss
+++ b/desktop/cmp/grid/impl/filter/ColumnHeaderFilter.scss
@@ -14,10 +14,6 @@
     width: 1.5em;
     font-size: 12px;
     color: var(--xh-grid-header-icon-color);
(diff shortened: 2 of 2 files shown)


lbwexler:
I think the layout tweaks make a big difference.  they look good, and get you a few more characters .

Not sure about the filter hover -- think it gives too much emphasis to that, and the difference in behavior with sorting (which you allude to) is a little confusing.  Also, the name never hovers., so the filter menu becomes weirdly more prominent than the menu.

TomTirapani:
Agree that the layout changes make a huge difference 👍 

I actually like the hover on the filter icon, I think pre-feedback on interactive elements makes a huge difference, but also find the discrepancy between it and the sort icon strange. How about applying hover to the text **and** any present sort icon on sortable columns whilst hovering over any part of the header that isn't the filter icon? Will communicate click to sort, and will make any columns with `sortable: false` less surprising since they'll appear "disabled".

TomTirapani:
(I don't feel strongly about the above, happy to see what others think and go with majority vote)

petradish:
I'll play with highlighting the text + sort icon and see if that works, otherwise will take out the hover color for consistency!

lbwexler:
Great.  Can we put hover changes in a separate pr so we can merge this one
now?

petradish:
yes!

PR #2674, opened 2021-09-17
Reduce default grid group row heights

### LINKED ISSUE #2668: Provide sizingMode aware control over grid group row heights, default to shorter group rows
Grids using full-width row grouping get group rows that are the same height as data rows by default. These often seem overly tall / heavy - we should review and determine if defaulting to shorter group rows (as we do with header rows) looks better overall.

We should also allow apps to customize in a sizingMode aware manner w/o needing to use agOptions - I'd suggest another set of overrideable values hung off of `AgGrid` as with headers + standard data rows. 

Finally, we should account for apps that use the grid's ability to group data (distinct from treeMode) and end up with group rows that are _not_ full-width - those rows should presumably not be affected by this new constant, as they are much more like normal data rows than the full-width group rows are.

saba-mo:
Closes #2668 

- Reduces height of full-width group rows based on grid sizing mode
- Group rows that do not use the full width continue to use the height of the non-group rows
- No change to mobile or DataView

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnHeader.js
+++ b/cmp/grid/impl/ColumnHeader.js
@@ -64,7 +64,7 @@ export const columnHeader = hoistCmp.factory({
         };
 
         const expandCollapseIcon = () => {
-            if (!xhColumn?.isTreeColumn || !xhColumn?.headerHasExpandCollapse || !impl.rootsWithChildren) {
+            if (!xhColumn || !xhColumn.isTreeColumn || !xhColumn.headerHasExpandCollapse || !impl.rootsWithChildren) {
                 return null;
             }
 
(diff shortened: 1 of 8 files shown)


lbwexler (approved):
Code looks great to me

saba-mo:
I pushed changes to make xhColumn null safe since it will be null when groupUseEntireRow is false.

cmp/grid/impl/ColumnHeader.js
```diff
@@ -64,8 +64,9 @@ export const columnHeader = hoistCmp.factory({
         };
```
lbwexler, line 65: Is the change to this file an artifact of a bad merge?

Issue #2681, opened 2021-09-21
GridFindField cmp


TomTirapani:
Similar to the StoreFilterFilter, except would be used to select matching rows rather than filter out unmatched rows. Useful for finding records in large dataset without actually changing the dataset by filtering. Would provide affordances to step backwards and forwards through the matching rows, selecting each in the grid.

For the UI, I'm thinking this would be function similar to the browser find (Crtl+F):
+ A text field to enter a string to match against.
+ Numbers of matches, and currently selected match.
+ Up / down buttons to step through the matches.

<img width="331" alt="Screenshot 2021-09-21 at 13 17 22" src="https://user-images.githubusercontent.com/3017757/134169349-a25b756d-9ffe-4716-856c-773ac7c963c7.png">

API can be designed to mirror StoreFilterField, with the same support for `gridModel`, `matchMode`, `includeFields` and `excludeFields`, etc.

PR #2686, opened 2021-09-23
Add GridFindField component

TomTirapani:
A first pass at the GridFindField component. Currently desktop only - I thought about making it cross-platform like the StoreFilterField, but didn't think it seemed as useful for mobile where selection is disabled by default.

See issue: https://github.com/xh/hoist-react/issues/2681
See toolbox PR: https://github.com/xh/toolbox/pull/534

Hoist P/R Checklist
-------------------


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### 🎁 New Features
 
+* New `GridFindField` component that enables users to search through a Grid and select rows that
+  match the entered search term.
```
amcclain, line 8:
```suggestion
  match the entered search term, _without_ applying any filtering. Especially useful for grids with aggregations
  or other logic that preclude client-side filtering of the data.
```

desktop/cmp/grid/find/GridFindField.js
```diff
@@ -0,0 +1,120 @@
...
+                    model: impl,
+                    bind: 'query',
+                    commitOnChange: true,
+                    leftIcon: Icon.search(),
+                    enableClear: true,
+                    placeholder: 'Find...',
+                    width: null,
+                    flex: 1,
+                    onKeyDown: (ev) => {
+                        if (ev.key === 'Enter') impl.selectNext();
```
amcclain, line 67: Will be discovered/used by power users only, but we should `selectPrev()` if `ev.shiftKey`

desktop/cmp/grid/find/GridFindFieldModel.js
```diff
@@ -0,0 +1,285 @@
...
+        const regex = this.getRegex(query),
+            valGetters = flatMap(activeFields, (fieldPath) => this.getValGetters(fieldPath));
+
+        this.results = records.filter(rec => {
+            return valGetters.some(fn => regex.test(fn(rec)));
+        }).map(rec => rec.id);
+    }
+
+    @action
+    updateRecords() {
```
amcclain, line 153: Pretty hardcore - this was a major subtlety (oxymoron?) that I had not considered - but of course it makes sense that would expect to walk through the records in order. I'm impressed that you worked this out w.r.t group sorting and all - nice!

I bit concerned that it could be an intensive operation for large recordsets, notably in our current client project where we were looking to use this for large grids. Might be worth profiling to ensure we are not paying more of a price than expected - could consider doing this lazily when the user first goes to actually use the control and see how that feels.

amcclain (commented):
I find this to be very usable - especially impressive when used in the treegrid example. Great stuff!

Some minor notes below. I also wanted to suggest that we auto-select the first hit if there's not an existing query or match - like IJ text find does.

TomTirapani:
Great - thanks for taking a look! Great suggestions above - I've implemented them in an update.

TomTirapani:
Also pushed a fix to the sorting so that it recursively handles tree-grid nesting

desktop/cmp/grid/find/GridFindFieldModel.js
```diff
@@ -0,0 +1,324 @@
...
+        this.matchMode = matchMode;
+        this.queryBuffer = queryBuffer;
+        this.includeFields = includeFields;
+        this.excludeFields = excludeFields;
+
+        throwIf(!gridModel, "Must specify 'gridModel' in GridFindField.");
+
+        this.addReaction({
+            track: () => [
+                this.hasQuery,
```
amcclain, line 89: I know that `hasQuery` came in at my suggestion around readying this model's view on records in a lazy manner, but I think these reactions could be optimized a bit further. Maybe I am making too much out of this, but:

- We don't need to rebuild the records in the (common) case where the user enters a query, blanks it out, then enters another query.
- These two reactions call `updateResults` twice whenever the query goes from nil -> something.

Feels like there might be a clearer way to handle - potentially by having a lazily cached field/getter for the processed records that is cleared out whenever the store data changes and regenerated again whenever it's asked for by a call to updateResults() with a non-empty query.

TomTirapani:
I made another pass at the record processing optimisation based on your suggestions above Anselm

PR #2704, opened 2021-09-29
Do not save SNAPSHOT versions to xhLastReadChangelog preference

### LINKED ISSUE #2695: Don't save SNAPSHOT versions in xhLastReadChangelog
Seems to me that we wouldn't want snap versions to be saved in this key - they are always at the next major version, which I think means that if a user reads a CL for e.g. 2.0-SNAP they will then never see any notifications for non-snap 1.x releases.

Because we typically don't deploy SNAPs in a prod environment, the current behavior is not a big deal, but still I believe this would be an improvement. This change would also mean that any SNAP version would always show the notification button, but I think that is more of a feature than a bug - if the user is looking at a SNAP version and that version has CL entries, it's all a work-in-progress and it's good to keep the log out in front of any testers or reviewers.

saba-mo:
Closes #2695 

Console warns 'Unable to mark changelog as read - latest version or required pref not found.'

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/ChangelogService.js
+++ b/svc/ChangelogService.js
@@ -61,7 +61,7 @@ export class ChangelogService extends HoistService {
 
     /** @return {?string} */
     get latestAvailableVersion() {
-        if (!this.enabled || includes(this.versions[0].version, 'SNAPSHOT')) return null;
+        if (!this.enabled) return null;
         return this.versions[0].version;
     }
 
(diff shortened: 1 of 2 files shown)


svc/ChangelogService.js
```diff
@@ -61,7 +61,7 @@ export class ChangelogService extends HoistService {
 
     /** @return {?string} */
     get latestAvailableVersion() {
-        if (!this.enabled) return null;
+        if (!this.enabled || includes(this.versions[0].version, 'SNAPSHOT')) return null;
```
amcclain, line 64: So I do find this a bit confusing to have the getter return null in the snapshot case - snapshots are valid versions, and there can be a CL entry for a snapshot, so it's "available" as far as this getter and its area of responsibility are concerned (which are limited to checking that the service is generally enabled and then returning the first version string from the list).

Instead I would put a new conditional within `markLatestAsRead()` since that's the specific behavior we want to adjust.  That method could check if `latestAvailableVersion` includes "SNAPSHOT" and return early - might as well log a slightly different console message in that case since we're bothering to log one in the existing case within that method already.

Since that getter is not used by any other methods at the moment the end functional result is the same, but going forward we might find it helpful to have the getter continue to return a version when enabled, even a snapshot. Hope that rationale for changing 'markLatestAsRead()' vs the more general getter sounds sensible / meaningful all around.

saba-mo: That is a much more sensible approach!

amcclain:
Thanks for getting this in!

PR #2714, opened 2021-10-03
Improvements to GridFindField

lbwexler:
Suggesting the following changes:

+ Store the value on a specifiable 'model', or the grid itself.  This is what StoreFilterField does, and it's important if you ever want to control/persist this value, but also because local state can easily be "lost" in toolbar controls.  (Toolbar will remount more than it should -- a separate issue).

+ Added keyboard handling for up/down arrow.

+ Stacked the buttons -- looking just to make the thing visually a bit tighter.

+ Hide the buttons when no search in place -- again, looking to make things visually tighter, especially when control is not in use.

+ SelectOnFocus : true, for this control. Don't feel strongly about this, but seemed like generally a worthwhile default.

+ Minor re-org to hide implemenation class

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/find/GridFindField.js
+++ b/desktop/cmp/grid/find/GridFindField.js
@@ -6,15 +6,15 @@
  */
 import {hoistCmp, useContextModel, useLocalModel} from '@xh/hoist/core';
 import {GridModel} from '@xh/hoist/cmp/grid';
-import {hbox, span} from '@xh/hoist/cmp/layout';
+import {hbox, vbox, span} from '@xh/hoist/cmp/layout';
 import {textInput} from '@xh/hoist/desktop/cmp/input';
 import {button} from '@xh/hoist/desktop/cmp/button';
 import {Icon} from '@xh/hoist/icon';
 import {splitLayoutProps} from '@xh/hoist/utils/react';
-import {errorIf, withDefault} from '@xh/hoist/utils/js';
+import {errorIf, withDefault, consumeEvent} from '@xh/hoist/utils/js';
 import PT from 'prop-types';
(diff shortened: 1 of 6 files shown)


desktop/cmp/grid/find/GridFindField.js
```diff
@@ -61,33 +61,47 @@ export const [GridFindField, gridFindField] = hoistCmp.withFactory({
...
+                            case 'ArrowDown':
+                                impl.selectNext();
+                                consumeEvent(e);
+                                break;
                         }
                     },
                     ...restProps
                 }),
                 hbox({
+                    omit: !countLabel,
```
TomTirapani, line 86: Unrelated to this change, but I wonder if we can show / hide the controls based on focus rather than results? I feel like that might be the most natural time to reveal them

TomTirapani:
Overall, these a great enhancements and I'm eager to get them in, particularly the binding to not lose the local state.

The only part I'm not sure about is the 'Up' / 'Down' keys for navigation - these feel pretty non-standard to me. I'm used to using the arrow keys to control the caret within editable text, and noticed that other implementations of similar controls (e.g. browser find) don't support it.

If we keep them, we could consider *not* doing anything if you're holding the Shift key. Shift+Up is a selection shortcut I use pretty instinctively to select-all then retype a field, and I immediately noticed its absence here.

lbwexler:
Intellij find supports up down key nav.  Try it!.  Seems natural to me.

TomTirapani:
Okay, fair enough :)

Still think we shouldn't swallow the shift+arrow key selection though (unless I'm the only person who uses those shortcuts)

PR #2771, opened 2021-11-04
Viewport orientation

TomTirapani:
Add observable viewport size and orientation getters. Primary motivation is allow an easier way for mobile / tablet apps to react to orientation changes.

Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/ViewportSizeModel.js
+++ b/appcontainer/ViewportSizeModel.js
@@ -1,0 +1,62 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2021 Extremely Heavy Industries Inc.
+ */
+import {HoistModel} from '@xh/hoist/core';
+import {action, computed, observable, makeObservable} from '@xh/hoist/mobx';
+import {debounced} from '@xh/hoist/utils/js';
+
+/**
+ * Track observable width / height of the browser viewport, and provide observable
+ * access to device orientation
+ *
+ * @private
+ */
+export class ViewportSizeModel extends HoistModel {
+
+    /** @member {Number} */
+    @observable width;
+
+    /** @member {Number} */
+    @observable height;
+
+    /** @returns {Object} */
+    @computed.struct
+    get size() {
+        const {width, height} = this;
+        return {width, height};
+    }
+
+    /** @returns {boolean} */
+    @computed
+    get isPortrait() {
+        return this.width < this.height;
+    }
+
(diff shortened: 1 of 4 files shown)


core/XH.js
```diff
@@ -345,6 +345,24 @@ class XHClass extends HoistBase {
         return this.acm.sizingModeModel.sizingMode;
     }
 
+    //------------------------
+    // Viewport Size
+    //------------------------
+    /** @return {Object} - current viewport width / height. */
```
lbwexler, line 351: should mention that these are observable

amcclain:
Looks great - nice use of a supporting model class. Agreed it's helpful to call out the XH pass-through getters as observable. 

Also worth a one-line CHANGELOG entry in case e.g. any client developers are tracking the same within app code, then let's get it merged.

appcontainer/ViewportSizeModel.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Track observable width / height of the browser viewport, and provide observable
+ * access to device orientation
+ *
+ * @private
+ */
+export class ViewportSizeModel extends HoistModel {
+
+    /** @member {Number} */
+    @observable width;
```
lbwexler, line 19: since these change atomically and we are only actually revealing "size" let's just simplify things by having a single @observable.ref property that is the object.  Probably doesn't matter, but its going to be less code/ observables and complexity, and this thing may actually change very rapidly when window is being resized.

appcontainer/ViewportSizeModel.js
```diff
@@ -0,0 +1,60 @@
...
+
+    /** @returns {Object} */
+    @computed.struct
+    get size() {
+        const {width, height} = this;
+        return {width, height};
+    }
+
+    /** @returns {boolean} */
+    @computed
```
lbwexler, line 32: fantastic use of computed here.

appcontainer/ViewportSizeModel.js
```diff
@@ -0,0 +1,60 @@
...
+
+    /** @returns {boolean} */
+    get isLandscape() {
+        return !this.isPortrait;
+    }
+
+    constructor() {
+        super();
+        makeObservable(this);
+        window.addEventListener('resize', () => this.setViewportSize());
```
lbwexler, line 45: I think you ought to do all the work in the constructor -- or all the work in init().  The init phase only really there for supporting cross model initialization work.

Probably best to have all in constructor -- mobx might be able to less change handling for the observable (and its less code)

TomTirapani:
Thanks for taking a look guys! I've pushed an update to address your feedback.

Issue #2780, opened 2021-11-16
fmtNumber output not correct on decimal portion of large numbers with high precision


cnrudd:
This can be seen in toolbox's https://toolbox.xh.io/app/other/numberFormats page:

Format `123456000000000.1` with precision `6` and zeroPad `true`.

Expected result: `123,456,000,000,000.100000`
Actual result: `123,456,000,000,000.093750`

Petra and I think the root cause is a bug in the [numbro](https://github.com/BenjaminVanRyseghem/numbro) lib that hoist-react uses.

It looks like other users have come across this issue with [numbro](https://github.com/BenjaminVanRyseghem/numbro), and submitted PRs to fix this.
(one example: https://github.com/BenjaminVanRyseghem/numbro/pull/580)
but numbro is generally slow with PRs & releases.

Most solutions point to incorporating [bignumber.js](https://mikemcl.github.io/bignumber.js/) to fix this.  Numbro already uses bignumber, but not to fix this issue (yet).

amcclain:
You can confirm it's numbro @ https://numbrojs.com/try-me.html?lang=en-US&value=123456000000000.1&format=%7B%22mantissa%22%3A6%2C%22optionalMantissa%22%3Afalse%7D

cnrudd:
I fmtNumber is called without `formatConfig`, we only end up specifiying three 
numbro configs:
```
    config.thousandSeparated 
    config.mantissa 
    config.trimMantissa 
```

I think it's worth trying to do those 3 functions ourselves, or pull in bigNumber to do them.  I already have thousandSeparated done.

lbwexler:
I am going to close this ticket, as I think formatting issue is really just a single  manifestation of the bigger issue that Hoist does not provide support for BigNumber (now a built-in JS class).  This includes the data package, as well as number formatting. 

 Its not something clients have needed or wanted yet, although cyrpto projects are very likely  to want it.

PR #2785, opened 2021-11-18
Fix fmtNumber's support for large ints and floats.

cnrudd:
This PR resolves this issue: https://github.com/xh/hoist-react/issues/2780

This corresponding toolbox branch https://github.com/xh/toolbox/pull/541 should also be merged, it adds a few tests that help confirm this fix is working.

Some helpful reference links:
https://mikemcl.github.io/bignumber.js/#
https://stackoverflow.com/questions/45929493/node-js-maximum-safe-floating-point-number

Hoist P/R Checklist
-------------------

- [x] Reviewed and tested on Mobile: did not yet test on mobile, not expecting any issues.  
- [x] Created Toolbox branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### ⚙️ Technical
 
-* The `formatConfig` property of the fmtConfig methods's config argument has been deprecated.
+* The `formatConfig` property of the `fmtConfig` method's `config` argument has been deprecated.
 
 ## v44.1.0 - 2021-11-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
 * `DashContainerModel` now supports an observable `showMenuButton` config which will display a
   button in the stack header for showing the context menu
 
+### 🐞 Bug Fixes
+
+* `fmtNumber` now correctly rounds and zero pads very large floats.
+
+### ⚙️ Technical
+
+* The `formatConfig` property of the fmtConfig methods's config argument has been deprecated.
+
 ## v44.1.0 - 2021-11-08
 
 ### 🎁 New Features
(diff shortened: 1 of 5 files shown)


cnrudd:
### Instances of formatConfig used in client apps:
**Easy replacements:**
formatConfig: {thousandSeparated: true}        // not needed.  withCommas defaults to true
formatConfig: {mantissa: 0, thousandSeparated: false}    // precision: 0, withCommas: false
formatConfig: {thousandSeparated: false, mantissa: 1} // precision: 1, withCommas: false

**might need to add our own 'average' config to replace this: http://numbrojs.com/format.html#average**
formatConfig: {average: true, mantissa: 3, trimMantissa: true}
formatConfig: {average: true, mantissa: 1, optionalMantissa: true}

cnrudd:
I've confirmed that this change is working well, producing correct numbers in client app.

cnrudd:
Hi @petradish  & @lbwexler ,  
after discussions about this PR with each of you last night, I made a number of changes this morning that I think address the concerns each of you brought up.

Petra: I've taken your suggestion of declaring the full FORMAT config object and modifying its comma key.  The code also no longer modifies the FORMAT of a BN instance.  The format is now passed as an argument to the toFormat function.   This change improves in the process in 2 ways:
1. There is no risk of some earlier modification to a BN instance's FORMAT polluting our own format requirements
2. We do not modify any BN instance's FORMAT.

Lee: I've found that it wasn't necessary to do any constructor cloning, since the format can be passed as an argument.  So, the use of BN is more straightforward and easier to understand now.


Another change I did was work the use of BN more completely into all of the implementation functions.  This makes the diff look more significant, but I think this is a necessary change if we're going to go this way at all.

lbwexler:
Thanks Colin.  Would like to hold off on further work on this, until we decide that we actually want to support BigNumber in Hoist or just used by clients that really need it directly in their applications.

Issue #2716, opened 2021-10-04
Harmonize mobile button props with desktop where possible


amcclain:
Mobile button props are currently driven by the underlying Onsen component and are somewhat obscure. Would like to take any easy wins at supporting desktop props where we can translate into corresponding onsen button props - e.g. minimal, outlined.

TomTirapani:
Had a couple of spare cycles so took a stab at this - it's a long-standing gap that would be great to harmonize.
https://github.com/xh/hoist-react/pull/2792
https://github.com/xh/toolbox/pull/542

PR #2792, opened 2021-11-23
Harmonize mobile button props with desktop

TomTirapani:
+ Deprecate 'modifier' in favour of 'minimal' and 'outlined' props
+ Add support for intents to mobile buttons
+ Refactor implementation and styling for higher consistency with desktop

See issue: https://github.com/xh/hoist-react/issues/2716
See Toolbox PR: https://github.com/xh/toolbox/pull/542

Hoist P/R Checklist
-------------------


amcclain:
I'm eager to review this as soon as we are ready to merge in changes for our next major release. Thanks!

mobile/cmp/button/Button.scss
```diff
@@ -1,10 +1,171 @@
...
+  }
+
+  // Outlined mode
+  &--outlined,
+  &--outlined.xh-button--enabled:active {
+    background-color: transparent;
+    border: 1px solid var(--xh-button-bg);
+
+    &.xh-button--intent-none {
+      color: var(--xh-button-bg);
```
amcclain, line 132: Believe we want button-text var here, not bg

mobile/cmp/button/Button.scss
```diff
@@ -1,10 +1,171 @@
...
+      border-color: var(--xh-button-bg);
+      &.xh-button--enabled.xh-button--active {
+        background-color: var(--xh-button-bg);
+      }
+    }
+
+    &.xh-button--intent-primary {
+      color: var(--xh-intent-primary-darker);
```
amcclain, line 140: This looked a bit off when in a toolbar, especially when in dark mode, where the darker color appears to be more muted relative to the borders. Looks like outlined desktop buttons w/intents use the same color for text + border

amcclain (commented):
Thanks for running through all these changes - great to clean up this old inconsistency.

+ It's a bit of a pain to setup, but a mobile buttons page similar to what we have on desktop would be great to have in terms of quickly scanning through all the button styling variations. It's proven very useful there - especially for combinations like outlined/intent, toolbars, and button groups. Would you mind creating such a page?
+ Wondering if the grid helper buttons (autosize, colChooser) should default to minimal or outlined? Not sure about it, but they always seem a bit heavyweight.

mobile/cmp/button/Button.scss
TomTirapani: On desktop, the text color is the `-darker` variant of the border's intent color like here - but only for the light theme. I'll adjust mobile to use the `-lighter` variant for both text and border in dark theme, which will get us full parity with desktop.

TomTirapani:
Thanks for taking a look!

Great idea about the buttons page - I've added it to Toolbox, and it was a great help to spot inconsistencies with desktop. I've pushed a number of updates to address issues and the dark theme styling, and think it's ready for another review.

TomTirapani:
Regarding defaulting the helper buttons to `minimal` / `outlined` - I quite like the heavy weight buttons on mobile, where you don't have hover / cursor pointer to communicate interactivity and thus less room for subtlety. But I don't feel too strongly about it, could be convinced otherwise :)

amcclain:
Sorry to have missed you earlier today and to hold this up for one more cycle, but looks like we lost the "pressed" or `:active` styling. IMO that's an important bit of UI feedback and something we want to maintain / improve (similar to your recent updates to menu items, pending updates for grid rows).

It's confusing because of the dual-use of the term "active" for both "depressed to indicate something is turned on" buttons (`--active` suffix) and "currently held down by the user" buttons (`:active` modifier).

I spent a little while tonight seeing if I could make some minimal / obvious changes to get what I was looking for, but decided that I should touch base with you before going any further down the rabbit hole.  Let's sync up tomorrow (Tuesday) AM on it. 

As part of this, I think we'll want/need mobile-specific overrides for `--xh-button-bg-darker/darkest` to go along with the override for `--xh-button-bg` - believe we'll use the "darkest" modifier for `:active` as we do on desktop.

(Also note I pushed a93b7354262f0e00c42cd981c6485790463918fe to add support for a dedicated border-radius bar for buttons and applied it to the mobile buttons)

amcclain (approved):

PR #2793, opened 2021-11-29
Auto reload app for non-SSO auth situations

lbwexler:
The current proposal here just makes a very specific patch for our handling of our own forms based auth in Hoist.

Did not see a clear path or need to come up with a more general API at this time.  However, this should not cause problems for Colin's use case at client.

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/FetchService.js
+++ b/svc/FetchService.js
@@ -27,6 +27,11 @@ import {never} from '@xh/hoist/promise';
  *
  * Note that the convenience methods `fetchJson`, `postJson`, `putJson` all accept the same options
  * as the main entry point `fetch`, as they delegate to fetch after setting additional defaults.
+ *
+ * Note: For non-SSO apps, FetchService will automatically trigger a reload of the app if a
+ * 401 is encountered from a local (relative) request.  This default behavior is designed to allow
+ * more seamless re-establishment of timed out authentication sessions, but can be turned off
+ * via config if needed.
  */
 export class FetchService extends HoistService {
 
@@ -243,15 +248,18 @@ export class FetchService extends HoistService {
     }
 
     async maybeReloadForAuthAsync() {
-        // Reload app to renew auth. Be careful not to trigger during
-        // initialization, or to get caught in a tight reload loop
-        const {appState, localStorageService} = XH;
-        if (appState !== AppState.RUNNING && appState !== AppState.SUSPENDED) return;
-        if (!olderThan(localStorageService.get('xhLastAuthReload', null), ONE_MINUTE)) return;
+        const {appState, configService, localStorageService} = XH;
 
-        localStorageService.set('xhLastAuthReload', Date.now());
-        XH.reloadApp();
-        await never();
+        // Don't interfere with initialization, avoid tight loops, and provide kill switch
+        if (
+            configService.get('xhReloadOnFailedAuth', true) &&
+            appState === AppState.RUNNING &&
+            olderThan(localStorageService.get('xhLastFailedAuthReload', null), ONE_MINUTE)
+        ) {
+            localStorageService.set('xhLastFailedAuthReload', Date.now());
+            XH.reloadApp();
+            await never();
+        }
     }
 
     async sendJsonInternalAsync(opts) {
--- a/svc/FetchService.js
+++ b/svc/FetchService.js
@@ -4,14 +4,14 @@
  *
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
-import {HoistService, XH} from '@xh/hoist/core';
+import {HoistService, XH, AppState} from '@xh/hoist/core';
 import {Exception} from '@xh/hoist/exception';
-import {isLocalDate} from '@xh/hoist/utils/datetime';
+import {isLocalDate, SECONDS, ONE_MINUTE, olderThan} from '@xh/hoist/utils/datetime';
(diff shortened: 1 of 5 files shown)


svc/FetchService.js
```diff
@@ -232,12 +232,28 @@ export class FetchService extends HoistService {
...
 
+    async maybeReloadForAuthAsync() {
+        // Reload app to renew auth. Be careful not to trigger during
+        // initialization, or to get caught in a tight reload loop
+        const {appState, localStorageService} = XH;
+        if (appState !== AppState.RUNNING && appState !== AppState.SUSPENDED) return;
+        if (!olderThan(localStorageService.get('xhLastAuthReload', null), ONE_MINUTE)) return;
+
+        localStorageService.set('xhLastAuthReload', Date.now());
+        XH.reloadApp();
```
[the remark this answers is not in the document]
cnrudd: Maybe the default should be to XH.reloadApp() if not alternate requireAuthHandler method provided?

svc/FetchService.js
```diff
@@ -232,12 +232,28 @@ export class FetchService extends HoistService {
 
         if (!ret.ok) {
             ret.responseText = await this.safeResponseTextAsync(ret);
-            throw Exception.fetchError(opts, ret);
+            const e = Exception.fetchError(opts, ret);
+            if (!XH.isSSO && isRelativeUrl && e.httpStatus === 401) {
```
cnrudd, line 236: I see you're restricting this to a pretty narrow use case.  The client app I worked on that has an auth requirement will still need to extend fetchService.

But what specifically is the app/use case we're solving for here?  maybe you can tell me over chat?

svc/FetchService.js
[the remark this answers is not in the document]
lbwexler: valid point about the big form.  but I hope in practice that's not going to be any more common than, say, losing your work because the app went to sleep.  If the app is not asleep, the session should be kept alive by version checking and other timed processes.   

This code is really designed to catch the case on mobile, where the app has not gone to sleep, but the session with the server has somehow died.  So presumably the case where the user has left their phone app unused for a while. Not sure we need to have high expectations about preserving any data at that point.

svc/FetchService.js
lbwexler: yes, agreed.  Use case discussed above -- Non-SSO mobile app that has lost its session and needs to immediately proceed to re-initialization of app.

svc/FetchService.js
```diff
@@ -232,12 +232,28 @@ export class FetchService extends HoistService {
...
+        // initialization, or to get caught in a tight reload loop
+        const {appState, localStorageService} = XH;
+        if (appState !== AppState.RUNNING && appState !== AppState.SUSPENDED) return;
+        if (!olderThan(localStorageService.get('xhLastAuthReload', null), ONE_MINUTE)) return;
+
+        localStorageService.set('xhLastAuthReload', Date.now());
+        XH.reloadApp();
+        await never();
+    }
+
```
amcclain, line 256: What about showing the user a brief message to indicate what's happening - no action required, and might be a bit less mysterious. 

I think this pattern of reloading to prompt re-auth is relatively common, so it shouldn't be a shock in the intended use case where the user is returned to a login prompt, but in case anything doesn't work quite as planned this could also help us understand what a user is reporting around unexpected reloads.

```suggestion
    // Reload app to renew auth.
    async maybeReloadForAuthAsync() {
        const {appState, localStorageService} = XH;
        // Be careful not to trigger during initialization...
        if (appState !== AppState.RUNNING && appState !== AppState.SUSPENDED) return;
        // ...and avoid any chance of a tight reload loop.
        if (!olderThan(localStorageService.get('xhLastAuthReload', null), ONE_MINUTE)) return;

        XH.alert({
            title: 'Authentication Expired',
            message: fragment(
                `We're sorry, but you must log on again to continue using ${XH.clientAppName}.`,
                br(),
                `This page will reload momentarily...`
            ),
            confirmProps: {text: 'Reloading...', disabled: true}
        });
        await wait(5000);

        localStorageService.set('xhLastAuthReload', Date.now());
        XH.reloadApp();
        await never();
    }
```

svc/FetchService.js
```diff
@@ -4,14 +4,14 @@
...
+import {HoistService, XH, AppState} from '@xh/hoist/core';
 import {Exception} from '@xh/hoist/exception';
-import {isLocalDate} from '@xh/hoist/utils/datetime';
+import {isLocalDate, SECONDS, ONE_MINUTE, olderThan} from '@xh/hoist/utils/datetime';
 import {throwIf} from '@xh/hoist/utils/js';
 import {StatusCodes} from 'http-status-codes';
 import {isDate, isFunction, isNil, omitBy} from 'lodash';
 import {stringify} from 'qs';
-import {SECONDS} from '@xh/hoist/utils/datetime';
+import {never} from '@xh/hoist/promise';
```
amcclain, line 14: For the real suggestion below (my use of fancy suggestion feature a bit sub-optimal here...)

```suggestion
import {never, wait} from '@xh/hoist/promise';
```

amcclain (approved):
Looks good. One suggestion below to consider.

Also wondering if we would ever need a way to disable, and if we could add a public comment anywhere to document re. this behavior  - e.g. in FetchService header doc comment.

PR #2794, opened 2021-11-30
Use Export formatting for Clipboard

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   button in the stack header for showing the context menu
 * Added `GridAutosizeMode.MANAGED` to autosize Grid columns on data or `sizingMode` changes, unless
   the user has manually modified their column widths.
+* Copying from Grids to the clipboard will now use the formatting provided by the `exportValue` and
+  `exportFormat` properties on the column.
 
 ### ⚙️ Technical
 
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -774,8 +774,6 @@ class GridLocalModel extends HoistModel {
         this.model.agGridModel.agApi.setFilterModel(filterState);
     }
 
-    // Underlying value for treeColumns is actually the record ID due to getDataPath() impl.
-    // Special handling here, similar to that in Column class, to extract the desired value.
     processCellForClipboard = ({value, node, column}) => {
         const {model} = this,
             recId = node.id,
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -776,8 +776,21 @@ class GridLocalModel extends HoistModel {
 
     // Underlying value for treeColumns is actually the record ID due to getDataPath() impl.
     // Special handling here, similar to that in Column class, to extract the desired value.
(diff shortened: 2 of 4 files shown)


cmp/grid/Grid.js
```diff
@@ -776,8 +776,21 @@ class GridLocalModel extends HoistModel {
 
     // Underlying value for treeColumns is actually the record ID due to getDataPath() impl.
```
haynesjm42, line 777: Comment is irrelevant now

svc/GridExportService.js
```diff
@@ -131,6 +142,58 @@ export class GridExportService extends HoistService {
...
+        // Get cell-level format if function form provided
+        let {exportFormat} = column,
+            cellHasExportFormat = isFunction(exportFormat);
+
+        if (cellHasExportFormat) {
+            exportFormat = exportFormat(value, {record, column, gridModel});
+        }
+
+        // Enforce date formats expected by server
+        if (exportFormat === ExportFormat.DATE_FMT) value = fmtDate(value);
```
haynesjm42, line 187: Should these only be applied if `forServer === true`? I'm thinking no - I would want a formatted value in my clipboard, not the `Date.toString()` value, so maybe the comment above is just confusing? This is also inconsistent in that we apply date formatting to date fields when copying, but not to other field types - though I think it does make sense.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
   button in the stack header for showing the context menu
 * Added `GridAutosizeMode.MANAGED` to autosize Grid columns on data or `sizingMode` changes, unless
   the user has manually modified their column widths.
+* Copying from Grids to the clipboard will now use the formatting provided by the `exportValue` and
```
haynesjm42, line 12: I don't think this is correct since we will never apply the `exportFormat` on the value we put in the clipboard (nor can we since it is an excel-specific format string in most cases)

svc/GridExportService.js
```diff
@@ -255,48 +318,14 @@ export class GridExportService extends HoistService {
...
     getRecordRow(gridModel, record, columns, depth) {
-        let aggData = null;
-        if (gridModel.treeMode && record.children.length) {
-            aggData = gridModel.agApi.getRowNode(record.id).aggData;
-        }
-        const data = columns.map(it => this.getCellData(gridModel, record, it, aggData));
+        const node = gridModel.agApi.getRowNode(record.id),
+            forServer = true,
+            data = columns.map(column => {
+                return this.getExportableValueForCell({gridModel, record, column, node, forServer});
```
haynesjm42, line 324: Why not just do `forServer: true` and remove the extra variable above?

cmp/grid/Grid.js
lbwexler: great catch

saba-mo (approved):

CHANGELOG.md
lbwexler: we do apply in the most common cases.  Understand you point bug going to let this stand..

Issue #2798, opened 2021-12-01
FilterChooser renders inner scrollbar even when value tags do not wrap


cnrudd:
This appears to be a css issue in Chrome 96+.  Does not happen in Chrome 95.
This does not happen in our select component.

![image](https://user-images.githubusercontent.com/2573928/144311867-c1254c0f-0560-4562-aed9-ccc1ad0c254a.png)
![image](https://user-images.githubusercontent.com/2573928/144312087-467e59f0-1b23-46ca-a80e-18a4e9ccccc6.png)

amcclain:
Reported by @haynesjm42 today - I've seen this also and have been meaning to file this ticket. Colin / John - if either of you have a few cycles to look this week please comment or chat so we don't dupe up the work - hopefully a quick style fix, but with React Select you never know...

cnrudd:
This issue exists on Safari 15.1 too.  So, it is not strictly a Chrome version bug, as I previously thought.
The artifact is the vertical scrollbar.

Definitely, if these lines are tweaked:

```
hoist-react/desktop/cmp/input/Select.scss
// Ensure multi-value selects don't overflow vertically in the toolbar
.xh-toolbar {
  .xh-select {
    &__value-container--is-multi {
      height: 28px;
      overflow-y: auto !important;
    }
  }
}
```
The scrollbars go away.  But those lines are needed for their specified purpose.
It's the lineheight of the element inside that is making the element too big.
![image](https://user-images.githubusercontent.com/2573928/144464814-99c03b8e-9109-44f7-afbc-d65c33946dbf.png)
and by just a smidge over 28.

Lineheight is coming from blueprint.css:
![image](https://user-images.githubusercontent.com/2573928/144464975-de30371d-96d0-460f-a48d-14c8b561254e.png)
Will look into whether that changed recently.

PR #2805, opened 2021-12-02
Ensure scrollbar does not appear on mulit-select in toolbar when not needed.

cnrudd:
This fixes https://github.com/xh/hoist-react/issues/2798

Hoist P/R Checklist
-------------------


desktop/cmp/input/Select.scss
```diff
@@ -182,6 +182,9 @@
     &__value-container--is-multi {
       height: 28px;
       overflow-y: auto !important;
+      div {
+        line-height: 14px;
```
cnrudd, line 186: I considered and tested `line-height: 1;`
but I think if we're going with a px height for the parent, px line-height is more appropriate, more predictable.
Also, line-height: 1;`makes the field label even MORE jumpy than it already is.  Wiggles twice as selections are added, as opposed to just one wiggle now in production and with the px value.

amcclain: Hi - thanks for setting this up and for the comment re. the fixed pixel value here. I tried to reproduce / understand the "wiggling" you're seeing, but I could not.

The bug and fix are both scoped to multi-selects within Toolbars. In Toolbox we have examples of that on the Dashboard in the GitHub activity widget, and in the Portfolio example app in the Orders detail panel.  Are there others examples you were looking at?

If you can show me or help me repro the issue with `line-height:1` that would be great (also want to see what is wiggling now prior to any fix).  

Finally, I tried applying the line-height rule directly to the `&__value-container--is-multi` selector - so no nested div rule - and that seemed to also work fine.

cnrudd: @amcclain Here is a screen capture of the "wiggle" in production.  The label jumps up 1 or 2 pixels when a state is selected, and jumps down a state is remove.
https://user-images.githubusercontent.com/2573928/144713280-74bcf416-8026-4c57-bc1c-31e18a26efe5.mov

line-height:1 caused a 2nd wiggle when enough states were added that scrolling was necessary
line-height:14px did not cause a 2nd wiggle in that case.

Funny enough, today, in my local instance, I do not get the wiggle AT ALL, with the fix applied.

I have confirmed that the fix works in the Portfolio example.

I have not confirmed that the fix works in the GitHub acitivity widget, because I do not have any data loading in there.
But the bug does not appear in that widget to begin with, because that widget is using the "compact" toolbar, and resulting calculated lineheight is less than 14px.  This makes me think that we might want to scope the fix to "not compact", or try with line-height:1, which may adapt better to modifiers like "compact".

Your suggestion of putting the line-height:n setting directly on the &__value-container--is-multi selector makes sense, since it would cascade down.  I haven't yet had a chance to test that though.

Where can I find docs on how to get the git data in my local instance?

amcclain: Great thanks, exactly what I needed to know - and that compact example on the activity tab was a concern of mine re. the fixed-pixel height (despite it seeming to be OK).  

Revised approach looks simpler and will merge it now, appreciate it.

PR #2813, opened 2021-12-06
ViewportSizeModel prefers window.orientation, if available

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/ViewportSizeModel.js
+++ b/appcontainer/ViewportSizeModel.js
@@ -6,6 +6,7 @@
  */
 import {HoistModel} from '@xh/hoist/core';
 import {action, computed, observable, makeObservable} from '@xh/hoist/mobx';
+import {isFinite} from 'lodash';
 
 /**
  * Track observable width / height of the browser viewport, and provide observable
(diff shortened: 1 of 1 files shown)


appcontainer/ViewportSizeModel.js
```diff
@@ -18,10 +19,15 @@ export class ViewportSizeModel extends HoistModel {
...
     @observable.ref size;
 
+    /** @member {int} - orientation in degrees [0, 90,180, 270] */
+    @observable.ref orientation;
+
     /** @returns {boolean} */
     @computed
     get isPortrait() {
-        return this.size.width < this.size.height;
+        const {orientation, size} = this;
```
lbwexler, line 28: this is a bit of a fragile `computed`- - it is relying on the fact that we dereference the size on line 28.  If we moved that down past the next line, our observability would break.  Probably need a comment if we want to let size drive the observability.

In general, am wondering two things: 

1) Should we skip the public observable 'orientation'? Seems like it kind of conflicts with isPortrait, and its a deprecated API.

2) Perhaps instead within isPortrait, we can just check the new orientation, and then the deprecated orientation, and then the width < height

TomTirapani: Great suggestions - will make these change 👍

lbwexler:
Closing this -- was replaced with a simpler solution using screen co-ordinates

Issue #2802, opened 2021-12-02
Improve discoverability of FilterChooser fields


haynesjm42:
It would be nice if the filter field list/info was available somehow without needing to type anything into the filter chooser, especially for cases where the chooser is not directly tied to columns in a grid. Have been asked multiple times over the last few months to remind (non-business) users which fields are supported.

One option would be to display all of the example options as default options when you focus onto the FilterChooser. This may further complicate the already complicated implementation though.

Another option may be to build a help popup which displays the same example options in a list for reference. This could be tied to a separate component that apps could display next to the filter chooser (a ? icon with a popover) or elsewhere in the app if they wanted to. Might also consider adding some description/additional text to be displayed to describe the fields if the display name is not sufficient.

amcclain:
Hi - I like this suggestion, as I've had similar usability concerns and think that many users might not realize / be able to use the full power of the component. Would love to brainstorm more on this.

amcclain:
Would like to discuss this when we have a chance / let us know if this has come up again. We can up the priority or look for incremental wins.

TomTirapani:
I'm also interested in this one, and happy to take it on. Available to discuss options any time.

TomTirapani:
I took a pass at determining the feasibility of the first suggested solution (showing the list fields before typing a query), and it turned out to be easier and less invasive that I expected. Was able to implement with minimal code changes and no breaking changes here:

https://github.com/xh/hoist-react/pull/2829

I know we wanted to brainstorm solutions, and am still open to different approaches - just thought I might as well open the PR for discussion since it was relatively low hanging fruit.

lbwexler:
thanks tom -- all about the low-hanging fruit. 🍌

PR #2829, opened 2021-12-21
Display list of fields in FilterChooser when focused.

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/2802
See toolbox branch: https://github.com/xh/toolbox/tree/filterChooserDiscoverability

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/FilterChooserModel.js
+++ b/cmp/filter/FilterChooserModel.js
@@ -48,6 +48,9 @@ export class FilterChooserModel extends HoistModel {
     /** @member {FilterChooserFieldSpec[]} */
     @managed fieldSpecs = [];
 
+    /** @member {boolean} */
+    suggestFieldsWhenEmpty;
+
     /** @member {number} */
     maxTags;
 
@@ -93,6 +96,8 @@ export class FilterChooserModel extends HoistModel {
      *      FieldFilters, to be 'AND'ed together.
      * @param {(Filter[]|function)} [c.initialFavorites] - initial favorites as an array of filter
      *      configurations, or a function to produce such an array.
+     * @param {boolean} [c.suggestFieldsWhenEmpty] - true to offer all field suggestions when the
+     *      control is focussed with an empty query, to aid discoverability.
      * @param {number} [c.maxTags] - maximum number of filter tags to render before disabling the
      *      control. Limits the performance impact of rendering large filters.
      * @param {number} [c.maxResults] - maximum number of dropdown options to show before
@@ -106,6 +111,7 @@ export class FilterChooserModel extends HoistModel {
         valueSource = bind,
         initialValue = null,
         initialFavorites = [],
+        suggestFieldsWhenEmpty,
         maxTags = 100,
         maxResults = 50,
         persistWith,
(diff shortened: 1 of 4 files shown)


haynesjm42:
Pulled down the branch, I think this is a perfect solution to the issue of discoverability. I think an argument could also be made for enabling this be default. 

Also could see wanting this to show only when you click on the filter chooser, but not necessarily when it is focused - the list of fields may be distracting if it pops up while tabbing through inputs or if the chooser is focused on render in a dialog or something.

cmp/filter/FilterChooserModel.js
```diff
@@ -106,6 +111,7 @@ export class FilterChooserModel extends HoistModel {
         valueSource = bind,
         initialValue = null,
         initialFavorites = [],
+        suggestFieldsWhenEmpty,
```
lbwexler, line 114: I think worth it to specify '= false' in situations like this for documentation/clarity

cmp/filter/impl/QueryEngine.js
```diff
@@ -66,7 +67,15 @@ export class QueryEngine {
     }
 
     //------------------------------------------------------------------------
-    // 1) No op yet, so field not fixed -- get field or value matches.
+    // 1) No query -- return all field suggestions
+    //------------------------------------------------------------------------
+    suggestAllFields() {
+        // Suggest all for the user to select on their way to a more targeted query.
```
lbwexler, line 73: for clarity/robustness think we could call this method `whenNoQuery` `noQueryPresent` etc.  and then within we can say return suggestFieldsWhenEmpty ? this.getFieldOpts() : []

lbwexler:
Very strongly agree with John's idea that we should NOT show it on focus. 

Don't feel strongly about whether it should be the default.  I suppose it probably should,

TomTirapani:
Great - thanks for the feedback guys! Just pushed an update with the following:

+ Don't show suggestions on keyboard focus
+ Default suggestFieldsWhenEmpty to true
+ Tweak QueryEngine code for clarity

lbwexler:
Merging this -- note that its very lightweight @atm, so  easy to tweak (or revert) before we ship if you have any additional feedback

amcclain:
Didn't have a chance yet to look at the code change - would still like to, but I did preview in place in Toolbox. Very happy to take any such improvements to usability.

Two quick thoughts:
* All of the leading "e.g."s are a bit weird, especially since they are rendered in the highest contrast color. I'd vote for removing or modifying those.
* Fields must be shown in the order in which they are defined within the fields config. That's probably right, in terms of giving the developer the most flexibility in ordering - e.g. most common fields up top. That said, my hunch is that this ordering will often be (or at least appear to be) random. Any thoughts as to whether it would be better to sort by default? Perhaps even worth an option?

lbwexler:
agree with removing the  multiple e.g. for sure.  Was going to say as much!

TomTirapani:
Thanks for the suggestions - agree with both.

Regarding the 'e.g.' string, I think we should probably a) remove it entirely for the empty query case here and b) keep it for the partial field match case (it may be shown alongside value suggestion), but mute the e.g. string. The same suggestion rows is used in both cases.

Regarding the sort, agree it makes sense to sort them. I'll make it an option defaulted to true.

TomTirapani:
Opened a new PR to address the feedback above: https://github.com/xh/hoist-react/pull/2840

Issue #2803, opened 2021-12-02
Charts that re-render while in fullscreen shrink in size


saba-mo:
As seen in a client app, viewing a chart in fullscreen mode and then changing the date range and re-rendering the chart causes it to shrink down to the dimensions of its parent element. A contrived toolbox example can be found in the portfolio app example on this branch (use range selector buttons to see behavior): https://github.com/xh/toolbox/tree/fullscreenChart

![Screen Shot 2021-12-02 at 12 12 55 PM](https://user-images.githubusercontent.com/67703233/144495816-6faa5a94-1f01-47f7-b31b-18900d8f57f4.png)

amcclain:
Completing #2753 would allow this menu option to be omitted for use cases that trigger this bug. We don't want to invest too much in trying to change/patch the built-in behavior here.

PR #2835, opened 2021-12-23
#2803 - Chart does not re-render well in fullscreen mode so just close fullscreen mode if it's open.

cnrudd:
Simple mitigation strategy for the issue described in #2803.
If a chart in fullscreen mode re-renders for some reason, (usually because new data is loaded in to accommodate a date range change) the chart will now exit fullscreen.  Users can get back to fullscreen with the new date range if they want to.

There are other more complex solutions, such as "fade document.body opacity, exit fullscreen before re-render, then re-open fullscreen right after render, then restore document.body opacity" but this creates some UI "jank" and might add more to maintenance complexity than we'd like.

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 ### 🐞 Bug Fixes
 
 * Fixes issue displaying the DashContainer overflow menu when the menu button is enabled
+* A chart in fullscreen mode now gracefully exits fullscreen mode before re-rendering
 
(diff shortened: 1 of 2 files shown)


lbwexler:
Merging this -- we may definitely revisit, but this is a relatively simple mitigation of the problem.  Might be a bit annoying if the user gets here, but much better than the current behavior.  Thanks Colin!

PR #2839, opened 2021-12-28
Delete logs

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/server/logViewer/LogViewerModel.js
+++ b/admin/tabs/server/logViewer/LogViewerModel.js
@@ -117,7 +117,7 @@ export class LogViewerModel extends HoistModel {
             const recs = this.filesGridModel.selectedRecords,
                 count = recs.length;
             if (!count ||
-                !(await XH.confirm({title: 'Confirm Delete?', message:`Delete ${count} files?`}))) {
+                !(await XH.confirm({title: 'Confirm Delete?', message: `Delete ${count} files?`}))) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


admin/tabs/server/logViewer/LogViewerModel.js
```diff
@@ -103,6 +112,26 @@ export class LogViewerModel extends HoistModel {
         }
     }
 
+    async deleteSelectedAsync() {
+        try {
+            const recs = this.filesGridModel.selectedRecords,
+                count = recs.length;
+            if (!count ||
+                !(await XH.confirm({title: 'Confirm Delete?', message: `Delete ${count} files?`}))) {
```
TomTirapani, line 120: I haven't seen this pattern of using await within the condition before - cool!

TomTirapani: Although I must admit I find it a bit hard to parse, given that it may or may not be async depending on the previous clause. Might be clearer to split them up into two different checks, one for the early out and one for async user confirmation?

amcclain:
Thanks - how very convenient that this was already available as a hoist-core endpoint!

Pushed minor tweaks in 330040a97cc3b2b3ef3e1f7175300588cc85ed15 to use RecordActions and have both delete and download in the context menu + bottom toolbar.  I had been meaning to move the download option for a while now due to space constraints on the top toolbar, so this was a nice prompt to do so.

Issue #2843, opened 2021-12-29
Grid ValuesTabModel extracts values from store, even when not enabled


amcclain:
We've discussed this when we first implemented grid filters, I believe, but now might be a decent time to revisit.

Within `ValuesTabModel`, we are doing some non-trivial work to extract and compute a list of values from the source store/cube, reconcile those with current filter values, and load into an inner managed GridModel.

All of this happens even when `fieldSpec.enableValues == false`. From what I can see, this flag only effects the default tab shown + visibility of the tab switcher within the header dialog. 

Can we (safely) avoid doing this values extraction - or even avoid creating a `ValuesTabModel` altogether within `ColumnHeaderFilterModel` - when this feature is disabled? We don't have any expectation that it can be "switched on" post-construction, so that should ease matters a bit.

TomTirapani:
PR here: https://github.com/xh/hoist-react/pull/2844

Issue #2841, opened 2021-12-29
Support values autocompletion within GridFilter custom tab


amcclain:
When the user is working with a grid column filter in the "Custom" tab and adding an equals/does not equal clause on a column/field with `enableValues:true`, we could consider autocompleting from the available values, as we do in `filterChooser`.

TomTirapani:
PR here: https://github.com/xh/hoist-react/pull/2846

PR #2844, opened 2021-12-30
Avoid extracting values from store when GridFilterFieldSpec.enableValues is false

TomTirapani:
It was straightforward enough to simply not create the ValuesTabModel if enableValues is false.

See issue [#2843](https://github.com/xh/hoist-react/issues/2843)

Note that this implementation is likely to change with https://github.com/xh/hoist-react/issues/2841, since the value extraction code will likely be refactored to be shared by both tabs. But I still think its worth taking this improvement in the interim.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/impl/filter/ColumnHeaderFilterModel.js
+++ b/desktop/cmp/grid/impl/filter/ColumnHeaderFilterModel.js
@@ -36,6 +36,10 @@ export class ColumnHeaderFilterModel extends HoistModel {
         return this.fieldSpec.field;
     }
 
+    get enableValues() {
+        return this.fieldSpec.enableValues;
+    }
+
     get currentGridFilter() {
         return this.gridFilterModel.filter;
     }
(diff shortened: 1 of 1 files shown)


desktop/cmp/grid/impl/filter/ColumnHeaderFilterModel.js
```diff
@@ -74,7 +78,7 @@ export class ColumnHeaderFilterModel extends HoistModel {
         this.column = column;
         this.fieldSpec = filterModel.getFieldSpec(column.field);
 
-        this.valuesTabModel = new ValuesTabModel(this);
+        this.valuesTabModel = this.enableValues ? new ValuesTabModel(this) : null;
```
lbwexler, line 81: seems like this could be the only place we reference enableValues?   And just don't include the tab in the container.  Then null checks everywhere else on the valuesTabModel, and the activeTab can never be 'valuesFilter'

PR #2847, opened 2022-01-04
Make LocalDate.get() accept hyphens.

lbwexler:
Not the final word on this but this bit of work could be done easily now without any API changes, or required changes to Hoist Core.

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/datetime/LocalDate.js
+++ b/utils/datetime/LocalDate.js
@@ -26,10 +26,6 @@ export class LocalDate {
     static _instances = new Map();
     static VALID_UNITS = ['year', 'quarter', 'month', 'week', 'day', 'date'];
 
-    // Very basic preliminary regex to partially validate input to LocalDate.get().
-    // Input fully validated as a date when passed to moment in constructor.
-    static fmtRegEx = new RegExp(/^\d{8}$/);
-
     _isoString;
     _moment;
     _date;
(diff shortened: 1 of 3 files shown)


utils/datetime/LocalDate.js
```diff
@@ -41,29 +41,22 @@ export class LocalDate {
...
-            ret = _instances.get(s);
-
-        if (!ret) {
-            ret = new LocalDate(s);
-            _instances.set(s, ret);
-        }
-
-        return ret;
+        if (!isString(s)) this.throwInvalidArgument(s);
+        s = s.replaceAll('-', '');
```
cnrudd, line 50: if here the code now took just the first 8 chars, it would open up easily converting many time strings to LocalDate.
Is there a safety/integrity use case where if a developer deliberately tries to convert a time string that has time or zone info to a LocalDate they should be blocked?

utils/datetime/LocalDate.js
```diff
@@ -313,8 +306,8 @@ export class LocalDate {
     //-------------------
     /** @private - use one of the static factory methods instead. */
     constructor(s) {
-        const m = moment(s, 'YYYYMMDD');
-        throwIf(!m.isValid(), `Invalid argument for LocalDate: ${s}`);
+        const m = moment(s, 'YYYYMMDD', true);
```
cnrudd, line 309: Nice using string here rather than the regex.

cnrudd (approved):
I think this is a step in the right direction.  Happy to go with incremental steps.  I'll work on the next step, if no one gets there first. :)

utils/datetime/LocalDate.js
lbwexler: You can use .from if you want that kind of parsing

cnrudd: Sure, but isn't the point of this change to make it easier to feed values into LocalDate when the feeding is done by an intermediary component, like the column parser, which uses only .get?

Issue #2845, opened 2022-01-03
`LocalDate.toString()` should be yyyy-mm-dd


cnrudd:
This change will bring the Hoist-React framework's LocalDate implementation in line with other implementations that follow the ISO-8601 standard:
https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html
https://nodatime.org/2.2.x/api/NodaTime.LocalDate.html
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

Default output format will change to yyyy-mm-dd
Default input formats accepted by the class' get method will be: yyyy-mm-dd && yyyymmdd

lbwexler:
Found the java link I looked at that made me feel that the format with out the hyphens was fully standard

BASIC_ISO_DATE is first, just by alphabetization, but the name is certainly as inviting as the ISO_LOCAL_DATE (that we are proposing moving toward)

https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

PR #2854, opened 2022-01-11
Change LocalDate's default output fmt to yyyy-mm-dd

cnrudd:
This is for issue #2845.
It is a breaking change.

Apps that take this change will need to use Hoist-Core v13.0.5 or later, as mentioned in the changelog entry.

Toolbox PR (with updated LocalDate tests): https://github.com/xh/toolbox/pull/546

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ### 💥 Breaking Changes
 
-* `LocalDate.toString()`, `LocalDate.toJSON()` and `LocalDate.valueOf()` all return a string with dashes: `YYYY-MM-DD`.  
-  Prior versions returned the string without dashes: `YYYYMMDD`.
+* `LocalDate` methods `toString()`, `toJSON()`, `valueOf()`, and `isoString()` now all return the
+standard ISO format: `YYYY-MM-DD`. (Prior versions returned`YYYYMMDD`).
+This is consistent with the built-in javascript `Date.toISOString()`
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ### 💥 Breaking Changes
 
 * `LocalDate.toString()`, `LocalDate.toJSON()` and `LocalDate.valueOf()` all return a string with dashes: `YYYY-MM-DD`.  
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,10 +3,17 @@
...
 
 ### 🎁 New Features
+
 * `FilterChooser` has new `menuWidth` prop, allowing you to specify as width for the dropdown
   menu that is different from the control.
 
+### 💥 Breaking Changes
+
+* `LocalDate.toString()`, `LocalDate.toJSON()` and `LocalDate.valueOf()` all return a string with dashes: `YYYY-MM-DD`.  
+  Prior versions returned the string without dashes: `YYYYMMDD`.  Requires Hoist-Core v13.0.5 or later.
```
lbwexler, line 13: if this is just for the parsing of the value in the admin activity/errors,  we could redo those to submitting to submit the old YYYYMMDD format temporarily, to avoid requiring taking a hoist core release to take this new hoist-react version

lbwexler: The main point is that applications need to be warned that if they were depending on the format of dates being submitted to the server as being YYYYMMDD they will have to change to accommodate that.

utils/datetime/LocalDate.js
```diff
@@ -313,7 +313,7 @@ export class LocalDate {
             !m.isValid(),
             `Invalid argument for LocalDate: ${s}.  Use 'YYYYMMDD' or 'YYYY-MM-DD' format.`
         );
-        this._isoString = s;
+        this._isoString = s.slice(0, 4) + '-' + s.slice(4, 6) + '-' + s.slice(6, 8);
```
lbwexler, line 316: seems like if we do this, we might as well cache the YYYY-MM-DD value,  and use that as the "main" format all around.  Basically need to reverse this login and add *in* the dashes in get()

This will work better if paired with a hypenated json serialization in Hoist Core as well.

lbwexler: Also --- there should be a cleanup to the rendering of the cube label in Hoist Activity Monitor -- we were jumping through some hoops to get the dashes in there.

cnrudd: Both suggestions applied in latest commits.

CHANGELOG.md
cnrudd: I've fixed the submits on the activity/error models and updated the changelog notice.

PR #2858, opened 2022-01-11
New `Store.loadTreeDataFrom` config

amcclain:
+ Allow developer to override path to children on raw data objects.
+ No change to `StoreRecord.children` (post-processing), defaults remain the same.

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -860,7 +860,7 @@ export class Store extends HoistBase {
 
             recordMap.set(id, rec);
 
-            if (loadTreeData && loadTreeDataFrom && raw[loadTreeDataFrom]) {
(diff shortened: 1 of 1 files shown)


amcclain:
This would be helpful for a client project - name is not the most elegant, and it does overlap somewhat with the existing boolean (i.e. they could be combined if we really wanted), but wanted to propose this incremental tweak w/o getting too much further into it.  

Happy to discuss the use case (involves loading custom aggregations of class-based objects, where the aggs should become parent records and the class instances child records, but where the class instances have their own `children` getter which would be difficult to change and which we don't want loaded into the store...)

PR #2859, opened 2022-01-11
Make tooltip more robust.

lbwexler:
Avoid ugly errors from casual tooltip implementors

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/columns/Column.js
+++ b/cmp/grid/columns/Column.js
@@ -20,9 +20,10 @@ import {
     isNil,
     isNumber,
     isPlainObject,
-    isString
+    isString,
+    toString
 } from 'lodash';
-import {forwardRef, useImperativeHandle, useState, createElement} from 'react';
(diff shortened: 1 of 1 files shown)


cmp/grid/columns/Column.js
```diff
@@ -651,7 +652,7 @@ export class Column {
                     tooltipSpec(val, {record, column: this, gridModel, agParams}) :
                     val;
 
-                return ret ?? null;
+                return isValidElement(ret) ? ret : toString(ret);
```
TomTirapani, line 655: Don't we still have to explicitly return null here rather than an empty string if ret is null/undefined? 

My understanding (haven't tested) is that this will display an empty tooltip, rather than no tooltip.

lbwexler:
Thanks for the careful look!  I did think of this and tested both
explicitly -- null and undefined are considered a valid element and so
still returned, and no tooltip returned.

The acceptance of undefined did surprise me, and I would like to understand
it better.  Perhaps this code is too clever to rely on that.

Issue #2815, opened 2021-12-06
ExceptionHandler should remove auth token headers from fetch stacktraces


amcclain:
Stacktraces serialized by Hoist's `ExceptionHandler` can contain an OAuth JWT, commonly in a `fetchOptions.headers.Authorization` key. 

While these tokens should have a relatively tight expiry, it would be best for our `stringifyErrorSafely` util to redact these if found, replacing them with a placeholder string (e.g. `"[redacted]"`) so it's clear that there _was_ a token that was deliberately stripped out.

We could consider exposing an overridable static property on the `ExceptionHandler` class to allow apps to customize this behavior / mark additional keys for checking and redaction.

Issue #2772, opened 2021-11-04
Add XH.handleException option to alert user w/toast instead of modal


amcclain:
For some apps we have a distinct preference for avoiding modal alerts, and yet we still want to ensure that users see error messages.

We could offer an option to `XH.handleException()` util that would show any errors in a toast. Would not be bad if we lost the "add details and report" flow supported by the modal dialog - that's very rarely used.

Error toasts could also have a longer timeout and should definitely get danger toast styling.

PR #2865, opened 2022-01-14
Exception handling

TomTirapani:
+ ExceptionHandler should redacts auth token headers. [#2815](https://github.com/xh/hoist-react/issues/2815)
+ Add XH.handleException 'showWithToast' option, to alert user w/toast instead of modal. [#2772](https://github.com/xh/hoist-react/issues/2772)
+ Add support for showing action buttons within toasts.

Hoist P/R Checklist
-------------------


core/ExceptionHandler.js
```diff
@@ -50,6 +55,8 @@ export class ExceptionHandler {
      *      'isAutoRefresh' fetch exceptions.
      * @param {boolean} [options.showAlert] - display an alert dialog to the user. Default true,
      *      excepting 'isAutoRefresh' and 'isFetchAborted' exceptions.
+     * @param {boolean} [options.showWithToast] - if `showAlert`, display the error message in a toast
```
lbwexler, line 58: How about alertMode= dialog|toast, defaulting to dialog?  Might be more extensible, and clearer w.r.t showAlert property.

desktop/appcontainer/ExceptionDialogDetails.js
```diff
@@ -29,7 +30,7 @@ export const exceptionDialogDetails = hoistCmp.factory(
 
         if (!detailsIsOpen || !exception) return null;
 
-        const errorStr = stringifyErrorSafely(exception);
+        const errorStr = stringifyErrorSafely(exception, ExceptionHandler.REDACT_PATHS);
```
lbwexler, line 33: why not just build this in to stringifyErrorSafely?  Seems to be totally appropriate.

core/ExceptionHandler.js
TomTirapani: Great idea, will do 👍

desktop/appcontainer/ExceptionDialogDetails.js
TomTirapani: I wanted to, but you get a circular import between Stringify.js and ExceptionHandler.js that breaks the app

TomTirapani: We could put the overridable static property on the Exception class to get around it?

core/ExceptionHandler.js
```diff
@@ -16,6 +16,11 @@ export class ExceptionHandler {
 
     #isUnloading = false;
 
+    // Error property paths to replace with "[redacted]"
+    static REDACT_PATHS = [
+        'fetchOptions.headers.Authorization'
```
lbwexler, line 21: would add jsdoc here, and add to changelog, so application can enhance this

desktop/appcontainer/ExceptionDialogDetails.js
lbwexler: we can declare the list in Stringify.js, was not sure it needed to be on ExceptionHandler anyway.

core/ExceptionHandler.js
lbwexler: cool, (or alertType)

core/ExceptionHandler.js
```diff
@@ -62,7 +69,15 @@ export class ExceptionHandler {
 
         this.logException(exception, options);
         if (options.showAlert) {
-            XH.appContainerModel.exceptionDialogModel.show(exception, options);
+            if (!options.showWithToast) {
+                XH.toast({
+                    message: exception.message,
+                    intent: 'danger',
+                    timeout: 10000
```
lbwexler, line 76: + Can we provide a link in the toast to bring up the full dialog with the show call on line 79?
+ struggling with whether we need to customize this toast?  Seems like we could do it with a global setting on ExceptionHandler in a future release if we ever needed to do that.  
+ Seems like we need a bundle of options we can set on ExceptionHandler globally when the app is loading ("ExceptionHandlerOptions"), this could include, say, the REDACT text, but also whether toast or dialog is the default, toast defaults, etc. etc.

TomTirapani:
Thanks for taking a look Lee!

I've updated the branch with your feedback, including:
+ New `ExceptionHandlerDefaults` to set app-wide defaults
+ Open the exception dialog by clicking on button within toast
+ `showWithToast` > `alertType`

lbwexler:
Thanks for getting this done so fast -- looks great -- just a matter of the "best" way to specify the configuration class now -- should it be statics on the main class?    Properties on an anonymous object?

core/ExceptionHandler.js
```diff
@@ -62,7 +68,19 @@ export class ExceptionHandler {
 
         this.logException(exception, options);
         if (options.showAlert) {
-            XH.appContainerModel.exceptionDialogModel.show(exception, options);
+            const alertType = options.alertType ?? ExceptionHandlerDefaults.ALERT_TYPE;
+            if (alertType === 'toast') {
+                XH.toast({
+                    message: exception.message,
```
lbwexler, line 74: Think we do need a bit more structure, including the title "Exception" -- Raw Exception messages are just two varied and 

Can we have the toast show a styled component with a title of Exception, and then the message.  Think this might look a little more sane and controlled....

core/ExceptionHandlerDefaults.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+/**
+ * Provides a centralized location to set Exception Handling defaults.
+ * Can be overwritten to apply throughout the application.
+ */
+export class ExceptionHandlerDefaults {
+
+    /**
+     * Property paths within error details JSON to replace with REDACT_TEXT
```
lbwexler, line 15: Am trying to see about getting these on to ExceptionHandler itself for API simplicity.  Would like the exception package *not* to depend on core.  So we can go back to your original method of passing the redaction to the stringify method, or perhaps just remove the stringify method from the public API.

TomTirapani:
Thanks Lee.

I took another pass at it, and brought the static options onto ExceptionHandler. I ended up moving the stringify method to ExceptionHandler as well, to resolve the circular dependency while still keeping it available for application code.

Also styled and titled the toasts a bit, think they look a bit more structured now:
Screenshot 2022-01-18 at 11 47 39

lbwexler:
Looks fantastic!

amcclain:
That does look great - eager to start using it. 👍 

Thoughts on going with more user-friendly "Error" vs. "Exception" ?

lbwexler:
Yes,  Let's move to 'Error' for consistency.  Happy to merge this.

amcclain:
That `showAsError` flag has always struck me as a bit confusing, but perhaps it will help to clarify if setting that flag to false has an obvious effect on the toast intent and title.

I would suggest that when false we default to no title and either the warning (orange) or primary (blue) intent. Leaning towards the latter, but defer to you guys re. what you think the most useful default would be.

TomTirapani:
All makes sense to me - just pushed changes.

I noticed we already have the title / message logic in `parseOptions`, so I changed it to use the parsed options rather than the exception (keeping the dialog and toasts consistent): https://github.com/xh/hoist-react/blob/develop/core/ExceptionHandler.js#L190-L191

Also picks intent (danger / primary) based on `showAsError`

core/XH.js
```diff
@@ -458,6 +458,8 @@ class XHClass extends HoistBase {
      * @param {?number} [config.timeout] - time in ms (default 3000) to show before auto-dismissing
      *      the toast, or null to keep toast visible until manually dismissed.
      * @param {string} [config.intent] - one of [primary|success|warning|danger].
+     * @param {Object} [config.actionButtonProps] - if provided, will render a button within the
```
lbwexler, line 461: This is a huge new feature that should be advertised in the changelog

core/ExceptionHandler.js
```diff
@@ -14,6 +17,35 @@ import {XH} from './XH';
...
+     */
+    static REDACT_PATHS = [
+        'fetchOptions.headers.Authorization'
+    ];
+
+    /**
+     * Text used to replace values that match REDACT_PATHS
+     * @type {string}
+     */
+    static REDACT_TEXT = '[redacted]';
```
lbwexler, line 32: thinking this can just be fixed as *****.  No need to allow customization -- we don't elsewhere in the API

mobile/appcontainer/Toast.js
```diff
@@ -22,7 +22,7 @@ export const toast = hoistCmp.factory({
     model: uses(ToastModel),
 
     render({model}) {
-        const {icon, message, intent} = model,
+        const {icon, message, intent, actionButtonProps} = model,
```
lbwexler, line 25: Think this should just be `actionButton` -- can be a button element, or props for one.

TomTirapani: `actionButtonProps` is the same API we use for banners and error panels - I think better to be consistent

PR #2866, opened 2022-01-14
Support for in-app docked Admin Console

petradish:
Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/AppContainerModel.js
+++ b/appcontainer/AppContainerModel.js
@@ -4,7 +4,11 @@
  *
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
+import {DockContainerModel} from '@xh/hoist/cmp/dock';
+import {filler, iframe} from '@xh/hoist/cmp/layout';
 import {HoistModel, managed, RootRefreshContextModel, TaskObserver, XH} from '@xh/hoist/core';
+import {button} from '@xh/hoist/desktop/cmp/button';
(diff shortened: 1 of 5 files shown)


amcclain:
Wild! It certainly is a cool idea, and makes for a pretty impressive demo.

I can think of a few prefs / capabilities we might wish to add if we want to push forward here - 2-3 sizing modes, swap to a standard browser tab, etc. Would be interesting to discuss how we think it might be useful IRL. Could be a fun topic for our Friday standup...

petradish:
Yeah! Would love to hear others' input on this. One thing I would like to figure out is if we can reload the main app to read in any admin changes without hard refreshing the tab (i.e. refresh everything except the admin dock?)

lbwexler:
Its definitely a lot to think about -- am curious about the use-case you are tying to optimize -- monitoring tracking? Changing config/preferences and seeing behavior on the fly?  

Seems like an interesting option would also be to expose certain admin tasks (config and prefs?  monitoring? ) in a more minimal popup/modal view in the main app.  Imagine you could even flip between key views in that tool just using a select affodance, rather than the full tabsets.

Think my my concerns are:
 1) anything that will complexify our runtime behavior for initialization.  (e.g. the smart refresh you describe)
 2) It seems like our admin UI would need some additional adjustments -- right now it looks too much like the actual app with the two level tabs, and all of the other app chrome of its own.  Its like looking into a mirror facing another mirror....
 
 Looking forward to discuss more...

PR #2868, opened 2022-01-17
Fix to incorrect viewport orientation reporting

amcclain:
+ Toolbox mobile usage (to hide app bar) has consistently/incorrectly been flipped on my device (after the initial, correct, setting on load).
+ Local experimentation indicated laggy behavior on both window.size and DOM orientation API reporting, accounted here with dual debounces.

Hoist P/R Checklist
-------------------


amcclain:
We should ensure that others can replicate the same behavior I was seeing - on updated iPhone with latest Chrome/Safari - and that this fix does not negatively impact client mobile apps in secure browsers that rely on orientation detection. (I am a bit mystified why we haven't gotten error reports on this...)

TomTirapani:
I'm also able to reproduce this on my device (iOS 15.2.1, latest) in Chrome, but not in Safari. I am also seeing it in the BB secure browser. Will review the PR shortly, just wanted to corroborate the issue.

amcclain:
Thanks for the additional testing. You are correct - on iOS Safari I see the current dev site working very smoothly and correctly.  The exception though is if you create a homescreen shortcut, where the app loads in fullscreen mode on the phone. That's broken for me and is where I first noticed the issue.

My hunch is that the fullscreen mode + Chrome are using the same underlying version/flavor of the webkit browser component as required by Apple for apps, whereas the proper Safari browser has optimizations not otherwise available...

appcontainer/ViewportSizeModel.js
```diff
@@ -42,27 +34,41 @@ export class ViewportSizeModel extends HoistModel {
...
+            // lag observed on mobile device responses to DOM orientation APIs.
+            debounce: 100
+        });
     }
 
     //---------------------
     // Implementation
     //---------------------
-    @action
+    @debounced(100)  // Debounced to account for slow inner window resize on mobile.
```
TomTirapani, line 51: FWIW, I've found you can use both @debounced and @action annotations on the same method, as long as you do so in that order (if you wanted to avoid the runInAction below)

appcontainer/ViewportSizeModel.js
```diff
@@ -42,27 +34,41 @@ export class ViewportSizeModel extends HoistModel {
...
         window.addEventListener('resize', () => this.setViewportSize());
         this.setViewportSize();
+
+        this.addReaction({
+            track: () => this.size,
+            run: () => this.updateOrientation(),
+            fireImmediately: true,
+            // Further debounce (on top of debounced setViewportSize) to allow for even greater
+            // lag observed on mobile device responses to DOM orientation APIs.
+            debounce: 100
```
TomTirapani, line 44: Curious about this second level of debouncing here - is this just to allow for the window.screen / orientation APIs to update? Wondering if we should drop them and only calc isPortrait from aspect ratio, if using them is actively degrading the experience like this?

TomTirapani (commented):
This looks good to me - just a shame about the visible debouncing. It's not the end of the world, but just makes feel less snappy. Also wondering where these 100ms values came from, and if there's any chance we could get away with reducing it (i.e. maybe we just need 1 frame)

lbwexler:
@TomTirapani  -- @atm and I discussed this one.  Please feel free to try your alternative to remove the second debounce, or whatever else you think makes the most sense?  Note that Anselm did mention that he tried waiting the single tick and that was not enough....

amcclain:
So I spent some more time with this at lunch out of curiosity - very interesting. One thing I learned is that iOS Chrome supports chrome://inspect to view console logs on device - much better than alerts!

First, I confirmed that iOS Chrome falls back to `window.orientation` in our `updateOrientation` method.

I then modified ViewportSizeModel to log window (inner) sizing and `window.orientation` at a few distinct points:

1. In raw `window.onResize` handler (triggered by rotating my phone from portrait > landscape)
2. +100ms (i.e. debounced `setViewportSize()`
3. +200ms (i.e. debounced `updateOrientation` - which you will see was not debounced _enough_)
4. +300ms
5. +400ms

IMG_61253405B27B-1

As per log output:
1. Size wrong - height is updated (334) but width has yet to update (390)
2. Size correct - but orientation still reports 0 (portrait)
3. Orientation still wrong
4. Orientation still wrong
5. Finally orientation reports correctly...

Repeated reliably several times, in this example and across app refreshes. Pretty crazy, eh? 

So based on this, seems like we should:
* Continue to debounce the onResize event
* Use the aspect ratio instead of messing with this weird/laggy API.

That said, didn't we start using the orientation API because the onscreen keyboard can change screen size in some cases? I was not seeing that on my phone in Chrome - the keyboard "slid over" the bottom of the screen, and my resize loggers didn't fire, but want to ensure we think through any gotchas here.....

amcclain:
Also note that in the trace above I flipped my phone four times - red box is first flip from P>L, then L>P, etc

On the second flip the first log prints the correct size right at the top of the cycle, but for the other three you can see the size update itself is laggy.

Finally, I tried again (not pictured here) logging with a 1ms debounce after the window event - to see if _any_ debounce was enough to catch the right size. It was not - still saw outdated size very consistently.

amcclain:
Last note - for the record - this is on my iPhone 13, iOS 15.2.1, Chrome 97.0.4692, testing on Toolbox mobile home screen immediately after a full refresh of the app/tab.

lbwexler:
We were originally using viewport, not screensize, hence the original bug.
Our first fix was just to move to screen size.

At some point during the research of how to get the screen size, I learned
more about the legacy and new "orientation" apis -- that seemed more
appropriate and elegant.

good example of where 'clean code' can be deceiving!  Happy to jettison it
and go back to using the aspect ratio on the screensize.  Need to make sure
we get the screen appropriately for the legacy BB Access browser...

lbwexler:
Final point here -- while it might be tempting to go back to Tom's original fix, it actually did not work for some reason on my or saba's phone!  That is to say, I did not do the orientation refactoring for fun!   So we need to be careful -- there was *some* issue -- maybe was the same lag anselm saw?  Really can't remember, but let's be sure to test carefully in bb- access.

Issue #2216, opened 2020-12-01
Support "force reload" option from Admin > General > Websockets


amcclain:
For websocket connected users / browsers, we should be able to leverage the socket connection to prompt a connected app to reload itself in the browser. This can be useful for rare-but-real cases where we want to ensure a user refreshes, e.g. to pick up a new set of configs or preferences or clear some other client-side error we have reason to believe exists.

lbwexler:
This came up at a client, where we changed some client configs and really wanted to pick them up.   
+1 for this important feature.  

HOWEVER -- Please keep in mind that using sockets for this core feature will make it harder for us to go multi-instance.

lbwexler:
I was thinking that we might be better off using the '/version` endpoint timer based pull  we have in place for a broader set of applications.  It already is in the business of getting the client to refresh when needed due to a version upgrade.  We could rename it to something like 'environmentCheck' or `clientPing`, but we could also leave the name as is, as it is a form of version checking.

Cons: It might be slightly slower to force the refresh, and would not allow us to easily target individual users.
Pros:  It could be broader, and would work multiple-instance as well.

A simple implementation right now would be to add a date (in memory) on the server of "refreshRequired", which had the latest date, if any, that a refresh was required.

The timer in environmentService could consult this data, and if any client was loaded after the date of `refreshRequired`, it would need to be reloaded.

lbwexler:
Adding this ticket for Dan -- seemed like it might be an interesting one for you Dan.

amcclain:
Discussed further with @lbwexler, and we determined that it would still be worth it to implement this feature as originally spec'd. There can be some cases where targeting a particular user is helpful, and the current websocket services are already setup nicely to handle this.

We also discussed a potential force-reload enhancement to the version checking routine that would _not_ rely on websockets and could cover many of the cases where we want to ensure clients refresh ASAP after a release. Will ticket that in a separate issue.

PR #2871, opened 2022-01-25
Support per-user app reload via web sockets

petradish:
Resolves [websocket implemention of force reload](https://github.com/xh/hoist-react/issues/2216)

- Admin console button to force reload on app instance (currently single select only)
- Suspends client activity and alerts user that app must be refreshed in order to continue (allows user to potentially see Version banner in backdrop?)

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/appcontainer/SuspendPanel.js
+++ b/desktop/appcontainer/SuspendPanel.js
@@ -27,18 +27,21 @@ export const suspendPanel = hoistCmp.factory({
             flexDirection: 'column',
             className: 'xh-suspend-viewport',
             item: panel({
-                title: `Restart ${XH.clientAppName}`,
+                title: `Reload Required`,
                 icon: Icon.refresh(),
                 className: 'xh-suspend-panel',
-                item: div(
-                    p({item: message, omit: !message}),
-                    p('Your application requires a restart.'),
-                    p('Please click restart to continue.')
-                ),
+                item: div({
+                    className: 'xh-suspend-panel__inner',
+                    items: [
+                        p({item: message, omit: !message}),
+                        p(`${XH.clientAppName} must be reloaded to continue.`)
+                    ]
+                }),
                 bbar: [
                     filler(),
                     button({
-                        text: 'Restart now',
+                        text: 'Reload now',
+                        icon: Icon.refresh(),
                         intent: 'primary',
                         minimal: false,
                         autoFocus: true,
--- a/desktop/appcontainer/SuspendPanel.js
+++ b/desktop/appcontainer/SuspendPanel.js
@@ -20,7 +20,7 @@ export const suspendPanel = hoistCmp.factory({
     displayName: 'SuspendPanel',
 
     render() {
-        const message = XH.suspendData?.message ?? 'Your application requires a restart.';
+        const message = XH.suspendData?.message
         return viewport({
             alignItems: 'center',
             justifyContent: 'center',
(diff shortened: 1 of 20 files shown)


svc/WebSocketService.js
```diff
@@ -269,6 +273,24 @@ export class WebSocketService extends HoistService {
         });
     }
 
+    async forceAppReloadAsync() {
+        // Suspend activity without setting AppState.SUSPENDED (i.e. don't show Idle panel backdrop)
```
lbwexler, line 277: Like the idea of putting the user in to the suspended state, but I think for state consistency, would like this to call `XH.idleService.suspendApp().` and have the app *actually* be in a suspended state.   Can we modify that state if needed to display alternate message text, if the suspension was forced?  We could add a method, or flags to suspendApp()?

petradish: I couldn't find a clear path to switch on SUSPENDED state flavors, so I created a new FORCE_RESTART state and mirrored the implementation of SUSPENDED --> idlePanelHost so that Apps could customize the forceRestartPanel to their liking.

lbwexler: discussed with Petra and am looking at a few more changes now to push this over the finish line.

lbwexler:
@petradish -- let me know what you think of the changes?  Hope this will be more extensible, has a more minimal API surface, and sets us up nicely for the additional changes to optionally suspend the app during version checking.  

Good idea you had putting the extra state on XH -- that was the key.

lbwexler (approved):

desktop/appcontainer/SuspendPanel.js
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Generic Panel to display when the app is suspended.
+ * @private
+ */
+export const suspendPanel = hoistCmp.factory({
+    displayName: 'SuspendPanel',
+
+    render() {
+        const message = XH.suspendData?.message ?? 'Your application requires a restart.';
```
petradish, line 23: There is no API for setting the suspendData.message property - should we make the WebsocketsPanel in the admin console more like the test message (where the message is defaulted in the prompt)?

petradish (commented):
I like it! I pushed a cleanup commit to remove the old "force restart" references.
Also, what do you think about supporting the custom suspend message?

desktop/appcontainer/SuspendPanel.js
lbwexler: So I had that message teed up for when we suspend based on a min version.  I was thinking it was overkill to tie it to the socket based kill for now.....  Realize I probably could have left the message out until we needed it, but also wanted to make it clear how I saw the need for a `suspendData` rather than just a `suspendReason`

lbwexler: Great catch on the extra state

svc/WebSocketService.js
```diff
@@ -206,9 +206,10 @@ export class WebSocketService extends HoistService {
                 case this.REG_SUCCESS_TOPIC:
                     this.installChannelKey(data.channelKey);
                     break;
-                case this.TEST_MSG_TOPIC:
-                    this.showTestMessageAlert(data);
-                    break;
+                case this.FORCE_APP_SUSPEND_TOPIC:
+                    console.log(data);
```
petradish, line 210: stray log

amcclain (approved):
Great - thanks for getting this long-desired admin feature in there.

PR #2876, opened 2022-01-29
Support for readonly form fields without children.

lbwexler:
Hope this moves the ball forward for Forms, and gets us beyond the silliness, providing a good readonly story, with minimal disruption to existing apps.

Agree with Tom -- a general binding label/value mechanism could be useful as a future iteration.
  
Also agree with Anselm that a readOnlyField() might be useful for the developer.  If we felt that we wanted that, could just be 
syntactic sugar on what we have.  Would like to avoid adding extra components in this space, if possible, as the relationship between the input/FormField/FormModel/FieldModel/Form is already incredibly intricate and somewhat duplicated between mobile and desktop. Again, could do in a future iteration. 

Also fixed an annoying issue preventing us from wrapping HoistInputs.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/form/FormField.js
+++ b/desktop/cmp/form/FormField.js
@@ -291,11 +291,11 @@ const editableChild = hoistCmp.factory({
             }
         }
 
-        if (displayNotValid && propTypes.leftIcon && leftErrorIcon) {
+        if (displayNotValid && propTypes?.leftIcon && leftErrorIcon) {
             overrides.leftIcon = Icon.warningCircle();
         }
 
-        if (propTypes.commitOnChange && !isUndefined(commitOnChange)) {
+        if (propTypes?.commitOnChange && !isUndefined(commitOnChange)) {
             overrides.commitOnChange = commitOnChange;
         }
         return cloneElement(child, overrides);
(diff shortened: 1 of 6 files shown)


desktop/cmp/form/FormField.js
```diff
@@ -306,9 +312,19 @@ const editableChild = hoistCmp.factory({
 const blockChildren = ['CodeInput', 'JsonInput', 'Select', 'TextInput'];
 
 function getValidChild(children) {
+    const count = Children.count(children);
+    if (count === 0) return null;
+    if (count > 1) {
+        throw XH.exception('Add a single HoistInput child to FormField, or no children and a readonlyRenderer.');
```
cnrudd, line 318: Not major, but is this error message leaving out some other options?   The single child is no longer required to be a HoistInput, and if no child is present, is a readOnlyRenderer really required?   A readOnlyRenderer is not a requirement right? because if not present the field will fall back to: defaultReadonlyRenderer right?

Maybe: 'Add a single input child to FormField, or if no children you may specify a custom readonlyRenderer'.

cnrudd (commented):
I think this a great change.  Just one suggestion/comment.

desktop/cmp/form/FormField.js
lbwexler: great point!

amcclain (approved):
Will make for some nice app cleanups removing those meaningless RO inputs, thanks!

PR #2879, opened 2022-02-03
Alternative id handling

lbwexler:
Restoring our ability to have integer ids in store by adding StoreRecord.agId

+ Fixes to AgGridModel code and docs to reflect new string ids.

Please Merge also toolbox branch with the same name

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -762,11 +762,10 @@ class GridLocalModel extends HoistModel {
     }
 
     processCellForClipboard = ({value, node, column}) => {
-        const {model} = this,
-            recId = node.data.id,
-            colId = column.colId,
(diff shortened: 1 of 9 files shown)


cmp/grid/Grid.js
```diff
@@ -763,7 +763,7 @@ class GridLocalModel extends HoistModel {
 
     processCellForClipboard = ({value, node, column}) => {
         const {model} = this,
-            recId = node.id,
+            recId = node.data.id,
```
haynesjm42, line 766: `node.data` could potentially be null/undefined here for group rows, I think the code was guarding against this by getting the id and looking up the record below. We should early-out if the node has no data, and just use the `node.data` as the record below to avoid the store lookup

Issue #2884, opened 2022-02-05
FormField renders anonymous inner div w/o className if `!childIsSizeable`


amcclain:
This div:

https://github.com/xh/hoist-react/blob/955cc42f57fcf0665ccd4b51c44448b9d479d661/desktop/cmp/form/FormField.js#L153

commonly appears in generated formField DOM. I am always thrown off by it, and find myself having to target it via `.xh-form-field > div` to e.g. apply a flex to formField internals.

Would like to review the approach here and see if we could either skip this layer when not rendering a sizable child or at least add a dedicated classname under that condition to make it easier to get a handle on / help explain why it exists.

(Can review in CryptIQ detail forms - that is where I ran into this)

PR #2885, opened 2022-02-07
Give FormField's anonymous inner div a className.

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/2884

It turns out that anonymous div is serving the necessary function of breaking out of flexbox for unsizeable children, so I couldn't remove it entirely. Instead, I gave it an explicit class name that hopefully helps clarify the intentions there.

Hoist P/R Checklist
-------------------


amcclain (changes requested):
Thanks for this cleanup - definitely helpful to have this consistent `-inner` class with the two variations and their more expressive names.

I see there is still a reference to `.xh-form-field-fill` in the stylesheet - looks like we had an override related to toolbars. Can you please check that out and (presumably) update to use the new classname?

Can you also please add a changelog entry?  While I don't think this rises to the level of an official 💥 breaking change, we can add a ⚠️ callout as we sometimes do noting that `.xh-form-field-fill` has gone away and that apps should check for any styles that selected on that class and replace as needed with `.xh-form-field-inner--flex`.

With those two tweaks we should be all set to go, thanks again.

TomTirapani:
Thanks for the detailed look Anselm - I've just pushed an update with the requested changes

amcclain (approved):
Thanks much

PR #2905, opened 2022-03-09
Fix autosizing headers when `headerName` is a react element

haynesjm42:
This seems to work well, tested in client app where I ran into this limitation.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnWidthCalculator.js
+++ b/cmp/grid/impl/ColumnWidthCalculator.js
@@ -223,7 +223,7 @@ export class ColumnWidthCalculator {
         if (isNil(headerClass)) return '';
 
         if (isFunction(headerClass)) {
-            headerClass = headerClass(column, gridModel);
+            headerClass = headerClass({column, gridModel});
         }
 
         const ret = [];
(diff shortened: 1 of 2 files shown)


cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -150,10 +153,23 @@ export class ColumnWidthCalculator {
...
     // Autosize header cell
     //------------------
+    getHeaderHtml(gridModel, column) {
+        const {headerName} = column,
+            headerValue = isFunction(headerName) ? headerName({column, gridModel}) : headerName;
+
+        if (isString(headerValue)) {
+            return headerValue;
+        } else if (isValidElement(headerValue)) {
+            return renderToString(headerValue);
```
TomTirapani, line 163: Could we use `renderToStaticMarkup` instead? Probably being paranoid, but it seems that would make it less likely for React to confuse the autosize header cell as part of its world

TomTirapani (commented):
Kind of a separate matter, but we should probably be processing and passing `Column.headerClass` in the same way we include `cellClass` and `rowClass`

cmp/grid/impl/ColumnWidthCalculator.js
haynesjm42: Nice find! Definitely seems like we can and should use that instead so react isn't adding anything of its attributes to the elements

amcclain:
Thank you for this! I had a nagging sense that something was up with autosize and rich headers, but wasn't about it.  Will try the latest snapshot in CryptIQ to see how it performs.

Issue #2911, opened 2022-03-16
Filterable Select inputs commit `null` when clearing filter text


TomTirapani:
Select inputs with `enableFilter` (both desktop and mobile versions) commit a value of `null` when the entered filter text is cleared. The expected behaviour is for the input to only commit a value when an option is selected, or the input is cleared using the `enableClear` affordance.

To reproduce:
1) Visit https://toolbox-dev.xh.io/app/forms/inputs
2) In "Select 1", pick any option.
3) Re-focus the select, and click within the text to clear the selection. Use the keyboard and backspace to clear the entered text.
4) As soon as the entered text has been cleared entirely, the committed value becomes `null`.

amcclain:
@TomTirapani if you are able to look at this, we can do a minor/patch release early next week or anytime.  I would give this higher priority than the webpack work.

TomTirapani:
PR opened here: https://github.com/xh/hoist-react/pull/2912

The default behaviour of react-select is to display the current value as a placeholder when the input text is empty. We added a line to commit `null` in this commit https://github.com/xh/hoist-react/commit/5f4b911ab26505241b71dca472385e0277b83d05 to override this default behaviour and show an empty input.

However, we can achieve the same result by toggling `controlShouldRenderValue` off while the input has focus, without the side-effect of changing the external value.

PR #2912, opened 2022-03-18
Fix Selects to not commit null when clearing filter text

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/2911

Hoist P/R Checklist
-------------------

### DIFF
--- a/mobile/cmp/input/Select.js
+++ b/mobile/cmp/input/Select.js
@@ -288,7 +288,6 @@ class Model extends HoistInputModel {
             if (action === 'input-change') {
                 this.inputValue = value;
                 this.inputValueChangedSinceSelect = true;
-                if (!value) this.noteValueChange(null);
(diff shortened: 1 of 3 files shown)


lbwexler:
Just saw this Tom -- thanks for jumping on it --- would definitely like to discuss

mobile/cmp/input/Select.js
```diff
@@ -288,7 +288,6 @@ class Model extends HoistInputModel {
             if (action === 'input-change') {
                 this.inputValue = value;
                 this.inputValueChangedSinceSelect = true;
-                if (!value) this.noteValueChange(null);
```
lbwexler, line 291: do we have any sense of what we were trying to accomplish with noteValueChange(null)?? IIRC we were trying to restore null when user when back to empty string.  Does not seem to be an issue after this change, however.

PR #2914, opened 2022-03-22
Add support for "sync with system" as a theme option

### LINKED ISSUE #2703: Support "sync with system" as a theme option
Several sites (Github here included) provide users with an option to use their OS/device level "dark mode" setting to determine the theme. It would be great for us to provide the same as a third option, including making it the default.

Note that the system theme can be detected in JS via `window.matchMedia('(prefers-color-scheme: dark)').matches`.

A listener can be registered like so to ensure an already loaded site switches when the OS changes: 
```
window.matchMedia('(prefers-color-scheme: dark)').addListener((e) => console.log(`dark mode: ${e.matches}`))
```

ghsolomon:
Fixes #2703 

Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/ThemeModel.js
+++ b/appcontainer/ThemeModel.js
@@ -13,10 +13,8 @@ import {action, observable, makeObservable} from '@xh/hoist/mobx';
  *  @private
  */
 export class ThemeModel extends HoistModel {
-
-    @observable darkTheme = false;
-    @observable themePref = 'light';
-    systemDarkMode = window.matchMedia('(prefers-color-scheme: dark)');
+    /** @member {boolean} */
+    @observable darkTheme;
 
     constructor() {
         super();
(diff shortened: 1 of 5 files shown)


ghsolomon:
See also https://github.com/xh/hoist-core/pull/236

amcclain (approved):
Looks great - @lbwexler leaving open for you to have a look / catch up with this change, but all good to go from my perspective.

appcontainer/ThemeModel.js
```diff
@@ -32,10 +32,25 @@ export class ThemeModel extends HoistModel {
...
         classList.toggle('bp3-dark', value);
         this.darkTheme = value;
-        XH.setPref('xhTheme', value ? 'dark' : 'light');
+
+    }
+
+    @action
+    setThemePref(value) {
+        if (value === 'system') {
+            this.setDarkTheme(window.matchMedia('(prefers-color-scheme: dark)').matches);
```
amcclain, line 41: FYI the `matchMedia` API seems well-established / supported as per https://caniuse.com/matchmedia, so I don't _think_ we need to defend against mobile or other wonky browsers throwing here.

appcontainer/ThemeModel.js
```diff
@@ -32,10 +32,25 @@ export class ThemeModel extends HoistModel {
         classList.toggle('xh-dark', value);
         classList.toggle('bp3-dark', value);
         this.darkTheme = value;
-        XH.setPref('xhTheme', value ? 'dark' : 'light');
+
+    }
+
+    @action
+    setThemePref(value) {
+        if (value === 'system') {
```
lbwexler, line 40: think a switch with three clauses would actually be clearer at this point.

lbwexler, line 40: think a switch with three clauses would actually be clearer at this point.

core/XH.js
```diff
@@ -324,9 +324,20 @@ class XHClass extends HoistBase {
         return this.acm.themeModel.toggleTheme();
     }
 
-    /** Enable/disable the dark theme directly (useful for custom app option controls). */
+    /**
+     * Enable/disable the dark theme directly (useful for custom app option controls).
+     * @param {boolean} value
+     */
     setDarkTheme(value) {
```
lbwexler, line 331: Think we should deprecate or remove `setDarkTheme()` and make "setTheme()" (without the suffix) the main entry point.  It's too confusing, and the 'pref' bit is an implementation detail.

lbwexler, line 331: Think we should deprecate or remove `setDarkTheme()` and make "setTheme()" (without the suffix) the main entry point.  It's too confusing, and the 'pref' bit is an implementation detail.

appcontainer/ThemeModel.js
lbwexler: also should throw if presented with a unknown value

lbwexler: also should throw if presented with a unknown value

Issue #2925, opened 2022-03-28
Column Filter throws error when run on a new record where `committedData` === undefined


cnrudd:
In a client's app, there is an editable grid that uses our ColumnFilters.
If the grid is filtered, adding a new record immediately throws because the new record is added with property `committedData === undefined`.

the `getVal` function built here:
https://github.com/xh/hoist-react/blob/develop/data/filter/FieldFilter.js#L90
assumes `committedData` will always be defined.

There was a deliberate change to get values from committedData, rather than from record.data https://github.com/xh/hoist-react/commit/9933a23c9e8bd761388ace6fdf69bf05281dcf59
Getting the value from record.data would work in this use case of a newly added record in an editable grid.

Should the fix be to:
1. fall back to record.data if committedData is not defined?
2. return null if committedData is not defined?

The [PR](https://github.com/xh/hoist-react/pull/2742) for [9933a](https://github.com/xh/hoist-react/commit/9933a23c9e8bd761388ace6fdf69bf05281dcf59) does not state the motivation for making that change.

cnrudd:
I spoke with Tom about the reasons for https://github.com/xh/hoist-react/commit/9933a23c9e8bd761388ace6fdf69bf05281dcf59.  It was this use case:

> User filters grid, user edits visible row, uncommitted changes to row should not  be caught by the filter and cause the row to be filtered out.

The use case my client app has to support is:

> User filters grid, user adds uncommitted clone of an existing row.   The new clone row should be visible.  The new clone row record will have these properties:  committedData === undefined, data={cloned values...}

cnrudd:
I've put in a [PR](https://github.com/xh/hoist-react/pull/2927) with a change that supports both use cases mentioned above.

PR #2927, opened 2022-03-29
FieldFilter ignores record if no committedData

cnrudd:
Resolves https://github.com/xh/hoist-react/issues/2925

Hoist P/R Checklist
-------------------

- [x] Created Toolbox branch: Can be tested on these 2 toolbox branches:
https://github.com/xh/toolbox/tree/filterCommitted  (Filter Chooser)
https://github.com/xh/toolbox/tree/columnFiltersOnInlineEditingPanel  (Column Filters)

### DIFF
--- a/data/filter/FieldFilter.js
+++ b/data/filter/FieldFilter.js
@@ -80,7 +80,7 @@ export class FieldFilter extends Filter {
 
         if (store) {
             const storeField = store.getField(field);
-            if (!storeField) return () => true; // Ignore if field not in store
+            if (!storeField) return () => true; // Ignore (do not filter out) if field not in store
 
             const fieldType = storeField.type;
             value = isArray(value) ?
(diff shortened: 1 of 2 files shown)


data/filter/FieldFilter.js
```diff
@@ -96,48 +96,66 @@ export class FieldFilter extends Filter {
         switch (op) {
             case '=':
                 return r => {
+                    if (isNil(r.committedData)) return true; // Ignore (do not filter out) record if not committed
```
TomTirapani, line 99: I think we need to change these conditions to `if (store && isNil(r.committedData))` - filters are supposed to be able to work against anonymous objects as well as StoreRecords (see [getTestFn docs](https://github.com/xh/hoist-react/blob/develop/data/filter/Filter.js#L28-L30)). As part of that change, we may as well put the condition into a named function e.g. `isUncommittedRecord(r)`.

data/filter/FieldFilter.js
```diff
@@ -80,7 +80,7 @@ export class FieldFilter extends Filter {
 
         if (store) {
             const storeField = store.getField(field);
-            if (!storeField) return () => true; // Ignore if field not in store
+            if (!storeField) return () => true; // Ignore (do not filter out) if field not in store
```
TomTirapani, line 83: Great to clarify what 'ignore' means in this context 👍

TomTirapani (changes requested):

data/filter/FieldFilter.js
cnrudd: re `store`: yes, of course, thanks for pointing that out.  I have made the fix and moved it all into a new local function called `mustIgnore`.

lbwexler:
Looks great guys, thanks!

Issue #2406, opened 2021-04-20
Improve TabContainer LoadSupport


TomTirapani:
TabContainers should be more intuitive / easier to integrate with load support

My suggestion would be to add `TabContainerModel.loadAsync()` / `TabContainerModel.refreshAsync()`. These methods respect the refresh mode of the tabs and should be equivalent to refreshing the outer RefreshContext. Typically, this would mean loading the visible tab, and having the other tabs refresh lazily.

Curious to hear other suggestions for how we can improve this crucial part of the API.

PR #2930, opened 2022-03-31
Add methods to load / refresh a TabContainer directly

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/2406

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/tab/TabContainerModel.js
+++ b/cmp/tab/TabContainerModel.js
@@ -4,7 +4,7 @@
  *
  * Copyright © 2021 Extremely Heavy Industries Inc.
  */
-import {HoistModel, managed, PersistenceProvider, RefreshMode, RenderMode, XH} from '@xh/hoist/core';
+import {HoistModel, managed, RefreshContextModel, PersistenceProvider, RefreshMode, RenderMode, XH} from '@xh/hoist/core';
 import {action, observable, makeObservable} from '@xh/hoist/mobx';
 import {ensureUniqueBy, throwIf} from '@xh/hoist/utils/js';
 import {find, isString, isUndefined, without, difference} from 'lodash';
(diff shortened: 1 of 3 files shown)


cmp/tab/TabContainerModel.js
```diff
@@ -269,6 +274,22 @@ export class TabContainerModel extends HoistModel {
         if (nextTab) this.activateTab(nextTab.id);
     }
 
+    //-----------------------------
+    // Trigger load / refreshes of TabContainer directly.
+    // Note that we are intentionally *not* implementing LoadSupport here.
+    //-----------------------------
+    async loadAsync(loadSpec) {
```
lbwexler, line 281: Don't fully understand this change, and would like to discuss, but it does seem to me that you should only need to override loadAsync and everything else should follow.

lbwexler:
@TomTirapani -- Looks appealingly simple!  Looking forward to discussing.  Am a little concerned about unexpected effect of adding so many new RefreshContext's in, but it makes a lot of sense.

amcclain:
Tried (and failed) to get the bundled yarn to stop showing up here as a diff, but I don't think anything actually changed.  (Some line ending thing, maybe?)

TomTirapani:
I fixed it with `git checkout 2e9b578a4db1a4e9a7c1a3f75cd6a0101234cb91 -- .yarn/releases/yarn-1.22.19.cjs`

lbwexler:
I still have a lingering question about whether we should have TabContainerModel delegate its doLoadAsync to this model.  Seems like as an API user I would expect to just call refresh methods directly on the TabContainerModel, but am concerned we are somehow going to get double loading.

lbwexler:
tom reminded me that we discussed this at length in the spring, and we actually decided *not* to do that delegation, after originally trying it.  

We agreed to merge this as is so we can get the benefits of the new refreshContextModel, and shake out any issues.  Can put a comment in the code, about potentially taking it further with a delegation.

Issue #2932, opened 2022-04-01
Upgrade to React v18


lbwexler:

TomTirapani:
Opened a draft PR for this, since there are console warnings from Blueprint about it using `ReactDOM.render` for context menus and toasts.

lbwexler:
we may want to link this to blueprint 4.x

lbwexler:
We decided to wait for v50 for this

TomTirapani:
Do we still want to wait for this, given that Blueprint has delayed fixing the warnings until v5:

https://github.com/palantir/blueprint/issues/5212

FWIW, the warnings go away when using the prod build of React.

lbwexler:
I would be up for merging this now.   Not a lot of other major change right now in Hoist React, and might be a good opportunity to get this done.

lbwexler:
@amcclain -- thoughts here?

PR #2934, opened 2022-04-05
Upgrade React to v18

TomTirapani:
Keeping this PR as a draft, since there are console warnings from Blueprint about it using `ReactDOM.render` for context menus and toasts. The app still works and these seem to be dev-mode only warnings, so we could release like this, but I think its worth waiting for our dependencies to be v18 compliant.

See Blueprint issue re: ReactDOM.render: https://github.com/palantir/blueprint/issues/5212
See issue: https://github.com/xh/hoist-react/issues/2932
See Toolbox PR: https://github.com/xh/toolbox/pull/571

Hoist P/R Checklist
-------------------


amcclain:
Great - excellent that we're back on the latest!

We should call this out in the changelog though - apps will need to update their package.json files, and devs will likely want at least a short comment helping them understand what might need to change (or reassure them that nothing does).

Also if there will be new warnings in default react dev mode, another pointer to `reactProdMode` flag would likely be helpful and might avoid support requests.

PR #2963, opened 2022-04-19
Add new GridModel hideGroupedColumns and showUngroupedColumns config

### LINKED ISSUE #1369: GridModel should have a 'hideGroupByCols' option.
EXT has this, but ag-grid requires you to hide the columns manually after grouping. We can set hidden on our columns in the grid model, but would be nice is the groupBy method could do this for you automatically if you so choose.

petradish:
Resolves #1369 
+ This was a lot trickier than I already expected. Needs more work/consideration.
+ Setting the ag-grid flag `suppressMakeColumnVisibleAfterUnGroup`= `true` seems like the best way to allow us to control column visibility.
+ There is a degenerate case where the combination of hideGroupedColumns = true && showUngroupedColumns = false can permanently hide a column with excludeFromChooser = true && hidden = false. I added a warning for this case -- not sure what else to do there.

Hoist P/R Checklist
-------------------


lbwexler:
Closing as part of a stale-PR cleanup. This was flagged WIP in 2022, predates the TS migration, and is built on the now-deprecated `suppressMakeColumnVisibleAfterUnGroup` flag. ag-Grid (v31+, current v35) consolidated this into `suppressGroupChangesColumnVisibility`, which covers most of the intended behavior natively — so a fresh implementation against today's `GridModel` is the right path rather than reviving this branch.

The feature request remains open and tracked in #1369, where I've captured the ag-Grid native support and a pointer back to this PR. Thanks @petradish.

PR #2972, opened 2022-04-23
Added type field to be passed to server during export

jskupsik:
Hoist P/R Checklist
-------------------


lbwexler:
So I think we are moving in the right direction here, with defaulting the type we send to the type from the store field in the underlying column.  My concern is that the 'type' we are using here is some weird mix of our Store `FieldType` with some additional ad-hoc strings of 'date',  'datetime', 'longtext'.  

Where It seems like 

- 'date'  is actually `FieldType.localDate`
- 'datetime' is something like `FieldType.Date`
- 'longtext' is never used.

Could we standardize this such that `type` was always guaranteed to be a 'FieldType'?

petradish:
> So I think we are moving in the right direction here, with defaulting the type we send to the type from the store field in the underlying column. My concern is that the 'type' we are using here is some weird mix of our [Store `FieldType`](https://github.com/xh/hoist-react/blob/109effdfd7d79f7b48e9641ca901741d9bd10083/data/Field.js#L114) with some additional [ad-hoc strings of 'date', 'datetime', 'longtext'](https://github.com/xh/hoist-react/blob/dc8db9098496ba87d8e84392e9b094f2d05be5dd/svc/GridExportService.js#L253).
> 
> Where It seems like
> 
> * 'date'  is actually `FieldType.localDate`
> * 'datetime' is something like `FieldType.Date`
> * 'longtext' is never used.
> 
> Could we standardize this such that `type` was always guaranteed to be a 'FieldType'?

^ Agree with this - would be great to standardize to `FieldType`

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### 🐞 Bug Fixes
 
+* Fixed issue where Column renderers returning boolean values did not actually render the value.
+* Fixed issue with model lookups in HoistComponent's callRender method.
+* Fixed issue where exporting to excel file would occasionally coerce strings (like "1e10") into
```
petradish, line 9: Changelog entries above seem off!

svc/GridExportService.js
```diff
@@ -186,13 +189,13 @@ export class GridExportService extends HoistService {
 
         // 2) Dates: Provide date data expected by server endpoint.
         // Also, approximate formats for CSV and clipboard.
-        if (excelFormat === ExcelFormat.DATE_FMT) value = fmtDate(value);
-        if (excelFormat === ExcelFormat.DATETIME_FMT) value = fmtDate(value, 'YYYY-MM-DD HH:mm:ss');
+        if (excelFormat === ExcelFormat.DATE_FMT || type === LOCAL_DATE) value = fmtDate(value);
```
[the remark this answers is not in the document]
jskupsik: Yes! This line makes CSV and Clipboard of LOCAL_DATE / DATE be formatted reasonably (in addition to being the standard format that the Date and LocalDate objects over in groovy/java land expect, which the comment above says). If we do not handle it, the string we get from a LOCAL_DATE when we copy to clipboard or export to csv is of the format `"Fri Apr 22 2022 00:00:00 GMT-0700 (Pacific Daylight Time)"`, and we get the same format for DATE. I think this comes from the default javascript toString of the date/localDate objects.

petradish (changes requested):
Hi Jakub, great job - love how we are going to pull in Hoist field types through to the server. See comments here and the hoist-core PR!

CHANGELOG.md
petradish: Yeah I am not sure where lines 47 and 48 above came from?

svc/GridExportService.js
```diff
@@ -181,21 +189,26 @@ export class GridExportService extends HoistService {
...
-        return forExcel && cellHasExcelFormat ?
-            {value, format: excelFormat} :
+        const cellHasVariedType = type !== fieldType;
+
+        // Send format and/or type with the cell in an object only if it varies within the column
+        return (forExcel && (cellHasExcelFormat || cellHasVariedType)) ?
+            omitBy({
+                value,
+                format: cellHasExcelFormat ? excelFormat : null,
+                type: cellHasVariedType ? type : null
```
TomTirapani, line 208: I don't think we will ever have a cell with a different type to the other cells in the column. I think we can simplify considerably by not considering 'type' at the cell level at all.

This code anticipates a cell having a varied type if the column's `exportValue` points to another field - but those cases are essentially pointing the **entire column** to another field. The new field type would apply to **every** cell in the column, and we can send the other field's type with the column's metadata rather than here.

TomTirapani (approved):
Looks great! I pushed some small changes in accordance to my comments, and think this is ready to be merged :)

PR #2982, opened 2022-04-29
Add support for filtering on committed data, current data, or both

haynesjm42:
When working on this it seemed to me that for an inline editable grid, you probably would want to consider both committed and current values when filtering, so I ended up adding a new `FilterValueMode` enumeration which you can set on `Store`, and will be respected by `FilterField` when building the test function. 

Let me know if this seems over-engineered, could be simplified into a boolean probably. 

Also note that this config is only respected in FilterField. Was not sure if we wanted to handle this elsewhere also (such as in StoreFilterField) 

Toolbox branch with control for testing different modes: https://github.com/xh/toolbox/tree/storeFilterValueMode

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.js
+++ b/data/Store.js
@@ -979,7 +979,7 @@ export const FilterValueMode = Object.freeze({
 
     /**
      * Both the committed and current values are used to test the filter and the record passes if
-     * either value passes.
+     * either value passes. This is the recommended mode for inline-editing enabled Grids.
      */
     ANY: 'any'
 });
--- a/data/Store.js
+++ b/data/Store.js
@@ -114,7 +114,7 @@ export class Store extends HoistBase {
      * @param {boolean} [c.filterIncludesChildren] - true if all children of a passing record should
      *      also be considered passing (default false).
      * @param {FilterValueMode} [c.filterValueMode] - determines how record field values are
-     *      considered when processing filters. Default behavior is to only consider 'committed'
+     *      considered when processing `FieldFilter`s. Default behavior is to only consider 'committed'
      *      values.
      * @param {boolean} [c.loadTreeData] - true (default) to load hierarchical/tree data, if any.
      * @param {string} [c.loadTreeDataFrom] - the property on each raw data object that holds its
@@ -966,7 +966,7 @@ function forIn(obj, fn) {
 }
 
 /**
- * Modes for how record field values should be considered when filtering.
+ * Determines which Record field values should be used for matching against a `FieldFilter`.
  * @readonly
  * @enum {string}
  */
--- a/data/Store.js
+++ b/data/Store.js
@@ -46,6 +46,9 @@ export class Store extends HoistBase {
     /** @member {boolean} */
     @observable filterIncludesChildren;
 
+    /** @member {FilterValueMode} */
+    @observable filterValueMode;
+
     /** @member {boolean} */
     loadTreeData;
 
@@ -110,6 +113,9 @@ export class Store extends HoistBase {
      *      array, a single 'AND' filter will be created.
      * @param {boolean} [c.filterIncludesChildren] - true if all children of a passing record should
      *      also be considered passing (default false).
+     * @param {FilterValueMode} [c.filterValueMode] - determines how record field values are
+     *      considered when processing filters. Default behavior is to only consider 'committed'
+     *      values.
      * @param {boolean} [c.loadTreeData] - true (default) to load hierarchical/tree data, if any.
      * @param {string} [c.loadTreeDataFrom] - the property on each raw data object that holds its
      *      (raw) child objects, if any. Default 'children', no effect if `loadTreeData: false`.
@@ -138,6 +144,7 @@ export class Store extends HoistBase {
         processRawData = null,
         filter = null,
         filterIncludesChildren = false,
+        filterValueMode = FilterValueMode.COMMITTED,
         loadTreeData = true,
         loadTreeDataFrom = 'children',
         loadRootAsSummary = false,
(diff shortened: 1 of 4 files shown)


data/Store.js
```diff
@@ -947,6 +965,25 @@ function forIn(obj, fn) {
     }
 }
 
+/**
+ * Modes for how record field values should be considered when filtering.
```
amcclain, line 969: Maybe worth an extra emphasis on FieldFilter here:

```suggestion
 * Determines which Record field values should be used for matching against a `FieldFilter`.
```

data/Store.js
```diff
@@ -947,6 +965,25 @@ function forIn(obj, fn) {
...
+
+    /** The current value is used to test the filter. */
+    CURRENT: 'current',
+
+    /**
+     * Both the committed and current values are used to test the filter and the record passes if
+     * either value passes.
+     */
+    ANY: 'any'
+});
```
amcclain, line 985: Could we use the comments in this enum to offer a little bit of context or advice for when / why you might opt for one mode vs. another?  Mainly to point at inline editing as the case where you commonly want the default `COMMITTED` behavior. 

If there is a clear / typical use case for the `CURRENT` we could call that out as well.

data/Store.js
```diff
@@ -110,6 +113,9 @@ export class Store extends HoistBase {
      *      array, a single 'AND' filter will be created.
      * @param {boolean} [c.filterIncludesChildren] - true if all children of a passing record should
      *      also be considered passing (default false).
+     * @param {FilterValueMode} [c.filterValueMode] - determines how record field values are
+     *      considered when processing filters. Default behavior is to only consider 'committed'
```
amcclain, line 117:
```suggestion
     *      considered when processing `FieldFilter`s. Default behavior is to only consider 'committed'
```

data/filter/FieldFilter.js
```diff
@@ -87,79 +87,72 @@ export class FieldFilter extends Filter {
...
+                opTestFn = v => regExps.some(re => re.test(v));
+                break;
             default:
                 throw XH.exception(`Unknown operator: ${op}`);
         }
+
+        if (!store) return r => opTestFn(r[field]);
+
+        // Pass/fail depends on the filter value mode of the store
+        // Uncommitted records always pass the filter
```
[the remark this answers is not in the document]
amcclain: Assuming we keep the "adds always pass" behavior, should we move that check up and short-circuit the creation of the filter functions entirely?

amcclain (approved):
Great thank you - minor comments if you could please review / tweak, but looks good to me!

data/StoreRecord.js
```diff
@@ -102,6 +102,14 @@ export class StoreRecord {
         return this.data[fieldName];
     }
 
+    /**
+     * @param {string} fieldName
+     * @returns {*} - the committed value of a field.
+     */
+    getCommitted(fieldName) {
+        return get(this.committedData, fieldName);
```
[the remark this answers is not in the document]
lbwexler: agree

data/filter/FieldFilter.js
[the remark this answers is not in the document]
haynesjm42: Yeah I wasn't sure about this, I considered adding an additional boolean to Store to control this behavior but in the end after playing around with it in the inline editing example, I just found it clearly undesirable to ever click an 'Add Row' button and have nothing at all appear to happen in the grid, regardless of the filter value mode selected, and I couldn't think of a use-case where you are dealing with uncommitted records and would want them to be filtered out since I think that implies that a record was added interactively by the user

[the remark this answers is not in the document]
haynesjm42: Also I wonder if there are use-cases where an app would want to have modified records always pass the filter also?

data/Store.js
haynesjm42: Actually after playing around with it I think `ANY` is the best mode for inline editing, as it handles the 2 common inline editing use cases with filtering:

1. User modifies quantity field on record to equal 100
2. User adds field filter quantity==100 - modified record is still visible since current quantity == 100, even though committed quantity != 100
3. User commits record - record still visible

1. User adds field filter quantity==100
2. User modifies quantity field on record to equal 50 - record still visible since committed quantity == 100
3. User commits modified record - record disappears since committed and current quantity == 100

I think a case could be made for `ANY` to be the default mode actually, have a hard time coming up with uses for the other modes but felt it was best to leave it flexible for apps to configure so not really sure what we could recommend those modes for in the comments ...

lbwexler:
Hi -- really like your reasoning John, and agree that 'ANY' and the inclusion of added rows make sense as defaults --- seems like the most common/safest usage is to show more rather than less.  'ANY' is also least at odds with always keeping  the ADD records, which we want to do and is great.

On the fence about adding the other two modes now, and wondering if for now we should change the default  and not introduce mode until we *need* to.  Don't doubt  that a real intensive editing app may actually need some support along these lines -- just not sure we know exactly what it is, and hate to try to guess on this API.

Happy to go either way here.  Guess I would feel comfortable if we had a real world app that actually wanted something other then ANY

haynesjm42:
I would be fine with removing the mode entirely and just have FieldFilter use the  `ANY` mode behavior - I think that is what I would always want for an inline editable grid, and that is the only use-case I can think of for the different modes at all ...

data/filter/FieldFilter.js
```diff
@@ -87,79 +87,72 @@ export class FieldFilter extends Filter {
...
+        switch (store.filterValueMode) {
+            case FilterValueMode.COMMITTED:
+                // Pass only if the committed value passes
+                return r => r.isAdd || opTestFn(r.getCommitted(field));
+            case FilterValueMode.CURRENT:
+                // Pass only if the current value passes
+                return r => r.isAdd || opTestFn(r.get(field));
+            case FilterValueMode.ANY:
+                // Pass if either the current or committed value passes
```
lbwexler, line 153: running the test twice here in the any case, even if -- the most likely case --- there is no edit and the values are the same.  We may want to optimize that.

lbwexler:
just brainstorming here, but another option is we provide the ability to choose the records being excluded **seperately** from the choice of value, e.g

`alwaysInclude: (record:StoreRecord) => boolean`

or 

`alwaysInclude: 'MODIFIED'|ADDED'|'NONE'

PR #2984, opened 2022-05-01
Enhance NumberInput

lbwexler:
This fixes a number of issues with NumberInput. :-)

- Precision now enforced when first focused on overprecise source value.
- Precision now enforced when setting value back to source.
- Debounce provides less thrashing when user typing in numbers
- scaleFactor now clearly documented as being a factor of 10
- Max/min are enforced.  Workaround for related Blueprint errors with this prop handled more reliably 

Also provides some additional template Hooks in HoistInputModel for future use.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/input/HoistInputModel.js
+++ b/cmp/input/HoistInputModel.js
@@ -58,7 +58,7 @@ import ReactDOM from 'react-dom';
 export class HoistInputModel extends HoistModel {
 
     /**
-     * Does this input have the focus ?
+     * Does this input have the focus?
      * @type {boolean}
      */
     @observable hasFocus = false;
@@ -116,6 +116,7 @@ export class HoistInputModel extends HoistModel {
     @observable.ref internalValue = null;    // Cached internal value
     inputRef = createObservableRef();        // ref to internal <input> element, if any
     domRef = createObservableRef();          // ref to outermost element, or class Component.
+    isDirty = false;
 
     constructor() {
         super();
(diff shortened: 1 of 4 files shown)


TomTirapani:
This definitely fixes the issue.

However, I find the new behaviour about discarding out-of-bounds values a bit confusing. There's nothing to indicate *why* it was discarded, and this lack of user feedback, especially when going from a blank input back to a blank input, makes it feel broken.

Couldn't we just remove these props entirely, and say that if your app wants to enforce min/max values to use validation instead?

lbwexler:
Let's discuss -- I do find it useful, especially the ability to have a value of min = 0 or max = 0.  See the usage in LogPanel.

We can recommend validation and forms more strongly in the doc strincs for sure, but I think this is still useful for a raw numberInput.

cmp/input/HoistInputModel.js
```diff
@@ -116,6 +116,7 @@ export class HoistInputModel extends HoistModel {
     @observable.ref internalValue = null;    // Cached internal value
     inputRef = createObservableRef();        // ref to internal <input> element, if any
     domRef = createObservableRef();          // ref to outermost element, or class Component.
+    isDirty = false;
```
TomTirapani, line 119: I'm not clear on the need for this new flag? Under which circumstances would `doCommitInternal()` be called before `noteValueChange`?

TomTirapani:
Looks great, particularly the precisions enforcing! Eager to get this in 👍

cmp/input/HoistInputModel.js
lbwexler: when you tab through the form, without making any changes

lbwexler: or click in and out of it

PR #2985, opened 2022-05-03
Provide soft config to hide General > Users tab in Admin console

### LINKED ISSUE #2946: Provide option to hide the Admin > General > Users tab
The Admin console ships with a default, minimal, read-only users tab, nested under the General top-level tab.

For applications that need to provide more advanced user management within the admin console - e.g. adding and editing users - this built in tab can be confusing. In other scenarios, we might be loading users lazily from an API or other auth workflow, meaning this tab might not offer a complete listing of users and is misleading in another wya.

It would be good to provide an option to disable it. hiding it from the admin interface.

saba-mo:
Resolves #2946 

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/general/GeneralTab.js
+++ b/admin/tabs/general/GeneralTab.js
@@ -6,7 +6,7 @@
  */
 import {configPanel} from '@xh/hoist/admin/tabs/general/config/ConfigPanel';
 import {tabContainer} from '@xh/hoist/cmp/tab';
-import {hoistCmp} from '@xh/hoist/core';
+import {XH, hoistCmp} from '@xh/hoist/core';
 import {Icon} from '@xh/hoist/icon';
 import {aboutPanel} from './about/AboutPanel';
 import {alertBannerPanel} from './alertBanner/AlertBannerPanel';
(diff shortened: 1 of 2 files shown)


admin/tabs/general/GeneralTab.js
```diff
@@ -21,9 +21,14 @@ export const generalTab = hoistCmp.factory(
...
-                {id: 'users', icon: Icon.users(), content: userPanel},
+                {id: 'users', icon: Icon.users(), content: userPanel, omit: omitTab('hideUsersTab')},
                 {id: 'alertBanner', icon: Icon.bullhorn(), content: alertBannerPanel}
             ]
         }
     })
 );
+
+const omitTab = (field) => {
+    const conf = XH.getConf('xhAdminConsoleConfig', {});
```
lbwexler, line 32: Can we call it `xhAdminAppConfig`?  Can we add a default value in hoist core, so easy for new app users to find it?   

Also this function should just be called `hideUsersTab()` doesn't seem like it is ever going to be a general tab hiding mechanism

lbwexler, line 32: Can we call it `xhAdminAppConfig`?  Can we add a default value in hoist core, so easy for new app users to find it?   

Also this function should just be called `hideUsersTab()` doesn't seem like it is ever going to be a general tab hiding mechanism

amcclain: We discussed the merits of adding a bootstrapped value in hoist-core.  I voted against it for the moment, while we only have one little flag here, and a rarely used one at that. 

Don't feel strongly about it, but in general I don't mind the pattern of supporting some more rare edge-case configs with baked-in defaults and keeping the list in most apps a bit more easily scannable.

Agree re. `omitTab` - a pattern I've used in a few places is to have a getter on my *model* class like:

`get conf() {return XH.getConf('someJsonConfigForMyModel', {})}` 

and then in the component here you could go with `omit: model.conf.hideUsersTab`

amcclain: Woops - got caught up in another convo and submitted this too late - thanks for merging and getting this change in!

PR #2991, opened 2022-05-05
Improve handling when Local Storage not available.

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/LocalStorageService.js
+++ b/svc/LocalStorageService.js
@@ -19,7 +19,7 @@ export class LocalStorageService extends HoistService {
     constructor() {
         super();
         errorIf(this.isFake,
-            'Local Storage is *not* supported in this browser. Transient in-memory storage will ' +
+            'Local Storage is not supported in this browser. Transient in-memory storage will ' +
             'be used for various user state, and lost when the page is closed.'
         );
     }
--- a/svc/LocalStorageService.js
+++ b/svc/LocalStorageService.js
@@ -25,7 +25,7 @@ export class LocalStorageService extends HoistService {
             XH.handleException(
                 XH.exception(
                     'Local Storage is not supported in this browser. Transient in-memory storage ' +
-                    'will be used for various user state, and lost when the page is closed.'
+                    'will be used as a fallback.  All data stored will be lost when page is closed.'
                 ),
                 {showAlert: false}
             );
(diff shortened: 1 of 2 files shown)


cnrudd (approved):
I approve of the changes to LocalStorageService.js but don't have enough experience working with our auth process to know if the fetchservice change is the right one.

lbwexler:
colin and I discussed -- am sort of on the fence on this one, given that we only saw local storage not available once!

I suppose its less catastrophic, which is always a good thing.  We could also log the error on the server,  I suppose.... to make sure we address any such situation.

svc/LocalStorageService.js
```diff
@@ -16,6 +16,18 @@ import store from 'store2';
  */
```
cnrudd, line 16: I think another line or two here in the docs explaining that this service falls back silently to transient memory if localstorage is not available (thanks to default settings in the store2 lib) would be helpful to developers.

svc/LocalStorageService.js
```diff
@@ -16,6 +16,18 @@ import store from 'store2';
...
 
+    constructor() {
+        super();
+        errorIf(this.isFake,
+            'Local Storage is not supported in this browser. Transient in-memory storage will ' +
+            'be used for various user state, and lost when the page is closed.'
+        );
+    }
+
+    get isFake() {
```
cnrudd, line 27: Ok, it took me a while to understand why we're using `isFake`, and not `isUnavailable`:  Because we are falling back to `store`'s transient memory.    A doc comment above this function pointing this out might useful.  We're hoping the penny drops for all developers, if we don't document this.

svc/LocalStorageService.js
lbwexler: done

svc/LocalStorageService.js
```diff
@@ -16,6 +16,18 @@ import store from 'store2';
  */
 export class LocalStorageService extends HoistService {
 
+    constructor() {
+        super();
+        errorIf(this.isFake,
+            'Local Storage is not supported in this browser. Transient in-memory storage will ' +
+            'be used for various user state, and lost when the page is closed.'
```
cnrudd, line 23: the "for various user state"  part is a little confusing.
Could the whole message be:
'Local Storage is not supported in this browser. Transient in-memory storage will be used as a fallback, so data stored with this service will be lost when the page is closed.'

PR #2995, opened 2022-05-06
Column renderers should not be allowed to take down an app

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/format/FormatUtils.js
+++ b/format/FormatUtils.js
@@ -44,18 +44,3 @@ export function asElementDeprecationWarning(opts) {
         msg: 'Formatters return elements by default. You can use `asHtml` to return a HTML string'
     });
 }
-
-/**
- * Used to ensure grid renderers cannot bring down an application
- */
-export function makeRendererSafe(fn) {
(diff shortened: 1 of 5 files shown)


format/FormatUtils.js
```diff
@@ -43,3 +44,18 @@ export function asElementDeprecationWarning(opts) {
         msg: 'Formatters return elements by default. You can use `asHtml` to return a HTML string'
     });
 }
+
+/**
+ * Used to ensure grid renderers cannot bring down an application
+ */
+export function makeRendererSafe(fn) {
```
lbwexler, line 51: let's put this in the `/cmp/grid/impl`.  its going to be pretty specific, and no need to reveal it as a public API

lbwexler: + a bit curious about whether the spread and destructure adds any overhead (versus explicit arguments) 
+ might be helpful to include the name of the column (or 'GROUPROW') in the error message
+ we could consider whether to also wrap the export function, but I think better not to

Issue #2996, opened 2022-05-06
Improve performance of autosizing with AG27


febbraiod:
We are seeing an ugly delay in rendering a particular grid in a flagship client app with the AG27 upgrade. 

Appears to be related to autosizing a grid with a lot  of columns (29) and a lot of (although not a tremendous amount) of rows. Let's call it ~2500. The grid renders with the rows in place, but no visible data. After something like 3 full seconds the grid fills in.

When we turn autosizing off this problem goes away.

lbwexler:
@TomTirapani -- assigning you and I to see if we can do anything here.  Realize there may not be a simple solution here.  At the very least want to look at with you and understand the scope of the problem

TomTirapani:
Don showed me how to reproduce this, and I spent some time investigating - please see PR: https://github.com/xh/hoist-react/pull/3001

The "rows without data" was due to the autosize beginning before AG has finished rendering the rows. It appears that since v27, AG takes a little longer to render the rows. Increasing our [existing wait()](https://github.com/xh/hoist-react/pull/3001/files#diff-a2de0175ecc720a2fc3b534c0e1b01de3885368a1fe6b5a58a1f9f06589ee399R1223) allowed the rows to render prior to starting the autosize.

I also looked at the performance of the autosize in general. The bulk of the time is taken up by collecting the rendered values for records, in columns that display elements that need to be [converted to static markup](https://github.com/xh/hoist-react/blob/develop/cmp/grid/impl/ColumnWidthCalculator.js#L103).

Unfortunately, I'm not sure what we can do about this. Now that renderers can return elements, we need to do this conversion in order to be able to sample / size them later. Note that running react in prod mode improved the speed of this step slightly (~30% improvement), so it might be better in deployed apps.

I added [this line](https://github.com/xh/hoist-react/pull/3001/files#diff-679de26b40b407185f61bc73366c115fffb6abbf82057e1c35aa40e43592d798R112) to limit the number of records we sample, which halved the overall time for this largish grid. However, this change introduces the risk that a record with a long value is excluded. On the other hand, if your grid has more than a 1000 records, outliers are less likely to be encountered by the user, and an "approximate" width is more useful than an "exact" width. I'm not too sure about this last change, and happy to discuss.

lbwexler:
Closing this for now, reflecting Tom's substantial improvements for v49.  We will keep our eyes on it

PR #3001, opened 2022-05-10
Improve autosizing with AG27

TomTirapani:
See issue for more information, discussion and caveats: https://github.com/xh/hoist-react/issues/2996

+ Ensure current recordset has been rendererd prior to starting autosize.
+ Limit total number of records sampled.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnWidthCalculator.js
+++ b/cmp/grid/impl/ColumnWidthCalculator.js
@@ -102,7 +102,7 @@ export class ColumnWidthCalculator {
 
         records.forEach(record => {
             if (!record) return;
-            if (!renderedValues.size >= this.VALUE_SAMPLES && !record.isSummary) return;
+            if (renderedValues.size >= this.VALUE_SAMPLES && !record.isSummary) return;
 
             const ctx = {record, field, column, gridModel, store},
                 rawValue = getValueFn(ctx);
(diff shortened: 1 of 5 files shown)


lbwexler:
Would like to understand what the performance degradation is.  Of course, realize that is a little difficult for us, as we have already released v48 all the way out to prod in toolbox.

A couple of thoughts, regarding our pre-scan to find the longest bits of data

- For many cases (e.g. strings) the character length of the *raw* data is actually  what we want.
- For many cases (e.g. dates) there really are going to be many duplicates and we could memoize this computation.
- It seems like for formatted numbers, its a bit trickier, but we essentially are looking for the character formatted number
- We could offer some kind of hook for application to provide this efficient width info (e.g. `widthCalculator`) on the column or renderer.
- The current approach of limiting to 10% sounds interesting -- I wonder if we should be more random about it than taking the first 10%.  What if the first 10% are all found in the very first sub-node?

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -88,19 +92,35 @@ export class ColumnWidthCalculator {
...
             bufferPx = column.autosizeBufferPx ?? options.bufferPx;
 
         // 1) Get Map of rendered values to List of records that contain them
-        const recsByValue = new Map();
+        const recsByValue = new Map(),
+            renderedValues = new Map();
+
         records.forEach(record => {
             if (!record) return;
+            if (!renderedValues.size >= this.VALUE_SAMPLES && !record.isSummary) return;
```
lbwexler, line 105: this logic seems wrong to me.  Why the initial '!' on line 105

TomTirapani: Typo!

Issue #3014, opened 2022-05-13
Update to Onsen 1.12


amcclain:
Looks like they got around to a number of fixes, as well as react 17 support (just in time for 18...)

Note despite the minor version bump this is called out as a breaking change - https://github.com/OnsenUI/OnsenUI/blob/master/react-onsenui/CHANGELOG.md

lbwexler:
interestingly enough, colin tried this and ran into some basic issues with onsen's attempt to serialize our element trees.

TomTirapani:
So, good news and bad news...

Good news is that I was able to workaround the above the issue, and after actioning some small breaking changes on their side, toolbox mobile seems to working again.

Bad news is that the styling is completely messed up. They've started used CSS vars in their style sheets, which unfortunately match the names we provide for apps to override XH styles: e.g.
```
// Hoist vars.scss
--xh-text-color: var(--text-color, var(--xh-blue-gray-dark));

// Onsen
--text-color: #1f1f21;
```
I'm not sure what our options are for this, their CSS vars are coupled quite closely with their core styles

PR #3016, opened 2022-05-17
Upgrade Onsen to latest versions

TomTirapani:
onsenui: 2.11.1 > 2.12.0
react-onsenui: 1.11.5 > 1.12.0

See issue: https://github.com/xh/hoist-react/issues/3014

As discussed in the ticket, there is a required workaround for an issue with Onsen serialising components in kit/onsen/index.js. The PR works, but I wonder if there might be a better approach to ensuring HoistModels never get passed to Onsen components? Let's discuss.

Hoist P/R Checklist
-------------------


TomTirapani:
Updated this to the latest, and unfortunately its not compatible with latest webpack - the webpack sass-loader complains about typos in their CSS:

Screenshot 2022-08-11 at 11 48 42

lbwexler:
Now a Draft, as we need to look into the CSS issue.  

Regarding the custom elem factory, we could consider building in to ElementFactory the code to omit HoistModels from props whenever the component itself is not a HoistComponent.    

On the other hand, am loathe to add anything to ElementFactory that is more specific to Hoist -- really would like to see it stay as an independent layer that really could be broken out from Hoist 'one day'

lbwexler:
If we do decide to cleanse the props only for Onsen component, I wonder if we could come up with some kind of inner wrapper for all the Onsen Components that would not require us to publicize normalizeArgs and override elemFactory.

TomTirapani:
Great idea about using an inner wrapper to cleanse the props of model, can give that a try.

The CSS issue is harder, and I really don't see many options forward besides either submitting a PR to Onsen and / or using our own fork. :|

TomTirapani:
I've updated this branch to the latest hoist and onsen versions, and think its ready for review / merge

PR #3024, opened 2022-05-20
Allow enableClear button within mobile textInput

### LINKED ISSUE #3023: Mobile textInput should allow an enableClear button
The mobile textInput component should have an optional enableClear button within the input, as does desktop's textInput component.

saba-mo:
See https://github.com/xh/toolbox/pull/586
Resolves #3023 

Hoist P/R Checklist
-------------------

### DIFF
--- a/mobile/cmp/input/TextInput.js
+++ b/mobile/cmp/input/TextInput.js
@@ -6,9 +6,13 @@
  */
 import {HoistInputModel, HoistInputPropTypes, useHoistInputModel} from '@xh/hoist/cmp/input';
 import {hoistCmp} from '@xh/hoist/core';
+import {Icon} from '@xh/hoist/icon';
+import {hbox} from '@xh/hoist/cmp/layout';
 import {input} from '@xh/hoist/kit/onsen';
 import {withDefault} from '@xh/hoist/utils/js';
 import {getLayoutProps} from '@xh/hoist/utils/react';
+import {button} from '@xh/hoist/mobile/cmp/button';
+import {isEmpty} from 'lodash';
 import PT from 'prop-types';
 import './TextInput.scss';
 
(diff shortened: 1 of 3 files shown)


mobile/cmp/input/TextInput.scss
```diff
@@ -5,4 +5,15 @@
     height: 30px;
     padding: 0 var(--xh-pad-px);
   }
+
+  ons-input {
+    flex: 1;
+  }
```
saba-mo, line 11: If we don't want to take this route of adding a rule to the ons-input element, we can set the default width to 100% when configuring the input in TextInput.js.

lbwexler: will let tom comment on this

mobile/cmp/input/TextInput.js
```diff
@@ -96,31 +108,48 @@ const cmp = hoistCmp.factory(
...
+                        width: withDefault(width, null),
+                        textAlign: withDefault(props.textAlign, 'left')
+                    },
+
+                    onChange: model.onChange,
+                    onKeyDown: model.onKeyDown,
+                    onBlur: model.onBlur,
+                    onFocus: model.onFocus,
+                }),
+                model.showClearButton ? clearButton() : null
```
lbwexler, line 138: total nit, but might use omit for readability here.

saba-mo: Good point, change pushed.

mobile/cmp/input/TextInput.scss
TomTirapani: I think using flex is the right approach - however, I think it would be better to use a BEM className here rather than relying on Onsen's component name.

TomTirapani (approved):
Looks great! I pushed a couple of small tweaks - please review

saba-mo:
Thank you, both!

PR #3026, opened 2022-05-23
Fix focus issues with SelectEditor

lbwexler:
Hoist P/R Checklist
-------------------


haynesjm42:
This is unfortunate, as the extra keypresses required are not ideal either - but preferable to the grid losing focus after committing I suppose.

I did a quick experiment and it seems that calling `props.agParams.stopEditing()` instead of using `gridModel.stopEditingAsync()` does not cause the grid to lose focus so am thinking we should just call that instead to retain the single-press commit behavior. It seems that the ag-grid api method `stopEditing()` causes the grid to lose focus, so isn't something we are doing in our wrapper method.

Also of note here is that even with these changes hitting tab to commit a value will not immediately tab to the next cell as desired because that even must be getting eaten by the select, so you need to hit tab twice, and then manually begin editing the next cell since we stopped editing - not really the behavior you would want.

Issue #2959, opened 2022-04-19
DashCanvas bugs/enhancements


haynesjm42:
Some actionable enhancements/bug fixes to make to the new (beta) DashCanvas component, and other UX issues which may require further discussion.

## Bugs
- [x] The DashCanvas context menu is showing behind view contents context menus
- [x] Initial state is not set on `state` member for DashCanvasModel until 1 second after model creation due to debounce
- [x] In some cases the width is incorrect for ReactGridLayout, it doesn't seem to always update when parent container size changes - not clear on how to reproduce 100%
- [x] Add position from context menu is not accounting for scrolling, view are put in wrong place if you are scrolled down

## Enhancements
- [x] Add `emptyText` like we have in DashContainer
- [x] Add support for  hiding the panel headers on DashCanvasView (via view spec and via observable on DashCanvasViewModel)
- [x] Add ability to change the view contents to a different view spec
  * Would make it much easier to create balanced layouts by cloning a view and then changing to different contents
- [x] Add support for hiding the menu button on the view panel header
- [x] Menu items should be hidden instead of disabled when not applicable
  * Is confusing to see the disabled duplicate item when a view is marked as unique, etc. I think we should just always hide instead of disabling
- [x] Add built-in option to view a DashCanvasView in full-screen - similar to what we have for charts or code input
- [ ] Add support for min and max widths for DashCanvasViews
- [x] When duplicating a view, scroll the container so that the new view is centered
- [x] Add extra menu items that can be specified separately per viewSpec
- [x] Add header items that can be specified per viewSpec
- [ ] Add support for disabling duplicates on DashCanvasView (via view spec and via observable on DashCanvasViewModel)
- [ ] Make "ultra-compact" toolbar option - extend to Hoist framework as a whole?

## For Discussion
- [ ] Consider supporting initial width/height in pixels instead of columns/rows and doing the conversion when the view is added based on the current size/row height. Might be more intuitive when setting up view specs.
- [ ] Consider switching to the responsive version of ReactGridLayout 
- [ ] Consider deferring and/or staggering initial rendering of DashCanvasView contents for a short amount of time, showing something like the Blueprintjs skeleton styling. 
  * ag-grid in particular can be an expensive component to mount/initialize, so having a dashboard with several grids could result in periods of unresponsiveness
- [ ] Consider slowing down the css transitions for react-grid-item to make the initial render look smoother
- [ ] Consider adding option to copy/paste size of a view
  * It can be annoying to get views to the same exact size
[...]

PR #3033, opened 2022-06-01
Dash canvas bugs

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dash/canvas/DashCanvasModel.js
+++ b/desktop/cmp/dash/canvas/DashCanvasModel.js
@@ -4,9 +4,9 @@ import {DashCanvasViewModel, DashCanvasViewSpec} from '@xh/hoist/desktop/cmp/das
 import {Icon} from '@xh/hoist/icon';
 import {action, bindable, makeObservable, observable} from '@xh/hoist/mobx';
 import {ensureUniqueBy} from '@xh/hoist/utils/js';
+import {createObservableRef} from '@xh/hoist/utils/react';
 import {defaultsDeep, isEqual, find, without, times} from 'lodash';
 import {computed} from 'mobx';
-import {createRef} from 'react';
 import {throwIf} from '../../../../utils/js';
 
 /**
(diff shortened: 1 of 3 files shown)


desktop/cmp/dash/canvas/DashCanvasModel.js
```diff
@@ -168,6 +170,14 @@ export class DashCanvasModel extends HoistModel {
             track: () => [this.viewState, this.layout],
             run: () => this.publishState()
         });
+
+        this.addReaction({
+            when: () => this.ref.current,
+            run: () => {
+                const {current: node} = this.ref;
```
lbwexler, line 177: find the destructure with aliasing a little ponderous -- would like to discuss this vs.
`const node = this.ref.current;`

lbwexler: also, in this case, node can just be the argument to the run function.

desktop/cmp/dash/canvas/DashCanvasModel.js
```diff
@@ -338,6 +348,14 @@ export class DashCanvasModel extends HoistModel {
         layout = layout.map(({i, x, y, w, h}) => ({i, x, y, w, h}));
         if (!isEqual(this.layout, layout)) {
             this.layout = layout;
+
+            // Check if scrollbar visibility has changed, and force resize event if so
+            const {current: node} = this.ref,
```
lbwexler, line 353: are we sure that node exists here?

desktop/cmp/dash/canvas/DashCanvasModel.js
ghsolomon: Sounds good- I see what you're saying.  I think I tried using `node` as the argument to the `run` function and it not working, but will double-check.  I assumed that maybe it was only passed to `run` when using `track` (and not `when`)

ghsolomon:
#2959

PR #3034, opened 2022-06-01
Support for 'array' field type

haynesjm42:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/renderers/TagsRenderer.js
+++ b/cmp/grid/renderers/TagsRenderer.js
@@ -1,0 +1,12 @@
+import {isEmpty} from 'lodash';
+import {div, hbox} from '@xh/hoist/cmp/layout';
+import './TagsRenderer.scss';
+
+export function tagsRenderer(v) {
+    if (isEmpty(v)) return null;
+
+    return hbox({
+        className: 'xh-tags-renderer',
+        items: v.map(tag => div({className: 'xh-tags-renderer__tag', item: tag}))
+    });
+}
--- a/data/Field.js
+++ b/data/Field.js
@@ -10,7 +10,7 @@ import {isLocalDate, LocalDate} from '@xh/hoist/utils/datetime';
 import {withDefault} from '@xh/hoist/utils/js';
 import {Rule} from '@xh/hoist/data';
 import equal from 'fast-deep-equal';
-import {isDate, isString, toNumber, isFinite, startCase, isFunction} from 'lodash';
+import {isDate, isString, toNumber, isFinite, startCase, isFunction, castArray} from 'lodash';
 import DOMPurify from 'dompurify';
 
 /**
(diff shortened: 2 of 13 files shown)


data/filter/BaseFilterFieldSpec.js
```diff
@@ -151,7 +154,7 @@ export class BaseFilterFieldSpec extends HoistBase {
         return this.values &&
             this.enableValues &&
             this.supportsOperator(op) &&
-            (op === '=' || op === '!=');
+            (op === '=' || op === '!=' || op === 'includes' || op === 'does not include');
```
lbwexler, line 157: "not includes" to mirror "not like".  It is isn't grammatical but neither is "ends" or "starts"

data/filter/FieldFilter.js
```diff
@@ -157,6 +157,18 @@ export class FieldFilter extends Filter {
...
+                return r => {
+                    if (doNotFilter(r)) return true;
+                    const v = getVal(r);
+                    return !isNil(v) && v.some(it => value.includes(it));
+                };
+            case 'does not include':
+                return r => {
+                    if (doNotFilter(r)) return true;
+                    const v = getVal(r);
+                    return !isNil(v) && !v.some(it => value.includes(it));
```
lbwexler, line 170: seems to me this needs to be isNil(v) || !v.some(...).

data/Field.js
```diff
@@ -88,6 +88,13 @@ export function parseFieldValue(val, type, defaultValue = null, disableXssProtec
 
     const FT = FieldType;
```
lbwexler, line 89: just looking at this, and it occurs to me that we might be sanitizing *twice* if we send down a single string and cast it to an array.  Even worse, I think we are actually sanitizing string representations of dates/local dates/booleans, for no reason -- and we know this is expensive.  Should we have a method that bundles the sanitization + flag check and just conditionally call it  withing each block of the switch statement?

haynesjm42: Makes sense I think, since we only care about sanitizing values which would actually make it into record data (and then potentially back to a server), right?

cmp/grid/renderers/TagsRenderer.js
```diff
@@ -0,0 +1,12 @@
+import {isEmpty} from 'lodash';
+import {div, hbox} from '@xh/hoist/cmp/layout';
+import './TagsRenderer.scss';
+
+export function tagsRenderer(v) {
```
lbwexler, line 5: Should we take this on step further and add a `tags` column partial in `cmp/grid/columns/Core.js`?  Does not seem any less general that the columns we define there.

haynesjm42: Makes sense, should we also add some defaulting for TAGS fields in GridModel for columns to use the renderer? Or is the JS default toString of a csv a better default?

data/Field.js
lbwexler: great!

data/Field.js
```diff
@@ -88,6 +88,13 @@ export function parseFieldValue(val, type, defaultValue = null, disableXssProtec
...
     switch (type) {
+        case FT.TAGS:
+            val = castArray(val);
```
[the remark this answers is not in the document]
lbwexler: we actually don't typically throw in these cases, as john pointed out, we'll skip this.

PR #3057, opened 2022-06-20
Debounce Grid's data reaction + autosize improvements

lbwexler:
+ Solves a critical row-gap rendering problem jbar found in client app, which is reproduced in a toolbox branch with same name.
+ Improves autosizing to use asynchronous loops, throw away obsolete results, and start right away in managed mode.
+ I like the debounce change, because even if there is some fix to ag-Grid that would handle the "multiple data updates in single tick" we would not want to thrash in that way -- could be problematic in a number of ways for large data sets.
+ I like the async loop change because, it allows grid sizing to degrade better as grids get too large.  We should aim to never write code that can lock up browser.
+ One major client flagship app testing fine.  Want to test a bit more on a few more in-the-wild apps, and see if it (hopefully) solves other related row-gap rendering problems.  But think this can be merged now.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@
 
 * FilterChooser field suggestions now search within multi-word field names
 
+### 🐞 Bug Fixes
+* Fixes several issues where Grid would display rows gaps after operating on it programmatically,
+  This problem was introduced in ag-Grid v27.
+
+### 💥 Breaking Changes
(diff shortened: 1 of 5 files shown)


lbwexler:
Happy to merge this at anytime-- Tom and I are pretty comfortable with the autosize changes, which are the most extensive.

The data debounce change -- which I am most concerned about --  is a one liner that could be easily rolled-back

PR #3063, opened 2022-06-23
Fix bug where PanelHeader would undesirably render if showHeaderColla…

ghsolomon:
…pseButton=true

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/panel/impl/PanelHeader.js
+++ b/desktop/cmp/panel/impl/PanelHeader.js
@@ -19,13 +19,11 @@ export const panelHeader = hoistCmp.factory({
     className: 'xh-panel-header',
     render({className, ...props}) {
         const panelModel = useContextModel(PanelModel),
-            {collapsed, collapsible, isModal, hasModalSupport, vertical, side,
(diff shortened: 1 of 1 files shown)


ghsolomon:
Can set title to an empty string in order to force the PanelHeader to render

desktop/cmp/panel/impl/PanelHeader.js
```diff
@@ -19,16 +19,11 @@ export const panelHeader = hoistCmp.factory({
...
+            {collapsed, isModal, vertical, side} = panelModel,
             {title, icon, compact} = props,
             headerItems = props.headerItems ?? [];
 
-        if (!title && !icon && isEmpty(headerItems) &&
-            (!showHeaderCollapseButton || !collapsible) &&
-            (!showModalToggleButton || !hasModalSupport)) {
-            return null;
-        }
+        if (isNil(title) && isNil(icon) && isEmpty(headerItems)) return null;
```
lbwexler, line 26: I suppose if there is code in the wild that is using `title: ''`  to try to hide the panel, it would need to be changed to `title: null`.  I think that's better, just wondering if we should call this out in the changelog

PR #3064, opened 2022-06-23
Fix bug where ColumnHeaders would not respond to mouse events on tablets

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnHeader.js
+++ b/cmp/grid/impl/ColumnHeader.js
@@ -33,7 +33,7 @@ export const columnHeader = hoistCmp.factory({
 
     render({className, model, enableMenu, showColumnMenu, displayName}) {
         const {gridModel, xhColumn} = model,
-            {isDesktop} = XH;
+            hasMouseSupport = XH.isDesktop || XH.isTablet;
 
         const sortIcon = () => {
             const {abs, sort} = model.activeGridSorter ?? {};
@@ -75,8 +75,8 @@ export const columnHeader = hoistCmp.factory({
             return div({
                 className: 'xh-grid-header-expand-collapse-icon',
                 item: icon,
-                onClick: isDesktop ? model.onExpandOrCollapse : null,
-                onTouchStart: !isDesktop ? model.onExpandOrCollapse : null
+                onClick: hasMouseSupport ? model.onExpandOrCollapse : null,
+                onTouchStart: !hasMouseSupport ? model.onExpandOrCollapse : null
             });
         };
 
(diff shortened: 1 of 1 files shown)


cmp/grid/impl/ColumnHeader.js
```diff
@@ -75,8 +75,8 @@ export const columnHeader = hoistCmp.factory({
             return div({
                 className: 'xh-grid-header-expand-collapse-icon',
                 item: icon,
-                onClick: isDesktop ? model.onExpandOrCollapse : null,
-                onTouchStart: !isDesktop ? model.onExpandOrCollapse : null
+                onClick: hasMouseSupport ? model.onExpandOrCollapse : null,
+                onTouchStart: !hasMouseSupport ? model.onExpandOrCollapse : null
```
TomTirapani, line 79: Doesn't this mean that tablets no longer respond to touch events? Or do iPads fire both kind of events when interacted with via the touch screen?

ghsolomon: Seems tablets fire click events 300ms after touchStart, to make sure you keep your finger on the target and weren't simply touching a target to initiate a swipe (to scroll) or double-tap (to zoom).  I was able to run this on my iPad and get the desired behavior touching and mouse-clicking.

Here's a couple articles I found:
https://www.quirksmode.org/blog/archives/2014/04/suppressing_the.html
https://www.sitepoint.com/5-ways-prevent-300ms-click-delay-mobile-devices/

Another approach could be to bind both events and add preventDefault() to onTouchStart since that fires first.

In the future, may be able to use this proposed PointerEvent interface:
https://www.w3.org/TR/pointerevents/
"To reduce the cost of coding to multiple input types and also to help with the above described ambiguity with Mouse Events, this specifications defines a more abstract form of input, called a [pointer](https://www.w3.org/TR/pointerevents/#dfn-pointer). A pointer can be any point of contact on the screen made by a mouse cursor, pen, touch (including multi-touch), or other pointing input device."

TomTirapani: I wouldn't hold my breath for PointerEvents :)

I'd like to see if we can try the approach of binding both events and using `preventDefault()` on touch events. Then we can favour touch events, and keep click events as a fallback. Maybe even losing the conditional binding consistently.

It just seems strange to me route touch events through the click event methods on tablets in order to support a trackpad (i.e. a 'secondary' input method)

ghsolomon: Makes sense

amcclain:
Thanks guys for working through this one - @TomTirapani can you please merge if all looks good to you tomorrow AM? 

(I wanted to ensure that you had a chance to loop back on your earlier discussion above.)

TomTirapani:
Code looks great!

@ghsolomon - has this been tested on device? I'm not able to test with the trackpad, and want to make sure we're not getting both kinds of events for the interactions. If you could please just confirm that this approach doesn't introduce double events I'll be happy to merge.

ghsolomon:
@TomTirapani yes tested on an iPad using the trackpad and worked as intended! Also tested using touch on the iPad for safe measure.

PR #3071, opened 2022-07-01
DashCanvas fails gracefully on unknown viewSpecId

amcclain:
+ A persisted state could ref. an obsolete widget - currently that will throw the model ctor.

Hoist P/R Checklist
-------------------


amcclain:
There are two other assertions within `addViewInternal()` that throw - wasn't sure if they could be encountered when restoring saved state.  The one around `allowAdd` seemed sus -  especially given this doc comment: 

https://github.com/xh/hoist-react/blob/0d7a8d3fc718d6b91c7e85b9ea56bb1396478067/desktop/cmp/dash/DashViewSpec.js#L40

ghsolomon:
This change looks good to me, and agreed- seems we want to handle these cases better when loading from persisted state.  

allowAdd in particular like you mentioned seems like it should only apply to the user trying to manually add a widget, not when loading from a persisted or initial state.

Issue #3075, opened 2022-07-06
Desktop ContextMenu imported into cross-platform Chart component


TomTirapani:
The Desktop / Blueprint ContextMenu component is being imported into the cross-platform Chart component, via its use of the `useContextMenu()` hook: https://github.com/xh/hoist-react/blob/develop/cmp/chart/Chart.js#L11

cnrudd:
Thank you for taking this on @Ryanseanlee .  
I wonder if the [useContextMenu hook ](https://github.com/xh/hoist-react/blob/develop/desktop/hooks/UseContextMenu.js) can be rebuilt as our first example of a "dynamics" hook that, when called in mobile turns into a no-op?
### Hoist's dynamics:
https://github.com/xh/hoist-react/blob/develop/dynamics/desktop.js
https://github.com/xh/hoist-react/blob/develop/dynamics/mobile.js

lbwexler:
I think that's right Colin.  We might do the same with HotKeys at the same time for good measure.  Its the only other function in `@xh/hoist/desktop/hooks` and would probably be useful if it could be a no-op on mobile as well.

Issue #3074, opened 2022-07-06
Enforce platform specific imports


TomTirapani:
The framework should be able to detect and warn/throw if a platform specific hoist component has been imported into the incorrect platform. We can leverage `XH.isMobileApp` to know what platform the app expects.

PR #3076, opened 2022-07-06
Enforce importing platform specific components

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/3074

HoistComponents can be optionally tagged as being intended for use with either `desktop` or `mobile` toolkits by setting a flag in `hoistComponent`, e.g.
```
export const [Button, button] = hoistCmp.withFactory({
    displayName: 'Button',
    desktop: true,
   ...
});
```
If these components are imported into an app using the wrong platform, `XH.renderApp()` will throw. Note that the component only has to be imported, not necessarily rendered.

Also note that this change has surfaced an import issue, that should be resolved before this is merged: https://github.com/xh/hoist-react/issues/3075

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### 🎁 New Features
 
+* Hoist now enforces platform-specific component imports. Importing components from the `desktop`
+  or `mobile` packages into apps for the opposing platform (determined via `AppSpec.isMobileApp`)
+  will cause the app to throw.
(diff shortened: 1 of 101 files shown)


lbwexler:
I guess I worry that this adds a burden to the declaration of HoistCmp, and users will think that they also need to do that for their own Hoist Components (they typically won't).  I guess that could be considered a feature.

I was thinking about something more lightweight -- a single method call at the top of the file, that just sets a global variable.  Could be done for all files, not just components --- or alternatively could just be added to the index files.

PR #3082, opened 2022-07-08
Enforce platform-appropriate imports

Ryanseanlee:
Hoist P/R Checklist
-------------------


TomTirapani:
Looks great!

A minor (but important) thing, but could we ensure the new `xxx/register` imports are the first import, at the top of the list? It'll will help devs know at a glance if its always in the same place.

Ryanseanlee:
So we found two cases where the mobile app was using desktop components. The two files are Chart.js and DockViewModel.js. 
For chart.js it is importing the useContextMenu and Anselm said @cnrudd you might have some insight on this. 
For DockViewModel.js it is also using ModelSupport @ghsolomon Anselm said this was recently added and if we could consider moving the entire DockView component into the desktop package.

Ryanseanlee:
While we were troubleshooting this we changed the register file to throw so that we get a stack trace back to identify the file, removing the throw in the XH file. Should we use this approach? 
```
if(XH._desktopAPIRegistered){
    throw "`Desktop components imported into mobile app.  Please check your imports.`";
 }
 XH._mobileAPIRegistered = true;
```

TomTirapani:
Ah, so it gives a stack trace back to the offending file? Smart! I like that approach :)

There's a ticket for the Chart JS import: https://github.com/xh/hoist-react/issues/3075

For the DockView, I'm surprised that's being included in mobile at all. I would also vote for moving it to dekstop, we have no plan (nor desire) for a mobile implementation.

desktop/register.js
```diff
@@ -0,0 +1,6 @@
+import {XH} from '@xh/hoist/core';
+
+if(XH._mobileAPIRegistered){
```
lbwexler, line 3: fine, but we should use `XH.exception()` to create this exception for good measure.

lbwexler, line 3: add space after `if`

lbwexler:
I also agree that, would be nice for the register import to be first in the file, but also know that there is a desire to let the imports be managed by the IntelliJ.

Would a line break help?

Ryanseanlee:
Anselm and I also talked about having the import first and we found that the line break does not help as it just organizes it anyways. So, we decided to just let IntelliJ manage the imports.

lbwexler (approved):

lbwexler:
I addressed the problem with ContextMenu to provide a dynamic implementation, and am reviewing that with Ryan.
This just extends a pattern we are already doing with a number of hoist implementations, no need for any API changes, or introducing new APIs.

Ryan will move the DockViewModel and we should be good to go!

Ryanseanlee:
https://github.com/xh/toolbox/pull/594 This is the toolbox update to have the correct import location for the DockView.

PR #3099, opened 2022-07-29
Added enableMulti option to ButtonGroupInput

Ryanseanlee:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/ButtonGroupInput.js
+++ b/desktop/cmp/input/ButtonGroupInput.js
@@ -4,13 +4,14 @@
  *
  * Copyright © 2022 Extremely Heavy Industries Inc.
  */
+import {active} from '@xh/hoist/admin/columns';
 import {HoistInputModel, HoistInputPropTypes, useHoistInputModel} from '@xh/hoist/cmp/input';
 import {hoistCmp} from '@xh/hoist/core';
 import {Button, ButtonGroup, buttonGroup} from '@xh/hoist/desktop/cmp/button';
 import '@xh/hoist/desktop/register';
 import {throwIf, withDefault} from '@xh/hoist/utils/js';
 import {getLayoutProps, getNonLayoutProps} from '@xh/hoist/utils/react';
-import {filter} from 'lodash';
+import {castArray, filter} from 'lodash';
 import PT from 'prop-types';
 import {Children, cloneElement} from 'react';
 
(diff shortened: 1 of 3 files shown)


desktop/cmp/input/ButtonGroupInput.js
```diff
@@ -64,6 +78,23 @@ class Model extends HoistInputModel {
         const btns = this.domEl?.querySelectorAll('button') ?? [];
         return filter(btns, {disabled: false});
     }
+
+    active(value){
+        if (!this.enableMulti) return this.renderValue === value;
```
lbwexler, line 83: probably best to use "internalValue" to be consistent with usage below. (although will be the same for this control.  Also destructure `internalValue` and consider ternary!

desktop/cmp/input/ButtonGroupInput.js
```diff
@@ -64,6 +78,23 @@ class Model extends HoistInputModel {
...
+    onButtonClick(value){
+        if (this.enableMulti){
+            const current = this.internalValue ?? [];
+            value = this.active(value) ?
+                current.length === 1 ? null : without(current, value) :
+                [...current, value];
+        } else {
+            value = this.enableClear && this.active(value) ? null : value;
+        }
+        this.noteValueChange(value);
```
lbwexler, line 96: 👍🏻   This logic looks great now.

Ryanseanlee:
We have a problem initializing a value using the forms/ form model. When enableMulti is true we expect (other than null) an array of strings as the value of the button group. However, it may be common for the user to set the initial value of the form model to a string as that is how it works without enableMulti. We have no way of intercepting/ fixing this value because it is set directly and not set through one of the HoistInputModel functions.

https://github.com/xh/hoist-react/blob/998a64a555f37e417029bb62603b0f5c44548642/cmp/form/field/BaseFieldModel.js#L186-L190

mobile/cmp/input/ButtonGroupInput.js
```diff
@@ -25,14 +26,20 @@ export const [ButtonGroupInput, buttonGroupInput] = hoistCmp.withFactory({
...
     }
 });
 ButtonGroupInput.propTypes = {
     ...HoistInputPropTypes,
 
     /** True to allow buttons to be unselected (aka inactivated). Defaults to false. */
-    enableClear: PT.bool
+    enableClear: PT.bool,
+
+    enableMulti: PT.bool
```
lbwexler, line 42: need jsdoc here.  Make both mirror desktop at the same time

lbwexler:
This looks good Ryan.  Will cleanup the one docstring and merge.

Regarding the issue about a user setting the value type to something incorrect -- will move that text to another ticket -- think its a much more general bug that afflicts all of our controls.

PR #3105, opened 2022-08-01
Fix `ModalSupport` bugs in `CodeInput`

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/CodeInput.js
+++ b/desktop/cmp/input/CodeInput.js
@@ -5,7 +5,7 @@
  * Copyright © 2022 Extremely Heavy Industries Inc.
  */
 import {HoistInputModel, HoistInputPropTypes, useHoistInputModel} from '@xh/hoist/cmp/input';
-import {div, filler, fragment, frame, hbox, label, span, vbox} from '@xh/hoist/cmp/layout';
+import {box, div, filler, fragment, frame, hbox, label, span, vbox} from '@xh/hoist/cmp/layout';
 import {hoistCmp, managed, XH} from '@xh/hoist/core';
 import {button} from '@xh/hoist/desktop/cmp/button';
 import {clipboardButton} from '@xh/hoist/desktop/cmp/clipboard';
@@ -17,6 +17,7 @@ import '@xh/hoist/desktop/register';
 import {Icon} from '@xh/hoist/icon';
 import {textArea} from '@xh/hoist/kit/blueprint';
 import {action, bindable, makeObservable, observable} from '@xh/hoist/mobx';
+import {wait} from '@xh/hoist/promise';
 import {withDefault} from '@xh/hoist/utils/js';
 import {getLayoutProps} from '@xh/hoist/utils/react';
 import classNames from 'classnames';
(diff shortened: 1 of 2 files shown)


lbwexler:
@TomTirapani -- actually, this is the one I was hoping you could look at.

desktop/cmp/input/CodeInput.js
```diff
@@ -331,6 +332,13 @@ class Model extends HoistInputModel {
 
     toggleFullScreen() {
         this.modalSupportModel.toggleIsModal();
+
+        // 'Nudge' the mouse wheel to trigger CodeMirror to update scrollbar state
+        const scrollEvent = d => new window.WheelEvent('mousewheel', {deltaX: d, deltaY: d});
+        wait().then(() => {
+            this.editor.getScrollerElement().dispatchEvent(scrollEvent(2));
+            this.editor.getScrollerElement().dispatchEvent(scrollEvent(-2));
+        });
```
TomTirapani, line 341: This is an interesting workaround! Did you try `this.editor.refresh()`? It seems to me from the docs that that might do the trick in a simpler way.

desktop/cmp/input/CodeInput.js
```diff
@@ -426,18 +434,20 @@ class Model extends HoistInputModel {
...
-
-        return modalSupport({
-            model: model.modalSupportModel,
-            item: inputCmp(childProps)
+            item: modalSupport({
+                model: model.modalSupportModel,
+                item: inputCmp({
+                    width: '100%',
+                    height: '100%',
+                    className,
```
TomTirapani, line 446: Do we still want the `className` to go here? I think typically we want the `className` on the outermost container, so devs always have a styling 'hook' into the component

ghsolomon: That makes sense.  I guess I was thinking that moving the `className` one component up in the hierarchy might be more disruptive to any custom styling devs were applying?  But I agree with you, better to have it applied to the outermost container- will make that change and test the `refresh()` method you mentioned above

desktop/cmp/input/CodeInput.js
ghsolomon: `this.editor.refresh()` doesn't seem to refresh the scrollbars unfortunately.

desktop/cmp/input/CodeInput.js
TomTirapani: Sorry - upon taking a closer look at it, I realised this means our styling does not get applied **at all** in modal mode! Apologies for the mislead there.

I'll revert this change, and instead add a *new* className to the outer component.

ghsolomon: I had this thought too, but I think the styling does still get applied.  In modal mode, the `inputCmp` gets rendered inside a blueprint `Dialog` inside `box`.  So styling applied to `box` still would get applied to `inputCmp` I think, right?

TomTirapani: No, because the Dialog is shown in a portal nested at the bottom of `<body>`. So any inherited styles no longer apply.  You can see the difference if you compare e.g. the font size in toolbox-dev against this branch (prior to my latest push)

ghsolomon: Ah got it! That makes sense, thanks for clarifying

TomTirapani:
Looks good to me - will merge!

PR #3107, opened 2022-08-02
Fix `Grid` rendering glitches when used inside `Panel` with `ModalSup…

ghsolomon:
…port`

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -567,7 +567,8 @@ class GridLocalModel extends HoistModel {
 
         return {
             track: () => modalSupportModel.isModal,
-            run: () => wait().then(() => this.model.agApi.redrawRows())
+            run: () => this.model.agApi.redrawRows(),
+            delay: 1 // Wait one browser tick for grid to move to new part of DOM before redrawing
         };
     }
 
--- a/cmp/grid/Grid.js
+++ b/cmp/grid/Grid.js
@@ -12,7 +12,8 @@ import {hoistCmp, HoistModel, lookup, useLocalModel, uses, XH} from '@xh/hoist/c
 import {
     colChooser as desktopColChooser,
     gridFilterDialog,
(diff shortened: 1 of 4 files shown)


cmp/grid/Grid.js
```diff
@@ -552,6 +557,20 @@ class GridLocalModel extends HoistModel {
...
+     * Force Grid to redraw rows when switching between inline and modal views
+     * {@see modalSupport}
+     */
+    modalReaction() {
+        const modalSupportModel = this.lookupModel(ModalSupportModel);
+        if (!modalSupportModel) return null;
+
+        return {
+            track: () => modalSupportModel.isModal,
+            run: () => wait().then(() => this.model.agApi.redrawRows())
```
lbwexler, line 570: can also just use `delay: 0` on reaction for slightly flatter code

ghsolomon: Needed to do `delay: 1` (maybe MobX makes `delay: 0` a no-op?), but that worked and is cleaner

lbwexler:
got it!  could do debounce: 0, also, that works and perhaps more consistent.

I think the comment is not needed here as well

ghsolomon:
`debounce: 0` works.  I wonder if `delay` is clearer since we really are intending to delay the reaction in this case?  Not really worrying about it firing multiple times in succession.  But happy to update (and remove the comment either way)!

cmp/grid/Grid.js
```diff
@@ -177,6 +178,10 @@ class GridLocalModel extends HoistModel {
         this.addReaction(this.sizingModeReaction());
         this.addReaction(this.validationDisplayReaction());
 
+        // If ancestor of a modalSupport cmp:
```
lbwexler, line 181: no need for comment here,  the method itself is perfectly clear, and the benefits of having the method mean we don't need to distract in the constructor.

PR #3121, opened 2022-08-11
Enhance `AboutDialog` to display more dynamic custom properties.

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * `ButtonGroupInput` supports new `enableMulti` prop.
 * Admin activity tracking chart now has a checkbox to toggle weekends when viewing a time
   series.
+* `AboutDialog` can now display more dynamic custom properties.
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
 
 * `CodeInput` is now rendered within an additional `div` element.  Unlikely to cause issues, unless
   using targeted styling of this component.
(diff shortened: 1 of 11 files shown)


ghsolomon:
Would love to get everyone's feedback on this.  Currently, custom properties are added to the `AboutDialog` between `User Agent` and `WebSockets`.  @lbwexler and I spoke about the possibility of revisiting this / perhaps adding a dedicated section for app-specific properties below the generic Hoist properties but wanted to get everyone's opinion.

lbwexler:
Hi @ghsolomon   So @amcclain and I had a great convo about this, and decided we would like to take this change one step further, and both clean-it up and provide more functionality.  Essentially we would do the following:

+ Add an overrideable method to HoistAppModel of the form `HoistAppModel.getAboutMenuItems()`.  It can return a list of items of the form `[{label: [String], value: [Element]}, omit: [boolean, opt]}]`.   Items returned by this method would be displayed as the about dialog, in the order they are provided.  This method would be similar to the pattern that we have with getAppOptions()

+ Hard remove support for the existing config based extension.  We could even output a warning in the base implementation of getAboutMenuItems() if this config is actually found.  This should be an easy update for apps.

Advantages are that apps will have much more control over this menu, including sorting, displaying rich react elements, removing default elements that are not desired, and having different about menus for different client apps!

lbwexler:
@ghsolomon -- also, think maybe you were hinting at this kind of change in our convo yesterday -- I was probably the one with the lack of ambition here!

PR #3133, opened 2022-09-12
XH.getActiveModels()

lbwexler:
Hoist P/R Checklist
-------------------

- [X ] Caught up with `develop` branch as of last change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Admin detailed configuration comparison display now shows who updated each value and when.
 * `DashCanvasViewModel` now supports `autoHeight` to automatically resize a `DashCanvasView's`
     height to fit its contents
+* New method `XH.activeModels()`allows accessing all currently active models in the application.
 
 ### ⚙️ Technical
 * Fixed an issue where `HoistComponents` were mutating the props object passed to them in React
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 * `DashCanvasViewModel` now supports `autoHeight` to automatically resize a `DashCanvasView's`
     height to fit its contents
 * New method `XH.activeModels()`allows accessing all currently active models in the application.
+* `ModelSelector` used for model lookup/matching will now accept the class name of the model to
+ match.  Previously only a class reference could be provided.
 
 ### ⚙️ Technical
 * Fixed an issue where `HoistComponents` were mutating the props object passed to them in React
--- a/core/HoistModel.js
+++ b/core/HoistModel.js
@@ -4,7 +4,7 @@
  *
  * Copyright © 2022 Extremely Heavy Industries Inc.
  */
-import {forOwn, has, isFunction} from 'lodash';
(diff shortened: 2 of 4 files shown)


ghsolomon (commented):
Looks really good!  And I like your idea of updating `matchesSelector()` so you can more easily provide a string classname.  I played with that a bit- see what you think:

```
    matchesSelector(selector, acceptWildcard = false) {
        if (selector === '*') return acceptWildcard;
        if (isString(selector)) return this.constructor.name === selector;
        if (!isFunction(selector)) return false;

        // 1) selector is a constructor function/class
        if (selector.isHoistModel) {
            return this instanceof selector;
        }

        // 2) selector is a callable function, call for either a boolean or a constructor function/class
        const result = selector(this);
        if (isString(result)) return this.constructor.name === result;
        return isFunction(result) && result.isHoistModel ?
            this instanceof result :
            !!result;
    }
```

lbwexler:
what you propose looks good, but my only concern there was that what if we later introduce a name property on models and want to allow searching by name.  That will also be a string -- am nervous about right now saying that any string will just be considered to be a classname.

would it be too hacky to have a string be interpreted as either a name or a class name?  Or should we have some kind of object form e.g. {name: 'mymodel'}, {class: 'FooModel'}, etc. that supports a lodash like flexibility?

amcclain:
A few incremental additions I might propose:

* Should we nest this feature under a singleton class or service on which we could hang additional future functionality - something like `XH.ModelRegistry`?
* Could we add a ready-to-getter that returns models grouped by ctor name, especially for easy and repeated console access.
* Could we add a `stats` object that provided some simple counts of both total (life of app) and currently active models, also grouped by ctor name? I don't know if or how this would specifically be useful, but it would admittedly be interesting, and might help keep an eye out for memory leaks or things like reactions firing much more often than intended and creating huge numbers of models.
* What's the intention with the proposed name property?  Is this a class-level property, distinct from the ctor name, or more like an id or key to identify an individual model instance?

ghsolomon:
> what you propose looks good, but my only concern there was that what if we later introduce a name property on models and want to allow searching by name. That will also be a string -- am nervous about right now saying that any string will just be considered to be a classname.
> 
> would it be too hacky to have a string be interpreted as either a name or a class name? Or should we have some kind of object form e.g. {name: 'mymodel'}, {class: 'FooModel'}, etc. that supports a lodash like flexibility?

That makes sense - I had the same thought about being able to pass an object like you suggested.

Also thought we could provide a helper function like `const ofClassNamed = className => it => it.constructor.name === className` so you could just write `matchesSelector(ofClassNamed('MyClass'))` but that doesn't really help when doing this from the console

lbwexler:
@amcclain -- thanks for the comments -- Greg and I discussed in detail.
We think the ideas all make sense and could be future extensions, but none of them would be made impossible by the simple first step we have here.  We are only introducing a single API addition to XH right now, that will be easy to continue supporting no matter what direction we go

Regarding searching for models instances by 'name', we can discuss further what I am thinking, but its all pretty speculative  and I don't think it needs to hold up this PR

PR #3140, opened 2022-09-25
Hoist Inspector

amcclain:
Hoist P/R Checklist
-------------------


cmp/form/FormModel.js
```diff
@@ -264,7 +264,9 @@ export class FormModel extends HoistModel {
             get(target, name, receiver) {
 
                 const field = me.fields[name];
-                if (field) return field.getDataOrProxy();
+                if (field?.isFieldModel) {
+                    return field.getDataOrProxy();
+                }
```
amcclain, line 269: This was triggered by Inspector code looking to print the type of this `values` property via `values.constructor.name`. That threw in this code, which is assuming that all non-null properties accessed off of `values` will be a `FieldModel` - which is almost always true except for edge-cases like this one.

core/HoistModel.js
```diff
@@ -62,6 +62,7 @@ export class HoistModel extends HoistBase {
     @observable
     _componentProps = {};
     _modelLookup = null;
+    _created = Date.now();
```
amcclain, line 65: Seems like a useful / harmless-enough property to install generally. We could prefix with `xh` if that seemed safer, but we weren't doing that for the other `_` props we set here.

cmp/form/FormModel.js
```diff
@@ -264,7 +264,9 @@ export class FormModel extends HoistModel {
             get(target, name, receiver) {
 
                 const field = me.fields[name];
-                if (field) return field.getDataOrProxy();
+                if (field?.isFieldModel) {
```
lbwexler, line 267: interesting -- what else can it be?

core/HoistModel.js
lbwexler: agree

desktop/appcontainer/inspector/models/ModelsModel.js
```diff
@@ -0,0 +1,302 @@
...
+
+        this.propsWatchlist = currItem ?
+            without(propsWatchlist, currItem) :
+            [...propsWatchlist, {xhId, property}];
+    }
+
+    getModelInstance(xhId) {
+        if (!xhId) return null;
+        const matches = XH.getActiveModels(it => it.xhId === xhId);
+        return matches.length ? matches[0] : null;
```
lbwexler, line 110: consider `first(match)`  or `first(matches) ?? null`

desktop/appcontainer/inspector/models/ModelsModel.js
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    togglePropsWatchlistItem(xhId, property) {
+        const {propsWatchlist} = this,
+            currItem = this.getWatchlistItem(xhId, property);
+
+        this.propsWatchlist = currItem ?
+            without(propsWatchlist, currItem) :
+            [...propsWatchlist, {xhId, property}];
```
lbwexler, line 104: consider lodash concat here for symmetry

desktop/appcontainer/inspector/stats/StatsPanel.js
```diff
@@ -0,0 +1,55 @@
...
+                compact: true,
+                items: [
+                    button({
+                        text: 'Reset stats',
+                        icon: Icon.reset(),
+                        onClick: () => XH.inspectorService.clearStats()
+                    }),
+                    filler(),
+                    popover({
+                        target: span(Icon.info(), ' JS Heap'),
```
lbwexler, line 44: on the fence about including this, given not related to the current tab.

desktop/cmp/button/AppMenuButton.js
```diff
@@ -95,15 +101,18 @@ function buildMenuItems({
     hideChangelogItem,
     hideFeedbackItem,
     hideImpersonateItem,
+    hideInspectorItem,
```
lbwexler, line 104: I might be happy to leave this out in favor of a more obscure keyboard shortcut.  Let's discuss.  This to me seems like more of a developer's tool (unlike e.g. admin/impersonation which really are for business users as well).  Not sure I want to make promises that it is "part" of the app...

svc/InspectorService.js
```diff
@@ -0,0 +1,168 @@
...
+        });
+
+        this.modelInstanceStore.loadData(modelData);
+
+        const {totalJSHeapSize, usedJSHeapSize} = (window.performance?.memory ?? {}),
+            modelCount = modelData.length,
+            prevModelCount = this._prevModelCount,
+            now = Date.now();
+
+        if (modelCount !== prevModelCount || now - this._lastStatsUpdate > 30*SECONDS) {
```
lbwexler, line 150: consider `olderThan()` util function.

lbwexler:
Think this is great and eager to get it in.  Some additional thoughts:

+ Think we ought to put nearly *all* the code for this in its own root level `inspector` package, similar to admin.  The more we keep it separate, the better, for future moves and clarity, and code searching etc.  Also -- could this facilitate a move to import all of this code DYNAMICALLY in the future?  

+ Think we could put the service code in to a model, reserving the service to be more stateless and "on-demand" for usages, say, directly on the console. But curious about the decision to have it in the service. 

+ Do we need to adjust the app container so that the tool is "outside" of the apps load mask?

utils/js/LangUtils.js
```diff
@@ -72,24 +72,30 @@ export function deepFreeze(obj) {
...
+ * Output a deep copy of an object or array up to a given depth, beyond which nested contents will
+ * be replaced by a placeholder string.
  *
- * @param {Object} obj
+ * @param {Object|Array} obj
  * @param {number} depth - maximum depth within the object tree that will be returned.
  */
-export function trimToDepth(obj, depth) {
-    if (depth < 1) return null;
+export function trimToDepth(obj, depth = 1) {
```
lbwexler, line 81: nice to have the default

lbwexler (commented):
👍🏻

amcclain:
Updated to show services in grid and support showing lazily evaluated getters in props listing.

I think this is pretty much it for me on this first iteration!

amcclain:
Please merge with simple Toolbox panel to highlight the feature - https://github.com/xh/toolbox/pull/602

PR #3151, opened 2022-10-02
xhImpl flag

amcclain:
+ Partial application of new HoistBase.xhImpl for review, with WIP support in Inspector for filtering out xhImpl instances.
+ Rework InspectorService to process instances and stats into observable arrays rather than stores.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,13 +11,15 @@
   match. Previously only a class reference could be provided.
 * The Admin Console diff tool for Configs, Prefs, and JSONBlobs now displays who updated each value
   and when.
-* New "Hoist Inspector" tool available in Desktop apps for displaying and querying models running in
-  an application and their properties (especially observables).
+* New "Hoist Inspector" tool available in Desktop apps for displaying and querying Models and
+  Services within the application.
     * Powered by a new method `XH.activeModels()`, which supports listing and querying all models
       instantiated within a running app, and `XH.inspectorService`, which provides additional
       processing of that model data when active.
     * Admin/dev-focused UI available on Desktop, activated via discrete new toggle in the bottom
       version bar (magnifying glass icon), or via console with `XH.inspectorService.activate()`.
+    * Selecting a model or service instance allows a quick view at its properties, including
+      reactively updated observables. Useful for troubleshooting application state in realtime.
 
 ### 💥 Breaking Changes
 
@@ -29,6 +31,8 @@
 ### 🐞 Bug Fixes
 
 * Boolean filterChooser fields with enableValues=false will no longer suggest true for 'false'.
+* Change to `CompoundTaskObserver` to prioritize using specific messages from subtasks over the
+  overall task message.
 
 ### ⚙️ Technical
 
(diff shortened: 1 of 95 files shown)


amcclain:
@lbwexler thanks for your comments on the initial commit (d9e77b0ead44510da58c1086d08cb1ecdd1ca351)

I think the key thing for us to decide is "what does `xhImpl` true vs. false distinguish between?"  It could be:  

a) Classes / instances representing important public APIs that app devs must know about, work with, create, access, etc. 
b) Hoist internal classes which devs should mostly (or entirely) avoid creating or interacting with directly.

-OR-

a) Classes / instances created directly by the app developer for their app-specific features.
b) Classes created by Hoist to power the toolkit and its components - some of which are entirely "behind the scenes" but others which are generally important and public, but still created and managed by Hoist.

Not to put the cart before the horse, but another way to ask this question is - if I create the most minimal / "empty" Hoist app possible - an appContainer with a default appBar and a "Hello world" panel in the middle - should I see _anything_ in Inspector when xhImpl instances are filtered out? I.e. should an "empty" Hoist app have an empty Inspector, or is there no such thing as an "empty" Hoist app? 😄 

I was leaning a bit more towards the second meaning above in my first iteration, scanning Inspector for anything that wasn't the direct result of some recognizable, app-specific feature and then tracking it back to Hoist code and adding xhImpl flags on either the class level or within a ctor call. It's an interesting (and somewhat surprising!) exercise. If taken all the way, that would lead to an empty Inspector that starts to "fill in" as I begin to add services, models, components to my app.

But I am also very sympathetic to having some Hoist-created instances _not_ marked that way, so when you start up that "hello world" app you _do_ see some key services and app infrastructure models listed - ones that you should know about and which might have relevant data to inspect. E.g. with the services - there we really do seem to have two classes of service - `ConfigService` vs. `AlertBannerService` for instance. So maybe we want to shoot for a world in which case you see the app services you wrote plus a set of key Hoist services.

I think part of the problem is that the use of "impl" implies a bit more privacy than I really want to express with the concept, whereas really it's more of a curation process around where we want to raise the profile or some classes and lower that of others.  (Although I like "xhImpl" and am not actually suggesting we change it, or at least don't have a better suggestion.)

So another thing to think about is, how centered is this new flag on Inspector - should we make choices for it based on what we want to show in that tool, or are there other uses or implications we want it to have?

Let's look together interactively tomorrow, as it really all comes down to these concrete cases and I think looking at a set of those can help us narrow our focus here.

amcclain:
Discussed in depth yesterday with @lbwexler and @TomTirapani - some decisions I plan to implement:

* Focus on the first case / meaning of xhImpl above - i.e. important public Hoist services and models are _not_ xhImpl.
* Research if we can use model context for linked models to infer xhImpl from parent models in the hierarchy when available
* Potentially move xhImpl to a getter / setter pair to support the above.
* Look at additional feature of supporting a name / displayName property on app models to allow a developer to give a model of any class (e.g. a specific `GridModel`) a dev-friendly name for identification in Inspector (e.g. `PositionsGridModel`) and/or allow reliable programatic access to lookup the same via `getActiveModels()` or a new variant.

lbwexler:
👍💪

PR #3155, opened 2022-10-07
Filter chooser model now accepts shorthand number inputs

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/FilterChooserFieldSpec.js
+++ b/cmp/filter/FilterChooserFieldSpec.js
@@ -51,7 +51,7 @@ export class FilterChooserFieldSpec extends BaseFilterFieldSpec {
         super(rest);
 
         this.valueRenderer = valueRenderer;
-        this.valueParser = valueParser ?? (this.fieldType === INT || this.fieldType === NUMBER) ? (input, op) => parseNumber(input) : undefined;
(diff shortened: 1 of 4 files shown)


cmp/filter/FilterChooserFieldSpec.js
```diff
@@ -77,6 +81,10 @@ export class FilterChooserFieldSpec extends BaseFilterFieldSpec {
         let ret;
         if (isFunction(this.valueRenderer)) {
             ret = this.valueRenderer(value, op);
+            if (isValidElement(ret)) {
+                // Prevents [object Object] rendering
+                ret = renderToStaticMarkup(ret);
+            }
```
amcclain, line 87: I was initially concerned that we were going to emit markup somewhere in the stack, given that the rendering pipeline for `FilterChooser` is fully React (as opposed to the issue handled by #3154 where we pipe into ag-grid context menus that require strings).

But that point is rendered moot 😁  by the use of `stripTags` below, which is where we first ended up with [object Object] before this fix.

I thought maybe we could update that to let this method return either a stripped string _or_ an element, but we look to do string-based comparisons downstream so, yeah, we def. need this to return a string

PR #3163, opened 2022-10-14
Typescriptify Buttons, AppBars and Intents

TomTirapani:
+ Convert desktop Button package to TS.
+ Convert desktop AppBar package to TS.
+ Add 'Intent' type. This helped a lot in the Alert Banner admin, where we iterate over intents to create buttons.

Hoist P/R Checklist
-------------------


desktop/cmp/button/AppMenuButton.ts
```diff
@@ -169,16 +165,16 @@ function buildMenuItems({
     ]);
 }
 
-function parseMenuItems(items) {
-    return items
-        .filter(it => !it.omit)
+// Todo: Use MenuItemSpec type rather than Record<string,any>
```
lbwexler, line 168: this is ok, as we discussed -- but given that its just a helper method, we could also just continue to pass it as any, rather than this.

The calculus might be different when we have a `MenuItemLike` type

desktop/cmp/button/Button.ts
```diff
@@ -89,21 +91,18 @@ export const [Button, button] = hoistCmp.withFactory({
...
+    icon?: ReactElement;
+    intent?: Intent,
+    minimal?: boolean,
+    onClick?: () => void,
+    outlined?: boolean,
+    rightIcon?: ReactElement;
+    style?: CSS.Properties,
+    text?: ReactNode,
+    title?: string,
+    tooltip?: string /** Alias for title. */
```
lbwexler, line 107: think js doc needs to actually come *before* the thing  being documented.

PR #3164, opened 2022-10-17
Typescriptify GroupingChooser

TomTirapani:
Hoist P/R Checklist
-------------------


desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -51,7 +50,7 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory({
             ...layoutProps,
             item: popover({
                 isOpen,
-                popoverRef: (v) => model.popoverRef(v), // Workaround for #2272
+                popoverRef: (v) => (model.popoverRef as any)(v), // Workaround for #2272
```
TomTirapani, line 53: Not sure if this is the best way to do this, but Typescript wasn't happy with calling methods on the ref

desktop/cmp/button/Button.ts
```diff
@@ -98,7 +98,7 @@ export interface ButtonProps extends BoxProps {
     icon?: ReactElement;
     intent?: Intent,
     minimal?: boolean,
-    onClick?: () => void,
+    onClick?: (e: any) => void,
```
lbwexler, line 101: just caught this as well, but added it as type :MouseEvent.  Will merge

desktop/cmp/grouping/GroupingChooser.ts
lbwexler: I can look at the type we are passing back for createObservableRef now.  Maybe it needs to be improved.  Note that it is generic, so you can say indicate what type your Ref is going to eventually refer to/receive.

desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -88,41 +87,33 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory({
...
-    ...Button.propTypes,
-
+export interface GroupingChooserProps extends ButtonProps {
     /** Text to represent empty state (i.e. value = null or []) */
-    emptyText: PT.string,
+    emptyText?: string,
 
     /** Primary component model instance. */
-    model: PT.instanceOf(GroupingChooserModel),
+    model?: GroupingChooserModel,
```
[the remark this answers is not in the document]
lbwexler: its a good question --- we could go either way!  I think yes, better to pass it through as a general convention.  type could default to null or void, to indicate that most buttons don't actually have a model

lbwexler:
good catch, fixed.

PR #3172, opened 2022-10-19
Typescriptify FilterChooser

TomTirapani:
Hoist P/R Checklist
-------------------


cmp/filter/FilterChooserFieldSpec.ts
```diff
@@ -14,40 +15,42 @@ import {renderToStaticMarkup} from 'react-dom/server';
 
 const {INT, NUMBER} = FieldType;
 
+export interface FilterChooserFieldSpecConfig extends BaseFilterFieldSpecConfig {
```
lbwexler, line 18: That name is a bit of a mouthful -- not sure if anything to be done

cmp/filter/FilterChooserFieldSpec.ts
```diff
@@ -14,40 +15,42 @@ import {renderToStaticMarkup} from 'react-dom/server';
...
  * Apps should NOT instantiate this class directly. Instead {@see FilterChooserModel.fieldSpecs}
  * for the relevant config to set these options.
+ *
+ * @internal
  */
 export class FilterChooserFieldSpec extends BaseFilterFieldSpec {
 
-    /** @member {FilterChooserValueRendererCb} */
-    valueRenderer;
+    valueRenderer: FilterChooserValueRendererCb;
```
lbwexler, line 44: Seems to me the `Cb` for both of these is redundant.

cmp/filter/FilterChooserModel.ts
```diff
@@ -28,93 +46,95 @@ import {
...
+     * Store or cube View to be used to lookup matching Field-level defaults for `fieldSpecs` and to
+     * provide suggested data values (if so configured) from user input. Defaults to `bind` if provided.
+     */
+    valueSource?: Store|View,
+    /**
+     * Configuration for a filter appropriate to be rendered and managed by FilterChooser, or a function
+     * to produce the same. Note that FilterChooser currently can only edit and create a flat collection
+     * of FieldFilters, to be 'AND'ed together.
+     */
+    initialValue?: FilterChooserFilterLike|(() => FilterChooserFilterLike),
```
lbwexler, line 78: Add recursive function to type definition itself to simplify?

cmp/filter/impl/Option.ts
```diff
@@ -17,8 +34,8 @@ import {FieldType} from '../../../data';
  * Create an option representing a detailed field suggestion
  * @returns {FilterChooserOption}
```
lbwexler, line 35: all these returns are redundant now, have been removing if only thing it specifies is type

cmp/filter/impl/QueryEngine.ts
```diff
@@ -84,7 +89,7 @@ export class QueryEngine {
     //------------------------------------------------------------------------
     // 2) No op yet, so field not fixed -- get field or value matches.
     //------------------------------------------------------------------------
-    openSearching(q) {
+    openSearching(q): FilterChooserOption|FilterChooserOption[] {
```
lbwexler, line 92: Use Some<FilterChooserOption>

cmp/filter/impl/QueryEngine.ts
```diff
@@ -103,13 +108,12 @@ export class QueryEngine {
...
         return isEmpty(ret) ? msgOption(`No matches found for '${q.field}'`) : ret;
-
     }
 
     //--------------------------------------------------------
     // 3) Op is the psuedo 'is' operator and field is set
     //-------------------------------------------------------
-    withIsSearchingOnField(q) {
+    withIsSearchingOnField(q): FilterChooserOption|FilterChooserOption[] {
         const spec = find(this.fieldSpecs, s => caselessEquals(s.displayName, q.field));
```
lbwexler, line 117: use Some<>

data/filter/BaseFilterFieldSpec.ts
```diff
@@ -62,7 +65,7 @@ export abstract class BaseFilterFieldSpec extends HoistBase {
         enableValues,
         forceSelection,
         values
-    }: Partial<BaseFilterFieldSpec>) {
+    }: BaseFilterFieldSpecConfig) {
```
lbwexler, line 68: Thanks for fixing -- was Partial not working?

data/filter/Types.ts
```diff
@@ -48,4 +48,13 @@ export type FilterLike = Filter |
     FieldFilterSpec |
     FunctionFilterSpec |
     FilterTestFn |
-    FilterLike[];
\ No newline at end of file
+    FilterLike[];
+
+/**
```
lbwexler, line 53: Can use Omit<FilterLike, FunctionFilter> for clarity.  Also -- do really need this type? Can we use it inline in FilterChooserCode?

desktop/cmp/filter/FilterChooser.ts
```diff
@@ -16,21 +16,39 @@ import {withDefault} from '@xh/hoist/utils/js';
...
  */
-export const [FilterChooser, filterChooser] = hoistCmp.withFactory({
+export const [FilterChooser, filterChooser] = hoistCmp.withFactory<FilterChooserProps>({
     model: uses(FilterChooserModel),
     className: 'xh-filter-chooser',
     render({model, className, ...props}, ref) {
         const [layoutProps, chooserProps] = splitLayoutProps(props),
             {inputRef, suggestFieldsWhenEmpty, selectOptions, unsupportedFilter, favoritesIsOpen} = model,
-            {autoFocus, enableClear, leftIcon, maxMenuHeight, menuPlacement, menuWidth} = chooserProps,
+            {autoFocus, enableClear, leftIcon, maxMenuHeight, menuPlacement, menuWidth} = chooserProps as FilterChooserProps,
```
lbwexler, line 51: We can parameterize splitLayoutProps() so you don't need to do this -- I will take a look and circle back

cmp/filter/FilterChooserFieldSpec.ts
amcclain: This is probably where projects start using the leading `I`...  I'm still in favor of our trailing `Config` convention, and think these occasionally long names (with e.g. `SpecConfig`) are just the price we pay for it.

cmp/filter/FilterChooserFieldSpec.ts
```diff
@@ -14,40 +15,42 @@ import {renderToStaticMarkup} from 'react-dom/server';
...
+}
+
 /**
  * Filter field specification class for the typeahead `FilterChooser` component. Manages additional
  * configuration related to data values available for suggestion.
  *
  * Apps should NOT instantiate this class directly. Instead {@see FilterChooserModel.fieldSpecs}
  * for the relevant config to set these options.
+ *
+ * @internal
```
amcclain, line 40: Minor point, but I would move the `@internal` annotation to the constructor, as the class overall is (I think) part of the public API - we just don't want devs constructing these directly.

data/filter/Types.ts
TomTirapani: We also have to exclude `FilterTestFn` and `FilterLike[]` and add `FilterChooserFilterLike[]`, so not sure if the Omit syntax is any clearer in the end, but good to know about it! I'll move it to FilterChooserModel as a local type, agree it doesn't need to be here.

Issue #3185, opened 2022-10-24
Div generated by ColumnWidthCalculator can impact the layout of other portal elements


ghsolomon:
The div generated by `ColumnWidthCalculator.getRowEl()` can cause a gap between a `Select` and its menu.  See screenshot for detail:

<img width="1278" alt="Screen Shot 2022-10-24 at 2 50 54 PM" src="https://user-images.githubusercontent.com/88904581/197603161-254c7194-987c-4a08-a949-63fc1439f6dd.png">

https://github.com/xh/hoist-react/blob/f9ec3afa5b8e6cc66465106ac6928ec838f9390a/cmp/grid/impl/ColumnWidthCalculator.js#L327

PR #3187, opened 2022-10-25
Fix to gaps between select inputs and their dropdown menus

amcclain:
+ Caused by grid autosizing - `ColumnWidthCalculator` is supposed to clean up the classes on its top-level DOM elements after sizing. Change back in #3057 to make these methods async defeated the cleanup in the `finally` clause.
+ A rare example where `return await` is required, and not an anti-pattern!

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/ColumnWidthCalculator.js
+++ b/cmp/grid/impl/ColumnWidthCalculator.js
@@ -82,7 +82,7 @@ export class ColumnWidthCalculator {
                     levelMaxes = await Promise.all(levelTasks);
                 return max(levelMaxes);
             } else {
-                return this.calcLevelWidthAsync(gridModel, records, column, options);
+                return await this.calcLevelWidthAsync(gridModel, records, column, options);
             }
         } catch (e) {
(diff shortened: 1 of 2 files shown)


amcclain:
Fixes #3185

cmp/grid/impl/ColumnWidthCalculator.js
```diff
@@ -82,7 +82,7 @@ export class ColumnWidthCalculator {
                     levelMaxes = await Promise.all(levelTasks);
                 return max(levelMaxes);
             } else {
-                return this.calcLevelWidthAsync(gridModel, records, column, options);
+                return await this.calcLevelWidthAsync(gridModel, records, column, options);
```
ghsolomon, line 85: Is the await not also needed for the call to calcLevelWidthAsync on line 80?

ghsolomon: Never mind- I see why not. Shouldn’t have been looking at this on my phone :)

TomTirapani:
Thanks for catching and fixing this guys!

Issue #3148, opened 2022-09-30
DockContainer behavior unintentionally changed by use of ModalSupport


amcclain:
Docked `DockView` elements within a `DockContainer` were intentionally configured (via CSS) to not flex/shrink if they overflow the available width of the container - instead they could overflow off the edge and clip.  

This was a tradeoff to avoid designers needing to handle arbitrary shrinking of docked view contents. See:

https://github.com/xh/hoist-react/blob/aadb6a5d1dc2ed3196fa22fde63d493f0a0a780e/desktop/cmp/dock/impl/Dock.scss#L37

Introduction of ModalSupport wrapper has broken this, causing dock views to shrink vs. clip.

<img width="1407" alt="Screen Shot 2022-09-30 at 12 13 13 PM" src="https://user-images.githubusercontent.com/225806/193342465-00a8a278-f6e0-4c2e-baa2-625aad08bb3e.png">
<img width="1954" alt="Screen Shot 2022-09-30 at 12 15 40 PM" src="https://user-images.githubusercontent.com/225806/193342479-cce3d28e-44e9-47c7-9174-a07bf6e8207c.png">

@pauldelano also reports that the `docked:false` config is no longer respected.

pauldelano:
Also, the modal dialog mode constrains the width of the dialog to 500px instead of to the width of the contained component. Was able to work around this by defining this style.
.bp4-dialog {
  width: auto;
}

PR #3196, opened 2022-11-03
Dock Container fixes

TomTirapani:
+ Fix docked views appearing on the left due to 1 frame with `width: 100%` due to waiting for observable ref. See https://github.com/xh/hoist-react/blob/develop/desktop/cmp/modalsupport/ModalSupport.js#L59-L60
+ Don't flex-shrink docked views.
+ Remove unwanted default `500px` width for modal support dialogs. Width should either be configured, default to `90vw`, or be explicitly unset.
+ Add support for starting modal support in modal mode. Used to fix `DockView.docked = false`.

See issue: https://github.com/xh/hoist-react/issues/3148

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dock/DockViewModel.js
+++ b/desktop/cmp/dock/DockViewModel.js
@@ -111,7 +111,10 @@ export class DockViewModel extends HoistModel {
         this.refreshContextModel = new ManagedRefreshContextModel(this);
 
         this.modalSupportModel = new ModalSupportModel({
-            width: width ?? null, height: height ?? null, canOutsideClickClose: false
+            width: width ?? null,
+            height: height ?? null,
+            startModal: !docked,
+            canOutsideClickClose: false
         });
     }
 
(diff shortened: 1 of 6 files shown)


lbwexler:
+ Looks great to me -- only question in my mind is whether it would be more clear to have the option property be `isModal` rather than `startsModal`.  Sort of like `isOpen`.    Some properties on the interface become mutable when applied to the object, others are immutable.  Think that is consistent with a bunch of things that we do.

+ These files moved around a bit in typescript but I am happy to take care of that

lbwexler (approved):
Happy to have Greg take a look at this as well.

TomTirapani:
Yeah, I originally had `isModal`, but then was concerned that its not clear you're setting an initial state rather than permanent state, somehow forcing it to always be modal. `isOpen` is less problematic for me, perhaps because 'open' is more clearly a mutable state with an obvious opposite state. I think if this was still referred to as "fullscreen" I wouldn't have an issue :)

Happy to go either way, I can't really justify my naming decision here :)

lbwexler:
Made a great name change suggested by greg (matches `defaultCollpased`)

PR #3208, opened 2022-11-30
Update to Highcharts ~10.3.2

Ryanseanlee:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/chart/Chart.ts
+++ b/cmp/chart/Chart.ts
@@ -313,6 +313,7 @@ class ChartLocalModel extends HoistModel {
                 }
             },
             credits: false,
+            accessibility: {enabled:false},
             exporting
         };
     }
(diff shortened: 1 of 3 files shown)


amcclain:
@lbwexler - I was wondering if we really need to force v10, or if we could allow apps to continue on v9.  Not clear to me that this small change is incompatible with v9, and some devs might not wish to upgrade to reduce testing load (or perhaps for licensing reasons).

We would of course want to ensure that we didn't get any confusing or distracting error messages when using v9 and the configs added with this change - @Ryanseanlee could verify that if you think it's worth it to span support across both versions.

lbwexler:
I am fine either way on this one.

Ryanseanlee:
I can test to see if there are any error messages when supporting both and if we are in the clear I can make a PR to revert that change to force v10

PR #3227, opened 2022-12-20
Support Icon aliases

Ryanseanlee:
Hoist P/R Checklist
-------------------


icon/Icon.ts
```diff
@@ -292,6 +281,46 @@ export const Icon = {
     x(p?: IconProps)                {return Icon.icon({...p,  iconName: 'times'})},
     xCircle(p?: IconProps)          {return Icon.icon({...p,  iconName: 'times-circle'})},
 
+    /**
+     * Allows apps to override Hoist choices for given icons.
+     */
+    //Icons for column sorting
+    sortAsc: () => Icon.arrowUp(),
```
lbwexler, line 288: why can't these get props?  Also we are adding an extra wrapping layer.  

I think we should `apply` all these aliases to the object after it is constructed, and then it can be just a super simple alias -- `sortAbsAsc: Icon.arrowToTop`

lbwexler: And now, thinking about it, wonder if it would actually just be more straightforward to have *all* of the functions going directly to the raw Icon.icon call.   

The distinction between `Alias` and the core functions just becoming a bit thin.

So essentially this project is just about adding some new  functions that represent higher-level UI operations.

svc/GridExportService.ts
```diff
@@ -94,14 +94,14 @@ export class GridExportService extends HoistService {
         if (cellCount > withDefault(config.streamingCellThreshold, 100000)) {
             startToast = XH.toast({
                 message: 'Your export is being prepared. Due to its size, formatting will be removed.',
-                icon: Icon.download(),
+                icon: Icon.downloadIcon(),
```
lbwexler, line 97: this seems confusing to me --is the icon suffix here a mistake?

amcclain: Agreed we should not have a trailing `xxxxIcon` on any of these

icon/Icon.ts
amcclain: @Ryanseanlee / @lbwexler - sounds like we might want to check in together and confirm what Lee is saying here to better standardize these aliases and maybe cut down on some boilerplate wrapping that's repeated again and again. Not looking at the code directly at this moment, it's not 100% clear what the cleanest/most elegant approach would be here, but I am sure we could decide on it quickly if we touch base together today.  

Would love to get this wrapped up.

svc/GridExportService.ts
Ryanseanlee: Not sure how that slipped through... will correct it.

icon/Icon.ts
Ryanseanlee: Would love to sync up and get this tightened up.

Ryanseanlee: When applying the aliases to the Icon object after construction, Typescript doesn't like when the object shape changes (i.e., when applying the aliases after construction). One way is to type the `Icon` object as `PlainObject,` although we lose the TS help. Thinking the best solution is to still just apply the aliasing in the construction of the object

amcclain (approved):
This looks good to me, as discussed. 👍 

We can continue to look for opportunities to swap out icon usages within Hoist components for these new "functional" aliases - there are certainly more opportunities to do so - but I think this is a good point to put this down for a bit and merge.

icon/Icon.ts
```diff
@@ -72,225 +72,257 @@ export const Icon = {
...
+    analytics:                      (p?: IconProps) => Icon.mixedChart({...p}),
+    approve:                        (p?: IconProps) => Icon.userCheck({...p}),
+    attachment:                     (p?: IconProps) => Icon.paperclip({...p}),
+    close:                          (p?: IconProps) => Icon.x({...p}),
+    columnMenu:                     (p?: IconProps) => Icon.bars({...p}),
+    danger:                         (p?: IconProps) => Icon.xCircle({...p}),
+    delete:                         (p?: IconProps) => Icon.minusCircle({...p}),
+    detail:                         (p?: IconProps) => Icon.magnifyingGlass({...p}),
+    diff:                           (p?: IconProps) => Icon.arrowRightArrowLeft({...p}),
+    disabled:                       (p?: IconProps) => Icon.slashedCircle(),
```
lbwexler, line 298: I think there is an extra cloning here.  There should be no need to clone all these props here (`{...p}`).

Also, looks like the props not forwarded at all in the case of `disabled`

lbwexler: I'll fix this, eager to get this in

PR #3239, opened 2022-12-28
Add allowBillions flag to fmtQuantity, Fixes #3210

### LINKED ISSUE #3210: Quantity Formatter doesn't have a allowBillions option
We have a client app that has code that is written as if it's expects this formatter to have this option. But in fact when the flag is set to false nothing changes.

Obviously can just write a custom formatter in the app, but I thought this was worth considering in the framework.

febbraiod:
Hoist P/R Checklist
-------------------

- [C] Created Toolbox branch / PR, or determined not required.

### DIFF
--- a/format/FormatNumber.ts
+++ b/format/FormatNumber.ts
@@ -188,7 +188,7 @@ export function fmtBillions(v: number, opts?: NumberFormatOptions): ReactNode  {
  * Render a quantity value, handling highly variable amounts by using units of millions (m) and
  * billions (b) as needed.'
  */
-export function fmtQuantity(v: number, opts?: NumberFormatOptions) {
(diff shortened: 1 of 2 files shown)


format/FormatNumber.ts
```diff
@@ -188,7 +188,7 @@ export function fmtBillions(v: number, opts?: NumberFormatOptions): ReactNode  {
  * Render a quantity value, handling highly variable amounts by using units of millions (m) and
  * billions (b) as needed.'
  */
-export function fmtQuantity(v: number, opts?: NumberFormatOptions) {
```
lbwexler, line 191: I think we should have an  explicit QuantityFormatOptions type that extends NumberFormatOptions and includes a `useMillions` and a `useBillions` flag.

febbraiod: kk on it.

PR #3246, opened 2023-01-06
Remove Column.tooltipElement

### LINKED ISSUE #3198: Deprecate Column.tooltipElement
Right now, the difference between tooltipElement and tooltip on column is really a legacy from when ag-Grid did not handle react.  

The only difference is the styling that get applied.  Can we deprecate `toolipElement`?

saba-mo:
Resolves #3198 
See https://github.com/xh/toolbox/pull/634

Hoist P/R Checklist
-------------------


TomTirapani (approved):
Think this look good and works well.

However, I do wonder whether the effort to cleanly deprecate `tooltipElement` is worth it vs just removing it as a breaking change? I would be in favour in the latter if its not urgent and can wait for a major release.

cmp/grid/columns/Column.ts
```diff
@@ -606,8 +606,7 @@ export class Column {
         this.rendererIsComplex = rendererIsComplex;
         this.highlightOnChange = highlightOnChange;
 
-        this.tooltip = tooltip;
-        this.tooltipElement = tooltipElement;
+        this.tooltip = tooltip ?? tooltipElement;
```
saba-mo, line 609: If we are just deprecating and not removing, should this be tooltipElement ?? tooltip since the former is what currently takes precedence?

TomTirapani: It seemed wrong to me to prioritise a deprecated property over the correct one, even if that meant changing the current behaviour slightly.

I could go either way. You're not expected to provide both (we had a warning before), so I don't think the order matters too much.

saba-mo:
@lbwexler Lee, Tom raises a good question of whether this is worth deprecating or just removing with the next major release.

saba-mo:
@lbwexler This PR is now caught up with develop. Please see Tom's suggestion above that opts for a removal instead of deprecation.

lbwexler:
We have typically been favoring removal for something that is just a simple rename.

saba-mo:
I've updated the PR to remove tooltipElement instead of deprecating it.

@TomTirapani, can you please re-review?

TomTirapani (approved):
Looks great to me!

Happy that we were able to remove it rather then deprecate, much simpler 👍

PR #3251, opened 2023-01-17
Enhance DashViewSpec.omit to support a function returning a boolean

ghsolomon:
Hoist P/R Checklist
-------------------


lbwexler:
I feel like we should change all of the remaining uses of `omit` to take this form at the same time:

`TabModel`
`AboutDialogItem`
`ElementFactory`
`DockViewModel`
`Icon`
`PromiseLinkSpec`

lbwexler:
Wondering, again, if we could come up with a generic type to represent `T| () =>T`?

`Delayable<T>  ?  Dynamic<T>  ? Callable<T> ?  WhenNeeded<T> ? ValOrFn<T>? JIT<T>`

ghsolomon:
> Wondering, again, if we could come up with a generic type to represent `T| () =>T`?
> 
> `Delayable<T> ? Dynamic<T> ? Callable<T> ? WhenNeeded<T> ? ValOrFn<T>? JIT<T>`

`Thunk<T>`?

lbwexler:
Greg and I discussed this.  `thunk` is  a great find, and our usage would actually closely corresponds to the original definition of that word in CS history (see wikipedia).  I did not really understand the origin of the term .

`Thunk` also familiar in some form to redux API users, for what its worth.

Could be `Thunkable<T>` or in practice: `omit: Thunkable<boolean>`

Perhaps too wonky?  Not sure.  Would love to get opinions.

ghsolomon:
> I feel like we should change all of the remaining uses of `omit` to take this form at the same time:
> 
> `TabModel` `AboutDialogItem` `ElementFactory` `DockViewModel` `Icon` `PromiseLinkSpec`

Just updated the PR. Wonder if we should add a utility function to Hoist LangUtils `getOrCall<T>(val: T | (() => T)): T` to dry this up a bit. Could eventually be `getOrCall<T>(val: T | Thunkable<T>): T`

lbwexler:
was thinking along the same lines (and have even written such a thing in the past).  But how about something even more specific of the form `isOmitted(obj) {return isFunction(obj.omit) ? obj.omit() : obj.omit}`

lbwexler:
Greg and I discussed.  Need to just decide whether this promises something is more dynamic than it actually is.  E.g. in cases where it is immediately going to be evaluated anyway,  think the consistency probably overall worth it, but need to think about this a bit.

amcclain:
Great discussion.  I was initially put off by `thunk` due to generalized wonkiness. But after some more thought, I am on board.  `Thunkable` expresses the concept well and matches with our `Awaitable` (or the similar `Promisable` from type-fest).

I did check type-fest and didn't find anything that looked like this already there. Even looked at their issues to see if it had been proposed but didn't see it.

The only other suggestion I came up with is `IsOrReturns<T>` which might get votes for being more immediately understandable, but is admittedly somewhat of a mouthful...

amcclain:
Re. the risk of promising to me more dynamic than it is - this really comes down to a question of observability, is that right? 

(I.e. the risk that a developer will spec a function that returns / relies on an observable, with the expectation that changes to the observable will automatically be picked up and omit / un-omit the target element.)

ghsolomon:
> Re. the risk of promising to me more dynamic than it is - this really comes down to a question of observability, is that right?
> 
> (I.e. the risk that a developer will spec a function that returns / relies on an observable, with the expectation that changes to the observable will automatically be picked up and omit / un-omit the target element.)

That sums it up, yes :)

lbwexler:
I went ahead and added `Thunkable` based on our discussion, and also made a DRY cleanup suggested above. 

I don't think the concern about users making assumptions about when the function is re-called or if it is observable in all cases is worth worrying about.  Its good to be consistent, and its primary purpose is to  allow us to write declarative code whenever possible for various collections/children in code...

utils/impl/IsOmitted.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Copyright © 2022 Extremely Heavy Industries Inc.
+ */
+import { PlainObject } from '@xh/hoist/core/types/Types';
+import {isFunction} from 'lodash';
+
+/**
+ * @internal
+ */
+export function isOmitted(obj: PlainObject): boolean {
```
ghsolomon, line 13: Should the parameter type be more specific?  Also looks like the implementation allows for `obj` to be `null`, in which case I think this would return `undefined` instead of a `boolean `.

Would this be clearer?:

```
export function isOmitted(obj?: {omit?: Thunkable<boolean>}): boolean {
    return isFunction(obj?.omit) ? obj.omit() : !!obj?.omit;
}
```

PR #3263, opened 2023-01-25
Panel class name dropped bug

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/panel/Panel.ts
+++ b/desktop/cmp/panel/Panel.ts
@@ -4,7 +4,7 @@
  *
  * Copyright © 2022 Extremely Heavy Industries Inc.
  */
-import {box, vbox, vframe} from '@xh/hoist/cmp/layout';
+import {box, frame, vbox, vframe} from '@xh/hoist/cmp/layout';
 import {
     BoxProps,
     HoistProps,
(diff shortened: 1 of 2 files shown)


lbwexler:
cr: Tom

desktop/cmp/panel/Panel.ts
```diff
@@ -196,7 +196,13 @@ export const [Panel, panel] = hoistCmp.withFactory<PanelProps>({
 
         // 3) Wrap in modal support if needed
         if (modalSupportModel) {
-            item = modalSupport({model: modalSupportModel, item});
+            item = modalSupport({
+                model: modalSupportModel,
+                item: frame({
+                    item,
+                    className: model.isModal ? className : undefined
+                })
```
amcclain, line 204: This is another layer of DOM nesting we're introducing here?  Looks like this ends up as an anonymous frame without any class or identifying info if we aren't specifying a classname? 

Sorry to be a pain here, it's just that we do sometimes need to target CSS selectors pretty specifically so (changes to / extra) DOM layers do end up mattering. I would like to understand this better and am not 100% on what we're doing here.

TomTirapani: I understand the concerns here (and had the same), but couldn't think of any other option that would avoid having to remount the content.

There are already several DOM layers between the outer wrapper ("xh-panel") and the inner content ("xh-panel__content"), the number and composition of which vary depending on which features you are using (e.g. resizable, modal support, etc). None of our styling assumes that `xh-panel__content` is a direct child of `xh-panel` for this reason.

Given the above, I figured it was unlikely to break styling in client apps and worthwhile compromise to support modal support properly.

amcclain: Thanks for the extra context - appreciate it, and just walked through with Greg also.  We added an additional line to the CL entry to call out the DOM change for the record.

Key takeaway being that the DOM is not really an API and that developers should not assume an overly specific hierarchy. Some seemingly "anonymous" elements might be present for internal implementation purposes that won't be immediately obvious when looking at the DOM itself.

PR #3270, opened 2023-01-30
Add flag to enable tracking of grid exports

### LINKED ISSUE #3242: Add Flag to enable/disable tracking of grid exports
Client Request:  Easy way to turn on/off tracking of export of grid data.

Seems like a flag called "trackExport" (default `false`) could be implemented on the gridModel, and read by the JS [GridExportService](https://github.com/xh/hoist-react/blob/develop/svc/GridExportService.ts) track or not.

saba-mo:
Resolves #3242

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/GridExportService.ts
+++ b/svc/GridExportService.ts
@@ -62,7 +62,7 @@ export class GridExportService extends HoistService {
         throwIf(!(isFunction(columns) || isArray(columns) || ['ALL', 'VISIBLE'].includes(columns)),
             'Invalid columns config - must be "ALL", "VISIBLE", an array of colIds, or a function'
         );
-        throwIf(!(isBoolean(track)),
+        throwIf(!isBoolean(track),
             'Invalid track value - must be either true or false'
         );
 
--- a/svc/GridExportService.ts
+++ b/svc/GridExportService.ts
@@ -49,6 +49,7 @@ export class GridExportService extends HoistService {
             filename = 'export',
             type = 'excelTable',
             columns = 'VISIBLE',
+            track = false,
             timeout = 30 * SECONDS
         }: ExportOptions = {}
     ) {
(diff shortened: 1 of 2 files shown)


svc/GridExportService.ts
```diff
@@ -61,6 +62,9 @@ export class GridExportService extends HoistService {
         throwIf(!(isFunction(columns) || isArray(columns) || ['ALL', 'VISIBLE'].includes(columns)),
             'Invalid columns config - must be "ALL", "VISIBLE", an array of colIds, or a function'
         );
+        throwIf(!(isBoolean(track)),
+            'Invalid track value - must be either true or false'
+        );
```
saba-mo, line 67: I added this for symmetry but maybe i's self-explanatory?

svc/GridExportService.ts
```diff
@@ -141,6 +145,14 @@ export class GridExportService extends HoistService {
...
             await dismissStartToast();
             XH.successToast('Export complete.');
+
+            if (track) {
+                XH.track({
+                    category: 'Navigation',
+                    message: `Downloaded ${filename}${fileExt}`,
+                    data: {rows: rows.length, columns: exportColumns.length}
+                });
+            }
```
saba-mo, line 155: The basic message relies on a helpful filename to know which grid was exported/where it can be found. Do you think it's helpful enough? It might encourage providing specific filenames instead of using the default "export."

svc/GridExportService.ts
```diff
@@ -61,6 +62,9 @@ export class GridExportService extends HoistService {
         throwIf(!(isFunction(columns) || isArray(columns) || ['ALL', 'VISIBLE'].includes(columns)),
             'Invalid columns config - must be "ALL", "VISIBLE", an array of colIds, or a function'
         );
+        throwIf(!isBoolean(track),
```
lbwexler, line 65: not sure we need this runtime checking especially with typescript....  

I don't think users will find this confusing, and the risk is fairly low if they do!

svc/GridExportService.ts
lbwexler: Looks good.  Not 100% sure about the category.  It's not quite "Navigation", but am unsure what, if any, new category of message we might introduce.  "Data"?  "Export"?  "Grids?"

lbwexler:
Saba -- this looks great and I am going to merge it.  We might want to discuss with the group the issue about the correct category before release?

Would be interesting to hear what people think.  Perhaps it does deserve its own category?.

saba-mo:
Good idea. I'll bring it up with the group in the morning when more people are online.

PR #3272, opened 2023-02-02
+ Improve cube caching via updateQuery.

lbwexler:
+ Related cleanups to mark immutable objects as immutable, and avoid extra deep copies when updating queries.

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/cube/Query.ts
+++ b/data/cube/Query.ts
@@ -5,12 +5,12 @@
  * Copyright © 2022 Extremely Heavy Industries Inc.
  */
 
-import {XH} from '@xh/hoist/core';
 import {Filter, parseFilter, StoreRecord} from '@xh/hoist/data';
 import {isEqual, find} from 'lodash';
 import {FilterLike, FilterTestFn} from '../filter/Types';
 import {CubeField} from './CubeField';
 import {Cube} from './Cube';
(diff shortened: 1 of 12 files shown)


data/cube/Query.ts
```diff
@@ -119,19 +129,24 @@ export class Query {
     //------------------------
     // Implementation
     //------------------------
-    private parseFields(names: string[]): CubeField[] {
+    private parseFields(raw: CubeField[]|string[]): CubeField[] {
         const {fields} = this.cube;
-        return names ? fields.filter(f => names.includes(f.name)) : fields;
+        if (!raw) return fields;
+        if (raw[0] instanceof CubeField) return raw as CubeField[];
```
TomTirapani, line 135: Is it safe to assume that because of the first instance was a CubeField, all the others will be? I know that's how we've typed it, but in a non-TS environment devs might easily end up with mixed array. Same below for dimensions.

lbwexler: could alternatively just make the usage of passing a cube field array an undocumented internal feature, and not even include CubeField[] in the documented type....

TomTirapani: I'm sold :)

PR #3273, opened 2023-02-02
Replace CloneDeep with structuredClone

lbwexler:
It actually failed on one usage (in TreeMapModel) where lodash did not, so there is some risk here.
(left deepClone).  Would be good to understand why.

Think this is still worth taking, but need to test it carefully -- being sure to run every usage.

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/differ/DifferModel.ts
+++ b/admin/differ/DifferModel.ts
@@ -13,7 +13,7 @@ import {Icon} from '@xh/hoist/icon';
 import {bindable, makeObservable, observable, action} from '@xh/hoist/mobx';
 import {pluralize} from '@xh/hoist/utils/js';
 import {hbox} from '@xh/hoist/cmp/layout';
-import {cloneDeep, isEqual, isString, isNil, omit, remove, trimEnd} from 'lodash';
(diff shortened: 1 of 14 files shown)


PeteDarinzo:
All uses of `structuredClone` perform as expected except for those that attempt to clone a Highcharts config, found in the Hoist Chart and TreeMap components. The structured clone algorithm does not duplication function objects [more details](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone)

The Chart component has been reverted to use the lodash `cloneDeep` function.

PeteDarinzo (approved):
All uses of `structuredClone` perform as expected except for those that attempt to clone a Highcharts config, found in the Hoist Chart and TreeMap components. Those configs contain function properties that the structured clone algorithm does not duplicate, [more details](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone).

The Chart component has been reverted to use the lodash `cloneDeep` function.

PR #3276, opened 2023-02-03
Fix bug where `TileFrame` would not pass through the keys of its chil…

ghsolomon:
…dren

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/layout/TileFrame.ts
+++ b/cmp/layout/TileFrame.ts
@@ -8,7 +8,7 @@ import {hoistCmp, useLocalModel, HoistModel, BoxProps, HoistProps} from '@xh/hoi
 import {frame, box} from '@xh/hoist/cmp/layout';
 import {useOnResize} from '@xh/hoist/utils/react';
 import {useState, useLayoutEffect} from 'react';
-import {minBy, isEqual, isObject} from 'lodash';
+import {minBy, isEqual} from 'lodash';
 import composeRefs from '@seznam/compose-react-refs';
 import {Children} from 'react';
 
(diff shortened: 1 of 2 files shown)


cmp/layout/TileFrame.ts
```diff
@@ -101,7 +101,8 @@ export const [TileFrame, tileFrame] = hoistCmp.withFactory({
             childrenArr.map((item, idx) => box({
                 style: localModel.getTileStyle(idx),
                 className: 'xh-tile-frame__tile',
-                item
+                item,
+                key: isObject(item) && 'key' in item ? item.key : undefined
             })):
```
cnrudd, line 106: I think an explanation in the form of a comment is needed here, to explain why passing the item keys to the box is necessary.  What issues is this solving/avoiding?

lbwexler:
I am looking at this too.   Think it makes sense, but want to understand
the motivation here

ghsolomon:
Children passed to `TileFrame` were being individually wrapped in container elements which did not themselves inherit the keys of their child components. 

This caused Children with keys to be fully re-rendered unnecessarily, whenever their index changed.  In addition to this negatively impacting performance by triggering extra re-renders, it caused any transient uncontrolled state (such as an uncontrolled `input`) to be lost.

cnrudd:
I think another note explaining the importance of adding 'key' to this component's children is warranted, at the top of the component, inside the existing 'intro' comment.

cnrudd:
> I am looking at this too.   Think it makes sense, but want to understand
> the motivation here
> […](#)

Greg put together some really great examples in his local sandbox of how the component fails without this change.  Might be worth a screenshare.

lbwexler:
No worries -- I am convinced

PR #3279, opened 2023-02-07
Update generic HoistModelClass interface to accept abstract HoistModel's

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/model/HoistModel.ts
+++ b/core/model/HoistModel.ts
@@ -207,6 +207,7 @@ export abstract class HoistModel extends HoistBase implements Loadable {
     }
 }
 
-export interface HoistModelClass<T extends HoistModel> {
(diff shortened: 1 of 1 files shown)


ghsolomon:
Currently, an `abstract` class that extends `HoistModel` cannot be used as a `ModelSelector` because it does not conform to the generic `HoistModelClass` interface which specifies that it must be a `newable` type that returns a `HoistModel` (`abstract` classes are not `newable` since they cannot be constructed).

Here is an example screenshot:
Screen Shot 2023-02-07 at 10 06 02 AM

Proposed change to `HoistModelClass` comes from this article:
https://stackoverflow.com/questions/52349758/how-does-type-constructort-function-prototype-t-apply-to-abstract-c/52350147#52350147

PR #3286, opened 2023-02-09
Ag-Grid -> 29.1.0

### LINKED ISSUE #3285: Upgrade Ag-Grid to 29
https://www.ag-grid.com/changelog/?fixVersion=29.0.0

haynesjm42:
Fixes #3285 

Toolbox PR: https://github.com/xh/toolbox/pull/637

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Changelog
 
-## v56.0.0-SNAPSHOT - unreleased
+## v56.0.0-SNAPSHOT - under development
 
-Coming soon...
+* Coming soon....
+
+## v55.2.1 - 2023-02-24
+
+### 🐞 Bug Fixes
+
+* Fixed issue where a resizable `Panel` splitter could be rendered incorrectly while dragging.
 
 ## v55.2.0 - 2023-02-10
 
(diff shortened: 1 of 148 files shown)


haynesjm42:
Found 1 regression so far - looks like setting the header height to 0 is now interpreted as a falsey value and resulting in the header height using the default height from the theme - added handling to set this via css instead but we might consider reporting this and getting it fixed in 29.1.0 - does not seem like correct behavior

lbwexler:
So 29.1.0 is out with a large number of fixes, including one related to header height.  Would like to try this and see if we can undo John's workaround.

PeteDarinzo:
AG Grid 29.1.0 does fix the issue of 0 header height getting the default height. 

New regression: all rows now get a style attribute `--ag-line-height: -1px;`, causing all grids to appear to empty even though data is loaded. AG Grid 28's attribute was `--ag-line-height: 26px;`, still determining cause.

PeteDarinzo:
Row height regression corrected by removing `height: auto` from the AG Grid outer div in AgGrid stylesheet. This outer div has a default height of 100%, overriding to auto causes collapse of the content spans.

@haynesjm42 agrees that `height: auto` is obsolete and safe to remove. Testing done on all Toolbox desktop and mobile grids confirms that deletion is safe, and regression is fixed.

Detailed analysis:

When setting line height, AG Grid selects the minimum of the height returned from the grid's `getRowHeight` method and a default row height, then subtracts 2. The default row height is calculated from the theme, or returns a default value. The calculation method inserts the html element into the grid and uses `window.getComputedStyle` to determine the height property for the element. In the case of `height: auto`, the calculated value is `1px`. This results in the `--ag-line-height` being set to `-1px`. The height value is correctly calculated in the case of `height: 100%`. The conclusion drawn is that when `height: auto`, the grid html element is in a collapsed state due to no data being loaded, therefore no height required.

lbwexler:
Thanks for the careful analysis and testing here.  We will merge and try to shake this out further before our v56 release.  Easy enough to rollback if needed.

PR #3291, opened 2023-02-13
Give modalSupport the ability to turn off 'enforceFocus' in it's dial…

cnrudd:
…og.  This may be needed by components in the modal that have clickable popovers.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## v56.0.0-SNAPSHOT - unreleased
 
-Coming soon...
+### 🎁 New Features
+
+* `ModalSupportModel` now accepts `enforceFocus` flag (default: `true`).
 
 ## v55.2.0 - 2023-02-10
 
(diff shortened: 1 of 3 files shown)


ghsolomon (approved):

lbwexler (commented):
Want to understand this better.  As I discussed with colin, do we really need to the option to set this?  

Trying to keep ModalSupport as simple as possible, and not leak out its underlying blueprint implementation more than needed.

cnrudd:
> Want to understand this better. As I discussed with colin, do we really need to the option to set this?
> 
> Trying to keep ModalSupport as simple as possible, and not leak out its underlying blueprint implementation more than needed.

enforceFocus is inherited by blueprint dialog from blueprint overlay.   
I found seven instances of enforceFocus: false applied to overlays and popovers in client repos.  So, it has been necessary in some cases before the one I ran into with the RichTextEditor.  

But I think it is hard to tell if making this the default: `enforceFocus: false` in ModalSupport would break existing uses of modalSupport.  Can't tell if the current default `enforceFocus: true` is helpful/required or just unused.

lbwexler:
Closing this.  We don't have a demonstrated need for it, and its not clear to me whether the right solution to the problem colin encountered would be to expose this flag to the API, or simply fix it to false.

PR #3307, opened 2023-03-13
Typescript for Exception code

lbwexler:
Took the additional step to enforce that *all* errors  passed to the framework become HoistExceptions.  So a few runtime changes here that deserve a careful look.  Also some runtime changes required by the move to typescript.

Curious for any feedback, but hopefully this change gets consistent typing in place for our lowest-level code.

Hoist P/R Checklist
-------------------


core/exception/Exception.ts
```diff
@@ -150,19 +161,27 @@ export class Exception {
...
 
     //-----------------------
     // Implementation
     //-----------------------
-    private static createInternal(defaults, override = {}) {
-        if (isString(override)) {
-            override = {message: override};
-        }
-        return Object.assign(new Error(), defaults, override, {isHoistException: true});
+    private static createInternal(
```
ghsolomon, line 173: Not sure if this is any better, but was thinking this could also be made generic to avoid the casting.

```
private static createInternal<E extends HoistException = HoistException>(
    defaults: PlainObject,
    override: PlainObject = {}
): E
```

ghsolomon: Still requires specifying `createInternal<FetchException>(...`.  Would be nice if the argument types could inform the return type, but maybe I'm overthinking this for an internal private method.

lbwexler: Sure seems marginally better -- good idea

core/exception/Exception.ts
```diff
@@ -17,96 +19,103 @@ import {isString} from 'lodash';
...
      * @param fetchOptions - original options the app passed to FetchService.
      * @param e - Error thrown by native fetch
      */
-    static fetchAborted(fetchOptions: FetchOptions, e) {
+    static fetchAborted(fetchOptions: FetchOptions, e: any): FetchException {
         return this.createInternal({
             name: 'Fetch Aborted',
             message: `Fetch request aborted, url: "${fetchOptions.url}"`,
             isRoutine: true,
             isFetchAborted: true,
```
amcclain, line 113: Should we add `isFetchAborted` to the FetchException interface?

amcclain (approved):
Great thanks for taking this on and filling in another chunk of TS coverage!

lbwexler:
great idea -- will add!

core/exception/Exception.ts
lbwexler: Done

PR #3313, opened 2023-03-20
Need to parse values in FieldFilter before comparing to existing values

lbwexler:
So FieldFilters have "unparsed" values that can evaluated at the time of filtering/evaluations.  Its a bit surprising, but there are good reason for it.  

But it can trip us up, as there are a number of places we need to remember to do that before using the value.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * Hoist exceptions have been enhanced and standardized, including new TypeScript types.  The
  `Error.cause` property is now populated for wrapping exceptions.
 
+### 🐞 Bug Fixes
+* Fix grid bug whereby LocalDate filter entered via FilterChooser was causing column filtering
+to fail.
+
 ### 💥 Breaking Changes
 * "Local" Preference support in PreferenceService is no longer supported.  Application should use
   `LocalStorageService` instead. With v56, the `local` flag on any preferences will be ignored, and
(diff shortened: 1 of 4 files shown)


amcclain:
Thanks much for jumping on this. Wish I could better comment on / evaluate, but it is difficult to load all of this up in my mind and I confess I don't really have a good sense of what is going on.

But while testing, I could not trigger any problematic behaviors.  I did however notice that when I select a date from the grid header values list, the value of the resulting filter is a `LocalDate` instance.

To create the screenshot below I did the following, logging the filter value to the console after each step.

* Open grid col filter, uncheck all, check one date (LocalDate value)
* Open grid col filter, check a second date (two LocalDate values)
* Use FilterChooser to add a third date (now all string values)
* Open grid col filter, uncheck third date to remove (back to two LocalDate values)

Screenshot 2023-03-20 at 6 38 37 PM

Everything seemed to behave fine from a user perspective, but this inconsistency seems like it could bite us.  Does this resonate with what you looked at as part of this fix?

lbwexler:
I agree the inconsistency in how these two UIs create filters is a bit disappointing -- it certainly would make more sense for the FilterChooser to also create filters that contain the parsed LocalDate as the value.  It has to do with how FilterChooser needs to pull apart the existing filter and put it back together again with changes.

Having said that, I do think this fix is appropriate and reliably solves your.  The idea that FieldFilter values can be "pre-parsed" any's is indeed baked into the very nature of our filtering.

I am happy to port this as a v55 hotfix as well, if you like.  Let me know if you need it today, or if you want to wait for any other fixes....

PR #3337, opened 2023-04-26
Use scrolling optimization

lbwexler:
+ Use scrolling optimization
+ Tweak to vanilla AgGridComponent row sizing to use appropriate group sizes

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 56.1.1 - 2023-04-27
+
+### ⚙️ Technical
+* Optimize scrolling performance for Grids.
+
+
 ## 56.1.0 - 2023-04-14
 * Add support for new memory management diagnostics provided by hoist-core
  (requires hoist-core 16.1.0 for full operation).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 57.0.0-SNAPSHOT - unreleased
 
 ### ⚙️ Technical
-* Optimize scrolling performance for Grids.
+* Optimize scrolling performance for `Grid` and `DataView`
 
 ## 56.1.0 - 2023-04-14
 * Add support for new memory management diagnostics provided by hoist-core
--- a/cmp/ag-grid/AgGrid.ts
+++ b/cmp/ag-grid/AgGrid.ts
@@ -35,13 +35,18 @@ export interface AgGridProps extends HoistProps<AgGridModel>, GridOptions, Layou
  * via the `model` prop to control additional Hoist customizations.
  *
  * This component complements and contrasts with the primary Hoist `Grid` class, which provides a
- * significantly more managed and opinionated wrapper around ag-Grid and a number of Hoist-specific
+ * significantly more managed and opinionated use of ag-Grid and a number of Hoist-specific
  * extensions and customizations. That fully managed component is expected to cover the majority of
  * use cases within Hoist apps and is recommended as the primary grid class within the toolkit.
  *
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 57.0.0-SNAPSHOT - unreleased
 
+### ⚙️ Technical
+* Optimize scrolling performance for `Grid` and `DataView`
```
amcclain, line 6:
```suggestion
* Improve scrolling performance for `Grid` and `DataView`, reducing transient gaps / whitespace due to rendering delays when scrolling rapidly through a large dataset.
```

cmp/ag-grid/AgGrid.ts
```diff
@@ -141,15 +148,20 @@ class AgGridLocalModel extends HoistModel {
         }
     }
 
+    getRowHeight = ({node}) => {
+        const {sizingMode} = this.model,
+            {groupDisplayType} = this.componentProps,
+            AgGridCmp = AgGrid as any;
+        return node.group && groupDisplayType === 'groupRows'
```
amcclain, line 155: This `groupDisplayType` is not familiar to me - could we can add a small comment to explain this additional check?

cmp/grid/Grid.ts
```diff
@@ -367,20 +351,78 @@ class GridLocalModel extends HoistModel {
         };
     }
 
+    //----------------------
+    // Row Height Management
+    //----------------------
+    @computed
```
amcclain, line 357: Am I correct that we are using `@computed` specifically for its caching benefit?

cmp/grid/Grid.ts
```diff
@@ -367,20 +351,78 @@ class GridLocalModel extends HoistModel {
...
-                    maxBy(visibleCols, 'rowHeight')?.rowHeight
-                ]);
+            track: () => [
+                this.useScrollOptimization,
+                this.calculatedRowHeight,
+                this.calculatedGroupRowHeight
+            ],
+            run: () => {
+                const {agApi} = this.model;
+                if (!agApi) return;
```
amcclain, line 390: Are we checking for `!agApi` b/c this might run before we have it, or b/c it might run after it's been disposed somehow?

If the former I am wondering if we should add to track

cmp/grid/Grid.ts
```diff
@@ -367,20 +351,78 @@ class GridLocalModel extends HoistModel {
...
+                agApi.resetRowHeights();
+                this.applyScrollOptimization();
             },
-            fireImmediately: true
+            debounce: 1
```
amcclain, line 394: This is interesting - the `debounce: 1` - do we have that pattern elsewhere?  Is it clear why we're doing it?

cmp/ag-grid/AgGrid.ts
lbwexler: also not familiar -- just copied from the Grid getRowHeight function. Will look into it post merge

cmp/grid/Grid.ts
lbwexler: yes, exactly

cmp/grid/Grid.ts
lbwexler: agree, that is the normal pattern, but in this case, I think that case where agApi is filled in later, the data reaction will ensure the scroll optimization is applied at least once.

cmp/grid/Grid.ts
lbwexler: copied from the similar reaction in AgGrid component.  But will check and prefer a setting of zero here.

Issue #3338, opened 2023-04-27
FilterChooser and GridFilter should modify date values for certain filter operators by default 


saba-mo:
The current default behavior when filtering a field with fieldType of date is that the time is set to the beginning of the day. So if the user filters by greater than (after) a date, they still see that day's rows included. The default behavior for operators > and <= should be that the time is set to end of day.

We currently do this in the FilterChooserFieldSpec's valueParser in Admin/Activity's tracking and client error tabs. The tracking tab also has column filters enabled, so if the user were to filter by > a timestamp, for example, they would see different results depending on whether they used the column filter or filter chooser.

saba-mo:
Thank you, Tom. I'm looking into it.

saba-mo:
@lbwexler and @amcclain, I'm moving the conversation here to the original ticket. Tom and I discussed the approach suggested in #3339 and the issue we ran into is that FieldFilter takes its value directly off the record before filtering and assumes the fieldType matches that of the field in the store. FieldFilter doesn't currently have access to the filter's fieldType to detect a difference, and we thought implementing that might be a bigger change than we were looking to make. 

So we are now exploring the idea of detecting if the developer has indicated localDate on the fieldSpec while the store's field's fieldType is date, and then providing the valueParser logic of setting the time to end of day for > and <=. Not sure yet how that would look for GridFilter, which does not currently take a valueParser.

PR #3339, opened 2023-04-27
Improve GridFilter's filtering of date fieldType for specific operators

saba-mo:
See #3338 (partially resolves ticket)

I'm not sure if we need to consider changes to the Values tab as well. I don't believe it is used for the date fieldType. 

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/impl/filter/custom/CustomRowModel.ts
+++ b/desktop/cmp/grid/impl/filter/custom/CustomRowModel.ts
@@ -10,6 +10,7 @@ import {ColumnHeaderFilterModel} from '@xh/hoist/desktop/cmp/grid/impl/filter/Co
 import {bindable, computed, makeObservable} from '@xh/hoist/mobx';
 import {isArray, isNil} from 'lodash';
 import {CustomTabModel} from './CustomTabModel';
+import moment from 'moment';
 
 type OperatorOptionValue = 'blank' | 'not blank' | FieldFilterOperator;
 
@@ -28,7 +29,7 @@ export class CustomRowModel extends HoistModel {
     /** FieldFilter config output of this row. */
     @computed.struct
     get value(): FieldFilterSpec {
-        const {field} = this.fieldSpec;
+        const {field, fieldType} = this.fieldSpec;
 
         let op = this.op,
             value = this.inputVal;
@@ -39,6 +40,13 @@ export class CustomRowModel extends HoistModel {
         } else if (op === 'not blank') {
             op = '!=';
             value = null;
+        } else if (fieldType === 'date' && ['>', '<='].includes(op)) {
+            // Special handling for '>' & '<=' date queries.
+            let newVal = moment(value, ['YYYY-MM-DD', 'YYYYMMDD'], true);
+            if (!newVal.isValid()) return null;
+
+            newVal = moment(newVal).endOf('day');
+            value = newVal.toDate();
         } else if (isNil(value)) {
             return null;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### ⚙️ Technical
 * Optimize scrolling performance for `Grid` and `DataView`
+* Improve `GridFilter`'s filtering of date `fieldType` to default to end of day when
+  `FieldFilterOperator` is > or <=.
 
 ## 56.1.0 - 2023-04-14
 * Add support for new memory management diagnostics provided by hoist-core


desktop/cmp/grid/impl/filter/custom/CustomRowModel.ts
```diff
@@ -39,6 +40,13 @@ export class CustomRowModel extends HoistModel {
         } else if (op === 'not blank') {
             op = '!=';
             value = null;
+        } else if (fieldType === 'date' && ['>', '<='].includes(op)) {
+            // Special handling for '>' & '<=' date queries.
+            let newVal = moment(value, ['YYYY-MM-DD', 'YYYYMMDD'], true);
+            if (!newVal.isValid()) return null;
+
+            newVal = moment(newVal).endOf('day');
```
amcclain, line 48: No need to call `moment` again here, as you already have a moment.

TomTirapani:
Hi Saba - thanks for filing the issue and taking it on!

However, unless I'm misunderstanding something, the special handling for `>` & `<=` on Dates comes from `ActivityTrackModel`'s implementation of `FilterChooserModel` - the actual default behaviour for FilterChooser and Grid Filtering is the same. Merging this PR would actually be **introducing** a discrepancy rather than **resolving** one. A better solution would be to make sure `ActivityTrackModel`'s grid filtering uses the same fieldSpecs via `GridModel.filterModel`.

That said, I wonder if the special handling is something we want to always do for Dates in the FieldFilter's testFn? Wouldn't we always want to work with the end of day in those cases?

saba-mo:
Hi Tom, sorry for not providing the full context. Yes, the default behavior is currently the same, and yes, this PR would introduce a discrepancy that would require the developer to customize filterChooser to behave similarly if utilizing both filters. The end goal is to set this as the default behavior for both, but wanted to get in a partial fix as available. 

And yes, we always want to work with the end of day in these two cases. I will look into FieldFilter's testFn as you suggest - maybe that's the place to make the change for both!

amcclain:
Need to be careful that we don't prevent FieldFilter from doing true datetime comparisons.

The grid filter case is somewhat special as it presents a date input w/o time component when building a filter for a date column, so we know that the user can only provide date-level granularity and therefore we want to take this approach to comparisons.  

That said, FieldFilter could check and if its value (i.e. the provided value that should be checked against record-level values) was a `LocalDate` then it could act this way - which is maybe what you were thinking.  Then the grid filtering component could always emit LocalDates unless the developer customized it otherwise. Note that is not what it's currently doing:

https://github.com/xh/hoist-react/blob/b440cae9fdc19c5f43c54b692bdd6ed67531aa84/desktop/cmp/grid/impl/filter/custom/CustomRow.ts#L85

lbwexler:
Like the idea of recognizing LocalDate in the test function, and having the UI emit LocalDate whenever it is not displaying time --- think that really leverages LocalDate appropriately.

saba-mo:
I may be misunderstanding, but if we change the valueType to strictly 'localDate' in CustomRow and then have FieldFilter still call parseFieldValue with the column's fieldType of 'date' (like the activity tracking example), we get a Date back for the prior day. So checking if it was a localDate and, if so, setting it to end of day still leaves us at the end of the prior day.

TomTirapani:
Was thinking about this, and realised the root issue with this tab is that we're trying to filter a Date **as if it was a LocalDate**; I think that's what we should be trying to solve.

A dev should be able to leave the field as `Date` in the Grid and Store, but filter is as a `LocalDate` by setting `FilterChooserModel.fieldSpecs.fieldType` to `localDate` and it should work. We wouldn't need the valueParser or valueRenderer stuff as LocalDate already handles the text input & `>` & `<=` operators properly.

The tricky part would be safely and consistently parsing the Record's date into a localDate before handing off to filtering. This ability to 'cast' before filtering is a new feature, but one that I think makes sense for Date <> LocalDate, and maybe some other fields types.

What do you guys think? Is the ability to override and cast the field type via the FilterFieldSpec something that could be generally useful?

amcclain:
I'm totally open to whatever we think can move this ticket forward and get us to a less-broken default state.

The idea of supporting a `FilterFieldSpecConfig.fieldType` value that does not correspond with the value of the same field within a bound store seems to make sense in the Date > LocalDate case, where we have two types where one can be considered to be a more general version of the other. 

We could determine if there are any other such conversions we would want to support (I am not sure that there are), and I suppose we could validate the provided `fieldType` against a typed store field (if there was one) to determine if the config was valid - i.e. they either need to match, one or both needs to be unspecified, or it's this special store:date filter:localDate case.

I don't see any issue with converting an existing Date to a LocalDate - I don't think we would expect to handle any timezone other than the browser's current timezone, so `LocalDate.from(date)` should be all we need.

We will still need to decide if this is something we apply automatically for date fields - I'd like to avoid leaving the current default grid filter behavior in place as I don't think it's correct as is.

saba-mo:
Closing in favor of #3362

PR #3344, opened 2023-04-28
Added generic type to `XH.promise()`

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### ⚙️ Typescript API Adjustments
 
 * Improved the recommendation for the app `declare` statement within our [TypeScript migration docs](https://github.com/xh/hoist-react/blob/develop/docs/upgrade-to-typescript.md#bootstrapts--service-declarations).
-* Added generic type T to `XH.message()` and `XH.prompt()`, to enable return type of `Promise<T | false>`.
+* Added generic type T to `XH.message()` and `XH.prompt()`, to enable return type of `Promise<T | boolean>`.
 
 ## 56.2.0 - 2023-04-28
 * Expose `margin` property on DashContainerModel.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### ⚙️ Typescript API Adjustments
 
-* Added generic type T to `XH.prompt()`, to enable return type of `Promise<T | false>`.
+* Added generic type T to `XH.message()` and `XH.prompt()`, to enable return type of `Promise<T | false>`.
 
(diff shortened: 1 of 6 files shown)


jskupsik:
Toolbox branch: https://github.com/xh/toolbox/pull/646

core/XH.ts
```diff
@@ -484,7 +484,7 @@ export class XHApi {
      *
      * @returns value of input if user confirms, false if user cancels.
      */
-    prompt(config: MessageSpec): Promise<any> {
```
lbwexler, line 487: 1)  As part of this change I think message should also get a generic return type of `<T | boolean>` 
2) I think that T should default to `unknown`. does that work.

ghsolomon: Had the same thought about defaulting to `unknown`.  Doesn't really provide a whole lot (most likely you'll need to specify the type to actually work with the returned data), but I think it makes sense.

jskupsik: I think that defaulting to unknown encourages the user to use this type feature, which is nice because then it reminds the user to properly handle the `false` case in addition to their specified `T` case.
```
// myMessage is of type <string | false>
const myMessage = await XH.prompt<string>({
    ...
});

if (!myMessage) return;

// myMessage is now considered to be of type <string> by IntelliJ
```

jskupsik: How this looks in action:

Screenshot 2023-05-01 at 11 45 19 AM

Screenshot 2023-05-01 at 11 46 27 AM

amcclain: @jskupsik were you planning to push a commit with this change so we can close this out? Any more decision/discussion items here?

lbwexler: just having one additional thought about this.  Given that the input defaults to a TextInput, seems like it would nice to have the type of this thing actually default to 'string'.   That's what you will have most often

Seems like a case where the typescript to do that would be kind of nasty, not sure if it is worth it. 

Also, we could probably add that intelligence at any time.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 57.0.0-SNAPSHOT - unreleased
 
+### ⚙️ Typescript API Adjustments
+
+* Added generic type T to `XH.message()` and `XH.prompt()`, to enable return type of `Promise<T | false>`.
```
lbwexler, line 7: Actually message I think should be `Promise<T|boolean>`

core/XH.ts
lbwexler: went ahead and added the `unknown` default that we had agreed on

PR #3351, opened 2023-05-03
Fixed ColumnSpec.editor type

jskupsik:
- Now properly types the "params" variable of the given editor function

Hoist P/R Checklist
-------------------


amcclain:
@jskupsik should this still be a WIP draft, or is this ready for a review and (hopefully soon after) merge?

jskupsik:
Ready to be merged.

amcclain:
Thanks much Jakub.  @TomTirapani and @ghsolomon  - if either of you have a chance to review (and merge if looking good), would appreciate it.  Nice to keep the incremental TS improvements flowing.

jskupsik:
Note, ts will complain if you specify: 
```editor: props => selectEditor(props)```
due to missing `inputProps`. I think this is fine because we support the syntax:
```editor: selectEditor```

Alternatively, we can make `EditorProps.inputProps` optional and allow both types to work.

Thoughts?

**Comment by TomTirapani** [comment]
Thanks for putting this PR together, a couple of thoughts:

a) Should we consider adding the `FunctionComponent` type back in (or `FunctionComponent<T extends ColumnEditorFnParams>`)? I understand its not strictly necessary, but one role of Typescript is to help with documentation and I think it might be clearer. Currently in this branch the comments talk about providing either a "Cell editor Component or a function to create one", and at first glance it seems a little confusing.

b) Yes, I think we should make `inputProps` optional to allow both syntaxes above to work. Beyond that it just seems more correct - they are optional if you want to use the defaults.

**Comment by ghsolomon** [comment]
I agree with @TomTirapani but also see a couple other `Grid` types that are essentially aliases for `FunctionComponent`:

- GroupRowRenderer
- ColumnHeaderNameFn

It seems like maybe all of these should use the `FunctionComponent` generic, and we could continue to export our aliases as to not introduce a “breaking change”?

**Comment by jskupsik** [comment]
I have a different way that has it working:
```
    editor?: FunctionComponent<ColumnEditorProps> | ColumnEditorFn;
```
with the trick this time being:
```
export type ColumnEditorFn = (props: ColumnEditorProps, nil?:any) => ReactElement;
```

I was wrong, the issue actually was that ColumnEditorFn was `(a: ColumnEditorProps) => ReactElement` while FunctionComponent<ColumnEditorProps> had `(a: ColumnEditorProps, b: any) => ReactElement`.

Adding a dummy second argument to `ColumnEditorFn`, typed as `?:any` but which will always be `undefined`, allows for IntelliJ to resolve that the first argument (in EITHER mode) will be ColumnEditorProps.

Thoughts on this version?

amcclain:
Checking to see if we should catch up this PR and revisit?  Looks like we've made some changes to the related types on develop, but not quite what we have here?

TomTirapani:
I've caught this up with the latest - our types have moved along a bit in the meantime, reducing this PR to:

+ Typing the `FunctionComponent` form of `Column.editor`
+ Making `inputProps` optional, as discussed above.

I still think these are great changes and worth merging.

Note that as I resolved merge conflicts prettier decided to introduce parentheses to a bunch of unrelated ternaries, bloating the diff with noisy changes - it's only `Types.ts`, `Column.ts` and `EditorProps.ts` that are relevant.

PR #3362, opened 2023-05-12
Improve filtering dates for specific operators

### LINKED ISSUE #3338: FilterChooser and GridFilter should modify date values for certain filter operators by default

saba-mo:
Resolves #3338 

Hoist P/R Checklist
-------------------


saba-mo:
Hi Tom, this PR is far from complete but I wanted you to have something to look at. The issue I'm still facing is that the converted values in the FilterChooserFieldSpec are not the values used for comparison when filtering. I am checking FieldFilter's getTestFn to log the actual filter value and record value being compared. 

I think the getTestFn could be the place to implement this detection/conversion, and it would address both filterChooser and gridFilter. However, line 110 of FieldFilter sets the filter's fieldType based on the store's fieldType for that field, not taking into account what is set in the FieldSpec by the developer. I think if we could detect the correct fieldSpec, we could handle the implementation in FieldFilter by converting the value on line 175, but I'm not sure how big of a change it is/if we'd want to add fieldType when constructing a FieldFilter.

Another issue I anticipate, though am hopefully wrong, is that for GridFilter, in order to provide a fieldType that differs from the column's fieldType, the dev would need to specify the fieldSpec for all filterable columns by providing a filterModel config and no longer be able to simply set filterModel: true on the gridModel. Maybe an argument for making this the default behavior for date filtering.

data/filter/BaseFilterFieldSpec.ts
```diff
@@ -74,7 +75,9 @@ export abstract class BaseFilterFieldSpec extends HoistBase {
         this.forceSelection = forceSelection ?? false;
         this.values = values ?? (this.isBoolFieldType ? [true, false] : null);
         this.hasExplicitValues = !isEmpty(this.values);
-        this.enableValues = this.hasExplicitValues || (enableValues ?? this.isEnumerableByDefault);
+        this.enableValues =
+            !this.isMismatchedDateFieldType &&
+            (this.hasExplicitValues || (enableValues ?? this.isEnumerableByDefault));
```
saba-mo, line 80: I believe we said we don't want to enable the Values tab in this situation, right? It could be the dev's responsibility to set enableValues to false on the fieldSpec, but I think the default behavior of not showing it for dates makes sense.

data/filter/BaseFilterFieldSpec.ts
```diff
@@ -165,6 +189,7 @@ export abstract class BaseFilterFieldSpec extends HoistBase {
     private getDefaultOperators(): FieldFilterOperator[] {
         if (this.isBoolFieldType) return ['='];
         if (this.isCollectionType) return ['includes', 'excludes'];
+        if (this.isMismatchedDateFieldType) return ['>', '>=', '<', '<='];
```
saba-mo, line 192: This ensures the FilterChooser and GridFilter will offer the same operators for this field, which is not currently the case in the ActivityTracking example - only the GridFilter offers equals/does not equal.

desktop/cmp/grid/impl/filter/custom/CustomRow.ts
```diff
@@ -82,7 +82,9 @@ const inputField = hoistCmp.factory<CustomRowModel>(({model}) => {
     } else if (fieldSpec.isDateBasedFieldType) {
         return dateInput({
             ...props,
-            valueType: fieldSpec.fieldType as 'localDate' | 'date'
+            valueType: fieldSpec.isMismatchedDateFieldType
+                ? 'date'
+                : (fieldSpec.fieldType as 'localDate' | 'date')
```
saba-mo, line 87: I noticed that if you picked a date and then closed and reopened the popover, the dateInput wouldn't render the value, due to the difference in fieldTypes.

cmp/filter/impl/Option.ts
```diff
@@ -70,7 +70,12 @@ export function fieldFilterOption({filter, fieldSpec, isExact = false}): FilterC
         displayValue = filter.op === '!=' ? 'not blank' : 'blank';
     } else {
         displayOp = filter.op;
-        fieldType = fieldType === 'tags' ? 'string' : fieldType;
+        fieldType =
+            fieldType === 'tags'
+                ? 'string'
+                : fieldSpec.isMismatchedDateFieldType
+                ? 'date'
+                : fieldType;
```
saba-mo, line 78: If we use localDate as the fieldType, we end up passing a date to the localDate case in Field's parseFieldValue function and would then need to use LocalDate.from(val) instead of LocalDate.get(val).

saba-mo:
@TomTirapani FilterChooser and GridFilter now provide the same behavior as long as the developer indicates 'localDate' as the fieldSpec's fieldType. If we move forward with this approach, I wonder how we can make it more obvious that it's necessary, since I imagine it's a rather common use case. I believe this approach also still allows for true datetime comparisons.

desktop/cmp/grid/impl/filter/custom/CustomRow.ts
TomTirapani: Makes sense - we need to input to be able to read `dates` since that is what the filter is actually using 👍

data/filter/BaseFilterFieldSpec.ts
TomTirapani: It would be great to `enableValues` if we could find a clean way to restrict the values to the unique local dates in the dataset, rather than the full date timestamp. If this proves too complex however perhaps its not worth it.

TomTirapani (changes requested):
This looks and works great.

It's a shame we can't support the full "local date experience"; namely, the Values tab and `=` and `!=` operators. We'll need to assess the added complexity of supporting them and see if its worth it for this convenience feature - I hope we can find a way to support the former without too much work, but I suspect the latter is just not worth the complexity of producing compound filters.

saba-mo:
Regarding enabling the Values tab, I'm still investigating how to compare the correct values, but it looks like in order to get just the unique local dates to appear in the tab, we could check filterDateAsLocalDate in the ValuesTabModel's values getter and return the converted values there. We'd have to check it again in the value column renderer and override the date column's renderer which likely includes a timestamp (as is the case of the activity tracking grid). I can't think of a specific example right now, but it could be problematic to not allow the developer to provide a custom renderer.

TomTirapani (approved):

lbwexler (approved):

lbwexler:
@amcclain  wanted to take a look at this as well -- adding as a reviewer

lbwexler:
@amcclain -- do you want to take a look at this?  Would love to include this one in the current release if possible

amcclain:
@lbwexler yes, will do - caught it up with develop, will look further this morning / early afternoon.

lbwexler:
Superseded by #4439, which resolves #3338 with a fresh, smaller implementation - so closing this out.

The new approach keeps all calendar-day correctness in the comparison layer (`FieldFilter`): a `localDate` value over a `date` field is compared against full-day bounds for both range and equality operators, with no per-spec operator plumbing. It also folds in the admin Roles cleanup. Scoped to client-side filtering; the Values-tab question from #3389 remains out of scope.

Thanks @saba-mo for the original work here - the analysis in this thread directly informed the new fix.

PR #3376, opened 2023-05-23
Add column name output to column uniqueness validation

leogretz2:
Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/js/LangUtils.ts
+++ b/utils/js/LangUtils.ts
@@ -5,7 +5,7 @@
  * Copyright © 2023 Extremely Heavy Industries Inc.
  */
 import {Exception} from '@xh/hoist/core/exception/Exception';
-import _, {
+import {
     forOwn,
     isArray,
     isEmpty,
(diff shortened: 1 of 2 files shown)


utils/js/LangUtils.ts
```diff
@@ -199,7 +199,16 @@ export function ensureNotEmpty(obj: any, exceptionMessage?: string) {
...
  */
 export function ensureUnique(arr: any[], exceptionMessage?: string) {
-    exceptionMessage = exceptionMessage ?? 'All items in the provided array must be unique.';
+    let nonUnique = _(arr)
+        .groupBy()
+        .pickBy(x => x.length > 1)
+        .keys();
+
+    exceptionMessage = nonUnique.size()
+        ? `${nonUnique} colId${
```
lbwexler, line 208: According to the docs (and the file name) this is a general, public utility for *any* uniqueness check.  
We can't just hijack it for this particular usage.  

You could potentially make what you are trying to do more general by making the passed exception message something that is appended to a general "xyz is not unique" message,  but I think most appropriate to leave this method alone and simply take the code in to column where it is needed.

leogretz2: Agreed. I saw it was only used in this place, but it could be used elsewhere later. Do you think it's necessary to specify in the error message that the repeated ID's are colId's or groupId's? It would be a bit of extra labelling in the collectIds method and parsing for the error message. How big/confusing are the grids discussed in this issue?

leogretz2: Determined that is likely unnecessary to specify whether the repeated ID is a colId or a groupId since the user can quickly search through their code with the repeated ID either way.

PR #3377, opened 2023-05-26
Improvements to Exception when empty result or poor json returned

jskupsik:
Hoist P/R Checklist
-------------------


svc/FetchService.ts
```diff
@@ -286,6 +288,11 @@ export class FetchService extends HoistService {
         }
     }
 
+    private async jsonAsync(response: Response) {
```
amcclain, line 291:
```suggestion
    // Cleanly return undefined for an empty response body (i.e. avoid throwing).
    // Successful 2xx HTTP status codes (e.g. 202 ACCEPTED) might or might not return a body.
    private async jsonAsync(response: Response) {
```

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * Fixed layout bug where popovers triggered from a parent `Panel` with `modalSupport` active could
   render beneath that parent's own modal dialog.
 * Fixed broken `CodeInput` copy-to-clipboard feature.
+* Fixed `fetchJson` and related methods throwing on non-204 successful results without contents.
```
amcclain, line 29: 56.5.0 has sailed - need to move up to the snapshot block / catch up with `develop`

amcclain (commented):
Looks good to me, thanks - other than the CHANGELOG fix / catching up with develop.  I suggested a minor inline comment to explain rationale behind `jsonAsync` for future readers.

svc/FetchService.ts
```diff
@@ -167,9 +167,11 @@ export class FetchService extends HoistService {
 
             if (e.isHoistException) throw e;
 
-            // Just two other cases where we expect this to throw -- Typically we get a failed response)
+            // Just three other cases where we expect this to throw -- Typically we get a failed response)
             throw e.name === 'AbortError'
                 ? Exception.fetchAborted(opts, e)
+                : e.name === 'SyntaxError' // May be thrown by JSON.parse()
```
cnrudd, line 173: I like this change.

lbwexler: me too!

svc/FetchService.ts
cnrudd: Not sure about this.  In the past, I've gotten API devs to fix/change the returned status code to 204 when they insisted they could not return content.

I searched a couple http status docs sites and they did not suggest that content was optional for a 202.  Content was required.

Are we being flexible here to be nice?

lbwexler: wow -- this is clever and hardcore.  

Wondering how this differs from the native `.json()` and the way it should be implemented according to the spec.

We could also consider trying .json() first, and falling back on this.

lbwexler: also might call this `safeJsonAsync()` and reference in the comment that we are trying to reproduce a version of the built-in `.json()` that does not throw for empty bodies.....

svc/FetchService.ts
```diff
@@ -78,7 +78,7 @@ export class FetchService extends HoistService {
                 },
                 aborter
             );
-            return this.NO_JSON_RESPONSES.includes(r.status) ? null : r.json();
+            return this.NO_JSON_RESPONSES.includes(r.status) ? null : this.jsonAsync(r);
```
lbwexler, line 81: wonder if we should return `undefined` on line 81  or `null` on line 293-- seems inconsistent as it is

amcclain:
Thanks for having a look, guys!

Re. whether or not HTTP 2xx returns _must_ have a body, from https://httpwg.org/specs/rfc9110.html#status.202 I see

> The 202 response is intentionally noncommittal. Its purpose is to allow a server to accept a request for some other process (perhaps a batch-oriented process that is only run once per day) without requiring that the user agent's connection to the server persist until the process is completed. The representation sent with this response ought to describe the request's current status and point to (or embed) a status monitor that can provide the user with an estimate of when the request will be fulfilled.

Note the word "ought" there - seems distinctly different from "MUST" and implies some amount of wiggle room.

But more than a case of any particular code / spec, Jakub and I didn't see a good reason for us to _force_ an exception on a 2xx response.  Now if you're expecting the API to return something and it doesn't, your app code will presumably throw an NPE or some other exception when it attempts to process the return. But that's really a different case - more similar in my mind to an API returning an unexpected schema, or an array instead of an object, or what-have-you. 

Main point being, when is it useful to have Hoist step in and insist that something is wrong, vs. giving the app code some more flexibility.

(Note counter-case for 204 NO_CONTENT - there the spec doc says "A 204 response is terminated by the end of the header section; it **cannot** contain content or trailers." - emphasis mine on "cannot")

cnrudd:
@amcclain I think the "ought" can be read as recommending what the "representation sent" should do, not recommending that a representation "ought" be sent, rather than not be sent.  There is no question that a representation is sent, just some wiggle room on what it ought to describe.

But, I can see two reasons to go with the more flexible approach proposed in this PR:
1: some 202 "representations" take the form of a 'location' header with a URI at which the requester can check the status or their original request, with no accompanying body. Maybe a "header only, no body" response is a practice some groups use for other 2XXs.
2: we are probably not the only ones who have discussed whether 2XX (excluding 204) responses MUST or MAY include response bodies, so being flexible about it and accommodating both camps seems like a good choice.

cnrudd (commented):
I really like all the Doc improvements!

cnrudd (commented):
I really like all the Doc improvements!

PR #3378, opened 2023-05-30
Fix change handler for mobile's NumberInput, SearchInput and TextInput

### LINKED ISSUE #3353: Mobile textInput's commitOnChange prop failing to commit on change
This may be related to the onsen upgrade in hoist-react 55.1.0. 

The storeFilterField in toolbox's mobile Grids page has commitOnChange set to true. It is working in toolbox (hoist-react 54), but not toolbox-dev (hoist-react 57-SNAPSHOT).

saba-mo:
Resolves #3353 
Addresses a bug from onsenui when we took the 2.12.8 release and the react-onsenui 1.13.2 release.

Hoist P/R Checklist
-------------------


saba-mo:
@TomTirapani, these are the three mobile inputs that required the change. TextArea is working correctly as is.

TomTirapani (approved):
Looks great to me, thanks for fixing!

For additional context, it seems the Onsen upgrade changed these input's `onChange` behaviour to spread them directly onto the underlying `<input>`, presumably not realising the behaviour was different. See the below quote:

```This event is similar to the [oninput](https://www.w3schools.com/jsref/event_oninput.asp) event. The difference is that the oninput event occurs immediately after the value of an element has changed, while onchange occurs when the element loses focus, after the content has been changed. The other difference is that the onchange event also works on <select> elements.```
https://www.w3schools.com/jsref/event_onchange.asp

We need to use the `onInput` event, as we need to check if `commitOnChange` on each value change.

PR #3379, opened 2023-05-31
Fix mobile input refs

TomTirapani:
Mobile input refs were broken for `NumberInput`, `SearchInput`, `Checkbox` and `SwitchInput`.

You can easily see the following warning when navigating mobile toolbox when `reactProdMode: false` (prod mode squelches the warning, but the broken ref remains):

`Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

Note that this change necessitating changing the internal structure of these inputs. Styling has been updated to maintain our practice of applying `className` and `styles` to the outermost component.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## 57.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Fixed a bug where mobile `NumberInput`, `SearchInput`, `Checkbox` and `SwitchInput` could not
+  receive refs.
+
+* ### 💥 Breaking Changes
+
+* The internal DOM structure of mobile `NumberInput`, `SearchInput`, `Checkbox` and `SwitchInput`
+  has changed. You may need to update styling that targets the inner structure of these components.
+
 ## 56.5.0 - 2023-05-26
 
 ### 🎁 New Features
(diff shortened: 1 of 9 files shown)


lbwexler:
Tom -- did we introduce this when we created our wrapper component for onsen?  Should it just support forwardRef?  Would that be a less invasive change, and allow us to still provide refs to the actual underlying controls?

TomTirapani:
Good point -- I had the same thought at first, but dismissed it because it looked like we were already doing work to support forwardRefs in our wrapper:

https://github.com/xh/hoist-react/blob/develop/kit/onsen/index.ts#L60-L69

However, your comment prompted me to look closer, and it turns out that `isForwardRef(rawCmp)` is *never* true! Maybe we're using it wrong - perhaps it needs the element?

Removing that condition resolves the warnings, the problem is that the code is never called. Any reason not to *always* forward refs? e.g.
```
function safeCmp(rawCmp): FunctionComponent {
    return forwardRef((props, ref) => {
        const safeProps = omitBy(props, it => it instanceof HoistModel);
        if (ref) safeProps.ref = ref;
        return createElement(rawCmp, safeProps);
    });
}
```

lbwexler:
Tom -- I think you are right -- it does need the element, not the component definition!  And I think we can and should just always call it that way

TomTirapani:
Great - closing this PR in favour of this better one: https://github.com/xh/hoist-react/pull/3383

PR #3381, opened 2023-05-31
Banner presets

leogretz2:
Hoist P/R Checklist
-------------------


admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -104,6 +110,50 @@ export class AlertBannerModel extends HoistModel {
...
     }
 
+    loadPreset(preset: PlainObject) {
+        this.formModel.setValues({...preset, expires: null});
+    }
+
+    @action
+    addPreset() {
+        const {message, intent, iconName, enableClose} = this.formModel.values;
+        const dateCreated = fmtDateTime(Date.now());
```
amcclain, line 118: We just want to save the timestamp (`Date.now()`) here - that way we can display it in the user's browser timezone when we render it into the component.  

With the approach you have currently, if someone in NY saved a preset it would capture the time in NY - if I then saw that immediately in CA it would appear to be from the future!

Saving the timestamp lets us process it into a formatted date when we _load_ the saved presets - that way everyone will see the times in their expected local zone.

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -82,9 +85,12 @@ export class AlertBannerModel extends HoistModel {
     }
 
     override async doLoadAsync(loadSpec: LoadSpec) {
+        await this.loadPresetsAsync();
         const {formModel} = this;
         if (formModel.isDirty && loadSpec.isAutoRefresh) return;
 
+        formModel.fieldList.map(f => console.log(f.name, ': ', f.value));
```
amcclain: Zap this

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -104,6 +110,50 @@ export class AlertBannerModel extends HoistModel {
...
+        this.formModel.setValues({...preset, expires: null});
+    }
+
+    @action
+    addPreset() {
+        const {message, intent, iconName, enableClose} = this.formModel.values;
+        const dateCreated = fmtDateTime(Date.now());
+        this.savedPresets = sortBy(
+            [...this.savedPresets, {message, intent, iconName, enableClose, dateCreated}],
+            [preset => preset.intent, preset => preset.message]
```
amcclain, line 121: Believe you can use shortcut form of passing the field names as strings instead of arrow functions - i.e. second arg can be `['intent', 'message']` - this is common across many lodash utils

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -104,6 +110,50 @@ export class AlertBannerModel extends HoistModel {
...
+        );
+        this.savePresetsAsync();
+    }
+
+    @action
+    removePreset(preset: PlainObject) {
+        this.savedPresets = without(this.savedPresets, preset);
+        this.savePresetsAsync();
```
amcclain, line 129: Let's add a confirm here, so a user who accidentally clicks the (x) button doesn't immediately delete something that they can't get back.

The best way to do that is to make the overall method async, then at the top `await XH.confirm({...})` - if the value is false, return without modifying the presets. (Sub in an appropriate message / title for the confirm dialog)

admin/tabs/general/alertBanner/AlertBannerPanel.ts
```diff
@@ -188,3 +201,67 @@ const previewPanel = hoistCmp.factory<AlertBannerModel>(({model}) => {
...
+            items.push(...savedPresets.map(preset => presetMenuItem({preset})));
+        }
+
+        items.push(
+            menuDivider(),
+            menuItem({
+                icon: Icon.add({intent: 'success'}),
+                disabled:
+                    !model.formModel.fields.message.value ||
+                    model.isPreset({message, intent, iconName, enableClose}),
```
amcclain, line 226: We shouldn't need to pass this method the current values - it can be a getter on the model - something like `get currentValuesSavedAsPreset() {}` - it would a good use of the `@computed` decorator - we can review what that is and why.

admin/tabs/general/alertBanner/AlertBannerPanel.ts
```diff
@@ -188,3 +201,67 @@ const previewPanel = hoistCmp.factory<AlertBannerModel>(({model}) => {
...
+        const {iconName, intent, message, dateCreated} = preset;
+        return menuItem({
+            icon: iconName
+                ? Icon.icon({iconName, intent, prefix: 'fas', size: 'lg'})
+                : Icon.placeholder({size: 'lg'}),
+            text: vbox({
+                items: [
+                    truncate(message, {length: 50}),
+                    span({
+                        item: `Saved by: ${XH.getUser().username} (${dateCreated})`,
```
amcclain, line 250: This makes it look like the current user saved every preset, regardless of who actually did! 

You want to pass `createdBy: XH.getUsername()` along with the timestamp when saving a new preset.  That will capture the user who is saving that preset and persist it to the shared blob on the server, so everyone else will see that they were the ones who created it.

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -104,6 +108,85 @@ export class AlertBannerModel extends HoistModel {
...
+    async removePreset(preset: PlainObject) {
+        const confirmRemove = await XH.confirm({
+            message: fragment(
+                p(
+                    'You are removing a preset banner configuration. A user may have saved this, see the preset subtext for the creator.'
+                ),
+                p(
+                    'Choose below whether you would like to remove this configuration from the admin list of saved presets.'
+                )
+            ),
```
amcclain, line 140: Looks good on the overall use of the confirm here.  I'd simplify the message a bit - we know that some user definitely _did_ save the preset, so not getting too much from the longer phrasing.  Could go with "Are you sure you wish to this delete this preset?"

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 57.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+
+* Added `presetMenu` to the Admin `AlertBannerPanel` - enable the user to save preset banner message
+  configurations and load them later to be activated or edited.
```
amcclain, line 8: We usually reserve the backtick quoted names for Hoist components and APIs that we expect developers to know / interact with.  Here they are a bit of unnecessary detail - I would focus more on the functionality.  Something like:

> Enhanced the Admin alert banner feature with a new ability to save messages as presets. Useful for standardizing alert or downtime banners, where pre-approved language can be added as a preset and then easily loaded into a banner by members of an application support team."

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -7,13 +7,16 @@
...
+import {HoistModel, LoadSpec, managed, XH, Intent, PlainObject} from '@xh/hoist/core';
 import {dateIs, required} from '@xh/hoist/data';
 import {action, makeObservable, observable} from '@xh/hoist/mobx';
 import {AppModel} from '@xh/hoist/admin/AppModel';
+import {sortBy, without} from 'lodash';
+import {computed} from 'mobx';
 
 export class AlertBannerModel extends HoistModel {
     savedValue;
+    @observable savedPresets: PlainObject[] = [];
```
lbwexler, line 19: observable.ref

admin/tabs/general/alertBanner/AlertBannerModel.ts
lbwexler: also -- along the lines of keeping things simple -- don't think we need to customize the buttons.  This is not *that* dangerous.  Idea is, you need a good reason to go beyond the hoist defaults, which hopefully are good for most cases....

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -104,6 +108,85 @@ export class AlertBannerModel extends HoistModel {
...
+        if (confirmRemove) {
+            this.savedPresets = without(this.savedPresets, preset);
+            this.savePresetsAsync();
+        }
+    }
+
+    @computed
+    get currentValuesSavedAsPreset() {
+        const {message, intent, iconName, enableClose} = this.formModel.values;
+        return this.savedPresets.some(
```
lbwexler, line 162: here is a cool place where you can use a cool bit of  lodash:  `_(this.savedPresets).some({message, iconName, intent, enableClose})`

admin/tabs/general/alertBanner/AlertBannerPanel.ts
```diff
@@ -155,6 +160,14 @@ const formPanel = hoistCmp.factory<AlertBannerModel>(({model}) => {
             ]
```
lbwexler, line 160: Have we ensured that all of these buttons/menus are not available to HOIST_ADMIN_READER?

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -194,14 +277,18 @@ export class AlertBannerModel extends HoistModel {
...
-            params: {value: JSON.stringify(value)}
-        }).track({
-            category: 'Audit',
-            message: 'Updated Alert Banner',
-            data: {active, message, intent, iconName, enableClose}
-        });
+        // {active, message, intent, iconName, enableClose}
+        // DEBUG
+        console.log('in save internal async');
+        this.saveBannerSpecAsync(value);
```
lbwexler, line 283: this is *exactly* an example of where you do need to await per our discussion

lbwexler: also, please be more careful not to leave the logging/commenting in the pushed code for PR

admin/tabs/general/alertBanner/AlertBannerPanel.ts
lbwexler: We should be able to globally disable the entire FormModel.  Think that is probably a simpler way to do this.

PR #3386, opened 2023-06-06
Idle duration Flag added to IdleService

### LINKED ISSUE #3385: Add observable property to Idleservice: idleFor
The  observable `idleFor` property would be an int, and have the time in minutes since activity was last detected on the app.

This property could be used by apps who wanted to implement their own "lighter touch" "sleep" profile.

For instance, rather than shutdown all services, and force users to re-awaken the app by reloading it (thereby losing any current unsaved state), 
and app could, when idleFor > 30mn were detected, stop making the several recurring data calls that may fetch a lot of data.

When activity was detected again (idleFor would == 0) those recurring data calls would be resumed, without reloading the app.

cnrudd:
This resolves #3385.

Hoist P/R Checklist
-------------------

- [x] Updated doc comments: docs updated on IdleService.ts
- [x] Reviewed and tested on Mobile: Should run fine on Mobile. :)

### DIFF
--- a/core/XH.ts
+++ b/core/XH.ts
@@ -28,7 +28,8 @@ import {
     makeObservable,
     observable,
     reaction as mobxReaction,
-    when as mobxWhen
+    when as mobxWhen,
+    runInAction
 } from '@xh/hoist/mobx';
 import {never, wait} from '@xh/hoist/promise';
 import {
@@ -53,7 +54,15 @@ import {
 import {Timer} from '@xh/hoist/utils/async';
 import {MINUTES} from '@xh/hoist/utils/datetime';
 import {checkMinVersion, getClientDeviceInfo, throwIf} from '@xh/hoist/utils/js';
-import {camelCase, compact, flatten, isBoolean, isString, uniqueId} from 'lodash';
+import {
+    camelCase,
+    compact,
+    debounce as lodashDebounce,
+    flatten,
+    isBoolean,
+    isString,
+    uniqueId
+} from 'lodash';
 import {createRoot} from 'react-dom/client';
 import parser from 'ua-parser-js';
 import {AppContainerModel} from '../appcontainer/AppContainerModel';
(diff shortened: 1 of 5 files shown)


svc/IdleService.ts
```diff
@@ -15,20 +16,30 @@ import {MINUTES, olderThan} from '@xh/hoist/utils/datetime';
...
+ * it will cause users to lose unsaved state) can set up their own "lighter touch" shutdown of resource
+ * intensive recurring data calls at whatever threshold of `idleFor` needed.  Apps can resume those
+ * calls when the `idleFor` value becomes 0 as activity is detected.
  */
 export class IdleService extends HoistService {
     override xhImpl = true;
 
     static instance: IdleService;
 
+    @observable idleFor: number = 0;
```
amcclain, line 33: Need to add some sense of the units in here - looks like this is in minutes? 

If we keep it at minutes, we should consider if we want it reporting 1min when idle for 30 secs, and how often we need to run the timer, how to make it clear what the contract is.

core/XH.ts
```diff
@@ -255,7 +255,7 @@ export class XHApi {
     @observable
     appState: AppState = 'PRE_AUTH';
 
-    /** milliseconds since user activity / interaction was last detected. */
+    /** milliseconds timestamp at moment of last user activity / interaction was last detected. */
```
amcclain, line 258: To maybe get right to the bigger question, should we consider doing something like making the property on XH observable and update it with a debounced handler?  Would that be a simpler way to make the existing property more useful - an app could e.g. add an `@computed` to report when the diff between then and now crosses their threshold - without the extra complexity of another timer in IdleService?

It seems a bit confusing to have IdleService officially disabled, and yet reporting / maintaining an `idleFor` value

lbwexler:
Closing -- we learned a lot here, but decided that we could actually do what we wanted in our client with a simple check of the existing public API `XH.lastActivityMs` on a timer.

Also learned alot about the new pagelifecycle model, and Chrome's new memory saving features.

We'll hold off before anything to hoist.

PR #3389, opened 2023-06-08
Enable values tab when filtering dates as localDates

saba-mo:
Hoist P/R Checklist
-------------------

### DIFF
--- a/data/filter/BaseFilterFieldSpec.ts
+++ b/data/filter/BaseFilterFieldSpec.ts
@@ -199,7 +199,6 @@ export abstract class BaseFilterFieldSpec extends HoistBase {
     }
 
     private get isEnumerableByDefault(): boolean {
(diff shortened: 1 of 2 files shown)


TomTirapani:
Hi Saba,

Apologies for the delay in looking at this.

I started playing around with it, and then realised something I should have remembered much earlier - the values tab produces an `includes` or `excludes` FieldFilter, which is essentially an multi-value variant of `=` and `!=`! 😮‍💨 

Our approach of using Dates all the way down (which is fine for `'>', '>=', '<', '<='` filters) means we run into the same problem as supporting the `=` and `!=` operators - for a Date to "equal" a LocalDate, we'll need to produce a compound `between` filter for each value.

So, unfortunately, I don't think there's a clean way to support the Values Tab with our current approach.

PR #3404, opened 2023-06-26
New service: PageLifecycleService

cnrudd:
Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry
- [x] Updated doc comments / prop-types: not required
- [x] Reviewed and tested on Mobile: not yet tested on Mobile, do not yet know for sure that the events behave the same on Mobile, but I know that the events exist on Mobile.
- [x] Created Toolbox branch: no


amcclain (approved):
Looks good to me, thanks!

I did some formatting / comment updates + minor tweak to spec `@computed` on the boolean getters - not sure how necessary / helpful that is, but seemed reasonable.

lbwexler:
Am curious as to the support for this API outside.  Also would like to use this in our own implementation of PrefService, which adds a listener to `onbeforeunload`.  According to the document here, there are good reasons to use PageLifecycleService instead.

lbwexler:
Trying to take seriously the recommendations here. 

https://developer.chrome.com/blog/page-lifecycle-api/#developer-recommendations-for-each-state


This would suggest that we should 

a) persist our unsaved preferences when we become not *active*.  Think the reasoning is that it is better to get the persistence done early, when we know no more changes are pending, and stop using the (problematic?) `onbeforeunload` handler.

b) make the app idle when it goes to 'frozen'?   Would need to understand "frozen" a bit better, but it seems like it would be appropriate.  Need to be careful with certain clients that don't want the app to sleep.

PR #3406, opened 2023-06-30
New option TrackOptions.logData

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/TrackService.ts
+++ b/svc/TrackService.ts
@@ -26,7 +26,8 @@ export class TrackService extends HoistService {
             maxRows: {
                 default: 10000,
                 options: [1000, 5000, 10000, 25000]
-            }
+            },
+            logData: false
         });
     }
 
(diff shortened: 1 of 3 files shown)


svc/TrackService.ts
```diff
@@ -90,15 +91,7 @@ export class TrackService extends HoistService {
...
             if (options.category) params.category = options.category;
             if (options.data) params.data = JSON.stringify(options.data);
-
-            const {maxDataLength} = this.conf;
-            if (params.data?.length > maxDataLength) {
-                console.warn(
-                    `[TrackService] | Track log includes ${params.data.length} chars of JSON data | exceeds limit of ${maxDataLength} | data will not be persisted`,
-                    options.data
-                );
-                params.data = null;
```
amcclain, line 100: This avoided POSTing potentially very large blobs to the server (in the event of rogue track data looping in some huge object), in addition to providing some client-side visibility into this limitation. Seems like it would still be worth it for those reasons, especially the first one WDYT?

lbwexler: Yea, I figured as much, but didn't think it worth the complexity.  We don't generally do anything to protect users from creating giant requests.

Happy to restore -- still managed to get rid of some duplicate checks on the server-side.

lbwexler: Yea, I figured as much, but didn't think it worth the complexity.  We don't generally do anything to protect users from creating giant requests.

Happy to restore -- still managed to get rid of some duplicate checks on the server-side.

PR #3416, opened 2023-07-14
expand type of emptyText to allow all reactNodes

connor15mcc:
Redefine the type of `emptyText`, expanding from `string`s to `ReactNode`, allowing the embedded of more complex fallback messages. Mirrors the interface set in GridConfig.

This is a dependency of toolbox PR

-------------------

### DIFF
--- a/cmp/dataview/DataViewModel.ts
+++ b/cmp/dataview/DataViewModel.ts
@@ -28,6 +28,7 @@ import {
 import {bindable, makeObservable} from '@xh/hoist/mobx';
 import {throwIf} from '@xh/hoist/utils/js';
 import {isFunction, isNumber} from 'lodash';
+import {ReactNode} from 'react';
 
 /**
  * Configuration for a DataView.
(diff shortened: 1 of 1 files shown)


cnrudd (changes requested):
This is the logical change for sure.  But I wonder why emptyText needs to be listed in DataViewConfig at all, since it will now match what is in GridConfig, which DataViewConfig extends.

Can we just remove emptyText from DataViewConfig?

connor15mcc:
Good call out. Looks like there was a lot duplicated between the two. Wonder if it was an artifact from the move to Typescript -- I removed all properties from `DataViewConfig` that are already established in `GridConfig`.

lbwexler:
Well, I actually think the very first comment here

>  // TODO: Accept grid keys without publicizing them?
   
speaks to how we ended up here.  It looks like we don't really *want* to be promising the entire GridConfig interface, but were somehow forced to do it, to support the accepting of the rest args.  Were we worried about backward compatibility?

lbwexler:
Connor and I discussed -- we are going to rolli this back to *just* changing the signature of empty text to `ReactNode` A separate ticket is forthcoming to update the DataViewConfig so that instead of extending GridConfig, it contains an additional property of `gridOptions` which will contain any direct GridConfig that the applications wishes to make.

PR #3420, opened 2023-07-18
Update `GroupingChooser` with local model for transient UI state

### LINKED ISSUE #3398: GroupingChooser should keep transient UI state separate from the provided model
Currently `GroupingChooserModel` contains transient UI state, like whether the editor is open or closed. This results in the undesired behavior where if you have multiple `groupingChooser` components rendered bound to the same model, clicking on one chooser will open the editor in all choosers.

It seems like we should keep transient UI state in a local model for each component so instances like this behave as expected. 

I ran into this while using the panel modal support to show a larger version of a widget, so I have 2 grouping choosers rendered sharing the same model.

haynesjm42:
Fixes #3398 

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grouping/GroupingChooser.ts
+++ b/desktop/cmp/grouping/GroupingChooser.ts
@@ -17,7 +17,7 @@ import {menu, menuDivider, menuItem, popover} from '@xh/hoist/kit/blueprint';
 import {dragDropContext, draggable, droppable} from '@xh/hoist/kit/react-beautiful-dnd';
 import {splitLayoutProps} from '@xh/hoist/utils/react';
 import classNames from 'classnames';
-import {compact, isEmpty, sortBy} from 'lodash';
+import {isEmpty} from 'lodash';
 import './GroupingChooser.scss';
 
 export interface GroupingChooserProps extends ButtonProps<GroupingChooserModel> {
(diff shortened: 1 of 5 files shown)


desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -39,6 +40,126 @@ export interface GroupingChooserProps extends ButtonProps<GroupingChooserModel>
     styleButtonAsInput?: boolean;
 }
 
+class GroupingChooserLocalModel extends HoistModel {
```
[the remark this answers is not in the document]
lbwexler: you can put it in an /impl directory, and mark it as @internal.  We do this all the time!

desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -107,35 +230,53 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory<GroupingC
...
 
-const dimensionList = hoistCmp.factory<GroupingChooserModel>({
-    render({model, emptyText}) {
-        if (isEmpty(model.pendingValue)) {
+interface DimensionListProps extends WithLocalModelProps {
+    emptyText: string;
+}
+
+const dimensionList = hoistCmp.factory<DimensionListProps>({
+    render({model, emptyText, impl}) {
```
lbwexler, line 278: same as above

desktop/cmp/grouping/GroupingChooser.ts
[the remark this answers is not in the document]
haynesjm42: Thanks for the tip, I knew we must have a pattern for this (and do this all the time in apps). Was hoping for something that would hide the class to app code, doesn't seem like there is any enforcement of `@internal` . Looks like there are lots of proposals out there for this kind of thing, hopefully something makes it in eventually.

Looks like there is a way to strip these from d.ts output - https://www.typescriptlang.org/tsconfig#stripInternal - though since we are full typescript I don't think this does anything.

desktop/cmp/grouping/GroupingChooser.ts
haynesjm42: Suggesting to move this into a shared impl directory? Can't really do that unfortunately as it's child components use platform-specific components

lbwexler: discussed with John -- there was a misunderstanding -- just mean that we should be passing the state down to the local components as `dimensionList({imodel: impl})`.   That way they have typed access to everything.

PR #3421, opened 2023-07-18
DataViewConfig no longer extends from GridConfig.

### LINKED ISSUE #3419: DataViewConfig should not extend GridConfig
Update the DataViewConfig so that instead of extending GridConfig, it contains an additional property of gridOptions which will contain any direct GridConfig that the application wishes/dares to make.

This will be a breaking change that will need to be called out in the changelog, but should be rare and easily resolved in applications.

connor15mcc:
Resolves  #3419.

DataViewConfig no longer extends from GridConfig. Instead, any additional configs can be set through `gridOptions: GridOptions`, which is an interface of all `GridConfig` properties, less those specified in `DataViewConfig` (ie, those that were previously extended). 

This is a *breaking* change and all applications that rely on passing such properties will need to update their usage of DataViewConfig. An example of such a change is provided:

Before:
```ts
dataViewModel = new DataViewModel({
    itemHeight: 80,
    onKeyDown: () => {...}
});
```

After:
```ts
dataViewModel = new DataViewModel({
    itemHeight: 80,
    gridOptions: {
        onKeyDown: () => {...}
    }
});
```

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * `DataViewConfig` no longer directly supports the passing of `GridConfig` parameters.
   Instead, all direct `GridConfig` toggles must be passed through the new `gridOptions`
-  parameter. See PR ![3421](https://github.com/xh/hoist-react/pull/3421)
+  parameter.
 
 ## 58.0.1 - 2023-07-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 59.0.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes
+
+* `DataViewConfig` no longer directly supports the passing of `GridConfig` parameters.
+  Instead, all direct `GridConfig` toggles must be passed through the new `gridOptions`
+  parameter. See PR ![3421](https://github.com/xh/hoist-react/pull/3421)
+
 ## 58.0.1 - 2023-07-13
 
 ### 🐞 Bug Fixes
--- a/cmp/dataview/DataViewModel.ts
+++ b/cmp/dataview/DataViewModel.ts
@@ -33,12 +33,8 @@ import {ReactNode} from 'react';
 /**
  * Configuration for a DataView.
  *
- *  Additional properties not specified here will be passed to the underlying
- *  GridModel. Note this is for advanced usage - not all configs supported, and many will
- *  override DataView defaults in ways that will break this component.
  */
-export interface DataViewConfig extends GridConfig {
-    // TODO: Accept grid keys without publicizing them?
+export interface DataViewConfig {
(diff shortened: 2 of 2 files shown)


amcclain:
Am sure this is a good idea, but just for the record and in case we get any questions re the breaking change, what was the motivator for this change?

cmp/dataview/DataViewModel.ts
```diff
@@ -262,3 +265,11 @@ export class DataViewModel extends HoistModel {
...
     }
 }
+
+/**
+ * the `GridOptions` interface contains all `GridConfig` properties, less those
+ * specified in the above `DataViewConfig`, allowing an "escape hatch" for
+ * developers to alter the underlying `GridConfig`. Note that such alteration is
+ * not guaranteed by the DataView and may have unintended consequences.
+ */
+export interface GridOptions extends Omit<GridConfig, keyof DataViewConfig> {}
```
lbwexler, line 275: nice use of typescript here!  
Not sure the `Omit<>` truly needed (its already documented so well), but why not!
 
BUT, I would probably just inline this to avoid needing to add a new identifier and doubling the documentation

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## 59.0.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes
+
+* `DataViewConfig` no longer directly supports the passing of `GridConfig` parameters.
+  Instead, all direct `GridConfig` toggles must be passed through the new `gridOptions`
+  parameter. See PR ![3421](https://github.com/xh/hoist-react/pull/3421)
+
```
lbwexler, line 10: We don't typically link to PRS in the changelog -- we could consider changing that convention, but lets leave out for now.

cmp/dataview/DataViewModel.ts
```diff
@@ -105,6 +101,13 @@ export interface DataViewConfig extends GridConfig {
      * the row's data. (Note that this may be null - e.g. for clicks on full-width group rows.)
      */
     onRowDoubleClicked?: (e: any) => void;
+
+    /**
+     * "escape hatch" object to pass directly to GridModel. Note these options may be used
```
lbwexler, line 106: Capitalize start of jsdoc comments

connor15mcc:
> Am sure this is a good idea, but just for the record and in case we get any questions re the breaking change, what was the motivator for this change?

As I understand it, while `DataView` (previously) extends `GridConfig`, we redefine a number of properties in order to make clear those which we explicitly support vs others that exist for historical control over the underlying GridModel.

Separating the latter type of property allows us to more clearly designate them as unsupported/experimental. This arose from #3416, where we noticed an inconsistency between the extended interfaces.

PR #3430, opened 2023-07-26
Fixed fmtNumber returning '-0' when rounding small negative numbers

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/format/FormatNumber.ts
+++ b/format/FormatNumber.ts
@@ -53,8 +53,8 @@ export interface NumberFormatOptions extends Omit<FormatOptions<number>, 'toolti
     /** True to prepend positive numbers with a '+'. */
     withPlusSign?: boolean;
 
-    /** True to not display a signed zero -0.0 (and +0.0 withPlusSign) in cases where a small number is rounded to zero. */
-    strictZeroUnsigned?: boolean;
+    /** False to not display a signed zero (ie -0.0) in cases where a small number is rounded to zero. */
+    strictZero?: boolean;
 
     /** True to prepend an up / down arrow. */
     withSignGlyph?: boolean;
@@ -127,7 +127,7 @@ export function fmtNumber(v: number, opts?: NumberFormatOptions): ReactNode {
         ledger = false,
         forceLedgerAlign = true,
         withPlusSign = false,
-        strictZeroUnsigned = false,
+        strictZero = true,
         withSignGlyph = false,
         withCommas = true,
         omitFourDigitComma = false,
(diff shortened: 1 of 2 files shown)


jskupsik:
Before: 
```js
fmtNumber(-0.0004, {precision: 2}) === '-0.00'
```

After:
```js
fmtNumber(-0.0004, {precision: 2}) === '0.00'
```

Ryanseanlee (approved):

lbwexler:
The implementation of this seems good, although I might be inclined to keep the initial early test for a true zero, for clarity/efficiency.

My bigger question is just to be sure that this is what we want.  I *think* its correct and appropriate, but to be clear we are throwing away a bit of information as to the fact that it is a diminuous amount with a real sign

lbwexler:
@jskupsik -- we discussed this and came up with the following:

1) Would like this behavior to be goverened by a new flag.  Not sure about name.  `displayRoundToZeroAsZero`?  `strictZeroDisplay`?  `zeroDisplayMode`.
2) In looking at the implementation, it seems like the logic will not prevent a ledger formatted number of the form `(0.00)`, if there is no zero display.  That seems inconsistent. Can we fix?

jskupsik:
fmtNumber(-0.0004, {precision:2})
'-0.00'
fmtNumber(-0.0004, {precision:2, strictZeroUnsigned: true})
'0.00'

fmtNumber(0.0004, {precision:2, withPlusSign: true})
'+0.00'
fmtNumber(0.0004, {precision:2, withPlusSign: true, strictZeroUnsigned: true})
'0.00'

fmtNumber(-0.0004, {precision:2, ledger: true})
-> children: (3) ['(', '0.00', ')']

fmtNumber(-0.0004, {precision:2, ledger: true, strictZeroUnsigned: true})
-> children: (2) ['0.00', {…}]

format/FormatNumber.ts
```diff
@@ -53,6 +53,9 @@ export interface NumberFormatOptions extends Omit<FormatOptions<number>, 'toolti
     /** True to prepend positive numbers with a '+'. */
     withPlusSign?: boolean;
 
+    /** True to not display a signed zero -0.0 (and +0.0 withPlusSign) in cases where a small number is rounded to zero. */
+    strictZeroUnsigned?: boolean;
+
```
lbwexler, line 58: how about something like `strictZero`?  Captures the fact that this relates to interpretation of numbers rounding to zero generally, both for the purposes of zeroDisplay and the signing.   But I think the name should imply the reverse sense.  That is 'strict' means that only true zero gets treated like zero

lbwexler: other than that, implementation looks good!

jskupsik:
Ran the same 6 tests as above.

PR #3435, opened 2023-08-01
Grid with `hideHeaders:true` omits extraneous header border

Ryanseanlee:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -40,6 +40,11 @@
     cursor: pointer;
     width: 100%;
 
+    &--hidden .ag-header {
+      border-bottom-width: 0;
+      min-height: 0 !important;
+    }
+
     span {
       flex: 1;
       overflow: hidden;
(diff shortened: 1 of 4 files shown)


cmp/grid/Grid.scss
```diff
@@ -40,6 +40,11 @@
     cursor: pointer;
     width: 100%;
 
+    &--hidden .ag-header {
+      border-bottom-width: 0;
+      min-height: 0 !important;
+    }
```
amcclain, line 46: I'm assuming the `min-height` rule is doing something, since you've added it in here, but was a bit surprised as we set the header height to zero in code using the `headerHeight` getting in `AgGrid.ts`. What were you seeing without this rule?  Was there a 1px gap or something like that?

amcclain (commented):
Thanks for looking at this - one comment + one question below - let's get it merged in today.

cmp/grid/Grid.scss
Ryanseanlee: image
Right Without the min-height there is a 1px gap

PR #3436, opened 2023-08-02
React select5x upgrade

leogretz2:
Hoist P/R Checklist
-------------------

- [ x ] Caught up with `develop` branch as of last change.
- [ x ] Added CHANGELOG entry, or determined not required.
- [ x ] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.
- [ x ] Updated doc comments / prop-types, or determined not required.
- [ x ] Reviewed and tested on Mobile, or determined not required.
- [ x ] Created Toolbox branch / PR, or determined not required.

Comments:

Tested in the Toolbox Admin 'Tests > Select' suite and all selects function the same as in develop.


desktop/cmp/input/Select.ts
```diff
@@ -52,6 +53,12 @@ export interface SelectProps extends HoistProps, HoistInputProps, LayoutProps {
     /** True to focus the control on render. */
     autoFocus?: boolean;
 
+    /**
+     * Takes an object with keys for the various react-select inner components and values
+     * with a callback function to return the desired class.
+     */
+    classNames?: ClassNamesConfig;
```
lbwexler, line 60: If we are going to reveal this new property that is so closely tied to react-select, think it should be called `rsClassNames`.  Also -- documentation should be simpler when we know have a type to do the heavy lifting.  Documentation should delegate simply to the react-select documentation.

lbwexler: more importantly, think this should just be handled by the `rsProps` sub-property?  Was there some reason to promote this to a top-level property on our component?

amcclain: Hi - thanks for reviewing.  @haynesjm42 had called out that new prop as useful in the original issue - indicated it would help clean up some very specific app styles. Given that, we figured it was worth highlighting as an "officially approved" API while continuing to designate `rsOptions` as a "use at your own risk" escape hatch.

I think the comment landed in an OK place - helps you get some orientation to what you need to provide w/o having to inspect the type that closely - it's [a bit gnarly](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/styles.ts#L122)... We thought about linking to the docs page that lists the specific component keys, but then didn't want to make it longer.

lbwexler: temporarily holding off on the API addition, as discussed with ATM.   Let's focus on getting this major upgrade done.

PR #3439, opened 2023-08-07
Admin log viewer log root path

leogretz2:
Hoist P/R Checklist
-------------------

- [ x ] Caught up with `develop` branch as of last change.
- [ x ] Added CHANGELOG entry, or determined not required.
- [ x ] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.
- [ x ] Updated doc comments / prop-types, or determined not required.
- [ x ] Reviewed and tested on Mobile, or determined not required.
- [ x ] Created Toolbox branch / PR, or determined not required.


admin/tabs/server/logViewer/LogDisplayModel.ts
```diff
@@ -45,6 +45,10 @@ export class LogDisplayModel extends HoistModel {
     @persist
     caseSensitive: boolean = false;
 
+    @bindable
+    @persist
```
lbwexler, line 49: @persist not doing anything, but saving this thing as a preference when it should not be.

admin/tabs/server/logViewer/LogViewerModel.ts
```diff
@@ -72,11 +72,19 @@ export class LogViewerModel extends HoistModel {
...
-        const store = filesGridModel.store as UrlStore,
+        const store = filesGridModel.store,
             selModel = filesGridModel.selModel;
 
+        // Get the data from 'listFileInfo`.
+        const data = await XH.fetchJson({
+            url: 'logViewerAdmin/listFileInfo'
+        });
+
+        const {logRootPath} = data;
```
lbwexler, line 83: this variable here is just noise, used once on line below.

lbwexler: also , it just occurs to me that we would like to just keep the `listFiles` endpoint in hoist-core with the same name for backward compatibility.  (realize we had talked about changing the name for clarity).  If there is no `logRootPath` key present (i.e. for an earlier version of Hoist Core), let's just quietly display nothing.

leogretz2: Sounds good to me. At what point would a similar method's name get changed to avoid backward compatibility issues?

PR #3445, opened 2023-08-11
Admin console tab for Connection Pool monitoring

amcclain:
+ Requires `hoist-core >= v17.2` (https://github.com/xh/hoist-core/pull/307)

Hoist P/R Checklist
-------------------


amcclain:
Main goal with this is to get something lightweight out there so we can see what data collected looks like in practice and if it's helpful.  I recognize that the periodic snapshots might not be especially helpful in terms of tracking active connections, assuming that it will be catching pool activity at a pretty random point and it might be difficult to draw any overall conclusions around usage or capacity tuning.  But for an app where there is any reason to focus on connection pool behavior, the snap frequency could be ratcheted way up and hopefully more accurate/useful data would result.

I build snaps of both projects and am deploying a client project to dev with those now to see what I find.  My hope is that it's useful and we then flow any learnings back into these stats to better describe them, drop uninteresting ones, add others to the chart in a way that makes them useful, etc. - once we know what they actually look like.

Issue #3442, opened 2023-08-09
Resizing a window with a collapsed panel can cause an app to freeze


ghsolomon:
Resizing a window with a collapsed panel can cause an app to freeze.  (See screen recording for Toolbox example).

https://github.com/xh/hoist-react/assets/88904581/9e84129e-a4b6-41cf-97f8-54d3957b65a2

PR #3448, opened 2023-08-15
Vertical collapsed panel header fix

TomTirapani:
Reimplement vertical collapsed panel header styles to not use `writing-mode`, which was triggering severe performance degradation on resize. Instead, it now uses `transform: rotate()` and proportional margins to correctly orient and position strings of different lengths outside of flex layouts.

This solution required a couple of sacrifices in order to work... the inner DOM structure of panel titles now includes an extra span (targeted at string titles only), and the order of panel icon > title has changed for left collapsed panels. The latter issue could probably be avoided with more time if we deem it important.

See issue: https://github.com/xh/hoist-react/issues/3442

Hoist P/R Checklist
-------------------


lbwexler (commented):
If this is what is has to be, am ok with it.    Thanks for figuring this one out!!

The inner span does not seem to be a downside.

Losing the side-specific icon ordering  is more of a concern -- that was a nice bit of polish.  If we do merge this, would be nice to come back to that at some point.

TomTirapani:
I can take another look at the side specific icon spacing if its a concern - maybe something can be done there by going back to revering the flex ordering. It's a bit trickier as we don't have a handle on how long the text label is. Let me know if its worth taking the time

lbwexler:
Tom -- let's file a ticket and wait a bit and see how developers and users
respond to  the new alignment.

Thanks again for resolving this problem so quickly.

-Lee

Issue #3450, opened 2023-08-16
Typing decimal point in NumberInput clears value 


lbwexler:
Reproducible in Safari, on Device.
Chrome seems fine

lbwexler:
Further troubleshooting indicates that this appears to be a problem with Onsen -- if I swap in a standard html input of type =`number` it seems to work fine, and our usage certainly seems 

Re the fix tom presents:

+ Love the fact that we will stop running shorthand related code when not enabled.
+ Love the fact that this change makes shorthand mode and regular mode much closer
+ Am a *little* concerned about  giving up on the use of type = `number` when focused, and the use of our own pattern instead -- will these behave differently.
+ Didn't even realize that iOS providing the step affordances -- worried someone using that, although perhaps losing it consistently is a feature

PR #3454, opened 2023-08-16
Fix mobile number input.

TomTirapani:
Mobile number input was previously using `type=number` to constrain input to numbers. Unfortunately, browser validation interprets decimals as non-numeric and clears the input. My best guess as to why this is a recent regression is that browsers have gotten stricter about complying with the spec.

According to the docs, it is better to user `type=text` and a regex pattern to constrain valid inputs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs

We can continue to rely on `inputMode` to bring up the correct device keyboard, and use the pattern to only allow either numeric characters or numeric + shorthand units.

See issue [3450](https://github.com/xh/hoist-react/issues/3450)

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@
       specific CSS selectors or visual tests may be affected.
 * Fixed bug where `manuallySized` was not being set properly on column state.
 * Suppressed extra top border on Grids with `hideHeaders: true`.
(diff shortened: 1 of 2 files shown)


lbwexler:
I tweaked tom's PR a bit further to recognize that pattern and min/max are not really doing anything for us.
Please see comments on related ticket for more info

PR #3463, opened 2023-08-23
Ensure `DashCanvas` layout updated if window resized while hidden

### LINKED ISSUE #3215: DashCanvas does not update layout when window resized while not visible
If you have a tab with a DashCanvas which has been rendered, then switch to another tab and resize the window (easiest is to open dev tools docked and resize it), then navigate back to the tab with the DashCanvas, the DashCanvas has not resized to fit the new window width.

amcclain:
* `DashCanvas` now properly updates its layout when shown if the browser window had been resized while the component was hidden (e.g. in an inactive tab).
* Fixes #3215

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 * Flex inner title element added to `Panel` headers in v59.0, and set `display:flex` on the new
   element itself. Restores previous flexbox container behavior (when not L/R collapsed) for apps
   that are providing custom components as titles.
+* `DashCanvas` now properly updates its layout when shown if the browser window had been resized
+  while the component was hidden (e.g. in an inactive tab).
 
 ## 59.0.1 - 2023-08-17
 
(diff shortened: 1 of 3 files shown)


amcclain:
Two things I wasn't 100% sure about:

* Update to use `composeRefs` to compose `model.ref`, the `useOnVisibleChange` hook, and (this is the part I wasn't clear on) a forwarded ref.  I was looking at the pattern in `DashContainer` and went with that:

https://github.com/xh/hoist-react/blob/8ee88deb310829c6f3b4d47efabdcf682f29d86e/desktop/cmp/dash/container/DashContainer.ts#L34

* In the ticket @haynesjm42 filed, he has a delay in his workaround - in local development I was seeing the canvas resize OK without it.  Might be a timing difference introduced by the use of the hook?  Happy to start without it, but wanted to call that out.

haynesjm42 (approved):
Re: the timing, I bet you are correct and that now that we are using the hook and reacting on the visibility of the DashCanvas itself instead of the `activeTabId` that it is no longer needed

PR #3467, opened 2023-08-28
Support `testId` prop across most desktop + core components

Ryanseanlee:
Hoist P/R Checklist
-------------------


desktop/cmp/filechooser/FileChooser.ts
```diff
@@ -81,6 +82,7 @@ export const [FileChooser, fileChooser] = hoistCmp.withFactory<FileChooserProps>
 
         return hbox({
             ref,
+            'data-testid': testId,
```
amcclain, line 85: Not needed (and confusing to switch to data-testid here) - just let props flow through

amcclain (commented):
Hi - did not finish reading through (got has far as `FileChooser`) -  please see notes so far and then look over the other comps to see if the same themes apply elsewhere.

core/elem.ts
```diff
@@ -60,6 +60,8 @@ export type ElementSpec<P extends PlainObject> = P & {
     /** React key for this component. */
     key?: Key;
 
+    'data-testid'?: string;
```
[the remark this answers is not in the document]
amcclain: Took a stab at a comment - still want to review this with @lbwexler as it involves `elem` - at the very heart of it all.

desktop/cmp/appbar/AppBar.ts
```diff
@@ -102,7 +103,8 @@ export const [AppBar, appBar] = hoistCmp.withFactory<AppBarProps>({
                         })
                     ]
                 })
-            ]
+            ],
+            'data-testid': testId
```
amcclain, line 107: And in that same discussion with @lbwexler would like to review again if we could avoid having to flip back to `'data-testid'` here when passing to non-Hoist library wrappers.  IIRC we looked at having elem do the conversion, but then the issue is that you have to deal with `data-testid` instead of `testId` within your render methods (I think)

desktop/cmp/error/ErrorMessage.ts
```diff
@@ -5,15 +5,15 @@
...
 import {button, ButtonProps} from '@xh/hoist/desktop/cmp/button';
 import '@xh/hoist/desktop/register';
 import {isEmpty, isNil, isString} from 'lodash';
-import {isValidElement, ReactNode, MouseEvent} from 'react';
+import {isValidElement, MouseEvent, ReactNode} from 'react';
 
 import './ErrorMessage.scss';
 
-export interface ErrorMessageProps extends HoistProps {
+export interface ErrorMessageProps extends HoistProps, TestSupportProps {
```
amcclain, line 16: Here is a case where upgrading this comp to support `BoxProps` should be straightforward and potentially useful - I would vote for going ahead with that, as we strip back other comps that said they support BoxProps but actually don't.

desktop/cmp/panel/Panel.ts
```diff
@@ -206,16 +206,17 @@ export const [Panel, panel] = hoistCmp.withFactory<PanelProps>({
...
             });
         }
 
         // 4) Return wrapped in resizable affordances if needed, or equivalent layout box
         item =
             resizable || collapsible || showSplitter
-                ? resizeContainer({ref, item, className})
-                : box({ref, item, className, ...layoutProps});
+                ? resizeContainer({ref, item, className, testId})
+                : box({ref, item, className, testId, ...layoutProps});
```
amcclain, line 219: Looks like we will end up with the same testId installed twice when modalSupport active - this feature continues to be a bit tricky when it comes to matters of Panel's DOM layout and structure.  Let's review - important though to not allow dupes.

desktop/cmp/input/Checkbox.ts
```diff
@@ -71,8 +71,8 @@ const cmp = hoistCmp.factory<CheckboxInputModel>(({model, className, ...props},
         inline: withDefault(props.inline, true),
         label: props.label,
         tabIndex: props.tabIndex,
-
         id: props.id,
+        'data-testid': props.testId,
```
amcclain, line 75: We should add `testId` to `HoistInputProps` - I think our goal (hopefully we achieve it) is to support testId for all inputs, and adding it there means it's officially declared for every one of them.

desktop/cmp/input/DateInput.ts
```diff
@@ -4,30 +4,30 @@
...
 import moment from 'moment';
 import {createRef, ReactElement, ReactNode} from 'react';
-import './DateInput.scss';
-import {PopperBoundary, PopperModifiers} from '@blueprintjs/core';
-import {ITimePickerProps} from '@blueprintjs/datetime';
 import {DayPickerProps} from 'react-day-picker';
+import './DateInput.scss';
 
-export interface DateInputProps extends HoistProps, LayoutProps, HoistInputProps {
+export interface DateInputProps extends HoistProps, LayoutProps, HoistInputProps, TestSupportProps {
```
amcclain, line 30: After adding `testId` to `HoistInputProps` pls be sure to remove it from here and similar

desktop/cmp/input/DateInput.ts
```diff
@@ -483,6 +483,7 @@ const cmp = hoistCmp.factory<DateInputModel>(({model, className, ...props}, ref)
                     selectOnFocus: props.selectOnFocus,
                     inputRef: model.inputRef,
                     ref: model.textInputRef,
+                    testId: getTestId(props, 'dateTextInput'),
```
amcclain, line 486: Here we're *only* installing a generated variant, which seems weird.  

Let's put the unmodified testId here, since it's the primary comp users interact with and has the `xh-date-input` className.  But let's install generated sub-testIds on the clear and show-picker buttons above

desktop/cmp/input/RadioInput.ts
```diff
@@ -99,6 +99,7 @@ const cmp = hoistCmp.factory<RadioInputModel>(({model, className, ...props}, ref
             label: opt.label,
             value: opt.value,
             className: 'xh-radio-input-option',
+            'data-testid': getTestId(props.testId, `${opt.value}`),
```
amcclain, line 102: If we are generating a sub-testId here for radio inputs based on value, we should be doing it for analogous controls like `ButtonGroupInput` - but are we sure this is reliable and will work for all accepted values?  

Also - we should have the unmodified testId installed on the outer `radioGroup` unless there's some reason it can't take it.

desktop/cmp/input/TextInput.ts
```diff
@@ -134,7 +146,9 @@ const cmp = hoistCmp.factory<TextInputModel>(({model, className, ...props}, ref)
             placeholder: props.placeholder,
             rightElement:
                 props.rightElement ||
-                (props.enableClear && !props.disabled && isClearable ? clearButton() : null),
+                (props.enableClear && !props.disabled && isClearable
+                    ? clearButton({testId: getTestId(props, 'clearBtn')})
```
amcclain, line 150: We should take care to use consistent suffixes for these sub-components. We used `clear-btn` above in `Select`.  I don't have a strong opinion about it - clearBtn is more our overall style, but clear-btn tracks the CSS classnames more closely, and while testids are not classnames, they kind of live in the same world.

amcclain (commented):
Submitting a batch of comments for review - 80% done 😄

desktop/cmp/input/RadioInput.ts
Ryanseanlee: Would label be a viable option? It is typed as a string but is optional

cmp/badge/Badge.ts
```diff
@@ -37,9 +41,17 @@ export const [Badge, badge] = hoistCmp.withFactory<BadgeProps>({
             classes.push('xh-badge--compact');
         }
 
+        const divProps = merge(
+            {className: classNames(className, classes)},
+            {style: layoutProps},
```
ghsolomon, line 46: Seems maybe a little unfortunate to have to pass the style props here since `box` already handles that (and these are designated `BoxProps`).  Should these component be returning a `box` instead of a `div`?  By accepting `BoxProps`, I wonder if that's the implication anyway?

Also wonder if we should make a utility method `splitBoxProps`, which would essentially do what `splitLayoutProps` does but with the addition of `TestSupportProps`, which we would just pass down to the underlying `box` cmp in all of these cases rather than having to destructure the `testId` for every component to handle

cmp/chart/Chart.ts
```diff
@@ -85,6 +86,7 @@ export const [Chart, chart] = hoistCmp.withFactory<ChartProps>({
         const coreContents = box({
             ...layoutProps,
```
ghsolomon, line 87: See above comment. If we had a `getBoxProps` method, we could just spread the `boxProps` here and let `box` handle the `TestSupportProps` (presumably if we introduce more than just `testId`, that could come in handy, though I'm not sure we would)

cmp/clock/Clock.ts
```diff
@@ -54,11 +54,12 @@ export const [Clock, clock] = hoistCmp.withFactory<ClockProps>({
     displayName: 'Clock',
     className: 'xh-clock',
 
-    render({className, ...props}, ref) {
+    render({className, testId, ...props}, ref) {
         const impl = useLocalModel(ClockLocalModel);
         return box({
             className,
             ...getLayoutProps(props),
```
ghsolomon, line 61: See above comments re: `getBoxProps`

cmp/dataview/DataView.ts
```diff
@@ -22,7 +23,7 @@ import {isFunction, merge} from 'lodash';
 import './DataView.scss';
 import {DataViewModel} from './DataViewModel';
 
-export interface DataViewProps extends HoistProps<DataViewModel>, BoxProps {
+export interface DataViewProps extends HoistProps<DataViewModel>, LayoutProps, TestSupportProps {
```
ghsolomon, line 26: It seems like this does support `BoxProps` still, or am I missing something? `dataView` returns a `grid` which returns an `agGrid` which returns a `frame` which returns a `box`. It seems at first glance like the `BoxProps` are passed down, but I might be overlooking something?

cmp/grid/Grid.ts
```diff
@@ -38,19 +50,8 @@ import {GridModel} from './GridModel';
...
-    ColDef,
-    ColGroupDef,
-    GetContextMenuItemsParams,
-    GridOptions,
-    GridReadyEvent,
-    ProcessCellForExportParams
-} from '@xh/hoist/kit/ag-grid';
 
-export interface GridProps extends HoistProps<GridModel>, LayoutProps {
+export interface GridProps extends HoistProps<GridModel>, LayoutProps, TestSupportProps {
```
ghsolomon, line 54: Could this extend `BoxProps`? (or if we make `BoxComponentProps` generic, could this extend `BoxComponentProps<GridModel>`)?

desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -91,14 +96,21 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory<GroupingC
                         ),
                         minimal: styleButtonAsInput,
                         ...buttonProps,
-                        onClick: () => model.toggleEditor()
+                        onClick: () => model.toggleEditor(),
+                        testId
                     }),
-                    favoritesIcon()
+                    favoritesIcon({testId: favoritesIconTestId})
```
ghsolomon, line 102: Looks like currently `favoritesIcon` is dropping the `testId` prop getting passed here

desktop/cmp/grouping/GroupingChooser.ts
```diff
@@ -91,14 +96,21 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory<GroupingC
...
-                    ? favoritesMenu()
+                    ? favoritesMenu({testId: favoritesMenuTestId})
                     : editorIsOpen
-                    ? editor({popoverWidth, popoverMinHeight, popoverTitle, emptyText})
+                    ? editor({
+                          popoverWidth,
+                          popoverMinHeight,
+                          popoverTitle,
+                          emptyText,
+                          editorTestId
```
ghsolomon, line 112: Looks like `editor` is expecting a `testId` prop, but the prop being passed here is called `editorTestId`

desktop/cmp/form/FormField.ts
```diff
@@ -253,17 +264,20 @@ const editableChild = hoistCmp.factory<FieldModel>({
         disabled,
         displayNotValid,
         leftErrorIcon,
-        commitOnChange
+        commitOnChange,
+        testId
     }) {
         const {props} = child;
+        testId = props.testId ?? testId;
```
ghsolomon, line 271: Should this be done on line 280 (where it is assigned to `overrides`) so it is symmetrical with the `disabled` override?

desktop/cmp/form/FormField.ts
```diff
@@ -253,17 +264,20 @@ const editableChild = hoistCmp.factory<FieldModel>({
...
 
         // Overrides -- be sure not to clobber selected properties on child
         const overrides: DefaultHoistProps = {
             model,
             bind: 'value',
             id: childId,
             disabled: props.disabled || disabled,
-            ref: composeRefs(model?.boundInputRef, child.ref)
+            ref: composeRefs(model?.boundInputRef, child.ref),
+            testId
```
ghsolomon, line 280: See above comment. Should this be `testId: props.testId ?? testId`?

desktop/cmp/input/JsonInput.ts
```diff
@@ -25,6 +25,7 @@ export const [JsonInput, jsonInput] = hoistCmp.withFactory<JsonInputProps>({
             linter: linter,
             formatter: fmtJson,
             mode: 'application/json',
+            testId: props.testId,
```
ghsolomon, line 28: I don't think this is necessary since the props are being spread on the next line anyway

desktop/cmp/input/NumberInput.ts
```diff
@@ -275,6 +275,7 @@ const cmp = hoistCmp.factory<NumberInputModel>(({model, className, ...props}, re
             flex: withDefault(flex, null),
             textAlign: withDefault(props.textAlign, 'right')
         },
+        'data-testid': props.testId,
```
ghsolomon, line 278: Seeing this string in so many places, I wonder if we should make a constant for it, so you could write `[TEST_ID]: props.testId` and not run the risk of a typo

cmp/badge/Badge.ts
amcclain: The underlying issue (in my mind, if I recall correctly) is that `Box` gets a default flex layout, which was a mistake we made very early on and never revisited.  I really want it to simply be a `div` with some syntax candy, but instead (due to flex defaults) `box` is actually equiv to `hbox`.

https://github.com/xh/hoist-react/blob/d8319d3f28ff1aefd975d56a3624e245b305239b/cmp/layout/Box.ts#L37

For this badge case, maybe that would be fine, or even desirable, but it would be another kind of change that could break the layout of existing badges that have multiple child comps. I don't know how many of those there are - maybe none.  

So I would be open to it, but we should probably add a note to the CL.

lbwexler:
We are looking to merge this ASAP.    Greg is looking through his comments and will mark resolved if appropriate.  

Tom brought up the concern over the use of '-' as our delimiter when that is the conventional delimiter for the ids themselves.  So we have to just decide if we want to use something else like `--`? `_`? for our constructed ids in TabContainer and FormField.  Personally, I like the idea of just using the dash, for a consistent style of all ids, but willing to go along here.

desktop/cmp/grouping/GroupingChooser.ts
ghsolomon: @Ryanseanlee Wanted to make sure you saw this- everything else looks good!

core/elem.ts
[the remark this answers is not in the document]
lbwexler: looked at this -- was hoping we could do something like `[key: 'data-${string}']?: unknown` but was unable to get that to work.  Think that comment you added was fine

desktop/cmp/appbar/AppBar.ts
lbwexler: Was wondering along the same line if we wanted special handling for say `div` and `span` to take `testId`.  On the fence about this

desktop/cmp/input/RadioInput.ts
lbwexler: I guess we just want to make sure we never crash or fail.  Much less important if the user ends up with an unusable test-id of course for a non-primitive or unusual value.

PR #3478, opened 2023-09-13
Update handling for Panel.collapsedTitle and add support for Panel.collapsedIcon

### LINKED ISSUE #3477: Update to `Panel.collapsedTitle` handling + support `collapsedIcon`
The new `Panel.collapsedTitle` prop would be useful in more cases if you could spec it on a panel that didn't have a "non-collapsed" title (or any other elements that trigger the display of panel header). That way you could easily have a panel show a header only when collapsed to aid in discovery/expand.

Along with that change, supporting `collapsedIcon` would provide the most common other option you might want to conditionally set on a panel header.

Finally, I'm thinking that `collapsedHeaderItems` would somehow a stretch too far - why would a user want to trigger header items on a panel where they can't see the content?  We currently always hide headerItems when collapsed for this reason, which seems sound to me.  If we introduced `collapsedHeaderItems` I think we'd need it to follow the conventions of the other props and default to `headerItems` if not specified, which means that apps would need to change their code to maintain the current behavior.  (So tldr let's leaver headerItems handling alone)

saba-mo:
Resolves #3477 

See Toolbox PR https://github.com/xh/toolbox/pull/671
 
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/panel/impl/PanelHeader.ts
+++ b/desktop/cmp/panel/impl/PanelHeader.ts
@@ -27,7 +27,7 @@ export const panelHeader = hoistCmp.factory({
             displayedTitle = collapsed ? collapsedTitle : title,
             headerItems = props.headerItems ?? [];
 
-        if (isNil(title) && isNil(icon) && isEmpty(headerItems)) return null;
+        if (isNil(displayedTitle) && isNil(icon) && isEmpty(headerItems)) return null;
 
         const onDoubleClick = () => {
             if (isModal) {
@@ -42,7 +42,7 @@ export const panelHeader = hoistCmp.factory({
             compactCls = compact ? 'xh-panel-header--compact' : null;
 
         // 1) Classic "top" title bar
-        if (!collapsed || vertical || isModal) {
+        if ((!collapsed || vertical || isModal) && (title || icon || headerItems)) {
             return hbox({
                 className: classNames(className, compactCls),
                 items: [
--- a/desktop/cmp/panel/impl/PanelHeader.ts
+++ b/desktop/cmp/panel/impl/PanelHeader.ts
@@ -46,7 +46,7 @@ export const panelHeader = hoistCmp.factory({
             sideCls = `xh-panel-header--${side}`,
             compactCls = compact ? 'xh-panel-header--compact' : null;
 
-        // 2) Display the classic "top" title bar
+        // 2) Display the classic "top" title bar for expanded, top/bottom or modal panels
         if (!collapsed || vertical || isModal) {
             return hbox({
                 className: classNames(className, compactCls),
@@ -73,7 +73,7 @@ export const panelHeader = hoistCmp.factory({
             });
         }
 
-        // 3) Otherwise, display the collapsed/side vertical header
+        // 3) Otherwise, display the "side" title bar
         return vbox({
             className: classNames(className, sideCls, compactCls),
             flex: 1,
--- a/desktop/cmp/panel/impl/PanelHeader.ts
+++ b/desktop/cmp/panel/impl/PanelHeader.ts
@@ -25,9 +25,11 @@ export const panelHeader = hoistCmp.factory({
             {title, icon, compact} = props,
             collapsedTitle = withDefault(props.collapsedTitle, title),
             displayedTitle = collapsed ? collapsedTitle : title,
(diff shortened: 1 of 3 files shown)


desktop/cmp/panel/impl/PanelHeader.ts
```diff
@@ -42,11 +44,11 @@ export const panelHeader = hoistCmp.factory({
             compactCls = compact ? 'xh-panel-header--compact' : null;
 
         // 1) Classic "top" title bar
-        if (!collapsed || vertical || isModal) {
+        if ((!collapsed || vertical || isModal) && (title || icon || headerItems)) {
```
amcclain, line 47: What is this change doing here?  I don't think we need it - think we want to be checking off of `displayedTitle/Icon`, and the change to the early-out on line 32 above should take care of the case where there aren't any of these.

Although note that we deliberately don't render `headerItems` when collapsed - see below line 60/62. I'm just reading here in Github, but it seems to me that we could setup `displayedHeaderItems` above to be an empty array if `collapsed && !isModal` - then we really have all of our collapsed-sensitive state consistently available and ready to check.

Also - sorry just looking a bit more closely - we should be careful that the early-out above isn't violating the expectations/requirements of the modal support stuff - it's planning to put a button in the header.  

I think this all needs to be looked at a bit more carefully

saba-mo: Now that I'm looking at it more closely, I'm not sure why we are showing the modal button in collapsed headers - in general but also specifically collapsed headers of top/bottom panels but not left/right panels. Why would someone want to pop out a collapsed panel? If it is something we want to continue to offer, shouldn't we do the same for all panels regardless of side?

TomTirapani: I think the aim there was to show a collapsed top/bottom header the same as an expanded one, with the same header tools and / or modal button. We can be reasonably sure that we have the same amount of space to work with and didn't want things to jump about. You can think of it as "the **content** goes away, and the existing header moves down".

Collapsing left / right is a different beast, as it changes the dimensions and orientation of the header. We can't reliably expect the available space to be equivalent - typically (as in the Portfolio example app) we'll have a lot less space to work with. You can think of it as "the **panel** goes away, and is replaced by a minimal affordance". Rotated text is harder to read, and rotated header items / buttons are just not particularly usable. There are also technical styling reasons for not attempting to rotate header items.

At least that's how I think about it 🤷‍♂️

TomTirapani:
Regarding the modal button - it seems the question we need to ask is "should we show a header with *only* a modal button?".

Right now (both develop and this PR) the answer is no - you need at least one other item in your header for it to show at all. However, given that we are willing to show a header with **only** header items (and you can think of the modal button as built in header item), this is likely an oversight on our part.

I think we should aim for needing at least one of Title / Icon / Header Items / Modal Button required to display. What do you think?

saba-mo:
I agree if header items are enough, the modal button should be enough to show a header. But if you pop out a collapsed panel with only a modal button, you end up with this (bottom panel is modal), which definitely looks strange. However, in my opinions, it's not much stranger than the current ability to pop out a collapsed panel. I'm not sure how complicated this would be, but what do you think about toggling a panel's collapsed status when clicking the modal button so it's in an expanded state and shows its content? 
Screenshot 2023-09-14 at 10 45 05 AM
Screenshot 2023-09-14 at 10 45 11 AM

desktop/cmp/panel/impl/PanelHeader.ts
lbwexler: I like your way of thinking about it tom

amcclain: In new PR #3480 I went with consistently showing the modalToggle for both LR and TB collapsed panels - that seemed like a pretty clear choice to me, as long as we're OK with both buttons migrating to the bottom of the vertical header.

I still left out app-specific headerTools for collapsed panels, even TB-collapsed ones with the horizontal header. My thinking was that it's very likely that headerTools are only there to control something about the content of the panel, and it seemed potentially dangerous (or at least weird) to show them when you can't see that content.  

If we hit a specific use case where we do want them, we could consider adding a flag to control at that point - `showHeaderToolsWhenCollapsed` or something crazy like that :)

amcclain:
Closing in favor of #3480 which includes additional fixes and refactoring

PR #3481, opened 2023-09-15
Allow Banners to render basic markdown such as links and emphasis.

TomTirapani:
image

+ Added `reactMarkdown` component and dependency.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
    of any `Panel`.
 * Added new `Panel.headerClassName` prop for straightforward CSS manipulation of panel's header.
 * Improved styling for disabled `checkbox` inputs.
+* Added new `reactMarkdown` component for rendering Markdown formatted strings as markup.
+* Enabled Banners to render bold, italics and links using Markdown syntax.
 
 ## 59.0.3 - 2023-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,15 @@
    of any `Panel`.
 * Added new `Panel.headerClassName` prop for straightforward CSS manipulation of panel's header.
 * Improved styling for disabled `checkbox` inputs.
-* Added new `reactMarkdown` component for rendering Markdown formatted strings as markup.
+* Added new `Markdown` component for rendering Markdown formatted strings as markup. This includes
+  bundling `react-markdown` in Hoist. If your app already uses `react-markdown` or similar, you should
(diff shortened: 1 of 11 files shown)


amcclain:
Thank you for jumping on this - has been on my list for a while to discuss built-in markdown support, and this was a good prompt to do so.

* Should we consider publishing a `markdown` component outside of the `kit` directory? I am not sure - I can't think of anything really to add or normalize on the API, but at the same time we typically publish a non-kit cmp for components we expect developers to be using regularly. 
* Not for the banner usage, but thinking of other in-app usages like comments, we could consider adding https://github.com/remarkjs/remark-breaks as a preconfigured plugin. Github makes this modification to standard MD in a few places and I always find it intuitive:

By way of example
this is on another line
(but wouldn't be in standard markdown, I don't think)

* Changelog entry should include it in a Libraries section + a callout in the New Features section to recommend that apps cut over to the Hoist export + remove from their package.json to take advantage of our managed upgrade cycle (compatibility testing, etc) and to avoid potentially including multiple versions of the library should we have non-overlapping version specs.
* I'm all for adding this as a built-in, but we should check to how much this increases the bundle size of e.g. Toolbox. It looks like a fair amount of dependencies, and with the usage within the built-in banner, we know it will always be included in any/every app bundle.

amcclain:
Another thought - at some point it would be great to have a user-friendly `MarkdownInput` component with something like the formatting toolbar in this Github commenting box.

Codemirror has a markdown mode and we could certainly configure that, but then you are in monospace/non-wrapping "code editing" mode and that's not really the experience I think we would want for user (or even admin) facing MD editing.

Not something we need to do for this PR, but it could be a companion comp we could consider when we go to use MD more widely for use cases like in-app comments (something I'd like to offer in one of our current apps).

TomTirapani:
Thanks for taking a look Anselm - I've updated the PR. Including the `remark-breaks` plugin was justification enough to wrap this in its own `Markdown` component. I've updated the CHANGELOG entry, and ensured the "Read More" popup is also rendered as markdown.

TomTirapani:
According `toolbox/yarn startAndAnalyze` with the inlineHoist flag set to true, this PR increases the bundle size 23.81 MB > 24.39 MB

lbwexler:
This looks great to me ---  we might want to add some text in the banner admin discussing what is going to work, and what won't.  Happy to merge this now

amcclain (approved):
Thanks for getting those additional changes in 👍

PR #3484, opened 2023-09-21
Register active models by component `testId`

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/HoistComponent.ts
+++ b/core/HoistComponent.ts
@@ -4,13 +4,15 @@
  *
  * Copyright © 2023 Extremely Heavy Industries Inc.
  */
+import {instanceManager} from '@xh/hoist/core/impl/InstanceManager';
 import {
     CreatesSpec,
     UsesSpec,
     HoistProps,
     DefaultHoistProps,
     elementFactory,
-    ElementFactory
+    ElementFactory,
+    TestSupportProps
 } from './';
 import {
     useModelLinker,
@@ -24,7 +26,7 @@ import {
     HoistModel
 } from './model';
 import {apiDeprecated, throwIf, warnIf, withDefault} from '@xh/hoist/utils/js';
-import {getLayoutProps} from '@xh/hoist/utils/react';
+import {getLayoutProps, useOnUnmount} from '@xh/hoist/utils/react';
 import classNames from 'classnames';
 import {isFunction, isPlainObject, isObject} from 'lodash';
 import {observer} from '../mobx';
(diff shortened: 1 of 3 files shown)


core/impl/InstanceManager.ts
```diff
@@ -44,6 +47,20 @@ class InstanceManager {
...
+        this.modelsByTestId.set(testId, m);
+    }
+
+    unregisterModelWithTestId(testId: string) {
+        this.modelsByTestId.delete(testId);
+    }
+
+    getModelByTestId(testId: string): HoistModel {
+        const model = this.modelsByTestId.get(testId);
+        throwIf(!model, `No model found for component with testId: ${testId}`);
```
amcclain, line 60: Would suggest we return null vs throw here.

core/HoistComponent.ts
```diff
@@ -313,11 +315,16 @@ function wrapWithModel(render: RenderFn, cfg: Config): RenderFn {
...
             let ctx = localModelContext;
             try {
+                const {testId} = props;
                 props.model = model;
                 delete props.modelRef;
                 delete props.modelConfig;
                 ctx.props = props;
                 ctx.modelLookup = newLookup ?? modelLookup;
+                if (testId && model) {
+                    instanceManager.registerModelWithTestId(testId, model);
```
lbwexler, line 325: I think you are going to be registering the same thing over and over as testId and model are passed down through the hierarchy.  And it seems like the resolved model could be different for some intermediate components that are passing through the same test id?  

Also violating the "rules of hooks" that say you should *always* call the hook for any given component regardless of the state -- think may be ok, but only if we can guarantee that the result of that conditional to never change -- not sure we can. Let's discuss?

ghsolomon: First point, definitely true. My thought was that ultimately the deepest component with a given `testId` would register itself last, and that would end up being the associated model.  There would be several unnecessary calls to `instanceManager.registerModelWithTestId`, but that seemed like it could be ok given how lightweight it is?  Or maybe it’s something we only enable when in a test environment?  Would love to discuss all of this further.

Second point, I did consider this and thought it would be ok since the presence of `testId` *should* be immutable, though you could make an argument that it is a prop like any other and therefore doesn’t have to be (but that would sure be confusing!). Same thing with `model`- a component that has a model for one render *should* always have a model, just maybe not the same model.  Again could see this not being a guarantee.

Would you suggest moving the cleanup hook to the `finally` block so it always runs?  That said, if we’d be doing that to guard against a `testId` becoming `nil`, the hook would be ineffective anyway unless we cache the first `testId` which seems maybe overkill?  Would love to discuss!

core/impl/InstanceManager.ts
ghsolomon: Updated to return `undefined` in that case. Could also fallback on `null` if you think that's better than `undefined`?

lbwexler: for what its worth, I almost *always* prefer null to undefined.  Certainly the java programmer in me, but its also a practice in Hoist generallyt

PR #3492, opened 2023-10-09
Remove lodash import

leogretz2:
Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/general/alertBanner/AlertBannerModel.ts
+++ b/admin/tabs/general/alertBanner/AlertBannerModel.ts
@@ -11,7 +11,7 @@ import {HoistModel, LoadSpec, managed, XH, Intent, PlainObject} from '@xh/hoist/
 import {dateIs, required} from '@xh/hoist/data';
 import {action, makeObservable, observable} from '@xh/hoist/mobx';
 import {AppModel} from '@xh/hoist/admin/AppModel';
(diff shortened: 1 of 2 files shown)


leogretz2:
It seems that Lodash requires _ in most of its declaration files, specifically a number of the ones we are using. For example: object.d.ts begins with the line 'import _ = require("../index");' . In the Lodash version we are using, the lodash.js file adds _ to the window. There were tickets/[PRs](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/6088) to split Lodash into two parts: one with the global variable, another without, but Lodash has declared issue [bankruptcy](https://news.ycombinator.com/item?id=37538426), so it is unlikely this will happen. The way to remove it from global scope is to then add _.noConflict() in Bootstrap.js. @amcclain mentioned in a Slack discussion that this adds an entire static library to the window, but I think it is already added by lodash and this will remove it. Further discussion is required - these are my findings so far.

Screenshot 2023-10-09 at 15 34 27

amcclain:
Thanks - not sure what to make of it. Unclear where the code screenshot above comes from. 

I don't have / wasn't looking to claim any knowledge about the noConflict feature - not anything I have used or research.

I must have been wrong that previously we did not have _ on window, or something changed that we can't revert back. I don't have any further opinions on this, personally.

lbwexler:
Looking

Issue #3497, opened 2023-10-11
Right-pinned grid columns do not have left border


amcclain:
We apply a right border to the _last left-pinned_ column via:

https://github.com/xh/hoist-react/blob/bc6ff3ff4e7472f26ef26acf33495cad8f2715c2/cmp/ag-grid/AgGrid.scss#L190

We should do the same for the _first right-pinned_ column also.

amcclain:
Ensure we handle  #3330 for this new case also

Issue #3330, opened 2023-04-12
Missing grid cell border for focused + pinned column cell


amcclain:
<img width="1840" alt="Screenshot 2023-04-12 at 7 03 50 AM" src="https://user-images.githubusercontent.com/225806/231487024-bd012bdf-14b0-4a51-8956-72f7c8615a64.png">

Note that this issue only appears when you have clicked on the pinned cell, "invisibly" focusing it.

amcclain:
Pinned border should be applied by:

https://github.com/xh/hoist-react/blob/bc6ff3ff4e7472f26ef26acf33495cad8f2715c2/cmp/ag-grid/AgGrid.scss#L190

PR #3501, opened 2023-10-15
Fix `Grid` cell border styling inconsistencies

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/ag-grid/AgGrid.scss
+++ b/cmp/ag-grid/AgGrid.scss
@@ -128,16 +128,17 @@
   }
 
   &--no-row-borders {
-    .ag-row {
+    .ag-row,
+    .ag-cell {
       border-bottom: none;
       border-top: none;
+    }
 
-      // Deliberately keep border on full-width grouped rows even when we aren't adding row borders.
-      // Without this collapsed groups (which don't stripe) blend together in a solid block.
-      &.ag-row-group.ag-full-width-row {
-        border-bottom: 1px solid var(--xh-grid-group-border-color);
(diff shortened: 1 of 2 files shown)


ghsolomon:
Closes #3497

cmp/ag-grid/AgGrid.scss
```diff
@@ -5,6 +5,10 @@
  * Copyright © 2023 Extremely Heavy Industries Inc.
  */
 
+@mixin pinned-border($side) {
```
TomTirapani, line 8: Great use of a mixin! 👍

cmp/ag-grid/AgGrid.scss
```diff
@@ -121,12 +125,14 @@
 
   &--no-row-borders {
     .ag-row {
-      border-color: transparent;
+      border-bottom: none;
+      border-top: none;
```
TomTirapani, line 129: Removing the border (rather than making it transparent) has the unfortunate side-effect of causing the grid to shift slightly when toggling borders, which does not happen in current toolbox prod.

Not sure how much of an issue that is in practice (border flags are typically not dynamic), but if we can achieve this fix with transparent borders that would be preferable. Something like `border-bottom-color: transparent` etc

TomTirapani (changes requested):
Works great, and addresses both #3497 and #3330. I left one note about making borders transparent to keep the sizing static.

TomTirapani:
Looks good to me 👍

PR #3506, opened 2023-10-18
Add flag to Column to sort nulls to the bottom

TomTirapani:
I'm not 100% sure on the flag name `bottomNullSort` - suggestions welcome

Apps that implement custom Column.comparators should *not* need to make any adjustments, as long as they fall back to the provided defaultComparator.

See Toolbox example: https://github.com/xh/toolbox/pull/678

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/columns/Column.ts
+++ b/cmp/grid/columns/Column.ts
@@ -185,6 +185,11 @@ export interface ColumnSpec {
     absSort?: boolean;
 
     /**
+     * True to always sort null / undefined values to the bottom, regardless of sort order.
+     */
+    bottomNullSort?: boolean;
+
+    /**
      * Alternate field name to reference or function to call when producing a value for this column
      * to be sorted by.
      */
@@ -433,6 +438,7 @@ export class Column {
     rowHeight: number;
     sortingOrder: ColumnSortSpec[];
     absSort: boolean;
+    bottomNullSort: boolean;
     sortValue: string | ColumnSortValueFn;
     comparator: ColumnComparator;
     resizable: boolean;
@@ -504,6 +510,7 @@ export class Column {
             flex,
             rowHeight,
             absSort,
+            bottomNullSort,
             sortingOrder,
             sortValue,
             comparator,
@@ -598,6 +605,7 @@ export class Column {
         this.rowHeight = rowHeight;
 
         this.absSort = withDefault(absSort, false);
+        this.bottomNullSort = withDefault(bottomNullSort, false);
         this.sortingOrder = this.parseSortingOrder(sortingOrder);
         this.sortValue = sortValue;
         this.comparator = comparator;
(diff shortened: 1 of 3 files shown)


cnrudd:
maybe `bottomNilSort` since the check is isNil.

cmp/grid/columns/Column.ts
```diff
@@ -184,6 +184,11 @@ export interface ColumnSpec {
      */
     absSort?: boolean;
 
+    /**
+     * True to always sort null / undefined values to the bottom, regardless of sort order.
+     */
+    bottomNullSort?: boolean;
+
     /**
```
lbwexler, line 192: Would we want to extend this to NaN, '',  undefined?  We should consider which other sentinel values (other than zero) might warrant the same treatment?  We could even have a *list* of values, that should be put at the bottom, with the order they should appear there.  Would this be useful for values like 'CLOSED'?  Or other high-level ordering we want to impose independent of the sorting.

lbwexler: could be something like `sortToTop:[]` `sortToBottom:[]`

TomTirapani: I really like this idea - makes it more flexible and generalizable. We could support an array of values of equality checks, or boolean closures for more complex checks (e.g. `v => isNil(v) || itEmpty(v)`)

TomTirapani:
I've updated the PR to use `sortToBottom: []`, taking an array of values or closures.

cmp/grid/columns/Column.ts
cnrudd: For values like 'CLOSED', user expectations have been that the 'CLOSED' record be at the bottom no matter which column is sorted, and no matter which column has the 'CLOSED'.  So, not sure how useful this would be, unless the comparators allow getting the value from other fields.

TomTirapani: Yeah, I don't think we can use this for cross-column sorting like you describe. I think that's kind of an edge case and not necessarily something we should support with hoist out of the box, where the 'normal' use case for sorting is only concerned with the sorted column.

lbwexler:
Every column would probably need this set.  Can use defaulting on GridModel for that.

TomTirapani:
I've updated the PR to move the implementation of sortToBottom to an internal comparator in Column, as discussed yesterday

cmp/grid/columns/Column.ts
lbwexler: tom -- simplified slightly, tweaked performance, and also allowed it to take a single value.

Let me know if it makes sense?

cmp/grid/columns/Column.ts
```diff
@@ -939,7 +939,7 @@ export class Column {
                         recordB,
                         column: this,
                         gridModel,
-                        defaultComparator: (a, b) => this.defaultComparator(a, b),
+                        defaultComparator: this.defaultComparator,
```
TomTirapani, line 942: If we are changing this from a fat arrow function, I think we need to do `this.defaultComparator.bind(this)`, so that the scope is correct for the implementation there.

cmp/grid/columns/Column.ts
```diff
@@ -1055,6 +1049,11 @@ export class Column {
         return sortingOrder?.map(spec => (isString(spec) || spec === null ? {sort: spec} : spec));
     }
 
+    private parseSortToBottom(sortToBottom): Array<(v: unknown) => boolean> {
+        if (isNil(sortToBottom)) return null;
+        return castArray(sortToBottom).map(v => (isFunction(v) ? v : it => it === v));
```
TomTirapani, line 1054: Great idea to parse this into an array for functions ahead of time 👍

cmp/grid/columns/Column.ts
```diff
@@ -1000,26 +1000,20 @@ export class Column {
...
-
-        if (!isNil(v1ToBottom) && !isNil(v2ToBottom)) {
-            return (v1ToBottom - v2ToBottom) * (asc ? 1 : -1);
+        for (let fn of sortToBottom) {
+            const v1ToBottom = fn(v1),
+                v2ToBottom = fn(v2);
+            const isAsc = sortDir === 'asc';
+            if (v1ToBottom != v2ToBottom) {
+                return v1ToBottom ? (isAsc ? 1 : -1) : isAsc ? -1 : 1;
+            }
```
TomTirapani, line 1014: I don't understand how this accounts for the order of entries in sortToBottom? I'm talking about the case when both values return true for _different_ closures within the config, and we want to respect the order of the closures.

cmp/grid/columns/Column.ts
lbwexler: defaultComparator is already a fat arrow function

PR #3514, opened 2023-10-31
RelativeTimeStamp localDateMode v2

lbwexler:
This is a refactor of Colin's work in the related branch, that I think better clarifies what we are doing, and how it differs from Moment.

Am curious if folks think we really need the mode 'limited' -- could either discard completely, or fold in to default as Colin has done.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/relativetimestamp/RelativeTimestamp.ts
+++ b/cmp/relativetimestamp/RelativeTimestamp.ts
@@ -4,7 +4,7 @@
  *
  * Copyright © 2023 Extremely Heavy Industries Inc.
  */
-import {inRange} from 'lodash';
+import {inRange, isNil} from 'lodash';
 import moment from 'moment';
 import {box, span} from '@xh/hoist/cmp/layout';
 import {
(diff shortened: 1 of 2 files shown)


cnrudd:
I like the 24hr threshhold change
but I don't see why anyone would ever choose localDateMode 'off', or to put it your way: I think we should "fold limited into default [off]".


If the output offered more granular units when the diff spans days, like "1 day, 13hrs", then I could see wanting the choice between `off` and `limited`.

lbwexler:
@cnrudd and I discussed this -- just want to clarify that I think 'limited' is probably correct for most of our applications, and if we really wanted to simplify this API it could be built in.

Also wanted to clarify that I really don't like these names for the modes -- just was at a loss for something better, and wanted to get the PR out there.

cmp/relativetimestamp/RelativeTimestamp.ts
```diff
@@ -62,6 +63,16 @@ export interface RelativeTimestampOptions {
...
     /** Time to which the input timestamp is compared. */
     relativeTo?: Date | number;
+
+    /**
+     * Governs if calendar days should be used for computing string rather than the native
+     * moment 24-hour day. Set to 'limited' to ensure that any output that refers to 'days'
+     * (i.e. between 1 day and 26 days elapsed time) will refer to the difference in
+     * *calendar* days between the timestamps.   Set to 'full' to ensure that any output less
+     * than 25 days will refer to the difference in *calendar* days. Off (default)
+     * will indicate that the standard moment formatting should be used.
```
TomTirapani, line 73: I find this comment confusing - perhaps it needs more emphasis on the **difference** between full and limited (i.e. "In limited mode, output of less than a day will be left as is")

lbwexler: totally agree

PR #3517, opened 2023-11-02
Add experimental `enableFullWidthScroll` Grid feature

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/Grid.ts
+++ b/cmp/grid/Grid.ts
@@ -7,9 +7,9 @@
 import composeRefs from '@seznam/compose-react-refs';
 import {agGrid, AgGrid} from '@xh/hoist/cmp/ag-grid';
 import {getTreeStyleClasses} from '@xh/hoist/cmp/grid';
-import {gridScrollbar} from '@xh/hoist/cmp/grid/impl/GridScrollbar';
+import {gridHScrollbar} from '@xh/hoist/cmp/grid/impl/GridHScrollbar';
 import {getAgGridMenuItems} from '@xh/hoist/cmp/grid/impl/MenuSupport';
-import {div, frame, vframe} from '@xh/hoist/cmp/layout';
+import {div, fragment, frame, vframe} from '@xh/hoist/cmp/layout';
 import {
     hoistCmp,
     HoistModel,
@@ -107,10 +107,11 @@ export const [Grid, grid] = hoistCmp.withFactory<GridProps>({
             highlightRowOnClick ? 'xh-grid--highlight-row-on-click' : null
         );
 
-        const container = model.experimental['enableFullWidthScroll'] ? vframe : frame;
+        const {enableFullWidthScroll} = model.experimental,
+            container = enableFullWidthScroll ? vframe : frame;
 
-        return container(
-            frame({
+        return fragment(
+            container({
                 className,
                 items: [
                     agGrid({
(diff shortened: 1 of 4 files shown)


cmp/grid/Grid.ts
```diff
@@ -841,4 +870,83 @@ class GridLocalModel extends HoistModel {
...
+
+    readonly SCROLLBAR_SIZE = 10;
+    readonly scrollerRef = createRef<HTMLDivElement>();
+
+    @observable viewportWidth: number;
+    @observable private isVerticalScrollbarVisible = false;
+
+    private viewportResizeObserver: ResizeObserver;
+
+    @computed
```
lbwexler, line 886: why is this computed?  don't think we expect this to change,  be observable, or be expensive to check

cmp/grid/Grid.ts
```diff
@@ -841,4 +870,83 @@ class GridLocalModel extends HoistModel {
...
+    }
+
+    scrollViewport(left: number) {
+        this.agViewport.scrollLeft = left;
+    }
+
+    override afterLinked() {
+        if (!this.isFullWidthScrollEnabled) return;
+        this.addReaction({
+            track: () => this.model.isReady,
```
lbwexler, line 924: should this just be a `when:()`

cmp/grid/Grid.ts
```diff
@@ -108,13 +110,35 @@ export const [Grid, grid] = hoistCmp.withFactory<GridProps>({
...
-                    ...getLayoutProps(props),
-                    ...impl.agOptions
-                }),
+                items: [
+                    agGrid({
+                        model: model.agGridModel,
+                        ...getLayoutProps(props),
+                        ...impl.agOptions
+                    }),
+                    div({
```
lbwexler, line 121: I would make this its own HoistComponent to avoid extra re-rendering whenever the scrollbar calcs need to change.  ALSO, weird that the (experimental) scrollbar dominates the render method the way it does now.

Ideally put *all* the state for this in a dedicated GridScrollBarModel/GridScrollBar pair in their own files.

lbwexler: Especially nice to have these split out while this is experimental, but also will be useful even if this becomes built-in

lbwexler:
Greg and I reviewed this and he has just a few tweaks to make.  Hoping to merge this tomorrow!  👍🏻

cmp/grid/Grid.ts
```diff
@@ -107,14 +107,22 @@ export const [Grid, grid] = hoistCmp.withFactory<GridProps>({
             highlightRowOnClick ? 'xh-grid--highlight-row-on-click' : null
         );
 
-        return fragment(
+        const container = model.experimental['enableFullWidthScroll'] ? vframe : frame;
+
+        return container(
```
amcclain, line 112: Can we avoid the extra level of DOM nesting that this appears to introduce?  Previously we were returning a `fragment` at the top level, with a frame and supporting colChooser/filter dialogs inside.  

Could those dialogs be moved to nest within the frame/vframe items (alongside the agGrid) so we still have a single outer (v)frame?

ghsolomon: Sorry about that- actually didn't mean to change the outer `fragment`- this was in a bad state mid-refactor. Just updated the PR

PR #3521, opened 2023-11-08
Fix bug where a user could be prompted to log in to an app with SSO e…

ghsolomon:
…nabled

Hoist P/R Checklist
-------------------


amcclain:
Wondering how this was triggered for an SSO user (getting a 401 from a Hoist-server endpoint) and what then the recovery plan should be.  There's the matter of the message that's shown to the user ("please login again") but this behavior was also setting the requireReload flag. Do we still want that to force a full refresh of the app in this case, or should something else happen?

lbwexler:
greg and I discussed -- was coming via a proxy endpoint!   Greg mentioned
that we could potentially distinguish that specifically, but I thought the
first order/simpler fix was  to just not do it at all for SSO

amcclain:
Got it, thanks - would never have thought of that! For that reason, would be good to comment and or mention in changelog note, e.g.

> Ensure an unauthorized response from a proxy service endpoint does not prompt the user to refresh and log in again on an SSO-enabled application.

We do have a kind of gray area with apps that use OAuth flows. Looking at the two we work on here, one is configured with `isSSO: true` and the other as `isSSO: false`, but I believe the only reason we have false for the second app is that we currently support a "backup" form of login with local credentials. I don't think this change would interfere with anything there, although I confess I don't have a great mental model for the world of possible circumstances that could result in a Hoist 401 with integrations like this one.

PR #3528, opened 2023-11-16
Updated uses of `document.hidden`

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/async/AsyncUtils.ts
+++ b/utils/async/AsyncUtils.ts
@@ -15,7 +15,7 @@ import {wait} from '@xh/hoist/promise';
  * allowing ongoing rendering of UI updates (e.g. load masks) and generally keeping the browser
  * event loop running.
  *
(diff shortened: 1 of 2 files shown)


utils/async/AsyncUtils.ts
```diff
@@ -55,7 +55,7 @@ export async function whileAsync(
     const {waitAfter = 50, waitFor = 0} = opts ?? {};
 
     // Fallback to basic loop when doc hidden: no user benefit, and throttling causes outsize waits
```
lbwexler, line 57: why the inconsistency here?  I would expect this would also us XH.pageIsVisible to take advantage of the debouncing there.

amcclain: Had asked the same of @jskupsik and he was worried that there might be some extra performance overhead evaluating that getter, but I don't see how there would be.  I would also vote for using !pageIsVisible

Issue #3510, opened 2023-10-25
`GridFilterFieldSpec.inputProps` typed too generically as `HoistInputProps`


amcclain:
We store and will then spread on any `inputProps` provided to the type-specific input component that ends up being rendered within the grid filter header's "custom" tab:

https://github.com/xh/hoist-react/blob/c7283cbba9ac52ca86c2cd64a0e228f72d15128a/desktop/cmp/grid/impl/filter/custom/CustomRow.ts#L79

This means we support all props for e.g. `dateInput` or `numberInput`, but the typing displays as an error.  We should likely amend to a union type of the PropTypes for all supported inputs - I expect that's the best we can do.

lbwexler:
Would be nice to take this typescript off the board, if simple @ghsolomon -- thanks

ghsolomon:
One complication here - we have different input props for our desktop vs. mobile components, but `GridFilterFieldSpec` is shared between both.  We could make`inputProps` an exclusive union of both mobile and desktop prop types, but then I believe we would also need to break those type definitions out into their own modules so we aren't importing anything but the types into the shared module.

ghsolomon:
We ultimately decided against changing the `GridFilterFieldSpec.inputProps` type in order to avoid importing platform-specific code into this cross-platform module.  Applications wishing to ensure type-safety and avoid TS errors can define their input props separately and pass them as illustrated below:

```
private createGridModel() {
        const inputProps: DateInputProps = {
            maxDate: LocalDate.today()
        }

        return new GridModel({
            filterModel: {
                fieldSpecs: [
                    {
                        field: 'trade_date',
                        fieldType: 'localDate',
                        inputProps
                    }
                ]
            },
            ...
        })
}
```

PR #3529, opened 2023-11-17
Misc Typescript Fixes

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/filter/GridFilterFieldSpec.ts
+++ b/cmp/grid/filter/GridFilterFieldSpec.ts
@@ -6,6 +6,7 @@
  */
 import {ColumnRenderer} from '@xh/hoist/cmp/grid';
 import {HoistInputProps} from '@xh/hoist/cmp/input';
+import {PlainObject} from '@xh/hoist/core';
 import {FieldFilterOperator, parseFilter, View} from '@xh/hoist/data';
 import {
     BaseFilterFieldSpec,
@@ -41,7 +42,7 @@ export interface GridFilterFieldSpecConfig extends BaseFilterFieldSpecConfig {
 export class GridFilterFieldSpec extends BaseFilterFieldSpec {
     filterModel: GridFilterModel;
     renderer: ColumnRenderer;
-    inputProps: HoistInputProps;
+    inputProps: PlainObject;
     defaultOp: FieldFilterOperator;
     valueCount: number;
 
--- a/cmp/grid/filter/GridFilterFieldSpec.ts
+++ b/cmp/grid/filter/GridFilterFieldSpec.ts
@@ -6,7 +6,6 @@
  */
 import {ColumnRenderer} from '@xh/hoist/cmp/grid';
 import {HoistInputProps} from '@xh/hoist/cmp/input';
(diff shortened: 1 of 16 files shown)


cmp/grid/filter/GridFilterFieldSpec.ts
```diff
@@ -42,7 +41,7 @@ export interface GridFilterFieldSpecConfig extends BaseFilterFieldSpecConfig {
 export class GridFilterFieldSpec extends BaseFilterFieldSpec {
     filterModel: GridFilterModel;
     renderer: ColumnRenderer;
-    inputProps: PlainObject;
+    inputProps: HoistInputProps;
```
amcclain, line 44: See #3510 - @saba-mo and I just ran into that issue last week, would be great if we could address as part of this PR, taking this one step further to support a broader union type here and on the `GridFilterFieldSpecConfig` interface

cmp/zoneGrid/impl/ZoneGridPersistenceModel.ts
```diff
@@ -22,7 +23,14 @@ export class ZoneGridPersistenceModel extends HoistModel {
...
     @observable.ref
-    state: PlainObject;
+    state: {
+        columns?: Partial<ColumnState>[];
+        sortBy?: GridSorterLike;
+        groupBy?: Some<string>;
+        version?: number;
+        autosize?: any;
+        mappings?: Record<Zone, Some<string | ZoneMapping>>;
+    };
```
amcclain, line 33: I don't believe we save `columns` or `autosize` in this state block - N/A / replaced by `mappings`

data/Store.ts
```diff
@@ -219,7 +219,6 @@ export class Store extends HoistBase {
     _filtered: RecordSet;
 
     private _dataDefaults = null;
-    private _created = Date.now();
```
amcclain, line 222: Turns out we read and show this in the instances grid within Hoist Inspector. There are also `_created` member vars on HoistModel and HoistService, although you can tell how essential they are by the fact that the HoistModel one was broken until just now (https://github.com/xh/hoist-react/commit/b0418d03e4982d8d6a6d1b6f9125dfc3f88d6886)

I don't feel strongly about it, but if we keep the timestamp we should not mark as private, obvs, and if we don't think it's useful we might as well remove it from the other two base classes and the little grid in Inspector

amcclain (commented):
Nice set of fixes/cleanups/improvements all around 👍

cmp/zoneGrid/impl/ZoneGridPersistenceModel.ts
lbwexler: good catch

data/Store.ts
lbwexler: I'll revert!  Was thrown off by the mistaken use of private.

cmp/grid/filter/GridFilterFieldSpec.ts
lbwexler: will hold off on this change then, and take the other simpler changes tin this PR

PR #3530, opened 2023-11-20
Make ZoneGridModel more forgiving (handle possibly out of sync state)

febbraiod:
… state)

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/zoneGrid/ZoneGridModel.ts
+++ b/cmp/zoneGrid/ZoneGridModel.ts
@@ -657,7 +657,7 @@ export class ZoneGridModel extends HoistModel {
 
             if (!isEmpty(limit.only)) {
                 const offender = find(ret, it => !limit.only.includes(it.field));
-                throwIf(offender, `Field "${offender}" not allowed in zone "${zone}".`);
+                throwIf(offender, `Field "${offender?.field}" not allowed in zone "${zone}".`);
             }
         }
 
--- a/cmp/zoneGrid/ZoneGridModel.ts
+++ b/cmp/zoneGrid/ZoneGridModel.ts
@@ -49,7 +49,16 @@ import {
 } from '@ag-grid-community/core';
 import {Icon} from '@xh/hoist/icon';
 import {throwIf, withDefault} from '@xh/hoist/utils/js';
-import {castArray, forOwn, isEmpty, isFinite, isPlainObject, isString} from 'lodash';
+import {
+    castArray,
+    forOwn,
+    isEmpty,
+    isFinite,
+    isPlainObject,
+    isString,
+    dropRight,
+    remove
+} from 'lodash';
 import {ReactNode} from 'react';
 import {ZoneMapperConfig, ZoneMapperModel} from './impl/ZoneMapperModel';
 import {ZoneGridPersistenceModel} from './impl/ZoneGridPersistenceModel';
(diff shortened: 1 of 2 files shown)


cmp/zoneGrid/ZoneGridModel.ts
```diff
@@ -616,41 +625,65 @@ export class ZoneGridModel extends HoistModel {
...
+                            );
+                            offenders.push(it.field);
+                        }
                     });
+
+                    if (!isEmpty(offenders)) {
+                        remove(mapping, it => offenders.includes(it.field));
+                    }
+
+                    if ((zone == 'tl' || zone == 'tr') && isEmpty(mapping)) {
```
TomTirapani, line 675: Great catch! But I think we want to do this outside of the `!isEmpty(limit.only)` condition.

(I know enforcing `only` is currently the only place where we remove mappings, but I think the intent is clearer if we always do this as the final check of parseMappings)

TomTirapani (commented):
Thanks for catching and taking a look at this issues Don!

It's certainly clearly problematic if this thing throws because saved state is no longer valid! I also like how you are resolving back to the default state to get back to a good state 👍  But I realised that assumes the **default** state respects the constraints, and `parseMappings()` is used both to set up the initial default state **and** to handle changing the state.

I wonder if we should only be this lenient in the latter case? It seems wrong to allow apps to set up a Grid where the defaults are always wrong - especially if we expect loading the defaults to resolve the issue. Maybe we could add a second `strict` parameter to `parseMappings()`, and conditionally throw or warn based on that?

cmp/zoneGrid/ZoneGridModel.ts
febbraiod: Great note Tom. I considered the same, but apparently made the wrong choice ;)

cmp/zoneGrid/ZoneGridModel.ts
```diff
@@ -635,52 +637,58 @@ export class ZoneGridModel extends HoistModel {
                 if (col) {
                     mapping.push(ret);
                 } else {
-                    console.warn(`Column not found for field '${ret.field}'`);
+                    const message = `Column not found for field '${ret.field}'`;
+                    this.throwOrWarn(message, strict);
```
TomTirapani, line 641: We could also do something like this:
```
const throwOrWarn = strict ? throwIf : warnIf;
throwOrWarn(`Column not found for field '${ret.field}'`);
```

cmp/zoneGrid/ZoneGridModel.ts
```diff
@@ -611,46 +621,77 @@ export class ZoneGridModel extends HoistModel {
...
                 const ret = isString(it) ? {field: it} : it,
                     col = this.findColumnSpec(ret);
 
-                throwIf(!col, `Column not found for field ${ret.field}`);
-                return mapping.push(ret);
+                if (col) {
+                    mapping.push(ret);
+                } else {
+                    const message = `Column not found for field '${ret.field}'`;
+                    this.throwOrWarn(message, strict);
```
febbraiod, line 641: inline function that captures strict.

cmp/zoneGrid/ZoneGridModel.ts
```diff
@@ -611,46 +621,77 @@ export class ZoneGridModel extends HoistModel {
...
+                    mapping = this._defaultState.mappings[zone];
+                }
+
+                if (isFinite(limit.max) && mapping.length > limit.max) {
+                    const diff = mapping.length - limit.max,
+                        message = `Exceeds maximum ${limit.max} mappings in zone "${zone}". Dropping last ${diff} fields`;
+
+                    this.throwOrWarn(message, strict);
+
+                    mapping = dropRight(mapping, mapping.length - limit.max);
```
febbraiod, line 662: use diff var.

febbraiod: set to default here as well

cmp/zoneGrid/ZoneGridModel.ts
```diff
@@ -611,46 +621,77 @@ export class ZoneGridModel extends HoistModel {
...
+                            const message = `Field "${it.field}" not allowed in zone "${zone}. Removing"`;
+
+                            this.throwOrWarn(message, strict);
+
+                            offenders.push(it.field);
+                        }
                     });
+
+                    if (!isEmpty(offenders)) {
+                        remove(mapping, it => offenders.includes(it.field));
```
febbraiod, line 678: set to default again here. Only need to warn for one field and say we're going to the default immediately at that point.

cmp/zoneGrid/ZoneGridModel.ts
febbraiod: this can also do the reset to default. handleConstraint.

PR #3531, opened 2023-11-22
FieldFilter should use isEquals for = and != op comparisons

### LINKED ISSUE #3511: `FieldFilter` should use `isEquals` for `=` and `!=` op comparisons
@saba-mo and I were looking at a case where we wanted to support filtering on dates (with times). There is some overhead required with the filterchooser / grid filter spec configs to support parsing and rendering (another target for future improvements) but even if you implement those you get stuck on FieldFilter's equality comparison, which is done via `[].includes`:

https://github.com/xh/hoist-react/blob/c7283cbba9ac52ca86c2cd64a0e228f72d15128a/data/filter/FieldFilter.ts#L125

Would like to look at changing this impl to e.g. `value.some(it => isEqual(v, it))` - as well as the corresponding != case. I believe that should clear the way for Date comparisons to work as expected and cannot currently think of anything that would start comparing as true where we wish that it did not....

Note that our story around Date filtering is made even worse if you `enableValues:true` on a grid date col. The grid filter code extracts  Date instances from the records, then uniques them via getTime, meaning that you don't see dupes but if you have multiple rows with the same exact date and check that date you will only get one match in the grid - the record that happens to have _the_ Date instance that was extracted and then survived the uniq:

https://github.com/xh/hoist-react/blob/c7283cbba9ac52ca86c2cd64a0e228f72d15128a/cmp/grid/filter/GridFilterFieldSpec.ts#L134

amcclain:
+ Fixes #3511

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/filter/FieldFilter.ts
+++ b/data/filter/FieldFilter.ts
@@ -12,6 +12,7 @@ import {
     difference,
     escapeRegExp,
     isArray,
+    isEqual,
     isNil,
     isString,
     isUndefined,
@@ -122,13 +123,13 @@ export class FieldFilter extends Filter {
             case '=':
                 opFn = v => {
                     if (isNil(v) || v === '') v = null;
-                    return value.includes(v);
+                    return value.some(it => isEqual(v, it));
                 };
                 break;
             case '!=':
                 opFn = v => {
                     if (isNil(v) || v === '') v = null;
-                    return !value.includes(v);
+                    return !value.some(it => isEqual(v, it));
                 };
                 break;
             case '>':


amcclain:
Should resolve the current weird/broken behavior with Date fields with grid filter + enableValues referenced in #3511 

Main concern is around performance - any chance this will introduce some extra overhead that would be material when filtering large amounts of data? I did not test that at all - we should do some measure of smoke/sanity checking before merging!

lbwexler:
I would not worry about performance here -- the big expense from filtering/unfiltering large stores IIRC is just the cost to rebuild the collection of records, not the actual test.  I am pretty confident isEqual is fast for the primitives we are testing with in practice.

PR #3533, opened 2023-11-22
Enhance LogUtils

amcclain:
+ Optionally accept messages as an array of any objects. Continue to concat string messages together with standardized | separator, log any other objects in their natural form.
+ Add logError() and logWarn()
+ Convert (some) existing console.log/warn/error calls to use the new utils - we should sweep through and update the rest of them before merge.

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/js/LogUtils.ts
+++ b/utils/js/LogUtils.ts
@@ -4,10 +4,11 @@
  *
  * Copyright © 2023 Extremely Heavy Industries Inc.
  */
+import {Some} from '@xh/hoist/core';
 import {castArray, isString} from 'lodash';
 
 /**
- * Track a function execution with console.log.
+ * Time and log execution of a function to `console.info()`.
  *
  * This method will log the provided message(s) with timing information in a single message *after*
  * the tracked function returns.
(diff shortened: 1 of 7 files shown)


amcclain:
For review/discussion - if we want to take this change, can add CHANGELOG entry + we should sweep through and upgrade other raw calls to console loggers where appropriate.

utils/js/LogUtils.ts
```diff
@@ -103,7 +121,51 @@ function parseSource(source) {
     return '';
 }
 
-function writeLog(msg, source, level) {
-    if (source) msg = `[${source}] ${msg}`;
-    level === 'info' ? console.log(msg) : console.debug(msg);
+function writeLog(msgs: Some<any>, source, level: LogLevel) {
```
lbwexler, line 124: this is alway an array, fwiw

utils/js/LogUtils.ts
```diff
@@ -20,60 +21,77 @@ import {castArray, isString} from 'lodash';
  * @param fn - function to execute
  * @param source - class, function or string to label the source of the message
  */
-export function withInfo<T>(msgs: string[] | string, fn: () => T, source?: any): T {
+export function withInfo<T>(msgs: Some<any>, fn: () => T, source?: any): T {
```
lbwexler, line 24: Think we should type these as Some<unknown>

utils/js/LogUtils.ts
```diff
@@ -103,7 +121,51 @@ function parseSource(source) {
...
-    if (source) msg = `[${source}] ${msg}`;
-    level === 'info' ? console.log(msg) : console.debug(msg);
+function writeLog(msgs: Some<any>, source, level: LogLevel) {
+    msgs = castArray(msgs);
+    if (source) msgs = [`[${source}]`, ...msgs];
+
+    const logArgs = [];
+    let strMsg;
+
+    msgs.forEach((msg, idx) => {
```
lbwexler, line 131: I think this all would be a bit easier to build logArgs with reduce.  With every step you get the current state of the array and then the next element. For each step,  either push a "|" and your arg, or just append your arg to the string that is the last element

utils/js/LogUtils.ts
amcclain: Changed signature on public APIs. Ran into type errors if I changed impl methods that I did not understand.

utils/js/LogUtils.ts
amcclain: Updated to `msgs: Array<any>` and removed castArray

utils/js/LogUtils.ts
amcclain: Sorry couldn't follow, tried but did not find an easier path with reduce. I tend not to reach for reduce as I almost never find it intuitive or easy to read. More than happy to have a different impl here, just ran out of time to figure it out myself.

amcclain:
Replaced by #3534

PR #3534, opened 2023-11-23
Enhanced LogUtils 2

lbwexler:
Took this one step further, by making delegating methods on HoistBase.  The syntax has its pros and cons, but I would argue that overall its better.

More similar to hoist core, leverages HoistBase, and allows us to not have to surround multiple args in `[]` for most cases.  

Happy to discuss

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Enhanced `LogUtils` to support logging objects (and any other non-string values). Also
   added new exports for `logWarn()` and `logError()` with the same standardized formatting.
+* Added standardized `LogUtils` methods to `HoistBase`, for use within Hoist models and services.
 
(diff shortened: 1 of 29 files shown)


lbwexler:
I played with the types -- PR here is what is required for the full typing -- not sure how much better it is , but the key thing is that the `any`  is really blowing away *all* typing on the thing it touches, such that a Some<any> can be passed as an any[], even though it may actually not even be an array!     `any` is surprisingly toxic to the typing!

When we remove that, we see we can e.g. no longer reuse the `messages` variable to hold the raw and parsed form....


I also played around a lot with the code that puts in the delimiters between the logic -- had some ideas on *very* marginally improving what was there, but in the end really think the simple thing.  Would love to discuss more -- I am probably missing something.

PR #3540, opened 2023-11-28
Fixed the ReactionSpec type fix

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/HoistBase.ts
+++ b/core/HoistBase.ts
@@ -140,12 +140,12 @@ export abstract class HoistBase {
     addReaction(...specs: ReactionSpec<any>[]): IReactionDisposer | IReactionDisposer[] {
         const disposers = specs.map(s => {
             if (!s) return null;
-            let {track, when, run, debounce, ...opts} = s;
+            let {track, when, run, debounce, ...rest} = s;
             throwIf(
                 (track && when) || (!track && !when),
                 "Must specify either 'track' or 'when' in addReaction."
             );
-            opts = parseReactionOptions(opts);
+            const opts = parseReactionOptions(rest);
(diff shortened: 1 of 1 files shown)


jskupsik:
My original PR ended up not changing the ReactionSpec.equals type.

The derived type ended up being `IEqualsComparer<T> & (keyof typeof comparer | IEqualsComparer<T>)` which simplifies down to `IEqualsComparer<T>` - which is exactly what it was before my change.

The issue is that ReactionSpec does not extend IReactionOptions in the typescript sense. The fact that it accepts values for `equals` that IReactionOptions does not accept breaks inheritance- a ReactionSpec may not work if used as a IReactionOptions. We understand that and use the private `parseReactionOptions()` function to convert the ReactionSpec to IEqualsComparer.

The type needs to be more explicit about this lack of inheritance - "ReactionSpec is mostly of type 'IEqualsComparer' EXCEPT for the 'equals' property, which we override and then handle ourselves."

core/HoistBase.ts
```diff
@@ -293,7 +293,7 @@ export abstract class HoistBase {
 /**
  * Object containing options accepted by MobX 'reaction' API as well as arguments below.
  */
-export type ReactionSpec<T = any> = IReactionOptions<T, any> & {
+export type ReactionSpec<T = any> = Omit<IReactionOptions<T, any>, 'equals'> & {
```
lbwexler, line 296: might use an interface in this case, for consistency.

lbwexler:
So interesting -- I almost suggested the use of omit for the last go around, but had thought in my research that the & would override the type.  In any case, looks good to me.  I would probably make this interface rather than a type

PR #3554, opened 2023-12-14
Enhanced Role support - admin UI to manage persistent user roles

amcclain:
Hoist P/R Checklist
-------------------


admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    override async doLoadAsync(loadSpec: LoadSpec) {
+        try {
+            const {data} = await XH.fetchJson({loadSpec, url: 'roleAdmin/list'});
+            if (loadSpec.isStale) return;
+            this.setRoles(this.processRolesFromServer(data));
+        } catch (e) {
+            if (loadSpec.isStale) return;
+            XH.handleException(e);
+            this.clear();
```
amcclain, line 50: I have a longstanding item in my mind to try and work through the best patterns for doLoadAsync() w.r.t. exception handling / reporting and autorefresh.  The issue here is that an autorefresh that fails can silently clear the grid and you end up with a "clean" looking UI that reports no roles.

I'd suggest also returning as a no-op if `loadSpec.isAutorefresh` - happy to discuss some more thoughts around use of `lastLoadException`, but that is where it gets a bit bogged down.

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+    // -------------------------------
+
+    private groupByCategoryReaction(): ReactionSpec<boolean> {
+        const {gridModel} = this;
+        return {
+            track: () => this.groupByCategory,
+            run: groupByCategory => {
+                gridModel.setGroupBy(groupByCategory ? 'category' : null);
+                gridModel.setColumnVisible('category', !groupByCategory);
+                gridModel.autosizeAsync();
```
amcclain, line 81: Need to instruct gridModel to not persist grouping - otherwise restore defaults on grid gets you out of sync

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    private processRolesFromServer(
+        roles: Omit<HoistRole, 'users' | 'directoryGroups' | 'roles'>[]
+    ): HoistRole[] {
+        return roles.map(role => {
+            const membersByType = mapValues(groupBy(role.members, 'type'), members =>
+                members.map(member => member.name)
+            );
+            return {
+                ...role,
```
amcclain, line 132: Interesting general question of how concerned we are with getting the types exactly right - in this case note that nothing is converting the date fields, until they are loaded into store records, at which point the record data has a date. Could parse here for correctness, don't really think it matters more raising as a general discussion point....

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    private async editAsync(role: HoistRole): Promise<void> {
+        const updatedRole = await this.roleEditorModel.editAsync(role);
+        if (!updatedRole) return;
+        await this.refreshAsync();
+    }
+
+    private createGridModel(): GridModel {
+        return new GridModel({
+            autosizeOptions: {mode: 'managed'},
```
amcclain, line 155: This is set by default in `admin/AppModel` - we can remove from here

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+            emptyText: 'No roles found.',
+            colChooserModel: true,
+            sortBy: 'name|asc',
+            enableExport: true,
+            exportOptions: {filename: 'roles'},
```
amcclain, line 160: See other admin grids for exportOptions

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+            filterModel: true,
+            headerMenuDisplay: 'hover',
+            onRowDoubleClicked: ({data: record}) =>
+                record &&
+                this.roleEditorModel
```
amcclain, line 165: Use `this.editAsync()`?

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+                    },
+                    {name: 'effectiveRoleNames', displayName: 'Effective Roles', type: 'tags'},
+                    {name: 'members', type: 'json'}
+                ],
+                processRawData: raw => ({
+                    ...raw,
+                    effectiveUserNames: raw.effectiveUsers.map(it => it.name),
+                    effectiveDirectoryGroupNames: raw.effectiveDirectoryGroups.map(it => it.name),
+                    effectiveRoleNames: raw.effectiveRoles.map(it => it.name),
+                    inheritedRoleNames: raw.inheritedRoles.map(it => it.name)
```
amcclain, line 194: Here vs in `processRolesFromServer()`?

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+                'roles',
+                'inheritedRoleNames',
+                'effectiveUserNames',
+                'effectiveDirectoryGroupNames',
+                'effectiveRoleNames',
+                'lastUpdatedBy',
+                {
+                    field: 'lastUpdated',
+                    example: 'YYYY-MM-DD',
+                    valueParser: (v, op) => {
```
amcclain, line 262: It looks like we have three copies of this thing in admin console now - could we DRY up, pending better support for date filtering?

admin/tabs/roles/RolesModel.ts
```diff
@@ -0,0 +1,280 @@
...
+                            ret = moment(ret).endOf('day');
+                        }
+
+                        return ret.toDate();
+                    },
+                    valueRenderer: v => fmtDate(v),
+                    ops: ['>', '>=', '<', '<=']
+                }
+            ],
+            persistWith: {...RolesModel.PERSIST_WITH, path: 'mainFilterChooser'}
```
amcclain, line 277: I would not persist filters - I consider it an anti-pattern in an admin UI like this.  You always want to come back to it fresh seeing all the data.

admin/tabs/roles/details/RoleDetails.ts
```diff
@@ -0,0 +1,41 @@
...
+                    })
+                }),
+                tabContainer()
+            ]
+        });
+    }
+});
+
+const valOrNa = v => (v != null ? v : naSpan());
+const naSpan = () => span({item: 'N/A', className: 'xh-text-color-muted'});
```
amcclain, line 41: I see we have these in multiple places - do they make sense to add to stock Hoist renderers?

admin/tabs/roles/details/RoleDetailsModel.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    get selectedRole(): HoistRole {
+        return this.rolesModel.selectedRole;
+    }
+
+    override onLinked() {
+        this.addReaction({
+            track: () => this.selectedRole,
+            run: role => this.formModel.init(role ?? {})
+        });
```
amcclain, line 22: Little gotcha here - if selectedRole already set before linked, you won't get your form populated.  I noticed we got into this state in the last demo.  Collapse detail panel (currently persisted, prob shouldn't be) then select a role in the grid then expand detail.

admin/tabs/roles/details/RoleDetailsModel.ts
```diff
@@ -0,0 +1,65 @@
...
+                    name: 'lastUpdatedBy'
+                }
+            ],
+            readonly: true
+        });
+    }
+
+    private createTabContainerModel(): TabContainerModel {
+        return new TabContainerModel({
+            persistWith: {...RolesModel.PERSIST_WITH, path: 'roleMembersTabContainer'},
```
amcclain, line 51: Could nix this

admin/tabs/roles/details/members/RoleMembersModel.ts
```diff
@@ -0,0 +1,205 @@
...
+            store: {
+                fields: [
+                    {name: 'name', type: 'string'},
+                    {name: 'type', type: 'string'},
+                    {name: 'sources', displayName: 'Assigned Via', type: 'json'}
+                ],
+                idSpec: data => `${this.selectedRole.name}:${data.type}:${data.name}`
+            },
+            emptyText: 'This role has no members.',
+            filterModel: true,
```
amcclain, line 130: Thinking this is overkill to have grid filters on this grid, plus there is some issue with the sources JSON field - would remove, unless you think they would be especially useful and we can fix that issue

admin/tabs/roles/details/members/RoleMembersModel.ts
```diff
@@ -0,0 +1,205 @@
...
+                    flex: true,
+                    renderer: (sources: Array<{role: string; directoryGroup?: string}>) =>
+                        hbox({
+                            className: 'roles-renderer',
+                            items: uniqBy(
+                                sources.map(({role, directoryGroup}) => {
+                                    const isThisRole = role === this.selectedRole.name;
+                                    return {
+                                        className: 'roles-renderer__role',
+                                        intent: isThisRole ? null : 'primary',
```
amcclain, line 175: The use of the primary intent here makes it look a bit like these are links to jump to that role - maybe they should be?

admin/tabs/roles/editor/RoleEditorModel.ts
```diff
@@ -0,0 +1,109 @@
...
+    readonly rolesModel: RolesModel;
+    readonly savingTask = TaskObserver.trackLast({message: 'Saving Role'});
+    readonly deletingTask = TaskObserver.trackLast({message: 'Deleting Role'});
+
+    @managed readonly roleFormModel = new RoleFormModel();
+
+    @observable isOpen = false;
+    @observable role?: HoistRole;
+
+    private resolve: (role?: HoistRole) => void;
```
amcclain, line 20: I wasn't clear on how this is used

admin/tabs/roles/editor/form/RoleForm.ts
```diff
@@ -0,0 +1,132 @@
...
+                        },
+                        items: [
+                            hbox(
+                                formField({
+                                    disabled: model.isEditingExistingRole,
+                                    field: 'name',
+                                    flex: 1,
+                                    item: textInput({autoFocus: true})
+                                }),
+                                formField({
```
amcclain, line 41: Should we make category required?  If we want to keep optional, let's spec enableClear. Note that no-category roles end up in a slightly wonky "(Blanks)" group in the main grid

static/index-manifest.html
```diff
@@ -9,7 +9,7 @@
     <% } %>
     <title><%= htmlWebpackPlugin.options.title %></title>
 </head>
-<body class="xh-app" style="margin: 0">
+<body class="xh-dark" style="margin: 0">
```
amcclain, line 12: Intentional?  I am assuming not....

admin/columns/Roles.ts
```diff
@@ -0,0 +1,13 @@
...
+    field: {name: 'inherits', type: 'string'}
+};
+
+export const inheritedUsers: ColumnSpec = {
+    field: {name: 'inheritedUsers', type: 'string'}
+};
+
+export const directoryGroups: ColumnSpec = {
+    field: {name: 'directoryGroups', type: 'string'}
+};
```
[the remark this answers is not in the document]
ghsolomon: Ah yes, this is leftover- will remove

admin/tabs/roles/HoistRole.ts
```diff
@@ -0,0 +1,34 @@
+export interface HoistRole {
```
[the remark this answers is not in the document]
ghsolomon: Didn't see any, but I'll double-check.  Like the idea of admin/Types.ts for any global types, but since this is local to the roles package, maybe would prefer to rename in place

admin/tabs/roles/Roles.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+const detailsPanel = hoistCmp.factory<RolesModel>(({model}) =>
+    panel({
+        icon: Icon.idBadge(),
+        title: model.selectedRole?.name ?? 'Role Details',
+        item: roleDetails(),
+        compactHeader: true,
+        modelConfig: {
+            defaultSize: '50%',
```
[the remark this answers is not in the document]
ghsolomon: Tweaked this some. I was thinking 30% with a reasonable pixel minSize

admin/tabs/roles/Roles.ts
```diff
@@ -0,0 +1,78 @@
...
+            persistWith: {...RolesModel.PERSIST_WITH, path: 'detailsPanel'},
+            side: 'right'
+        }
+    })
+);
+
+const graphPanel = hoistCmp.factory<RolesModel>(() =>
+    panel({
+        item: roleGraph(),
+        modelConfig: {
```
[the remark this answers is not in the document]
ghsolomon: Agreed- already tweaked this to include a title (and the switcher to show inherited vs. effective)

admin/tabs/roles/details/members/RoleMembersModel.ts
ghsolomon: Ah yes agreed, will remove

admin/tabs/roles/details/members/RoleMembersModel.ts
ghsolomon: Had that thought too, will implement

admin/tabs/roles/editor/RoleEditorModel.ts
ghsolomon: It should've been called on save or cancel, but realize I missed calling it on cancel

admin/tabs/roles/editor/form/RoleForm.ts
ghsolomon: Don't think it should be required, but good point about the wonky blanks group...

static/index-manifest.html
ghsolomon: Definitely not!

static/index.html
```diff
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=0, shrink-to-fit=no">
     <title><%= htmlWebpackPlugin.options.title %></title>
 </head>
-<body class="xh-app">
+<body class="xh-dark">
```
lbwexler, line 8: what is this change

PR #3557, opened 2023-12-28
Role Improvements

ghsolomon:
Hoist P/R Checklist
-------------------


cnrudd (approved):
Reviewed and ran locally.  All good!

admin/tabs/general/roles/RoleModel.ts
```diff
@@ -16,6 +16,13 @@ import {HoistRole, RoleMemberType, RoleServiceConfig} from './Types';
 export class RoleModel extends HoistModel {
     static PERSIST_WITH = {localStorageKey: 'xhAdminRolesState'};
 
+    static fmtDirectoryGroup(name?: string): string {
```
lbwexler, line 19: don't think name is optional here

ghsolomon: Thanks - actually should be `string | null`, but I'll change to just `string`

admin/tabs/general/roles/details/members/RoleMembersModel.ts
```diff
@@ -175,13 +187,39 @@ export class RoleMembersModel extends HoistModel {
                 {
                     field: 'name',
                     autosizeMaxWidth: 300,
-                    renderer: (name, {record}) =>
-                        record.get('type') === types.DIRECTORY_GROUP
-                            ? this.fmtDirectoryGroup(name)
-                            : name,
-                    tooltip: true
+                    renderer: (name, {record}) => {
```
lbwexler, line 190: might help put this grid config on a diet if we were to define this renderer and the renderer below as functions at the bottom of the file.

admin/tabs/general/roles/editor/form/RoleFormModel.ts
```diff
@@ -226,4 +263,22 @@ export class RoleFormModel extends HoistModel {
             debounce: 250
         };
     }
+
+    private async lookupDirectoryGroupAsync(directoryGroup: string, recordId: string) {
+        try {
+            const {data} = await XH.fetchJson({
+                autoAbortKey: recordId,
```
lbwexler, line 270: nice to see this used -- was surprised looking at the FetchService that this abortKey is not actually specific to the the url!  Wondering if the best practice here is to include some more context in this key

lbwexler (approved):

PR #3562, opened 2024-01-12
Inline popup editor background fix

Ryanseanlee:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 61.0.0-SNAPSHOT - unreleased
+
+### 🐞 Bug Fixes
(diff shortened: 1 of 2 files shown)


Ryanseanlee:
Screenshot 2024-01-11 at 4 38 39 PM

This is what is currently happening when in dark mode.

I am not sure if this is the exact fix we want but this will give the popover a background

TomTirapani:
Thanks Ryan, this looks great! The CSS certainly looks like we're targetting the right selector and fixing the problem... however, I would recommend we use `var(--xh-grid-bg);` so it looks more part of the grid. `var(--xh-popup-bg);` default gray looks a bit like a disabled field to my eyes.

Issue #3570, opened 2024-01-30
`FileChooser` should accept a `maxFiles` prop


febbraiod:
This is part of the underlying component's api and I have use for it in a client app.

HarjobandeepSingh:
@febbraiod

ghsolomon:
Reopening since `maxFiles` is not a valid prop in the version of `react-dropzone` we are currently using. Requires a breaking change upgrade to the latest version of `react-dropzone`.

amcclain:
Unwound overly-eager addition of this prop in https://github.com/xh/hoist-react/commit/6bde8182f569974cfed6346f0484e416b7b322e7, filed linked ticket for the library update

PR #3573, opened 2024-02-01
fix/ #3570 FileChooser should accept a maxFiles prop

HarjobandeepSingh:
FileChooser will accept a maxFiles prop


HarjobandeepSingh:
#3570

amcclain:
Thanks, @HarjobandeepSingh - we saw this come in a while back but are just getting to merge it now. I'm not sure how you connected with XH + Hoist, but appreciate your interest. 😄

PR #3575, opened 2024-02-07
Flag and display Config overrides in the Admin Console UI

ghsolomon:
Hoist P/R Checklist
-------------------


amcclain:
Need to wrap it up for today but can pick up again tomorrow and I am confident we can finalize.

Not sure how to do it cleanly, but I think we need to handle display of overridden pwd configs in dialog - currently those show the unmasked digest value, which since it looks like your typical password is going to alarm people :) 

Screenshot 2024-02-07 at 5 32 34 PM

desktop/cmp/rest/impl/RestFormField.ts
```diff
@@ -29,12 +30,12 @@ export const restFormField = hoistCmp.factory({
             fieldModel = model.getFormFieldModel(field),
             fieldVal = fieldModel.value;
 
-        // Skip fields that are a) empty and b) readonly when c) adding a record. No point in
-        // showing these fields as they are not populated (nor are they expected to be), and they
-        // can't be edited. Common examples are metadata such as `dateCreated` and `lastUpdated`.
+        // Skip fields if explicitly requested via `omit`, or if they are a) empty and b) readonly
+        // when c) adding a record. No point in showing as they are not populated (nor are they
+        // expected to be), and they can't be edited - e.g. `dateCreated` and `lastUpdated`.
         if (
```
ghsolomon, line 36: Was thinking a bit more about this-- is there ever a time you'd want to override this "default omit" behavior? 

Rather than always checking for both conditions (omit and nil-readonly-add), we could provide a default closure to `omit` if none is specified (i.e. `omit = () => isNil(fieldVal) && fieldModel.readonly && model.isAdd`). That would maintain the current behavior for existing apps but allow `omit` to be the single source-of-truth for apps that specify it.

amcclain: Hi, yes I was wondering the same thing - a bit of an annoyance to have this new omit property and then also have omit logic hard-coded.  

Trying to think if we'd ever want a case where you both want some custom logic and the default case would also apply, and if we need to make that easy for a developer to implement.  (I.e. by providing a default omit fn you could call from within your custom one).  Not sure that there is.

amcclain (approved):
Looking great.  

I'm fine without further changes to the `omit` flag, but on board with some default impl if it seems warranted.

Do not believe this is a breaking change currently, which is nice - i.e. does not require HC 19 (pls correct me if I am wrong).

PR #3577, opened 2024-02-13
New `FilterChooserFilter` type, used for `FilterChooserModel.value`

### LINKED ISSUE #3331: FilterChooserModel.value's Filter type is not quite right
FilterChooserModel.value's `Filter` type is not quite right.

Screenshot of TS error:
image

This can be solved in app by typing value`as CompoundFilter|FieldFilter`
```
    get filter(): PlainObject {
        const ret = this.filterChooserModel.value as CompoundFilter|FieldFilter;
        return ret?.toJSON();
    }
```

but it might be worth considering a solution in Hoist.

amcclain:
+ FilterChooser supports only `FieldFilter` and `CompoundFilter`, make this explicit while also avoiding the use of the abstract `Filter` as its value type in favor of the two concrete implementations and their overlapping methods.
+ Fixes #3331

Hoist P/R Checklist
-------------------


amcclain:
Not sure about this change, but was looking at filter-related code and issues and figured I would take a run at it.

* Is it an improvement to forgo the abstract base class as the primary interface / value type?

cmp/filter/FilterChooserModel.ts
```diff
@@ -423,25 +423,25 @@ export class FilterChooserModel extends HoistModel {
 
     @action
     setFavorites(favorites: Filter[]) {
-        this.favorites = favorites.filter(f => this.validateFilter(f));
+        this.favorites = favorites.filter(f => this.validateFilter(f)) as FilterChooserFilter[];
```
amcclain, line 426: Had been hoping that `validateFilter` type guard signature would flow through the array `filter` here, but does not appear to be the case, hence the explicit `as`.  LMK if I'm doing it wrong or we can think of a more elegant solution 😄

ghsolomon: Looks like the issue here is that you are passing a closure (which calls the typeguard) rather than passing the typeguard directly to `filter`.  The closure isn't typed specifically, so TS seems to infer that its return type is `boolean`.

Possible solution would be to pass the typeguard directly, like this
`this.favorites = favorites.filter(this.validateFilter)`.

Would need to ensure `validateFilter` is still bound to the instance on which it's called, since it calls methods on `this`.
  
Screenshot 2024-02-15 at 2 37 43 PM

amcclain: Thank you for the explanation!  

Not in love with it, but for this little filter seemed worth it to swap to:

`this.favorites = favorites.filter(this.validateFilter.bind(this));`

to leverage the type guard directly (almost) as I had intended.

Issue #3578, opened 2024-02-13
`FilterChooserModel` can drop field type specific parsing in emitted values


amcclain:
* Construct a `FilterChooserModel` with a fieldSpec specifying a `localDate` field.
* Provide it with a value source populated with `LocalDate` values in that field
* Use the filter suggestions to select a specific localDate from the type-ahead
* Observe `filterChooserModel.value` - the emitted `Filter` has a string `value` instead of a `LocalDate`
* Call `filterChooserModel.value.getTestFn()` and attempt to use that function to test candidate objects (_not_ store records) - equality operators will not pass (although `< / >` comparators will).

This bug is not typically an issue in most common use case when a FCM is bound directly to a store and is applying its filter to that store. In that case, `RecordSet` calls `getTestFn()` with its store, which triggers type-aware handling of the filter within its generated function:

https://github.com/xh/hoist-react/blob/8db6db9d1b948809aeec48dbf626aaa6b51b752c/data/impl/RecordSet.ts#L120

But if you want to get a "raw" test function yourself and apply it to non-StoreRecord objects, you run into trouble.

Underlying cause appears to be round-tripping of possible filter options from `QueryEngine` through the select control.  QE emits filters with correctly typed values, using the `FilterFieldSpec` to parse a string query value into the actual type:

https://github.com/xh/hoist-react/blob/8db6db9d1b948809aeec48dbf626aaa6b51b752c/cmp/filter/impl/QueryEngine.ts#L160

But then when those (fully functional) `FieldFilter` candidates are used to construct a `FilterChooserOption` to be returned from the QE, they are stringified as JSON:

https://github.com/xh/hoist-react/blob/8db6db9d1b948809aeec48dbf626aaa6b51b752c/cmp/filter/impl/Option.ts#L82

When one of these options is selected by a user via the select control, `FilterChooserModel.setSelectValue` re-parses the JSON back into a `FilterLike` object:

https://github.com/xh/hoist-react/blob/8db6db9d1b948809aeec48dbf626aaa6b51b752c/cmp/filter/FilterChooserModel.ts#L307

But the nested `value` within each filter loses its type if it doesn't serialize "naturally" to/from JSON.  This is where our localDate values end up as strings.

This is a blocker for us in a relatively complex use-case within a client app, where we are relying on FCM to produce a filter testFn that we then use to test candidate objects before loading them into a store (i.e. we don't want to test store records here, we need to use a custom path).

[...]

PR #3579, opened 2024-02-15
Improve `FilterChooser` handling of `Date` and `LocalDate` value type…

ghsolomon:
…s (#3578)

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/filter/Utils.ts
+++ b/data/filter/Utils.ts
@@ -5,14 +5,16 @@
  * Copyright © 2024 Extremely Heavy Industries Inc.
  */
 
-import {castArray, flatMap, groupBy, isArray, isFunction} from 'lodash';
+import {FilterChooserFilterLike} from '@xh/hoist/cmp/filter';
+import {LocalDate} from '@xh/hoist/utils/datetime';
+import {castArray, flatMap, groupBy, isArray, isFunction, isNil, isString} from 'lodash';
 
 import {FilterLike} from './Types';
 import {Filter} from './Filter';
 import {FieldFilter} from './FieldFilter';
 import {FunctionFilter} from './FunctionFilter';
 import {CompoundFilter} from './CompoundFilter';
-import {Some} from '@xh/hoist/core';
+import {PlainObject, Some, XH} from '@xh/hoist/core';
 
 /**
  * Parse a filter from an object or array representation.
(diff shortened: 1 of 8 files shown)


amcclain:
This is looking very promising.

Pushed https://github.com/xh/toolbox/commit/ba20e90887e0d8806ecf992d4a359419a5b3e05b to Toolbox on `develop` to add an "Order Day" column to the Portfolio example orders detail grid - can use that to test the persistence of a LocalDate type value through the filter chooser favorites.

data/filter/Utils.ts
```diff
@@ -146,3 +148,51 @@ export function combineValueFilters(filters = []): Filter[] {
...
+    }
+}
+
+/**
+ * Deserializes a filter, respecting the value type information stored by `serializeFilter` for
+ * Date and LocalDate values. Note that this method will throw an exception if it encounters
+ * a filter it can in no way deserialize; however, the absence of an exception does not guarantee
+ * that the filter is valid.
+ */
+export function deserializeFilter(filter: string | PlainObject): FilterChooserFilterLike {
```
amcclain, line 177: Not sure about use of `FilterChooserFilterLike` here - or at least not sure if that's the name we want to be seeing here.  Is it too closely bound to that one component / implementation, or is it the thing we already have that nicely captures the compound+field filter only subset of filters.

Also, looking at that type itself, should we swap out `FilterChooserFilter` (the two concreete types) for `Filter`?  And should we consider a rename, or leave it be?

Finally, should the `serialize` method also take the same, given that it will throw if passed a function filter or spec? Wanted to talk through the relationship between say the type signature there and the final throw

amcclain:
Another thought - should we leverage toJSON() instead on the concrete classes for the serialization with type and then fold the value-aware handling into parseFilter + add the type field to the spec interfaces for inclusion in FilterLike?

Also - should we skip serializing type for primitives, where it's just extra weight?

data/filter/Utils.ts
ghsolomon: I agree the name isn't right here, but I went with it since it's the union type we have for serializable filters.  Maybe another type alias or a rename would make sense?

I'm not sure I quite follow the 2nd question.

I do think serialize should take the same. The tricky part is that our `CompoundFilter.filters` is typed as `Filter[]`, so when we recursively call `serializeFilter` with those nested filters, we'd get a TS error.  I suppose we could filter the `filters` first to make that type-safe. That probably is the cleanest approach and more clearly communicates the allowed types.

ghsolomon:
> Another thought - should we leverage toJSON() instead on the concrete classes for the serialization with type and then fold the type-aware handling into parseFilter + add the type field to the spec interfaces for inclusion in FilterLike?
> 
> Also - should we skip serializing type for primitives, where it's just extra weight?

Thought about that, but couldn't remember if we're always dealing with instances of these filters or if we're sometimes dealing with specs.  I can look into that though.

Makes sense to skip serializing type for primitives unless we think the consistency is cleaner? Happy to make that change

amcclain:
> Thought about that, but couldn't remember if we're always dealing with instances of these filters or if we're sometimes dealing with specs. I can look into that though.

Great, thanks - good for us to be entirely clear on that if we can, and hoping that the answer is (or can be) instances.

> Makes sense to skip serializing type for primitives unless we think the consistency is cleaner? Happy to make that change

Yes I think we should. Always in favor of consistency, but in this case I think it's sufficiently obvious why we would/would not choose to record the type.

PR #3581, opened 2024-02-16
Fix bug where a discarded browser tab could re-init an app to an obso…

ghsolomon:
…lete (cached) version.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@
   type, a union of `CompoundFilter | FieldFilter` - the two concrete filter implementations
   supported by this control.
 
+### 🐞 Bug Fixes
+
+* Fixed bug where a discarded browser tab could re-init an app to an obsolete (cached) version.
+
 ### 📚 Libraries
 
 * classnames `2.3 → 2.5`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@
 
 * `FilterChooser` now correctly round-trips `Date` and `LocalDate` values. Previously it emitted
   these as strings, with incorrect results when using the generated filter's test function directly.
-
-### 🐞 Bug Fixes
-
 * Fixed bug where a discarded browser tab could re-init an app to an obsolete (cached) version.
 
 ## 60.2.0 - 2024-02-16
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,10 @@
   type, a union of `CompoundFilter | FieldFilter` - the two concrete filter implementations
   supported by this control.
 
+### 🐞 Bug Fixes
+
+* Fixed bug where a discarded browser tab could re-init an app to an obsolete (cached) version.
+
```
amcclain, line 24: Need to move to new SNAP release before merging

lbwexler:
Hi Greg:   Love this, but can we try something like this API?

```
   const url = new URL(location.href);
        url.searchParams.set('xhCacheBuster', Date.now().toString());
        document.location.assign(url);
```
        
Seems to work well and would allow us to:
- Avoid problems when there is a '#'
- Allow us to check the date of any existing value and avoid getting into an endless loop.
- just set a single value in the URL

Also -- should we use this generally wherever we do reload()

lbwexler:
that is, we could just build this in to XH.reloadApp() and maybe avoid other caching issues as well.

ghsolomon:
> that is, we could just build this in to XH.reloadApp() and maybe avoid other caching issues as well.

Got it, yes that makes a lot of sense to me!

lbwexler:
greg -- think this looks much better two questions:

+ Do we know that the standard reload wouldn't have worked?  Just occurred to me that this was not clear and just  want to understand this better.
+ Should we check that the cacheBuster flag date to make sure that we are not in a death loop?   (Am fine if the answer is no)

ghsolomon:
> greg -- think this looks much better two questions:
> 
> * Do we know that the standard reload wouldn't have worked?  Just occurred to me that this was not clear and just  want to understand this better.
> * Should we check that the cacheBuster flag date to make sure that we are not in a death loop?   (Am fine if the answer is no)

I don't know that the standard reload wouldn't have worked.  This was belt-and-suspenders since the root issue had to do with the client holding onto an old version of the app for too long.  It might be unnecessary. Not sure the best way to test.

I don't think we _should_ end up in a death loop since `wasDiscarded` _shouldn't_ keep getting set to `true` after the first reload, but happy to put that check in to be safe.

PR #3582, opened 2024-02-21
More user friendly roles grid in admin

cnrudd:
Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry: yes.
- [x] Updated doc comments / prop-types: determined not required.
- [x] Reviewed and tested on Mobile: determined not required.
- [x] Created Toolbox branch: determined not required.

### DIFF
--- a/admin/tabs/general/roles/RoleModel.ts
+++ b/admin/tabs/general/roles/RoleModel.ts
@@ -18,6 +18,7 @@ import {action, observable, runInAction} from 'mobx';
 import moment from 'moment/moment';
 import {RoleEditorModel} from './editor/RoleEditorModel';
 import {HoistRole, RoleMemberType, RoleModuleConfig} from './Types';
+import {actionCol, calcActionColWidth} from '@xh/hoist/desktop/cmp/grid';
 
 export class RoleModel extends HoistModel {
     static PERSIST_WITH = {localStorageKey: 'xhAdminRolesState'};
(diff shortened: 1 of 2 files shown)


admin/tabs/general/roles/RoleModel.ts
```diff
@@ -270,6 +272,16 @@ export class RoleModel extends HoistModel {
...
             },
             columns: [
+                ...(this.readonly
+                    ? []
+                    : [
+                          {
+                              ...actionCol,
+                              actionsShowOnHoverOnly: true,
+                              width: calcActionColWidth(3),
+                              actions: [this.editAction(), this.cloneAction(), this.deleteAction()]
```
lbwexler, line 282: I personally think this is overkill -- if we think double-click editing not obvious, would just add an edit button

Dont't think a delete action or a clone action is warranted.

lbwexler: also -- if we do go forward with this PR, we conveniently have an `omit` property on column spec that would make this a lot neater!

lbwexler (approved):
Don't feel strongly but don't think this is really needed

admin/tabs/general/roles/RoleModel.ts
cnrudd: Thanks for the `omit` reminder.  Applied.

I trimmed the column to just the one action, `edit`.  Delete and Clone I don't think will be used too often, and can eventually be found in the context menu.

But for `edit`, I think its really important to make that more discoverable.
We have a new UI that is going to be used 90% of the time for adding/removing users to roles, and we've hidden the button that gets you there.

lbwexler: ok, thanks!

PR #3589, opened 2024-03-01
Adds support for app targeting for alert banners

### LINKED ISSUE #3587: Alert banners should support option to limit to one or more specific client apps
Currently the Admin Console allows you to display a configurable alert banner, but it will be shown across all client apps. We have some applications that use a single project / server to host distinct client apps with different purposes / user bases. For those projects, we might wish to display an alert banner to one client app but not another.

We should add a multi-select option to the alert banner configs to spec one or more client app codes.  If blank (default) would show to all client apps.  

A nice bonus would be to populate the select options with the list of all client app codes so admins don't need to know them.  I don't believe we have that now. Could use an optional config, but could also consider having our dev-utils [define a const](https://github.com/xh/hoist-dev-utils/blob/e83a9812a84e0d863519655bb96e7efd469e8efc/configureWebpack.js#L614) that apps could read - it knows the codes.

Do _not_ need the (much more complex) ability to show different banners at the same time to different apps.

minusleaf:
Fixes #3587

Hoist P/R Checklist
-------------------


admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -67,6 +73,13 @@ export class AlertBannerModel extends HoistModel {
         ];
     }
 
+    get clientAppOptions() {
+        return XH.getConf('xhClientApps', []).map(app => ({
```
amcclain, line 77: Don and I discussed the pros and cons of this approach - an alternative might involve having dev-utils inject a list, with some tradeoffs between the two options.  Let us discuss!

lbwexler: or have it specified in js code in the admin app model?

svc/AlertBannerService.ts
```diff
@@ -106,4 +108,25 @@ export class AlertBannerService extends HoistService {
     private onClose = () => {
         XH.localStorageService.set('xhAlertBanner.lastDismissed', Date.now());
     };
+
+    private isTargetedApp(clientApps: string[]): boolean {
+        return isEmpty(clientApps) || clientApps.some(targetApp => targetApp === XH.clientAppCode);
```
lbwexler, line 113: nit: use `Array.includes()`

lbwexler:
Looks great Don!   Happy to get your first PR in

lbwexler (approved):

admin/tabs/general/alertBanner/AlertBannerModel.ts
minusleaf: Anselm and I decided in the end to go the dev-utils injection approach. We thought this approach had the least annoying issues compared to the other two where it would force users to update dev-utils if they want the latest hoist-react and not get display names to populate dropdowns, using the app codes instead. It generally requires the least amount of additional effort in that if the clients have apps set up with clientAppCodes prior, then the select field would just automatically populate. This does assume the app codes are reasonably differentiating, like "contacts" vs "app12".

With the config approach, we run into the issue of user error being a bit punishing for when an admin misspells codes in spite of entering in the correct JSON format within the production app itself where a banner that was expected to appear just doesn't. Also, if clients want consistency in behavior between different stages, then the config would have to be copied at each level every time a change would be made (though this is infrequent).

Lastly with the app model approach, we thought that not all Hoist consumers would extend the exported AppModel off Admin, opting to just use the Admin as is given it's not an abstract class. By making it hardcoded, we'd force them to create a class that extends that if they already haven't, and it just seems like a clunky way to add in app codes, especially if the extension is only really just for the codes.

lbwexler: 👍🏻

PR #3602, opened 2024-03-14
Add support for Correlation ID's across fetch requests and error / ac…

ghsolomon:
…tivity tracking

Hoist P/R Checklist
-------------------


svc/FetchService.ts
```diff
@@ -72,6 +77,22 @@ export class FetchService extends HoistService {
         }
     }
 
+    /**
+     * Set the token (header name) to be used for correlationId tracking.  Set false to disable
```
amcclain, line 81: Rename to `setCorrelationIdHeaderName` ?

utils/js/LangUtils.ts
```diff
@@ -19,6 +19,7 @@ import {
     uniq,
     uniqBy
 } from 'lodash';
+import {v4} from 'uuid';
```
amcclain, line 22: Nice to have this as a util. 👍  We should add `uuid` to hoist-react `package.json` so it's an explicit dependency. Right now I think we're getting it as a transitive dev dependency only:

```
yarn why uuid
yarn why v1.22.19
[1/4] 🤔  Why do we have the module "uuid"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "uuid@8.3.2"
info Reasons this module exists
   - "@xh#hoist-dev-utils#webpack-dev-server#sockjs" depends on it
   - Hoisted from "@xh#hoist-dev-utils#webpack-dev-server#sockjs#uuid"
info Disk size without dependencies: "332KB"
info Disk size with unique dependencies: "332KB"
info Disk size with transitive dependencies: "332KB"
info Number of shared dependencies: 0
✨  Done in 0.18s.
```

svc/FetchService.ts
ghsolomon: Lee had the same thought.  Changed to `setCorrelationIdHeaderKey`. Does that sound ok to you, or would you prefer `name > key` ?

utils/js/LangUtils.ts
ghsolomon: Good call, meant to do this already. Updated the PR.  Also moved this method from `LangUtils` to `FetchService.generateCorrelationId(prefix?: string)` for various reasons.  But could keep it in `LangUtils` if you think it'd be useful in other contexts.

svc/FetchService.ts
amcclain: Great, thanks - it was mostly the word "token" that seemed too ambiguous as to whether we were talking about a key or a value.  I think with headers "name" is the common parlance for what we're talking about, but the net result is that this string is used as a key in an object we pass to the fetch API, so either term sounds fine and clear enough to me. 👍

svc/FetchService.ts
```diff
@@ -169,16 +216,20 @@ export class FetchService extends HoistService {
...
+        const headers = {
+            'Content-Type': isPost ? 'application/x-www-form-urlencoded' : 'text/plain',
+            ...defaultHeaders,
+            ...extraHeaders,
+            ...opts.headers
         };
+
+        const {correlationIdHeaderKey} = this;
+        if (opts.correlationId) {
+            throwIf(headers[correlationIdHeaderKey], 'Correlation ID already set via Fetch API.');
```
amcclain, line 228: I'd vote for warning here vs. throwing - does not seem like a critical enough conflict to warrant a throw. 

(Seems like there could be a chance that they are the same if a dev is doing something a bit weird/unnecessary). 

Error/warning message also seems a bit unclear to me - what about something like `Correlation ID already set in ${correlationIdHeaderKey} header within FetchOptions - ID specified directly at top level of FetchOptions will be ignored`. Case where FetchService is configured to auto-generate IDs might be worth calling out.

promise/Promise.ts
```diff
@@ -92,6 +92,11 @@ declare global {
          * @param options - TrackOptions, or simply a message string.
          */
         track(options: TrackOptions | string): Promise<T>;
+
+        /**
+         * @internal
```
amcclain, line 97: `@internal - set by FetchService to relay correlation IDs on responses to downstream error handling and tracking`

svc/FetchService.ts
```diff
@@ -79,24 +107,29 @@ export class FetchService extends HoistService {
...
+        opts = this.withResolvedCorrelationId(opts);
+        const ret = this.withDefaultHeadersAsync(opts, {Accept: 'application/json'}).then(opts =>
+            this.managedFetchAsync(opts, async r => {
+                if (this.NO_JSON_RESPONSES.includes(r.status)) return null;
+                return r.json().catchWhen('SyntaxError', e => {
+                    throw Exception.fetchJsonParseError(opts, e);
+                });
+            })
+        );
+        ret.correlationId = opts.correlationId as string;
```
amcclain, line 131: Do we have any sense that APIs might be returning correlation IDs in a response header and that we should provide a way to specify that header and have this code alternately check it?

I looked at one API in a client project that I know generates IDs without us specifying them (as we see them in error messages), and I do _not_ see those IDs returned in headers, so perhaps this is not a common pattern.  And in general if we are setting a corrId on the outbound request ourself, the entire idea is that the same ID should come back to us if an API were to be including them.  

So I guess I am imagining a case where we are not spec'ing IDs but the API is generating them _and_ returning them. Probably not anything we need to worry about, just wanted to flag as a possible consideration in case it ever comes up.

svc/FetchService.ts
```diff
@@ -314,6 +365,12 @@ export interface FetchOptions {
      */
     body?: any;
 
+    /**
+     * Unique identifier for this request, used for tracking and logging. If `true`, a UUID will be
+     * sourced from `LoadSpec.correlationId` or generated. If `false`, no correlationId will be set.
+     * */
```
amcclain, line 371: Don't think this is quite right - no defined behavior for true 

"Unique identifier for this request, used for tracking and logging. If unspecified, the service will check for a correlation ID on this config's `loadSpec` (if provided) or generate a new one if so enabled. If `false`, no correlationId will be set."

Exact wording / behavior might depend on our overall approach to enabling and originating these (see overall review comment)

amcclain (commented):
So the main questions I have are around where / how support for automatic ("on by default") correlation IDs should live.  Trying to reconcile some competing factors...

* Originating/generating earlier and automatically in `LoadSpec` seems powerful / attractive, as this supports a relatively common pattern of making multiple fetches to satisfy a single user request or action. 
    * It also gives you the option of getting a generated ID to pass to some other kind of non-FetchService based request or API that might take one - I don't have a specific example, but seems possible that an app might be using some other library - e.g. a redis or GRPC client  - that would be able to consume them. 
    * With the current impl, we can get a corrId on LoadSpec, but have to call the public load methods with a config object every time - no way to have it generated by default. 
* At the same time, I would still like auto ID support to be available on via direct uses of the FetchService API, in case you are calling it directly without having a LoadSpec on hand (common for the many calls apps might make made outside of `doLoadAsync`).
* Could we select someplace to default-enable and customize generally for both use cases, like the AppSpec or XH?
    * I'd also vote to enable by default, so we're talking about a place to disable and potentially override the generation routine.
    * FetchService API to customize the header key does of course make sense to live on FetchService.

package.json
```diff
@@ -84,7 +84,8 @@
     "router5-plugin-browser": "~7.0.2",
     "semver": "~7.6.0",
     "store2": "~2.13.1",
-    "ua-parser-js": "~1.0.2"
+    "ua-parser-js": "~1.0.2",
+    "uuid": "^9.0.1"
```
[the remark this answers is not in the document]
amcclain: They also did a v10 release last month

amcclain:
Related to decision on how / if to enable by default - what about for our own admin client?

CHANGELOG.md
```diff
@@ -6,6 +6,21 @@
 
 * Enhanced `markdown` component to support the underlying `components` prop from
   `react-markdown`. Use this prop to customize markdown rendering.
+* Added support for Correlation ID's across fetch requests and error / activity tracking:
+    * Requires `hoist-core >= v21.0.0`.
```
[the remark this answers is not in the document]
amcclain: Changed my mind on this - don't think it is required in general, tweaking CL now and will push

CHANGELOG.md
```diff
@@ -4,9 +4,28 @@
 
 ### 🎁 New Features
 
+* Added support for Correlation IDs across fetch requests and error / activity tracking:
+    * ⚠️ Upgrade to `hoist-core >= v20.4.0` to enable support for persisting correlation IDs to
```
amcclain, line 8: Expecting this can go out in core as the next minor release

ghsolomon:
> So the main questions I have are around where / how support for automatic ("on by default") correlation IDs should live. Trying to reconcile some competing factors...
> 
> * Originating/generating earlier and automatically in `LoadSpec` seems powerful / attractive, as this supports a relatively common pattern of making multiple fetches to satisfy a single user request or action.
>   
>   * It also gives you the option of getting a generated ID to pass to some other kind of non-FetchService based request or API that might take one - I don't have a specific example, but seems possible that an app might be using some other library - e.g. a redis or GRPC client  - that would be able to consume them.
>   * With the current impl, we can get a corrId on LoadSpec, but have to call the public load methods with a config object every time - no way to have it generated by default.
> * At the same time, I would still like auto ID support to be available on via direct uses of the FetchService API, in case you are calling it directly without having a LoadSpec on hand (common for the many calls apps might make made outside of `doLoadAsync`).
> * Could we select someplace to default-enable and customize generally for both use cases, like the AppSpec or XH?
>   
>   * I'd also vote to enable by default, so we're talking about a place to disable and potentially override the generation routine.
>   * FetchService API to customize the header key does of course make sense to live on FetchService.

All good points here.

I see the value in assigning the same Correlation ID to multiple, related requests that share a common `LoadSpec`. I also see the potential downside there- a `doLoadAsync` method which makes several calls and assigns the same correlation ID to all could make it harder to follow a single one of those requests.  Providing an option to enable either by default could make sense.  Point taken re: apps that choose not to use `FetchService`.  To take advantage of our Correlation ID support, those apps would need to lean entirely on `LoadSpec` as you pointed out or generate their own correlation IDs (presumably in some GRPCService or RedisService) and pass those along to `track`, `handleException`, etc.

Re: enabling by default, I agree that seems reasonable but have encountered client services that throw and return 500's when receiving unexpected headers. Would probably just want to note this as a breaking change if we enabled by default.

admin/tabs/activity/clienterrors/ClientErrorsModel.ts
```diff
@@ -66,13 +66,15 @@ export class ClientErrorsModel extends HoistModel {
                 {...Col.instance, hidden},
                 {...Col.day},
                 {...Col.dateCreatedWithSec, displayName: 'Timestamp'},
-                {...Col.impersonating, hidden}
+                {...Col.impersonating, hidden},
+                {...Col.correlationId, hidden}
```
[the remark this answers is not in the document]
ghsolomon: Seems reasonable to me

core/exception/Exception.ts
```diff
@@ -194,9 +194,15 @@ export class Exception {
     }
 
     private static createInternal(attributes: PlainObject, baseError: Error = new Error()) {
+        let correlationId: string;
+        const correlationIdHeaderKey = XH?.fetchService?.correlationIdHeaderKey;
+        if (correlationIdHeaderKey) {
+            correlationId = attributes.fetchOptions?.headers?.[correlationIdHeaderKey];
+        }
```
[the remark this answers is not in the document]
ghsolomon: I remember having this thought when reviewing with Lee and now can't think of a good reason not to move this to `createFetchException`.

svc/FetchService.ts
```diff
@@ -157,7 +190,21 @@ export class FetchService extends HoistService {
     //-----------------------
     // Implementation
     //-----------------------
-    private async withDefaultsAsync(
+
+    /** Resolve convenience options for Correlation ID to server-ready string */
+    private withResolvedCorrelationId(opts: FetchOptions): FetchOptions {
```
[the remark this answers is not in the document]
ghsolomon: Use of the word `resolved` here was to indicate that we're resolving the potential `boolean` values. Debated what to call this method. I'm good with dropping `resolved` if you think that's more confusing.

svc/FetchService.ts
ghsolomon: Yeah - I'm assuming this will mainly trip up apps that already implemented their own support for correlation ID's by setting default headers. In that case, I think we'd want to be pretty loud. But point taken... we could warn here. I like your message change.  If we warn instead of throw, maybe we still prefer the `opts.correlationId` and write over the previously set header? Not sure which is more likely to be "correct" in that case.

svc/FetchService.ts
ghsolomon: Thanks for flagging - we tweaked this behavior during CR (used to be as described in the docstring). Will update.

svc/FetchService.ts
ghsolomon: That's an interesting case you also mentioned in your original ticket.  I'm not sure what would be best, so just listing some scenarios to consider when discussing this:

Scenario A:
- Client makes a call and does not provide a Correlation ID.
- External API assigns a Correlation ID to the request, but client times out before the response is sent.

In the above case, the Correlation ID isn't very effective since we have no record of it on the client.  Seems like ideally we want the originator of a request to be assigning the correlation ID. As I write this, I'm realizing we should probably think about how correlation ID's might factor into our websocket service (if at all) where the roles are reversed.

Scenario B:
- Client makes a call and provides a Correlation ID
- External API responds with some other Correlation ID.

In the above case, do we warn? Track / log both IDs?

svc/FetchService.ts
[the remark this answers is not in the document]
amcclain: Yeah, it's a small thing but I do think it's more confusing than helpful in this context.

svc/FetchService.ts
amcclain: Agreed that in general the idea is that we originate correlation IDs (at least as far as fetchService is concerned). 

I don't think that this is likely, but it seems at least possible that we would have some project where, for whatever reason, the API wanted to take control of this and requested that we didn't send outbound IDs but did return them in a header. 

The scenario B above where the API misbehaves and returns an alternate ID seems like it's just an error - i.e. we are sending the outbound request in some other header than what the API expects.

More than anything, it felt like there was something to be said for reading and respecting any ID coming back on a request header.  BUT I then went and looked at two easy-to-check APIs  - the MSAL call to get tokens and a client-app specific API that I know uses recent .net and returns corrIDs with errors.  Neither of those had a correlation ID header on the response or returned one in the body, so maybe that's just "not a thing," or at least I'm no longer motivated.  We can leave as-is and revisit if we have a real use case.

svc/FetchService.ts
amcclain: It's such an edge case I think it's low-stakes either way, but I liked how you had it respecting the header - that feels like a more granular / specific / advanced usage, and the opts id could have been auto-generated.

lbwexler:
Hi -- Sorry for the delay here, and really don't have *that* much to say, other than to throw out the following "conclusions" that came out of Anselm and my discussion with Tom and John today.  Unless you strongly object, we are suggesting:

1) Having the auto installation of correlationIds apply *only* to LoadSpecs.  Its a simpler story to have that behavior in one place, and LoadSpec is already the higher order object where we are introducing our "low-code" magic.  If this change just makes the entire object more valuable, great.  On the other hand, FetchService is the lower-level API, and ok to be more explicit there.

2) Accordingly,  we want to move the global settings for generation/autogeneration *off* of FetchService.      Specifically, we were thinking of specifying them as follows:

- XH.correlationIdGenerator (getter/setter)
- XH.defaultCorrelationIdGenerator (getter)
Note We want to lose the concept of a built-in prefix -- you can do that if you want using your own generator function, and easy access to the default generator makes falling back to it easy.
 
- LoadSpec.autoGenerateCorrelationIds  (static property)
This could also be placed on XH, but I personally like placing it closer in the API to the thing it is acting on, and also think it might benefit conceptually from being seperated from the generator functions.  

3) We never want to throw in the case of conflict with specified ids from multiple sources -- we should just have a simple algorithm to resolve, and warn in the console at most.  We never want our instrumentation code to bring down the app.


Anselm please correct me if I missed anything, and happy to continue to hash this out tommorrow.

ghsolomon:
> Hi -- Sorry for the delay here, and really don't have _that_ much to say, other than to throw out the following "conclusions" that came out of Anselm and my discussion with Tom and John today. Unless you strongly object, we are suggesting:
> 
> 1. Having the auto installation of correlationIds apply _only_ to LoadSpecs.  Its a simpler story to have that behavior in one place, and LoadSpec is already the higher order object where we are introducing our "low-code" magic.  If this change just makes the entire object more valuable, great.  On the other hand, FetchService is the lower-level API, and ok to be more explicit there.
> 2. Accordingly,  we want to move the global settings for generation/autogeneration _off_ of FetchService.      Specifically, we were thinking of specifying them as follows:
> 
> * XH.correlationIdGenerator (getter/setter)
> * XH.defaultCorrelationIdGenerator (getter)
>   Note We want to lose the concept of a built-in prefix -- you can do that if you want using your own generator function, and easy access to the default generator makes falling back to it easy.
> * LoadSpec.autoGenerateCorrelationIds  (static property)
>   This could also be placed on XH, but I personally like placing it closer in the API to the thing it is acting on, and also think it might benefit conceptually from being seperated from the generator functions.
> 
> 3. We never want to throw in the case of conflict with specified ids from multiple sources -- we should just have a simple algorithm to resolve, and warn in the console at most.  We never want our instrumentation code to bring down the app.
> 
> Anselm please correct me if I missed anything, and happy to continue to hash this out tommorrow.

Got it. I’d love to talk more this morning, but here are some initial thoughts on having Correlation ID auto-generation live within `LoadSpec` and not `FetchService`:

- I believe many (most?) applications using Correlation IDs would need to write their own boilerplate to generate Correlation ID’s for calls made outside of `doLoadAsync` methods.  I imagine there will be a lot of these, since (at least in my experience) `doLoadAsync` tends to be used for fetching core model and service data and not for making `POST` / `PUT` / `DELETE` requests, which are arguably some of the most important calls to trace via Correlation ID in tools like Splunk.  I worry that having automatic Correlation ID generation as a feature of `LoadSpec` and not `FetchService` might confuse unfamiliar developers (Why do I sometimes need to generate a Correlation ID, and other times it’s generated for me? Oh, because it’s part of `LoadSpec`. Should I be creating a `LoadSpec` then?)
- I believe it would encourage developers to use the same Correlation ID for multiple calls within `doLoadAsync`, which are undoubtedly related as far as the UI is concerned, but are we sure this is a best practice? I can imagine a case where 3 calls are made, and one fails. Might it make it harder to trace the failed call downstream if it shares the same Correlation ID as other (successful) calls that may be related in the UI but not in the context of other downstream services?  Originally, I had imagined making Correlation ID a property of `LoadSpec` would be an advanced feature and not necessarily a standard practice we’d encourage. I’m definitely interested in hearing your thoughts on this, however, since it sounds like you feel differently.
- Understood about `FetchService` being a lower-level API. For what it’s worth, given that Correlation ID’s ultimately end up in request headers, and `FetchService` already has a method for modifying the default headers, it didn’t feel out of place to me that default Correlation ID generation could also live within `FetchService`, but understood if others feel differently.

lbwexler:
Greg and I had a great talk about this this morning and I think there are
some really good points here.  I mentioned to Greg the option of punting on
auto-generation completely in this round, until we get a bit more data.
It's really an additional feature on the

I agree that the id generation specification and code  could live on
FetchService, but to be clear we would still need an Alias on XH.  If I
want to generate one, I am definitely looking for the terse
`XH.generateCorrelationId()` or perhaps even terser....

ghsolomon:
Just pushed new changes here and for Toolbox:

- Removed global auto-generation of CID's
- Created `XH.genUUID()` (would we want this to just replace `XH.genID()`?
- Updated `FetchOptions.correlationId` to accept `true` for convenience and symmetry with `LoadSpec`

Issue #3572, opened 2024-01-31
Relay (at least some) Admin Activity Tracking filter params to server


amcclain:
Our Admin Console activity tracking functionality would be significantly enhanced if at least some key filtering criteria were passed to the server and folded into the query.  This would allow us to load longer date ranges for selected categories / messages / users, without running up against the overall row limits we require to be in place to avoid overly-large result sets.

Especially useful for relatively rare messages, which can be swamped by very chatty messages.

I don't think we would need to support every possible filter criteria on the server, but it raises interesting questions about our use of filter chooser.  Could be a nice baked-in example of a more advanced filter chooser case, where the output is observed and certain params are split off to a server-side query and others are applied locally.  (Unless we wanted to just commit all the way and make it an example of an all-server-side filter.)

amcclain:
Note that `TrackLogAdminController` in hoist-core supports some params already, but would need to be enhanced to support the different operators.  That could potentially prompt us to look again at https://github.com/xh/hoist-core/pull/282.

See #1981 for some additional notes on this.  I am going to close that older umbrella ticket in favor of this one.

PR #3608, opened 2024-03-21
Activity Tracking Panel Improvements to use server side filtering

jacob-xhio:
Activity Tracking Panel Refactor

Dependent on: https://github.com/xh/hoist-core/pull/341,

Addresses [xh/hoist-react#3572](https://github.com/xh/hoist-react/issues/3572)

Hoist P/R Checklist
-------------------


admin/tabs/activity/tracking/ActivityTrackingModel.ts
```diff
@@ -103,50 +102,35 @@ export class ActivityTrackingModel extends HoistModel {
...
-                        if (['>', '<='].includes(op)) {
-                            ret = moment(ret).endOf('day');
-                        }
-
-                        return ret.toDate();
-                    },
-                    valueRenderer: v => fmtDate(v),
-                    ops: ['>', '>=', '<', '<=']
+                    field: 'username',
+                    valueRenderer: v => v,
```
amcclain, line 114: Don't think you need these valueRendererers - that's the default, only need this prop when customizing

admin/tabs/activity/tracking/ActivityTrackingModel.ts
```diff
@@ -191,30 +175,40 @@ export class ActivityTrackingModel extends HoistModel {
         this.addReaction({
             track: () => {
                 const vals = this.formModel.values;
-                return [vals.startDay, vals.endDay, vals.maxRows];
+                return [vals.startDay, vals.endDay, vals.maxRows, this.filterChooserModel.value];
```
amcclain, line 178: It's a common pattern for us to define a getter on this model (eg `get query() {}`) then observe that here, pass it as `body` in `doLoadAsync` fetch call below.

admin/tabs/activity/tracking/ActivityTrackingModel.ts
```diff
@@ -191,30 +175,40 @@ export class ActivityTrackingModel extends HoistModel {
...
+                },
                 loadSpec
             });
 
+            const lookups = await XH.fetchJson({url: 'trackLogAdmin/lookups'});
+
+            filterChooserModel.fieldSpecs.forEach(spec => {
+                const {field} = spec;
+                if (lookups[field]) spec.values = lookups[field];
+            });
```
amcclain, line 210: We could run this lookups query just once, on init, especially if we had any reason to think it could be resource-intensive on the server. 

Reloading each time does have the advantage of picking up new discrete values that were added to the overall log since you first opened the tab. That's not something that's likely to happen _all_ that often, but of course it could. So I'm a bit on the fence.  I'd want to see how that call performs with a very large dataset, and we should generally be confident it performs well (e.g. those fields are marked for indexing)

If we keep reloading it each time, I'd suggest we kick that off in its own method and don't bother awaiting it here (or  letting it throw the main return for that matter, if it e.g. timed out).  While we don't want to plan for that call to fail, it is a bit ancillary. Certainly nice to not make the user wait on it before seeing the main results.

admin/tabs/activity/tracking/ActivityTrackingModel.ts
```diff
@@ -103,50 +102,35 @@ export class ActivityTrackingModel extends HoistModel {
...
+                    valueRenderer: v => v,
+                    enableValues: true
+                },
+                {
+                    field: 'browser',
+                    valueRenderer: v => v,
+                    enableValues: true
                 }
-            ],
-            persistWith: this.persistWith
```
amcclain, line 149: Thanks for removing the persistence here - just wanted to note that was a request I made as I think this is an example of a filter you likely apply once, while investigating something specific, and it's probably _not_ the case that you want to see it three days later when you next open up this tab in your app.

amcclain (commented):
Minor comments below.  We'll also need an update to the changelog marking this as a breaking change that will require the latest hoist-core.

If we felt that it was important, we could map out what it would take for this change to be backwards compatible with the current hoist-core.  Obviously the filtering wouldn't work as it without the updated back end.  And we don't want to keep / switch between both old and new implementations, but we could drop out the filter completely, so you would get a degraded functionality but still something.

All that said, we required a move to HC 18 with HR 60, so if you are going to take this next version of HR you have already been required to take the most recent breaking change on the core side.  So there's not really any advantage to us making it backwards compatible, I just wanted to talk through how I usually think about those decisions.

On the HC PR we can review if that change needs to be breaking and require this HR - i.e. are they linked together from both directions?

admin/tabs/activity/tracking/ActivityTrackingModel.ts
jacob-xhio: Going to opt to move this logic into initialization for now

PR #3627, opened 2024-04-10
Update tsconfig.json with `composite: true`

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "composite": true,
     "experimentalDecorators": true,
     "target": "ES2022",
     "module": "ES2022",
@@ -16,5 +17,6 @@
     },
     "lib": ["dom", "es2022"]
   },
+  "include": ["./**/*", "package.json"],
   "exclude": ["node_modules"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,22 +1,29 @@
 {
   "compilerOptions": {
     "composite": true,
-    "experimentalDecorators": true,
+    "emitDeclarationOnly": true,
+    "declarationDir": "./build/types",
+
     "target": "ES2022",
     "module": "ES2022",
+    "lib": ["dom", "es2022"],
+
+    "jsx": "react",
     "moduleResolution": "Node",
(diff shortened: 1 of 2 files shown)


amcclain:
This change is to support a client that wishes to use stricter `tsc` compilerOptions in their app code than we do in HR - specifically they have `noImplicitAny: true`.  Prior to this change, and due to the fact that we don't ship HR as a pre-compiled distribution, the stricter app-level tsconfig settings will override those of HR, causing many errors to be thrown in their app build.

The client is using dev-utils for the app build itself - so babel actually compiling the app for webpack not tsc - but still wants pre-commit tsc checks to return cleanly. They also use jest for automated testing, and something about that workflow also causes jest to fail due to the TS "errors" thrown by HR.

Adding `composite: true` to the HR tsconfig allows it to be referenced differently within an app - see [this change in Toolbox](https://github.com/xh/toolbox/commit/3e0763ee76730ae181ac23da8e4c0339b1d49bf6) for the app-level tsconfig modification required to reference HR in this way. (Further comments on that commit)

tsconfig.json
```diff
@@ -1,20 +1,27 @@
 {
   "compilerOptions": {
-    "experimentalDecorators": true,
+    "composite": true,
+    "emitDeclarationOnly": true,
+    "declarationDir": "./build/types",
```
amcclain, line 5: Participating in a composite build requires the project to emit something - can be declaration files only as we now have here.

tsconfig.json
```diff
@@ -1,20 +1,27 @@
...
+
     "paths": {
       "@xh/hoist/*": ["./*"]
-    },
-    "lib": ["dom", "es2022"]
+    }
   },
-  "exclude": ["node_modules"]
+  "include": ["./**/*", "package.json"],
+  "exclude": ["node_modules", "build"]
```
amcclain, line 26: New `include` now required, as with `composite:true` you need an include directive (or an even more explicit list of files).  Needed to add `package.json` due to import in EnvironmentService

Expanded `exclude` to include `build` as otherwise generated declarations from one run are attempted to be compiled by the next.  Build outputs clearly shouldn't be inputs.

PR #3638, opened 2024-04-23
Remove ping call from FetchService.ts; move 'ping check' suggestion to AppContainerModel.ts

cnrudd:
Remove ping call from FetchService.ts; move 'ping check' suggestion to AppContainerModel.ts.
This will help prevent misleading 'server unavailable' alerts if ping fails before an auth proxy session has been established.

Hoist P/R Checklist
-------------------

- [x] Updated doc comments / prop-types:  not required.
- [x] Created Toolbox branch: tested on TB, new branch not required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@
 * Provides admin support for Cluster-aware version of Hoist.
 * `Store` now supports multiple `summaryRecords`.
 
+### ⚙️ Technical
+
+* `ping` call removed from `FetchService.ts`.  `ping` check now suggested  in alert msg
+  if server unavailable error occurs during app initialization.
+
 ### 📚 Libraries
 
 * @ag-grid `30.x -> 31.x`
--- a/appcontainer/AppContainerModel.ts
+++ b/appcontainer/AppContainerModel.ts
@@ -175,7 +175,24 @@ export class AppContainerModel extends HoistModel {
             await this.completeInitAsync();
         } catch (e) {
             this.setAppState('LOAD_FAILED');
-            XH.handleException(e, {requireReload: true});
+
+            let ex = e;
+            if (ex.isServerUnavailable) {
+                const {baseUrl} = XH,
+                    pingURL = baseUrl.startsWith('http')
+                        ? `${baseUrl}ping`
+                        : `${window.location.origin}${baseUrl}ping`;
+
+                ex = XH.exception({
+                    name: 'Unable to Contact UI Server',
+                    message:
+                        'Client is unable to contact the UI server.  Please check the UI server at the ' +
+                        `following location: ${pingURL}`,
+                    details: ex
+                });
+            }
+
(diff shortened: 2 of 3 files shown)


amcclain:
I am in favor of removing that initial ping check - I can't really recall why we thought it was necessary.  Came up recently while looking at a client mobile app and where the time was going during initial load - it can be not-incredibly-fast in a mobile scenario, and in that case every second was counting.

That said, I don't know that it makes sense to ask a business user in an error message to visit some ping endpoint. What do we think anyone will actually do, and what would that tell them?  I'd vote for removing that part.

cnrudd:
> I am in favor of removing that initial ping check - I can't really recall why we thought it was necessary. Came up recently while looking at a client mobile app and where the time was going during initial load - it can be not-incredibly-fast in a mobile scenario, and in that case every second was counting.
> 
> That said, I don't know that it makes sense to ask a business user in an error message to visit some ping endpoint. What do we think anyone will actually do, and what would that tell them? I'd vote for removing that part.

Agreed.  I've made the msg more user friendly.  Also removed `Client` which always seemed a little wonky/techie to me.

CHANGELOG.md
```diff
@@ -22,6 +22,11 @@
 * Provides admin support for Cluster-aware version of Hoist.
 * `Store` now supports multiple `summaryRecords`.
 
+### ⚙️ Technical
+
+* `ping` call removed from `FetchService.ts`.  `ping` check now suggested  in alert msg
+  if server unavailable error occurs during app initialization.
```
amcclain, line 28: Need to fixup this message - we can clean it up before release as I am sure we'll loop over the CL

appcontainer/AppContainerModel.ts
```diff
@@ -175,7 +175,22 @@ export class AppContainerModel extends HoistModel {
...
+            let ex = e;
+            if (ex.isServerUnavailable) {
+                const {baseUrl} = XH,
+                    serverUrl = baseUrl.startsWith('http')
+                        ? baseUrl
+                        : `${window.location.origin}${baseUrl}`;
+
+                ex = XH.exception({
+                    name: 'Unable to Contact UI Server',
+                    message: `The web browser is unable to contact the UI server at ${serverUrl}`,
```
amcclain, line 188: How are we sure that a serverUnavailable exception is actually being thrown from the UI server?  It looks like this catch encompasses appModel.initAsync() which could presumably be making other calls.  Just checking that we should be this specific with our message and call out the UI server directly

CHANGELOG.md
cnrudd: done in new PR: https://github.com/xh/hoist-react/pull/3639

appcontainer/AppContainerModel.ts
cnrudd: @amcclain suggested changes are this new PR: https://github.com/xh/hoist-react/pull/3639

PR #3647, opened 2024-05-03
NumberEditor no longer activates on keypress of letter characters.

cnrudd:
I came across this issue while working on the agGrid 31 deprecations branch.

The problem:  in our NumberEditor in a grid, if you type a letter on the cell before the editor/field is rendered, the letter ends up in the number input field when the editor is rendered.

This is because we are using the BP NumericInput as a controlled component, and BP does not validate 'value' prop changes.  It only validates changes made from within the component (typing when the cursor is already in it).

Hoist P/R Checklist
-------------------

- [x] Reviewed and tested on Mobile: not required, desktop only.


desktop/cmp/input/NumberInput.ts
```diff
@@ -172,12 +177,18 @@ class NumberInputModel extends HoistInputModel {
         return val;
     }
 
-    override isValid(val: number): boolean {
+    override isValid(val: number | string): boolean {
```
amcclain, line 180: This new usage seems to be conflating two different concerns - whether an overall value is valid and whether or not a single initial keypress is valid to begin editing.  

I would like to better understand the job of this particular `isValid` implementation here, in the input model, and a) ensure this code is meeting that contract then b) determine if it does make sense to re-use for the new use case or if we should have a dedicated check there, perhaps a simpler one to confirm that the keystroke is 0-9?

When validating that _first_ keystroke, we aren't expecting any non-numeric characters even if we do have shortcuts or commas enabled. We also don't want to reject e.g. a first keystroke of `1` if a field has a min of `100` - reading the code here, looks like we could have that problem

cnrudd: @amcclain Great point about "min of 100".  So, I've rewritten the solution so that it applies only to the NumberEditor.  This solution is inspired by the ["custom numeric editor" example here](https://www.ag-grid.com/react-data-grid/cell-editors/).

I've changed the merge target to `[agGrid31x-reactiveCustomComponents](https://github.com/xh/hoist-react/tree/agGrid31x-reactiveCustomComponents)` since the callback used now is part of agGrid's new
reactiveCustomComponents: true setup.

desktop/cmp/grid/editors/NumberEditor.ts
```diff
@@ -19,6 +21,25 @@ export const [NumberEditor, numberEditor] = hoistCmp.withFactory<NumberEditorPro
...
+    // cancel the editing before it even starts if char is not a number.
+    const isCancelBeforeStart = useCallback(
+        () => eventKey?.length === 1 && '1234567890'.indexOf(eventKey) < 0,
+        [eventKey]
+    );
+
+    useGridCellEditor({
+        isCancelBeforeStart
+    });
+};
```
lbwexler, line 45: Might need a personal walkthrough of this.

desktop/cmp/input/NumberInput.ts
lbwexler: I am ok with this , as it seems to be a mod of ag-Grids own code around this--- thanks for the careful look anselm -- let us know if you are satisfied with the change.

PR #3656, opened 2024-05-09
Fix selectEditor to stop editing after value change

### LINKED ISSUE #3653: `SelectEditor` requires user to blur grid cell after input commit to set value on record
When editing a grid cell with a select editor, the user would naturally expect a just-selected value to commit to the StoreRecord - it's been "committed" to the input and in a form context would have been set.  But our inline editing grid requires the grid cell to be blurred for the new value to take.  

We would like to avoid this for select in particular.  (Could be an option if needed, but I don't know if there is any reason to maintain the current behavior.)

https://github.com/xh/hoist-react/assets/225806/97c3690a-ba71-40fc-b0f4-d68dfa803bf1

saba-mo:
Resolves #3653

Hoist P/R Checklist
-------------------


desktop/cmp/grid/editors/SelectEditor.ts
```diff
@@ -19,12 +19,19 @@ export const [SelectEditor, selectEditor] = hoistCmp.withFactory<SelectEditorPro
...
     render(props, ref) {
+        const quickEdit = !props.gridModel.fullRowEditing && !props.inputProps.enableMulti;
+
         props = {
             ...props,
             inputProps: {
                 hideDropdownIndicator: true,
                 hideSelectedOptionCheck: true,
                 selectOnFocus: false,
+                onCommit: () => {
```
lbwexler, line 30: just to be clear, if the application supplies an onCommit of its own, that will stomp on this behavior.  Maybe that 's ok, or maybe we want to rework this, so we always call both?

amcclain: I'd say that's tolerable - I don't imagine that would be common, and you could argue it falls under the general umbrella of "overriding some inputProps might break or change built-in behaviors" with these inputs.

Since it's non critical and is in the right direction (allowing app to override hoist) probably simplest to let it be.  If an app dev wonders why they aren't getting this subtle improvement, we can look here and know what to do :)

CHANGELOG.md
```diff
@@ -36,6 +36,9 @@ for more details.
 
 * Removed support for passing a plain object to the `model` prop of Hoist Components (previously
   deprecated back in v58). Use the `modelConfig` prop instead.
+* Improved `selectEditor` to stop editing when the value changes, rather than when the user clicks
+  away from the cell, making it more intuitive for users. This only occurs in the absence of
+  `GridModel.fullRowEditing` and `Select.enableMulti`.
```
amcclain, line 41: I'd vote for dropping this last sentence - it's a huge changelog entry already, that that's a bit more detailed than a developer needs in this context

PR #3666, opened 2024-05-20
Addition of global server disconnection banner

### LINKED ISSUE #3663: Improve Admin Console handling of server unavailable (e.g. over restart)
If you are browsing the Admin Console and lose your connection to the server - e.g. across a restart - the current handling is extremely scattershot.  Each tab or sub-tab fails in its own way, with a combination of modal alerts, inline error messages, and missing / mysterious exception handling.

For the admin console in particular, it would be great to have a more unified sense of "am I connected to the server in general?", managed more centrally, perhaps via a combination of websocket connection status with a fallback heartbeat / connectivity check of some kind (as websockets won't aways be enabled).

If the server is detected to be unavailable, we should show a clear alert that's visible across the entire admin app. Could consider if the app should remain browsable, in case you have some data already loaded, but I'm not too sure that would work out in practice, as many tabs blank out their data (again in an inconsistent way).  Might be better to replace the entire app with a placeholder indicating what's going on. Could even have a little countdown timer to the next liveness check or something cute like that.

I believe this would be some nice polish to the admin app - which is exactly the kind of app where you want to know really clearly and immediately if you've lost connectivity - and also provide an example strategy business apps could use themselves, if warranted.

oodarluz:
Added server connection service to ping end and display banner on threshold. Paired with  xhAppServerHealth config property.

Hoist P/R Checklist
-------------------


amcclain:
Thanks for putting this together Otito - I will catch up with you and take a look ASAP, by Friday for sure.

amcclain (commented):
Thanks for posting!  Apologies for my delayed review - and for the number of comments I managed to leave here 😄 

Happy to discuss more via phone or chat any time.

amcclain:
Also pls do the catch-up merge with `develop` when you have a chance 👍

PR #3672, opened 2024-05-23
Restore disabling of blueprint animations on popovers and tooltips (had b…

cnrudd:
…een disabled, but the animations came back on the BP5 upgrade)

This can be tested on the column chooser button in our toolbox demo grids, which opens a popover.
I also added a blueprint tooltip in my local, to test tooltips, and saw that they had no animations with this change.

The BP animations are very fast (300ms), but some users found them annoying anyways, and we had them turned off before, on BP 4.

This new css brings back the 'turn off' by matching and overriding the new BP5 css specificity without use of react-transition.  It might be an improvement, in that it turns off the opacity animation, too, not just the scale animation.

Fast animations are hard to inspect.  But Chrome Dev Tools has a tab for animations:
image

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry.
- [x] Reviewed and tested on Mobile: not required
- [x] Created Toolbox branch / PR: not required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 65.0.0-SNAPSHOT - unreleased
 
-### ⚙️ Technical
+### 🐞 Bug Fixes
 * Restore disabling of Blueprint animations on popovers and tooltips (regression in v63.0.0)
 
 ## 64.0.2 - 2024-05-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Changelog
 
 ## 65.0.0-SNAPSHOT - unreleased
-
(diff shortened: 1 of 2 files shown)


TomTirapani (approved):
I'll admit I can't really see the fast animations in develop, but the animation tab of the debugger confirms this does indeed work well :)

The code makes sense and the detailed comment does a great job explaining the rationale behind setting the "end of transition" values. Looks great!

ghsolomon:
> I'll admit I can't really see the fast animations in develop, but the animation tab of the debugger confirms this does indeed work well :)
> 
> The code makes sense and the detailed comment does a great job explaining the rationale behind setting the "end of transition" values. Looks great!

Looks great to me too!

PR #3686, opened 2024-06-10
Mobile orientation zoom fix

TomTirapani:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 65.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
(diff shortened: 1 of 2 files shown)


lbwexler:
[just mentioning to Tom the following idea]

it looks safe to me, but obviously I share your disappointment that it doesn't work unless in full screen mode.  Was wondering if worth trying to just reset the zoom level to 1 after every rotation (e.g. width = device-width or 100vw, height= device-height or 100vh)


also has the advantage I think of only having to touch this thing thing once.  The disadvantage of zooming out on orientation change does not seem like a big problem to me -- if you changing the orientation, you would not be surprised by that

PR #3702, opened 2024-06-25
Added new convenience method `XH.renderAdminApp()`

amcclain:
Hoist P/R Checklist
-------------------


amcclain:
I'm not overly attached to this if you think it's overkill.  Prompted by a prod issue that @jskupsik handled where a client admin app was still setup with `checkAccess: 'HOIST_ADMIN'` from before our introduction of `HOIST_ADMIN_READER`.  

Seemed like it was worth DRYing up this small set of almost-always-set-that-way configs for the admin console app.

In Toolbox, the admin appSpec can go from:

```
XH.renderApp({
    clientAppCode: 'admin',
    clientAppName: 'Toolbox Admin',
    componentClass: AppComponent,
    modelClass: AppModel,
    containerClass: AppContainer,
    isMobileApp: false,
    enableLogout: true,
    webSocketsEnabled: true,
    checkAccess: 'HOIST_ADMIN_READER',
    lockoutMessage: "Contact support@xh.io for information on Hoist's bundled Admin Console."
});
```

to:

```
XH.renderAdminApp({
    componentClass: AppComponent,
    modelClass: AppModel,
    containerClass: AppContainer,
    enableLogout: true,
    lockoutMessage: "Contact support@xh.io for information on Hoist's bundled Admin Console."
});
```

Unfortunately we still need the app devs to import and pass in the comp/container/model classes (even if they don't have a custom admin app model) as we can't have those imported by XH, but still better than nothing?

core/XH.ts
```diff
@@ -336,6 +337,27 @@ export class XHApi {
...
+     * code in XH, where it would be added to the bundles for all client apps.
+     */
+    renderAdminApp<T extends HoistAppModel>(
+        appSpec: SetOptional<AppSpec<T>, 'isMobileApp' | 'checkAccess'>
+    ) {
+        this.acm.renderApp({
+            clientAppCode: 'admin',
+            clientAppName: `${this.appName} Admin`,
+            isMobileApp: false,
+            webSocketsEnabled: true,
```
amcclain, line 355: Realizing this might not be appropriate to default due to need to opt-in on server - can check that tomorrow

core/XH.ts
```diff
@@ -92,8 +93,8 @@ export class XHApi {
     exceptionHandler: ExceptionHandler = new ExceptionHandler();
 
     //----------------------------------------------------------------------------------------------
-    // Metadata - set via webpack.DefinePlugin at build time.
-    // See @xh/hoist-dev-utils/configureWebpack.
+    // Metadata - the `xhFooBar` values on the right hand of these assignments are injected at build
```
lbwexler, line 96: prefer `xhXxx`

lbwexler:
+ I am fine with this -- agree.
+ We should also make websockets enabled by default on the server.  We do want them widely, if only for this admin, so why are we forcing every app to go through this.  Don't think that change is required for this change, but we should make it all the same.

core/XH.ts
amcclain: Will swap to that - it's what I typed first and then seemed like a lot of X's, but agree :)

amcclain:
> We should also make websockets enabled by default on the server.

Great - agree, will look at what that entails, also agree it doesn't need to be a blocker. Easy for an app w/o websockets to override, but combined with a change to enable by default will be a nice change.

amcclain:
Hoist Core default changed @ https://github.com/xh/hoist-core/commit/e59df229fb335be1fe84cb6bab5ad8dee7475f67

PR #3707, opened 2024-06-28
Overall type improvements

ghsolomon:
Hoist P/R Checklist
-------------------


cmp/badge/Badge.ts
```diff
@@ -5,14 +5,14 @@
...
-import {BoxProps, hoistCmp, HoistProps, Intent} from '@xh/hoist/core';
+import {BoxProps, hoistCmp, HoistPropsWithRef, Intent} from '@xh/hoist/core';
 import {TEST_ID} from '@xh/hoist/utils/js';
 import {splitLayoutProps} from '@xh/hoist/utils/react';
 import classNames from 'classnames';
 import {merge} from 'lodash';
 import './Badge.scss';
 
-export interface BadgeProps extends HoistProps, BoxProps {
+export interface BadgeProps extends HoistPropsWithRef<HTMLDivElement>, BoxProps {
```
amcclain, line 15: Realizing that I need to catch up with the `ref` handling I see we have throughout the toolkit. I had not realized that we were passing refs through on so many components, and I confess I don't know what the rules are here as to when / why we do that (or when/why app components should do the same).

cmp/filter/FilterChooserModel.ts
```diff
@@ -145,7 +146,7 @@ export class FilterChooserModel extends HoistModel {
     @observable.ref selectValue: string[];
     @observable favoritesIsOpen = false;
     @observable unsupportedFilter = false;
-    inputRef = createObservableRef<HTMLElement>();
+    inputRef = createObservableRef<HoistInputModel<null>>();
```
amcclain, line 149: Curious about the `null` passed to the generic here - what is that about / worth a comment?

cmp/form/Form.ts
```diff
@@ -21,7 +21,7 @@ import {FormModel} from './FormModel';
 /** @internal */
 export interface FormContextType {
     /** Defaults props to be applied to contained fields. */
-    fieldDefaults?: Partial<BaseFormFieldProps> & DefaultHoistProps;
+    fieldDefaults?: Partial<BaseFormFieldProps> & PlainObject;
```
amcclain, line 24: I'm assuming that use of `PlainObject` here means that you can put any key and TS will accept.  But if you specify a key that _is_ on `BaseFormFieldProps` we will get some validation on it, right?  

Is mobile vs. desktop split the thing that's driving this?

cmp/layout/Tags.ts
```diff
@@ -15,38 +15,39 @@ export const fragment = elementFactory(Fragment);
...
+export const pre = elementFactory('pre');
+export const span = elementFactory('span');
+export const strong = elementFactory('strong');
+export const table = elementFactory('table');
+export const tbody = elementFactory('tbody');
+export const td = elementFactory('td');
+export const th = elementFactory('th');
+export const thead = elementFactory('thead');
+export const tr = elementFactory('tr');
+export const ul = elementFactory('ul');
```
amcclain, line 53: Nice cleanup!

cmp/tab/TabModel.ts
```diff
@@ -38,7 +39,7 @@ export interface TabConfig {
     icon?: ReactElement;
 
     /** Tooltip for the Tab in the container's TabSwitcher. */
-    tooltip?: ReactNode;
+    tooltip?: JSX.Element | string;
```
amcclain, line 42: This is new to me - especially vs. `ReactElement` - would like to understand more

amcclain (commented):
I only read a little bit of the way through, but looks like a super in-depth and interesting change!  I left a few comments with questions - will catch up again when I return after OOO

cmp/filter/FilterChooserModel.ts
ghsolomon: I made HoistInputModel generic to provide a type for the inputRef. In cases where that isn't being used, I thought it made sense to pass null since inputRef.current will always be null.

cmp/form/Form.ts
ghsolomon: DefaultHoistProps was effectively an alias for PlainObject. Happy to revert this change, but in the process of reviewing some of this with Lee, we talked about the possibility of removing DefaultHoistProps entirely, so lots to consider here! I'm not entirely sure why this needs to be so generous-- might be desktop vs mobile, will investigate.

cmp/tab/TabModel.ts
ghsolomon: This change was to conform with Blueprint's accepted prop type (ReactNode was too broad)

lbwexler:
@ghsolomon -- we need to catch this up and give it another shot, or abandon it, I think.

lbwexler:
Closing this -- too difficult to maintain and reason about, and no immediate demand for such a breaking change.   

Cherry picked 1 bug fix, and created ticket https://github.com/xh/hoist-react/issues/4388 with the details on how to try again on this using an LLM.

PR #3708, opened 2024-06-28
Fix `datePicker` dates wrapping when used within a grid within a dialog

cnrudd:
…sed in a grid inside a dialogBody.

Resolves the issue described here:  https://github.com/xh/hoist-react/issues/862#issuecomment-2195113273

Hoist P/R Checklist
-------------------

- [x] Added CHANGELOG entry: added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 * Fix timing issue with missing validation for records added immediately to new store.
 * Mobile `TabContainer` now flexes properly within flexbox containers.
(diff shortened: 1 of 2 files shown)


amcclain (approved):
Thanks for tracking that down - added a comment to the original workaround as I needed to trace it back to understand why we are applying that `word-break` setting. I wonder if that was really the right approach - the bug here a great example of how it can go wrong - but given that this is the first concrete issue we've had reported, happy to keep it rolling.

amcclain:
Something is wrong with github at the moment and I can't merge from here - stuck on "checking for ability to merge automatically" for some reason.  Will check back in a bit, but to be clear am happy to have this merged any time

PR #3711, opened 2024-06-28
Infer prop types for components passed to elementFactory and createEl…

### LINKED ISSUE #3683: Add types to kit elementFactory wrappers
Currently, the wrappers exported from our `kit` package do not specify their props interface and therefore default to type `any`. 
 `elementFactory` is generic, so it will be easy to improve type-safety here.  Example change:

`export const alert = elementFactory(Alert)` becomes `export const alert = elementFactory<AlertProps>(Alert)`

A bigger lift - we could enhance `elementFactory` to infer the props interface for whatever component is passed to it.

For example:
```
export function elementFactory<
    T = any,
    P = T extends React.Component<infer P> | React.FunctionComponent<infer P> ? P : any
>(type: T): ElementFactory<P> {
```

ghsolomon:
…ement

closes #3683

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/elem.ts
+++ b/core/elem.ts
@@ -7,13 +7,15 @@
 import {TEST_ID} from '@xh/hoist/utils/js';
 import {castArray, isFunction, isNil, isPlainObject} from 'lodash';
 import {
+    ComponentType,
     createElement as reactCreateElement,
     isValidElement,
+    JSX,
     Key,
     ReactElement,
     ReactNode
 } from 'react';
-import {Some, Thunkable} from './types/Types';
+import {PlainObject, Some, Thunkable} from './types/Types';
 
 /**
  * Alternative format for specifying React Elements in render functions. This type is designed to
@@ -72,8 +74,17 @@ export type ElementSpec<P> = P & {
     $omit?: any;
 };
 
-export type ElementFactory<P = any> = ((...args: ReactNode[]) => ReactElement<P, any>) &
-    ((arg: ElementSpec<P>) => ReactElement<P, any>);
+/**
+ * Union type of all known React Component types supported by Hoist.
+ */
+export type ReactComponent = ComponentType | keyof JSX.IntrinsicElements;
+
+/**
+ * Factory function that can create a ReactElement from an ElementSpec or children.
+ * Hoist alternative to JSX.
+ */
+export type ElementFactory<P = any> = ((arg: ElementSpec<P>) => ReactElement<P, any>) &
+    ((...args: ReactNode[]) => ReactElement<P, any>);
 
 /**
  * Create a React Element from a Component type and an ElementSpec.
@@ -81,10 +92,13 @@ export type ElementFactory<P = any> = ((...args: ReactNode[]) => ReactElement<P,
  * This function is a thin-wrapper over `React.createChildren` that
  * consumes the ElementSpec format.
  *
- * @param type - React Component or string representing an HTML element.
+ * @param component - React Component or string representing an HTML element.
  * @param spec - element spec.
  */
-export function createElement<P = any>(type: any, spec: ElementSpec<P>): ReactElement<P, any> {
+export function createElement<C extends ReactComponent>(
+    component: C,
+    spec: ElementSpec<PropType<C>>
+): ReactElement<PropType<C>, C> {
     const {omit, item, items, ...props} = spec;
 
     // 1) Convenience omission syntax.
@@ -102,15 +116,17 @@ export function createElement<P = any>(type: any, spec: ElementSpec<P>): ReactEl
         }
     });
 
-    return reactCreateElement(type, props as P, ...children) as any;
+    return reactCreateElement(component, props, ...children) as ReactElement<PropType<C>, C>;
 }
 
 /**
- *  Create a factory function that can create a ReactElement from an ElementSpec.
+ *  Create a factory function that can create a ReactElement from an ElementSpec or list of children.
  */
-export function elementFactory<P = any>(type: any): ElementFactory<P> {
+export function elementFactory<C extends ReactComponent>(component: C): ElementFactory<PropType<C>>;
+export function elementFactory<P extends PlainObject>(component: ReactComponent): ElementFactory<P>;
+export function elementFactory(component: ReactComponent): ElementFactory {
     const ret = function (...args) {
-        return createElement<P>(type, normalizeArgs(args, type));
+        return createElement(component, normalizeArgs(args, component));
     };
     ret.isElementFactory = true;
     return ret;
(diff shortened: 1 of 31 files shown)


desktop/cmp/filechooser/FileChooser.ts
```diff
@@ -91,9 +87,10 @@ export const [FileChooser, fileChooser] = hoistCmp.withFactory<FileChooserProps>
                     accept,
                     maxSize,
                     minSize,
-                    maxFiles,
                     multiple: enableAddMulti,
-                    item: ({getRootProps, getInputProps, isDragActive, draggedFiles}) => {
+                    // Passing children directly since it is not possible to pass a function via
```
lbwexler, line 91: comment should be removed

desktop/cmp/grid/impl/filter/ColumnHeaderFilter.ts
```diff
@@ -38,7 +38,6 @@ export const columnHeaderFilter = hoistCmp.factory({
             ),
             popoverClassName: 'xh-popup--framed',
             position: 'right-top',
```
lbwexler, line 40: Do we need `rootBoundary: viewport` -- see link below

https://github.com/palantir/blueprint/wiki/Popover2-migration

ghsolomon: Good catch!

desktop/cmp/input/CodeInput.ts
```diff
@@ -473,7 +473,7 @@ const inputCmp = hoistCmp.factory<CodeInputModel>(({model, ...props}, ref) =>
                 className: 'xh-code-input__inner-wrapper',
                 item: textArea({
                     value: model.renderValue || '',
-                    ref: model.manageCodeEditor,
+                    inputRef: model.manageCodeEditor, // TODO - confirm this is correct change
```
lbwexler, line 476: indeed

desktop/cmp/input/DateInput.ts
```diff
@@ -71,7 +71,7 @@ export interface DateInputProps extends HoistProps, LayoutProps, HoistInputProps
      * Element to display inline on the right side of the input. Note if provided, this will
      * take the place of the (default) calendar-picker button and (optional) clear button.
      */
-    rightElement?: ReactNode;
+    rightElement?: JSX.Element;
```
lbwexler, line 74: ?

desktop/cmp/input/DateInput.ts
```diff
@@ -439,7 +439,7 @@ const cmp = hoistCmp.factory<DateInputProps & {model: DateInputModel}>(
                 enforceFocus: false,
                 modifiers: props.popoverModifiers,
                 position: props.popoverPosition ?? 'auto',
-                boundary: props.popoverBoundary ?? 'viewport',
+                boundary: props.popoverBoundary ?? 'clippingParents',
```
lbwexler, line 442: see comment above  on this-- do we use clippingParents elsewhere?

ghsolomon: No, `clippingParents` is the popover default for `boundary`, but I believe this should actually be `rootBoundary: props.popoverBoundary ?? 'viewport'` based on the migration wiki you found. That being said, it's currently not working, and I don't believe we've seen any issues?  Will look into this more.

desktop/cmp/input/NumberInput.ts
```diff
@@ -226,6 +226,9 @@ class NumberInputModel extends HoistInputModel {
     }
 }
 
+// Note: we don't use the `ref` here, but the presence of a second argument is required
```
lbwexler, line 229: Should be one-line comment at most, especially since repeated.  Does typescript complain if we were to take it away?

desktop/cmp/input/RadioInput.ts
```diff
@@ -112,7 +115,6 @@ const cmp = hoistCmp.factory<RadioInputModel>(({model, className, ...props}, ref
         inline: props.inline,
         selectedValue: model.renderValue,
         onChange: model.onChange,
-        testId: props.testId,
-        ref
+        [TEST_ID]: props.testId
```
lbwexler, line 118: nice catch!

desktop/cmp/input/DateInput.ts
ghsolomon: Changed since this gets passed to BP `InputGroup` which accepts `rightElement?: JSX.Element` (more constrained type than `ReactNode`)

desktop/cmp/input/NumberInput.ts
ghsolomon: No complaint from TS - just didn't want someone to see the unused second argument and remove it assuming it's cruft.

kit/react-select/index.ts
```diff
@@ -17,4 +17,5 @@ export const reactSelect = elementFactory(Select),
     reactCreatableSelect = elementFactory(Creatable),
     reactAsyncSelect = elementFactory(AsyncSelect),
     reactAsyncCreatableSelect = elementFactory(AsyncCreatable),
-    reactWindowedSelect = elementFactory(WindowedSelect);
+    // Typed as any due to issue with react-windowed-select types
```
lbwexler, line 20: how does this manifest?  Will typescript complain?  if so, comment not adding anything

desktop/cmp/input/DateInput.ts
lbwexler: is this a typescript error with BP's signature, or a runtime error?  If ReactNodes generally work fine, than we should just cast internally.

Don't want to limit ourselves based on BP, and typescript any more than neccessary

ghsolomon: I just tested `string`, `number`, `boolean` and `Iterable<ReactNode>` which are all valid `ReactNodes` but not `JSX.Elements`.  All work without causing runtime exceptions.

I hadn't noticed this before, but our type for `leftIcon` is `ReactElement`, whereas BP's is `IconName | MaybeElement`.  `IconName` is a union of supported string literals, and `MaybeElement = JSX.Element | false | null | undefined`.  It seems like if we're going to deviate from BP's types, these ought to be the same?

kit/react-select/index.ts
ghsolomon: Here's the TS error without casting `WindowedSelect`

Screenshot 2024-07-01 at 12 14 35 PM

core/elem.ts
```diff
@@ -65,13 +65,11 @@ export type ElementSpec<P> = P & {
...
     [TEST_ID]?: string;
-
+} & {
     //----------------------------
     // Technical -- Escape support
     //----------------------------
-    $items?: any;
-    $item?: any;
-    $omit?: any;
+    [K in keyof P & ('items' | 'item' | 'omit') as `$${K}`]?: P[K];
```
lbwexler, line 72: medicine worse than the cure here?

ghsolomon: Hmm not sure. Nice to have the types right for these "reserved" props for components that use them, though we don't expect that to happen often.

PR #3712, opened 2024-07-01
New Client-Side AuthModel

lbwexler:
See also toolbox branch with the same name

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/HoistAuthModel.ts
+++ b/core/HoistAuthModel.ts
@@ -27,12 +27,13 @@ export class HoistAuthModel extends HoistModel {
      *
      * It should return true if the user could be successfully authenticated, false if the user
      * is not authenticated, and throw in the case of technical failure. If false is returned and
-     * `AppSpec.enableLoginForm` is true the framework will enter the 'loginRequired' state and
-     * show the user a *manual* login form.
+     * `AppSpec.enableLoginForm` is true the framework will enter the 'LOGIN_REQUIRED' state and
+     * show the user a secure login form with username and password fields to be submitted securely
+     * to the server.
      *
-     *  The default implementation of this method simply checks the auth status on the server, which
-     *  may be appropriate for fully SSO (e.g. NTLM) based solutions.  Override to consult or
-     *  initialize third-party client resources such as OAuth.
+     * The default implementation of this method simply checks the auth status on the server, which
+     * may be appropriate for fully SSO (e.g. NTLM) based solutions.  Override to consult or
+     * initialize third-party client resources such as OAuth.
      */
     async completeAuthAsync(): Promise<boolean> {
         return this.getAuthStatusFromServerAsync();
--- a/core/HoistAuthModel.ts
+++ b/core/HoistAuthModel.ts
@@ -56,7 +56,7 @@ export class HoistAuthModel extends HoistModel {
      *
      * Return true if user is successfully logged in, otherwise false.
      */
-    async loginFromFormAsync(username: string, password: string): Promise<boolean> {
+    async loginWithCredentialsAsync(username: string, password: string): Promise<boolean> {
         const resp = await XH.fetchJson({
             url: 'xh/login',
             params: {username, password}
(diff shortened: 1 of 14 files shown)


core/XH.ts
```diff
@@ -326,6 +330,16 @@ export class XHApi {
         return this.identityService?.username ?? null;
     }
 
+    /**
+     * Logout the current user.
+     * @see HoistAuthModel.logoutAsync
+     */
+    async logoutAsync(): Promise<void> {
+        await this.authModel?.logoutAsync();
+        await this.appModel?.logoutAsync();
```
TomTirapani, line 339: Do we still want to support AppModel.logoutAsync(), given that apps can now provide their own HoistAuthModel implementation that handles any custom logic? Should that be the preferred method going forward?

core/HoistAuthModel.ts
```diff
@@ -0,0 +1,87 @@
...
+                throw e;
+            });
+    }
+
+    /**
+     * Process a manual login, submitted by user via form.
+     *
+     * Return true if user is successfully logged in, otherwise false.
+     */
+    async loginFromFormAsync(username: string, password: string): Promise<boolean> {
```
TomTirapani, line 59: Minor thing that I can't explain, but my brain is short-circuiting on the 'Form' + 'Form' combination :) Maybe rename this to e.g. `loginWithCredentials()`?

core/XH.ts
cnrudd, line 339: Can we explain in a comment why we are logging out twice here? So that app developers can understand whether or not they need to also log out twice in their own apps.

lbwexler: Yes, we can go ahead and remove that for clarity.

core/HoistAuthModel.ts
lbwexler: good suggestion, agree

core/HoistAuthModel.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ *  Base class for managing authentication lifecycle.
+ *
+ *  Hoist will consult this model early in  the initialization sequence, prior to
+ *  full Hoist initialization. This means that several core services (identity, configs, prefs)
+ *  will *not* be available, but it provides the app a hook to
+ *  do early service initialization or other work to support flows such as OAuth.
+ *
+ *  The base implementation of this class is minimal and would be adequate only for fully
+ *  transparent SSO based solutions such as NTLM.  Applications wishing to provide custom
```
cnrudd, line 18: This choice of words here, "fully transparent", doesn't clarify anything for me.  What would a "fully opaque" SSO based solution be? Or "half transparent"?  

I guess "transparent" is being used as a metaphor or an analogy for something, but I can't tell what.  Why is NTLM transparent?  Will all developers reading this understand why NTLM is transparent (except me)?

core/HoistAuthModel.ts
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Main Entry Point
+     *
+     * This method will be called *once* by Hoist early in the app loading lifecycle to determine
+     * if full app loading can proceed.
+     *
+     * It should return true if the user could be successfully authenticated, false if the user
+     * is not authenticated, and throw in the case of technical failure. If false is returned and
+     * `AppSpec.enableLoginForm` is true the framework will enter the 'loginRequired' state and
+     * show the user a *manual* login form.
```
cnrudd, line 31: All logins are "manual", even the ones where I have to point the phone to my face for biometric auth.
Can we say "show the user a login form with two fields, username & password, which when submitted will be checked against...."  What will these credentials be checked against?  "...JESPA/NTLM or credentials set in the YML config for local development"?

cnrudd (changes requested):
Looks like a solid change.   I left a few suggestions about how I felt the docs/comments could be clarified.

core/XH.ts
lbwexler: removed the second method, per toms suggestion

core/HoistAuthModel.ts
lbwexler: sorry or the hand waving

core/HoistAuthModel.ts
lbwexler: 👍🏻

PR #3720, opened 2024-07-09
Fixed numberInput losing quick edits due to debounce

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * Updated mobile `TabContainer` to flex properly within flexbox containers.
 * Fixed timing issue with missing validation for records added immediately to new store.
 * Fixed CSS bug in which date picker dates wrapped when `dateEditor` used in a grid in a dialog.
+* Fixed bug where inline grid edit of NumberInput was lost after quick navigation.
 
 ## 65.0.0 - 2024-06-26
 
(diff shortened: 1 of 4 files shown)


lbwexler:
wondering about the history of these, and what this debounce is doing

lbwexler:
Nervous about thrashing both the nuimberinput and our code

Also, just curious about what changed with our grid code/ag-Grid that is now making this problematic.  The debouncing of numberinput is fairly old.

If we wanted to keep the debounce generally, but felt that we needed to go without it in grid, we could expose it (privately or public) and have grid set it to `null`

amcclain:
Thanks @jskupsik and @lbwexler.  Agreed I would also like to better understand what caused this to regress.  

The bug up on a client call this morning (@minusleaf also attending) as a possible blocker. We've already merged the Hoist update on an app that needs to get out to production relatively soon.

For this case we need to provide the client with:

a) a better sense of the severity - are users likely to hit this in their normal work? Do we need to roll back the upgrade? 
b) timeline for a fix (or rollback) - we have some time pressure as their QA lead is heading OOO so they are looking for guidance on testing and to finalize the release this week

Thanks much guys - I know these kinds of bugs can be maddening

Issue #3723, opened 2024-07-12
DashContainer nested menus are incorrectly positioned


TomTirapani:
If you use groups for your DashViewSpecs they are grouped in the context menu into submenus. The second level of menus are incorrectly positioned - they seems to also open in the same position, rather than anchored to the prior menu item. This can make it impossible to select an item if the two menus do not align properly.
![Screenshot 2024-07-12 at 12 15 40](https://github.com/user-attachments/assets/b98927ad-d27f-4959-9435-72c7e8a1a797)

Note that this is not an issue with other forms of nested menus (e.g. grid context menus), which appear to be positioned correctly:

![image](https://github.com/user-attachments/assets/7a373e05-47ca-4054-8997-3c4bd9789581)

TomTirapani:
I have set up a Toolbox branch which demonstrates the issue:
https://github.com/xh/toolbox/tree/dashContainerMenuBug

PR #3724, opened 2024-07-15
Fix bug where `ContextMenu` items were being improperly positioned

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@
 * Improved redirect handling within beta `MsalClient` to use Hoist-provided blank URL (an empty,
   static page) for all iFrame-based "silent" token requests, as per MS recommendations. Intended to
   avoid potential race conditions triggered by redirecting to the base app URL in these cases.
+* Fixed bug where `ContextMenu` items were being improperly positioned.
+
+### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - uncommon use case)
+
+* `MenuItems` inside a desktop `ContextMenu` are now rendered in a portal, outside the normal
+  component hierarchy. This change was made to ensure that menu items are positioned properly
+  relative to their parent.  It should not affect the vast majority of applications, but could
+  impact apps that rely on specific CSS selectors targeting the previous DOM structure.
+
 
 ## 66.0.1 - 2024-07-10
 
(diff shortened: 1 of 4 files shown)


ghsolomon:
Fixes #3723 . Issue appears to be with the following CSS style:
https://github.com/xh/hoist-react/blob/c11785889a98fa533288b14da9e2cff0498535eb/kit/blueprint/styles.scss#L112-L117

By applying a `max-height` and `overflow-y: auto` to the `Menu` container, sub-menus which are rendered in `Popovers` do not get positioned properly.  This can be remedied by rendering the submenu popovers in portals. Per BP's docs:

> Using a Portal is necessary if any ancestor of the target hides overflow or uses very complex positioning.
https://github.com/palantir/blueprint/blob/9754f3d6e/packages/core/src/components/popover/popoverSharedProps.ts#L315

This is `true` by default for `Popovers`, but overridden by BP for `MenuItems` so that submenus will live in the same container per BP's docs:

> popoverProps: Props to spread to the submenu popover. Note that content and minimal cannot be changed and usePortal defaults to false so all submenus will live in the same container.
https://blueprintjs.com/docs/#core/components/menu.menu-item

An alternative solution would be to remove the `max-height` and `overflow-y: auto` styles we apply to `Menu`.  This would cause clipping when the menu height exceeds that of the viewport but would allow us to keep the existing `usePortal: false` behavior, which does have some benefits - 

> Not using a `Portal` can result in smoother performance when scrolling and allows the popover content to inherit CSS styles from surrounding elements

amcclain:
Thanks for diving into this, Greg.  I'm assuming since you setup the PR like this, you are thinking that the portal is the best overall option.  (I am inclined to agree.)

The style rules to ensure we don't overflow the viewport height are important for menus in general, as it is not difficult to end up with a big menu that can exceed the height of the viewport on a short screen or compressed window - we have seen this a number of times with the primary app menu, and it's one of those badly broken outcomes we really want to avoid (i.e. it looks really silly).  

We could consider not applying that rule to context menus in particular - those do seem less likely to overlfow vs. the main app menu - but if the portal approach works OK and we aren't seeing janky scrolling, etc. let's go for this.  I'll merge it now and we can keep an eye out.

ghsolomon:
Agreed - those were my thoughts as well.  It would be nice if BP built overflow handling into `Menu`, but I think this is a reasonable workaround.

TomTirapani:
Thanks for the research and for this fix, looks great!

PR #3735, opened 2024-07-26
Support deleting roles with dots or other special chars in their name

cnrudd:
Fixes issue described in this hois-core ticket: https://github.com/xh/hoist-core/issues/370

Breaking change due to corresponding required change in [hoist-core](https://github.com/xh/hoist-core/pull/376)

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes: BREAKING CHANGE
- [ ] Created Toolbox branch / PR: not required.


ghsolomon (approved):

amcclain (approved):
Checked and last HR release we had that required a HC change was v64, the big multi-instance upgrade.  Same story in the other direction (HC requiring HR).

Given small/targeted nature of this fix, it would be nice if we didn't need to force a breaking change, but at the same time the upgrade on both sides should be easy - if you have gotten to this point, both sides are trivial updates relative to what you have.  

If we really wanted to, we could have the delete method here check the HC version and swap its impl, likewise on the HC side we could look for body vs. params.  But given the above I do NOT think that's warranted.  Just wanted to talk it out here in the PR.

Let's check in with @lbwexler as we were planning releases for as soon as today to confirm if we're good with both sides bumping up.

cnrudd:
> Checked and last HR release we had that required a HC change was v64, the big multi-instance upgrade. Same story in the other direction (HC requiring HR).
> 
> Given small/targeted nature of this fix, it would be nice if we didn't need to force a breaking change, but at the same time the upgrade on both sides should be easy - if you have gotten to this point, both sides are trivial updates relative to what you have.
> 
> If we really wanted to, we could have the delete method here check the HC version and swap its impl, likewise on the HC side we could look for body vs. params. But given the above I do NOT think that's warranted. Just wanted to talk it out here in the PR.
> 
> Let's check in with @lbwexler as we were planning releases for as soon as today to confirm if we're good with both sides bumping up.

I don't think this fix is urgent.  I don't need it in the project I am working on right now.  No one else has found this bug.  I think we can wait and merge it in when the correlation id change is merged in, since that change also requires coordination between Hoist React and Hoist-Core.

amcclain:
> I think we can wait and merge it in when the correlation id change is merged in, since that change also requires coordination between Hoist React and Hoist-Core.

Great idea, all for it!

amcclain (approved):

PR #3737, opened 2024-07-29
Add `select.emptyValue` prop

Ryanseanlee:
Hoist P/R Checklist
-------------------


desktop/cmp/input/Select.ts
```diff
@@ -52,6 +52,13 @@ export interface SelectProps extends HoistProps, HoistInputProps, LayoutProps {
     /** True (default) to close the menu after each selection. */
     closeMenuOnSelect?: boolean;
 
+    /**
```
lbwexler, line 55: This docstring sounds a little too much like the spec ---

`Value to use when the input is empty. Defaults to null.
Typical usage is to set to [] for multi-selects.`

desktop/cmp/input/Select.ts
```diff
@@ -236,6 +243,10 @@ class SelectInputModel extends HoistInputModel {
         return !!this.componentProps.enableMulti;
     }
 
+    get emptyValue(): any {
+        return this.componentProps.emptyValue ?? null;
```
lbwexler, line 247: the only case where this *might* be a bit surprising is if you set emptyValue to `undefined`, but I think this is actually fine as is

desktop/cmp/input/Select.ts
```diff
@@ -449,7 +460,7 @@ class SelectInputModel extends HoistInputModel {
         if (isNil(internal)) return null;
```
lbwexler, line 460: shouldn't this also be `emptyValue`?   Also, I think we may need the *opposite* transform for an override to `toInternal`.

desktop/cmp/input/Select.ts
```diff
@@ -435,11 +434,11 @@ class SelectInputModel extends HoistInputModel {
     // (Exception for a null value, which we will only accept if explicitly present in options.)
     override toInternal(external) {
         if (this.multiMode) {
-            if (external == null) external = []; // avoid [null]
+            if (external == this.emptyValue) external = []; // avoid [null]
```
lbwexler, line 437: maybe we should be checking null OR emptyValue here?

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 * New `mergeDeep` method provided in `@xh/hoist/utils/js` as an alternative to `lodash.merge`,
   without lodash's surprising deep-merging of array-based properties.
 * Enhanced Roles Admin UI to support bulk category reassignment.
+* New `Empty Value` prop for select inputs, allowing for a custom value to be returned when select
```
amcclain, line 12: Suggest:

> Added new `select.emptyValue` prop, allowing a custom value to be....

In general for things like new props and configs, best to use the exact name of the prop

desktop/cmp/input/Select.ts
amcclain: I'm not sure we want to be comparing to `this.emptyValue` in general here.

* We def. want to keep the null check, as we might have new prop set to `emptyValue: []` but still need to take incoming external value of `null` to `[]` for the commented reason (avoid [null] in the map below).
* I think of new `emptyValue` as only applying for _outbound_ values - i.e. the control has ended up empty through one course or another, so return this value instead.  Comparing an _inbound_ value here is something different - more like "always treat this value as empty".  I wonder if we want that here (and below on 441 for the single-value case).  Was this part of the change necessary to avoid some obviously wrong behavior?
* `==` comparison won't return true if an array involved anyway (unlikely to be reference to the _exact same_ empty array instance)

amcclain:
Not sure where this left off or how to move forward.  Happy to find a time to discuss, or not overly involve myself. These classes are tricky and I don't want to put too many cooks in the kitchen.

haynesjm42 (approved):
Did some testing on this locally, everything looks good to me and is working consistently with other inputs in terms of how/when the external value is updated if the initial external value is different from what the input would set for it's empty/falsey state

PR #3738, opened 2024-07-29
Update fmtNumber so zeroPad would accept a number that would specify …

Ryanseanlee:
…the maximum number of decimal places a formated number should be zero padded

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/input/NumberInput.ts
+++ b/desktop/cmp/input/NumberInput.ts
@@ -8,7 +8,7 @@ import composeRefs from '@seznam/compose-react-refs';
 import {HoistInputModel, HoistInputProps, useHoistInputModel} from '@xh/hoist/cmp/input';
 import {hoistCmp, HoistProps, HSide, LayoutProps, StyleProps} from '@xh/hoist/core';
 import '@xh/hoist/desktop/register';
-import {fmtNumber, parseNumber} from '@xh/hoist/format';
+import {fmtNumber, parseNumber, Precision, ZeroPad} from '@xh/hoist/format';
 import {numericInput} from '@xh/hoist/kit/blueprint';
 import {wait} from '@xh/hoist/promise';
 import {TEST_ID, throwIf, withDefault} from '@xh/hoist/utils/js';
@@ -90,12 +90,8 @@ export interface NumberInputProps extends HoistProps, LayoutProps, StyleProps, H
      */
     valueLabel?: string;
 
-    /**
-     * True to pad with trailing zeros out to precision, default false.
-     * Can also be a number to specify a specific number decimal places out to which a formatted
-     * number should be zero-padded.
-     *  */
-    zeroPad?: boolean | number;
+    /** @see NumberFormatOptions.zeroPad */
+    zeroPad?: ZeroPad;
 }
 
 /**
@@ -133,12 +129,12 @@ class NumberInputModel extends HoistInputModel {
         throwIf(Math.log10(this.scaleFactor) % 1 !== 0, 'scaleFactor must be a factor of 10');
     }
 
-    get precision(): number {
-        return withDefault(this.componentProps.precision, 4);
-    }
-
     override get commitOnChange(): boolean {
         return withDefault(this.componentProps.commitOnChange, false);
+    }
+
+    get precision(): number {
+        return withDefault(this.componentProps.precision, 4);
     }
 
     get scaleFactor(): number {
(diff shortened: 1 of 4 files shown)


lbwexler:
Would be great to get one or two examples of how we see this being used

Ryanseanlee:
**Input value: .7
zeroPad: false (default)**
image
--------------------------------------------------------------------------------------------------------------
**zeroPad: true (takes precision zeroPad)**
image
--------------------------------------------------------------------------------------------------------------
**zeroPad: 3 (limits zeroPad to 3 but precision stays at 8 see next two screen shots)**
image
image

format/FormatNumber.ts
```diff
@@ -430,14 +430,19 @@ function calcStyleFromColorSpec(v: number, colorSpec: ColorSpec | boolean): CSSP
...
+        decimalCount = countDecimalPlaces(num);
 
     const config: Numbro.Format = {};
     let mantissa = undefined;
 
     if (precision % 1 === 0) {
         precision = precision < MAX_NUMERIC_PRECISION ? precision : MAX_NUMERIC_PRECISION;
-        mantissa = precision === 0 ? 0 : precision;
+        if (typeof zeroPad === 'number' && zeroPad < precision) {
+            mantissa = decimalCount < zeroPad ? zeroPad : decimalCount;
```
amcclain, line 442: We could compute `decimalCount` within this branch only. We want to ensure this method has as little overhead as possible - it can/will be called many, many, many times and we don't want to do any work that we don't need to do while formatting each number.

format/FormatNumber.ts
```diff
@@ -464,6 +469,16 @@ function buildFormatConfig(v, precision, zeroPad, withCommas, omitFourDigitComma
...
 }
 
+function countDecimalPlaces(number) {
+    const numberStr = number.toString(),
+        decimalIndex = numberStr.indexOf('.');
+
+    if (decimalIndex === -1) return 0;
+
+    const decimalStr = numberStr.slice(decimalIndex + 1);
+    return decimalStr.replace(/0+$/, '').length; // Remove trailing zeros then return count
```
amcclain, line 479: I think this could be faster/simpler with `return decimalIndex === -1 ? 0 : numberStr.length - decimalIndex - 1`. (without the slice or regex)

Pls also type this util fn args/return - good to be super careful / tight on typing in these widely used, library-level utils

desktop/cmp/input/NumberInput.ts
```diff
@@ -90,8 +90,12 @@ export interface NumberInputProps extends HoistProps, LayoutProps, StyleProps, H
...
     valueLabel?: string;
 
-    /** True to pad with trailing zeros out to precision, default false. */
-    zeroPad?: boolean;
+    /**
+     * True to pad with trailing zeros out to precision, default false.
+     * Can also be a number to specify a specific number decimal places out to which a formatted
+     * number should be zero-padded.
+     *  */
+    zeroPad?: boolean | number;
```
amcclain, line 98: There's a more important / direct `NumberFormatOptions` interface that needs to be updated.  

But agree it's good to update the doc here too, and on _mobile_ NumberInput.

format/FormatNumber.ts
```diff
@@ -430,14 +434,18 @@ function calcStyleFromColorSpec(v: number, colorSpec: ColorSpec | boolean): CSSP
...
+    const num = Math.abs(v),
+        config: Numbro.Format = {};
     let mantissa = undefined;
 
     if (precision % 1 === 0) {
         precision = precision < MAX_NUMERIC_PRECISION ? precision : MAX_NUMERIC_PRECISION;
-        mantissa = precision === 0 ? 0 : precision;
+        if (typeof zeroPad === 'number' && zeroPad < precision) {
+            const decimalCount = countDecimalPlaces(num);
+            mantissa = decimalCount < zeroPad ? zeroPad : decimalCount;
```
lbwexler, line 445: use lodash max() for readability?

format/FormatNumber.ts
```diff
@@ -464,6 +472,12 @@ function buildFormatConfig(v, precision, zeroPad, withCommas, omitFourDigitComma
     return config;
 }
 
+function countDecimalPlaces(number: number): number {
+    const numberStr = number.toString(),
+        decimalIndex = numberStr.indexOf('.');
```
lbwexler, line 477: tough for me to verify this just by looking

format/FormatNumber.ts
```diff
@@ -28,73 +38,87 @@ const UP_TICK = '▴',
         neutral: 'xh-neutral-val'
     };
 
+export type Precision = 'auto' | IntRange<0, 13>;
+export type ZeroPad = boolean | IntRange<1, 13>;
```
amcclain, line 42: We can decide if we want the range - wanted to try it out but could be annoying for apps that have a dynamic value they are looking to pass that is typed as `number` would require cast

PR #3739, opened 2024-07-29
Support passing configs through to OAuth client library constructors

### LINKED ISSUE #3732: Support passing configs through to OAuth client library constructors
Currently we hard-code the configuration for our AuthZeroClient and MsalClient, e.g. https://github.com/xh/hoist-react/blob/18aeddfa56e398e95429eed9a17824d1f181e6ac/security/authzero/AuthZeroClient.ts#L168, allowing only a small number of predefined configs to be set.

We should add an "escape hatch" to allow an admin to soft-configure arbitrary configuration values that get passed through to the ctor. As always, we can't guarantee that some combinations of custom options won't break our implementation, but we also can't anticipate (and don't want to replicate) every option that might need to be used across these SDKs.

amcclain:
* Fixes #3732

Hoist P/R Checklist
-------------------

### DIFF
--- a/security/authzero/AuthZeroClient.ts
+++ b/security/authzero/AuthZeroClient.ts
@@ -4,18 +4,25 @@
  *
  * Copyright © 2024 Extremely Heavy Industries Inc.
  */
-import {Auth0Client} from '@auth0/auth0-spa-js';
+import {Auth0Client, Auth0ClientOptions} from '@auth0/auth0-spa-js';
 import {XH} from '@xh/hoist/core';
 import {never, wait} from '@xh/hoist/promise';
 import {Token, TokenMap} from '@xh/hoist/security/Token';
 import {SECONDS} from '@xh/hoist/utils/datetime';
-import {throwIf} from '@xh/hoist/utils/js';
+import {mergeDeep, throwIf} from '@xh/hoist/utils/js';
 import {flatMap, union} from 'lodash';
 import {BaseOAuthClient, BaseOAuthClientConfig} from '../BaseOAuthClient';
 
 export interface AuthZeroClientConfig extends BaseOAuthClientConfig<AuthZeroTokenSpec> {
     /** Domain of your app registered with Auth0 */
     domain: string;
+
+    /**
+     * Additional options for the Auth0Client ctor. Will be deep merged with defaults, with options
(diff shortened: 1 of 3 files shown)


security/authzero/AuthZeroClient.ts
```diff
@@ -4,18 +4,25 @@
...
 export interface AuthZeroClientConfig extends BaseOAuthClientConfig<AuthZeroTokenSpec> {
     /** Domain of your app registered with Auth0 */
     domain: string;
+
+    /**
+     * Additional options for the Auth0Client ctor. Will be deep merged with defaults, with options
+     * supplied here taking precedence. Use with care, as overriding defaults may have unintended
+     * consequences or fail to work with Hoist's expected usage of the client library.
+     */
+    authZeroClientOptions?: Partial<Auth0ClientOptions>;
```
amcclain, line 25: Note name of config primarily inspired by `agOptions`

security/msal/MsalClient.ts
```diff
@@ -58,6 +58,13 @@ export interface MsalClientConfig extends BaseOAuthClientConfig<MsalTokenSpec> {
 
     /** The log level of MSAL. Default is LogLevel.Warning. */
     msalLogLevel?: LogLevel;
+
+    /**
+     * Additional options for the MSAL client ctor. Will be deep merged with defaults, with options
+     * supplied here taking precedence. Use with care, as overriding defaults may have unintended
+     * consequences or fail to work with Hoist's expected usage of the client library.
+     */
+    msalClientOptions?: Partial<msal.Configuration>;
```
lbwexler, line 67: Am a little concerned about the 'client' containing a 'client', and this sort of puts a fine point on it.  (The internal `client` property had been private.)  Sort of wish we had called our classes "XXXProvider" or (really anything else). 

Not sure what else we would at this point.  Would just calling it `msalOptions`/`auth0Options` help?

amcclain: Let's consider renaming the classes then - because we also have the top-level interfaces hosting these new properties - `MsalClientConfig.msalClientOptions` 🤯 .

If you think that would be manageable at our clients where we've accumulated a few AuthModels already, now is the time to pull the trigger on any better name options.  We DID say it was beta after all! 😄

PR #3743, opened 2024-07-31
Add `GroupingChooserModel.setDimensions()`

Ryanseanlee:
Hoist P/R Checklist
-------------------


cmp/grouping/GroupingChooserModel.ts
```diff
@@ -183,10 +183,21 @@ export class GroupingChooserModel extends HoistModel {
...
 
+    @action
+    setDimensions(dimensions: Array<DimensionSpec | string>) {
+        this.dimensions = this.normalizeDimensions(dimensions);
+        this.dimensionNames = keys(this.dimensions);
+    }
+
     @action
     setValue(value: string[]) {
         if (!this.validateValue(value)) {
```
Ryanseanlee, line 203: Should we have a "louder" error handle? Maybe a toast along with the logWarn on line 204.

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -183,10 +183,21 @@ export class GroupingChooserModel extends HoistModel {
...
+
         this.setValue(value);
         this.setFavorites(favorites);
     }
 
+    @action
+    setDimensions(dimensions: Array<DimensionSpec | string>) {
+        this.dimensions = this.normalizeDimensions(dimensions);
+        this.dimensionNames = keys(this.dimensions);
+    }
```
amcclain, line 199: We should use this in the ctor vs. duplicating code

cmp/grouping/GroupingChooserModel.ts
amcclain: We discussed showing a toast within setValue prob not a great idea - could be called rapidly / repeatedly by app reaction, etc. At the same time, we talked about how current handling is pretty subtle, with warning log only. I am not sure there is a better option here, or that we want to get into it with this change.

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -308,7 +319,7 @@ export class GroupingChooserModel extends HoistModel {
         return sortBy(
             this.favorites.map(value => ({
                 value,
-                label: this.getValueLabel(value)
+                label: this.getValueLabel(value) ?? value
```
amcclain, line 322: Why not have `getValueLabel` handle the edge case ?

cmp/grouping/GroupingChooserModel.ts
amcclain: It would be safer to use `_dimensions` as a private property to store the dims, with this setter and a public `get dimensions()` getter, to ensure this method is called when changing - i.e. they are not changed directly.  

Then could remove the new reaction and call `cleanStaleDims` here - more direct / synchronous

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -284,12 +299,14 @@ export class GroupingChooserModel extends HoistModel {
         return {displayName: genDisplayName(src.name), ...src};
     }
 
-    getValueLabel(value: string[]) {
-        return value.map(dimName => this.getDimDisplayName(dimName)).join(' › ');
+    getValueLabel(value: string[]): string {
+        return (
+            value.map(dimName => this.getDimDisplayName(dimName)).join(' › ') ?? value.toString()
```
amcclain: Output of `map` will never be null - `?? value.toString()` not getting us anything here

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -84,8 +84,9 @@ export class GroupingChooserModel extends HoistModel {
 
     @observable.ref favorites: string[][] = [];
 
-    dimensions: Record<string, DimensionSpec>;
-    dimensionNames: string[];
+    @observable.ref private _dimensions: Record<string, DimensionSpec>;
+    @observable.ref private dimensionNames: string[];
```
amcclain, line 88: Happy to discuss making these private - I asked Ryan to rename/make private `_dimensions` so it was clear that you must call `setDimensions` with its validation / pruning logic.

I then made `dimensionNames` private - did not seem likely that it needed to be public, although if we are concerned apps might be relying on it for some reason...  Wanted to keep some symmetry here.

Not sure how to best to tighten down privacy settings like this and harden classes - is an incremental approach helpful, or does it start to add more confusion than it is worth, without taking on a full review of each property and method ?

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -343,4 +336,41 @@ export class GroupingChooserModel extends HoistModel {
...
+
+    private cleanStaleDims() {
+        const {value, dimensionNames, allowEmpty} = this,
+            cleanValue = value.filter(dim => dimensionNames.includes(dim));
+
+        if (isEqual(value, cleanValue)) return;
+
+        if (isEmpty(cleanValue) && !allowEmpty) {
+            cleanValue.push(dimensionNames[0]);
+        }
```
amcclain, line 372: We made the choice here to reset the value to "the first dimension in the list" in the event that a dim change would leave it an unsupported state.  

We don't think that this will happen often, but felt that we needed to do *something* in this case.

It was difficult to know how to make handling of possible value changes and possible dimension changes consistent.  E.g. if an invalid value is set - where value contains dim not in dimensions - we just log a warning (invisible to user) and reject the value change.  

Rejecting the dim change (because value held something not in the new list) did not seem appropriate in the same way.  We could throw an exception instead, but in the end felt better to prioritize accepting and proceeding with the control's overall "setup" at the expense of any given/transient value it might be holding.

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -187,6 +190,15 @@ export class GroupingChooserModel extends HoistModel {
         this.setFavorites(favorites);
     }
 
+    @action
+    setDimensions(dimensions: Array<DimensionSpec | string>) {
+        throwIf(isEmpty(dimensions), 'Must provide valid dimensions available for selection.');
```
haynesjm42, line 195: I have a use-case where the user defines which fields are available as dimensions, and it is valid to have no dimensions defined. It would make my app code cleaner if this component and model could at least be instantiated/rendered with no dimensions configured when `allowEmpty == true` - avoiding the need to re-create models in cases like this is always a nice cleanup

haynesjm42 (approved):
Tested this out for my use case and worked perfectly.

Am now able to create the GroupingChooserModel as a `readonly` member of my model with no dimensions, and I can set the dimensions when I load initial state in my model or when the relevant state changes to determine the dimensions - no more re-creating models and much less/simpler code!

haynesjm42:
@Ryanseanlee I think we just need a changelog entry and this can be merged

amcclain:
Great thanks much - I'll add that in now and merge it. 🚀

PR #3744, opened 2024-07-31
Improve `HoistException` to correctly pass exception message to under…

ghsolomon:
…lying `Error` instance

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/exception/Exception.ts
+++ b/core/exception/Exception.ts
@@ -193,17 +193,15 @@ export class Exception {
         }) as FetchException;
     }
 
-    private static createInternal(
-        attributes: PlainObject,
-        baseError: Error = new Error(attributes.message)
-    ) {
+    private static createInternal(attributes: PlainObject, baseError?: Error) {
+        const {message, ...rest} = attributes;
         return Object.assign(
-            baseError,
+            baseError ?? new Error(message),
             {
                 isRoutine: false,
                 isHoistException: true
             },
-            attributes
+            rest
         ) as HoistException;
     }
 }
--- a/core/exception/Exception.ts
+++ b/core/exception/Exception.ts
@@ -193,7 +193,10 @@ export class Exception {
         }) as FetchException;
     }
 
-    private static createInternal(attributes: PlainObject, baseError: Error = new Error()) {
+    private static createInternal(
+        attributes: PlainObject,
+        baseError: Error = new Error(attributes.message)
+    ) {
         return Object.assign(
             baseError,
             {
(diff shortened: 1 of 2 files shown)


amcclain:
How about that!  Would not have guessed at the behavior of Console and Error here - had to play with it a bit in console to understand the difference:

Screenshot 2024-07-31 at 3 32 16 PM

core/exception/Exception.ts
```diff
@@ -193,7 +193,10 @@ export class Exception {
         }) as FetchException;
     }
 
-    private static createInternal(attributes: PlainObject, baseError: Error = new Error()) {
+    private static createInternal(
+        attributes: PlainObject,
+        baseError: Error = new Error(attributes.message)
```
cnrudd, line 198: After reading the docs at the bottom of this page:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
I am curious to know for what reasons our `Object.assign` approach was chosen over the 
`HoistException extends Error {` approach.

With the extends approach we would still be able to set:
this.isRoutine = false;
this.isHoistException = true;

Was the `extends` approach tried?

ghsolomon:
@amcclain, I was playing with this in the console as well and found that the message passed to the ctor also gets prepended to the `stack`, which is what I believe is being outputted to the console by default (at least in Chrome):
Screenshot 2024-08-01 at 10 23 17 AM

ghsolomon:
According to MDN, the `Error` object is largely non-standard (only the `cause`, `message`, `name`, `constructor` and `toString` members are standardized). 

For `stack` specifically:

> Note: The stack property is de facto implemented by all major JavaScript engines, and [the JavaScript standards committee is looking to standardize it](https://github.com/tc39/proposal-error-stacks). You cannot rely on the precise content of the stack string due to implementation inconsistencies, but you can generally assume it exists and use it for debugging purposes.

> Because the stack property is non-standard, implementations differ about where it's installed.

> In Firefox, it's an accessor property on Error.prototype.

> In Chrome and Safari, it's a data property on each Error instance

core/exception/Exception.ts
ghsolomon: I'm interested in promoting `HoistException` to its own class as well. Have you found any risks?

Curious about deleting the message key though for our current approach. I don't think that's ever necessary since `Exception.create(src: Error)` would just return (and mutate) the same error with 2 additional properties:
```
isRoutine: false,
isHoistException: true
```
No other attributes are passed in this case:
https://github.com/xh/hoist-react/blob/87173ce1b34baee1f8549e9e8e25f9a6ebbca014/core/exception/Exception.ts#L32

PR #3750, opened 2024-08-08
New handlers for FetchService

lbwexler:
This provides support I need for being able to catch 401s, in non-SSO based scenario, and respond appropriately by bringing up.

The handler for the successful response seemed like a good thing to add at the same time for symmetry, and we have seen a couple of places where it might have been useful, and avoided the need for  overriding FetchService to look for e.g. 200 returns with `success: false` 

Hoist P/R Checklist
-------------------

- [X ] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,7 @@
       tracking and error reporting and are available for review in the Admin Console.
     * New `XH.genUUID()` method to generate UUIDs for use as Correlation IDs or elsewhere.
 * `GridModel.contextMenu` will now accept `false` as a value to omit context menus.
-*  New global event handlers on `FetchService`. See `FetchService.addExceptionHandler()` and
-   `FetchService.addSuccessHandler()`.
-
+*  New global exception handlers on `FetchService`. See `FetchService.addExceptionHandler()`.
 
 ### 🐞 Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
       tracking and error reporting and are available for review in the Admin Console.
     * New `XH.genUUID()` method to generate UUIDs for use as Correlation IDs or elsewhere.
 * `GridModel.contextMenu` will now accept `false` as a value to omit context menus.
-* New global exception handlers on `FetchService`. See `FetchService.addExceptionHandler()`.
+* New global interceptors on `FetchService`. See `FetchService.addInterceptor()`.
 * New property `FetchOptions.asJson` to instruct `FetchService` to decode an HTTP response as JSON.
   Note that `FetchService` methods suffixed with `Json` will set this property automatically.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@
       tracking and error reporting and are available for review in the Admin Console.
     * New `XH.genUUID()` method to generate UUIDs for use as Correlation IDs or elsewhere.
 * `GridModel.contextMenu` will now accept `false` as a value to omit context menus.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
ghsolomon, line 1: Code all looks good to me.  Out of curiosity, what is the need for multiple success and exception handlers? As an alternative approach, was wondering if we could support a single `interceptor` object with this interface:

```
interface Interceptor {
    onFulfilled: (opts: FetchOptions, r: FetchResponse) => Promise<FetchResponse>;
    onRejected: (opts: FetchOptions, e: unknown) => Promise<FetchResponse>;
}
```

Then we could pass `onFulfilled` and `onRejected` as the 2 arguments to `then`.

ghsolomon: Not sure how much of a concern this is, but wanted to flag--

I wonder if individually pushing handlers onto the arrays could lead to some confusing bugs? I'm imagining an app that calls `FetchService.addExceptionHandlers` in a couple places where it's not obvious (or possibly even non-deterministic) which call would execute first (i.e. a couple services in `initAsync`).  Since `FetchService` invokes the handlers in order, you could get very different outcomes.  Wondering if it would be safer to require setting a new array of handlers rather than appending?

lbwexler: So, you raise interesting points, and your API looks like it would be more elegant/powerful/easier to reason about.  I am all over that, and will be interesting to see how to implement calling `then()` in a dynamic loop

 Definitely not wedded to what we have here, and actually I need something fairly limited for my use case (namely add a single exception handler to catch 401, idle the app and return never()).  Some thoughts:

Support for multiple handlers mainly for the 'onSuccess` case, where I imagined you might very likely have say,  different plugins/ app modules wanting to generally enhance fetch requests/responses with different, largely non-competing functionality.  You are right about the ordering mattering, but I was thinking/hoping it was not worth the complexity of making any guarantees there.   Hoping for the most part the ordering would be well-determined 
for applications that needed.

lbwexler:
@ghsolomon -- thanks again for taking a look, and your comments -- refactoring to try to take account of your comments, I realized that I have a problem with the `onResolved` side.   In particular, for some requests, we will be resolving to a processed `any` and for others to a `FetchResponse`

lbwexler:
ok @ghsolomon -- can you take another look?  Simplified this to deal with just the exception case.
Also -- made some orthogonal refactors/cleanups while working through this code that I think we should take.

ghsolomon:
Looks good to me.  One other thought - since we're supporting multiple exception handlers and only allowing developers to **append** their own, should we move our built-in library handlers out of `managedFetchAsync` and put them in that array initially?  Could be a nice way to demonstrate the pattern and simplify `managedFetchAsync` a bit.

I.e.  `private _exceptionHandlers: FetchExceptionHandler[] = [this.TIMEOUT_HANDLER, this.ABORTED_HANDLER, this.SERVER_UNAVAILABLE_HANDLER]`

svc/FetchService.ts
```diff
@@ -452,7 +464,7 @@ export interface FetchOptions {
     autoAbortKey?: string;
 
     /**
-     * True to decode the HTTP response as JSON.
+     * True to decode the HTTP response as JSON. Default false.
```
ghsolomon, line 467: Nit - unless I'm missing something, I don't think we default this to `false` anywhere? Maybe more accurate to say "Unset by default"?

lbwexler: Well, its a technical detail that its not defaulted to `false`, and we could potentially remedy that.  But from the users perspective is very much `falsy`!

ghsolomon (approved):

PR #3752, opened 2024-08-10
Improve `NumberFormatOptions.zeroPad` handling of unexpected precision

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/format/FormatNumber.ts
+++ b/format/FormatNumber.ts
@@ -498,11 +498,10 @@ function buildFormatConfig(
 
     // Calculate numbro mantissa and trimMantissa options based on precision and zeroPad settings.
     if (isNumber(zeroPad)) {
-        // Specific zeroPad set - find a specific mantissa value between zeroPad and precision.
-
-        // Calculate the required precision of the number after rounding (since rounding can change
-        // the non-zero decimal places of a number, especially in cases of js floating point
-        // arithmetic errors). This is already guaranteed to be less than or equal to precision.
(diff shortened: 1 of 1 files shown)


jskupsik:
I found a fairly-comprehensive list of all of the various techniques for "rounding js float numbers" to avoid the arithmetic error: https://stackoverflow.com/questions/11832914/how-to-round-to-at-most-2-decimal-places-if-necessary/48764436#48764436

I've ended up using the lodash `_.round()` function due to its simplicity, speed (faster than my rounding code), and the fact we have an upper bound of precision anyways.

amcclain (approved):
This looks great to me - took me a bit to grok, but that's more about me than anything with the code...

It's kinda funny how we were not rounding directly before (ourselves) - instead letting numbro do the work there.  I did check on their site and their `mantissa` arg does _round_ vs truncate - I wasn't 100% sure but they have some examples showing as much.

I'm OK with merging this, but given how this code is run EVERYWHERE would appreciate another pair of eyes. That said, it is strictly limited to the newer support for numeric `zeroPad`, which greatly limits exposure if, oh, I don't know, this new rounding ended up problematic at some scale.  I really don't see how it could be, I guess I'm just feeling a bit timid about this formatting code. 😄

amcclain:
Looked again and moving ahead - this is a solid change and again it is limited only to the newer usage of `zeroPad` as number.

PR #3763, opened 2024-08-22
feat(AppContainerModel): expose bindable var to allow custom init mes…

jacob-xhio:
…sages for hoist apps

Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/AppContainerModel.ts
+++ b/appcontainer/AppContainerModel.ts
@@ -19,6 +19,7 @@ import {Icon} from '@xh/hoist/icon';
 import {action, bindable, when as mobxWhen} from '@xh/hoist/mobx';
 import {never, wait} from '@xh/hoist/promise';
 import numbro from 'numbro';
+import {ReactNode} from 'react';
 import {createRoot} from 'react-dom/client';
 import {
     AlertBannerService,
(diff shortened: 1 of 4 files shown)


jacob-xhio:
Screenshot 2024-08-22 at 11 03 09 AM

appcontainer/AppContainerModel.ts
```diff
@@ -72,10 +72,16 @@ export class AppContainerModel extends HoistModel {
     appSpec: AppSpec = null;
     appModel: HoistAppModel = null;
 
+    //---------------------------------
+    // Configurable Application State
+    //--------------------------------
+    @bindable initializingLoadMaskMessage;
```
amcclain, line 78: Would add the type here - `ReactNode`

appcontainer/AppContainerModel.ts
```diff
@@ -316,6 +322,10 @@ export class AppContainerModel extends HoistModel {
         return !isEmpty(this.aboutDialogModel.getItems());
     }
 
+    setInitializingLoadMaskMessage(message: string) {
+        this.initializingLoadMaskMessage = message;
+    }
```
amcclain, line 327: Sorry as I think I had this in the spec, but realized that we've recently trended more towards exposing `@bindable` properties directly for use by callers - it's reasonable for apps to set them directly, and we can remove the setter here.  That tracks more with some other updates done recently.

To emphasize though that the property is a public API, please add a proper `/** */` doc string, in addition to the type.  Can replace that banner-style block comment

amcclain (commented):
Great thanks - small comments below, also pls add the same to mobile

appcontainer/AppContainerModel.ts
jacob-xhio: Resolved, added docstring explaining the purpose of the bindable.

PR #3764, opened 2024-08-22
feat(AlertBanner): allow banner visibility filtering by role and app …

jacob-xhio:
…version

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/general/alertBanner/AlertBannerModel.ts
+++ b/admin/tabs/general/alertBanner/AlertBannerModel.ts
@@ -16,6 +16,7 @@ import {isEqual, isMatch, sortBy, without} from 'lodash';
 
 export class AlertBannerModel extends HoistModel {
     savedValue;
+    roles;
     @observable.ref savedPresets: PlainObject[] = [];
 
     @managed
@@ -43,8 +44,8 @@ export class AlertBannerModel extends HoistModel {
                 initialValue: true
             },
             {
-                name: 'clientApps',
-                displayName: 'Client Apps',
+                name: 'limitTo',
+                displayName: 'Limit To',
                 initialValue: []
             },
             {name: 'created', readonly: true},
@@ -83,7 +84,7 @@ export class AlertBannerModel extends HoistModel {
                 formModel.values.intent,
                 formModel.values.iconName,
                 formModel.values.enableClose,
-                formModel.values.clientApps
+                formModel.values.limitTo
             ],
             run: () => this.syncPreview(),
             fireImmediately: true
@@ -92,6 +93,7 @@ export class AlertBannerModel extends HoistModel {
 
     override async doLoadAsync(loadSpec: LoadSpec) {
         await this.loadPresetsAsync();
+        this.roles = await this.getRolesAsync();
         const {formModel} = this;
         if (formModel.isDirty && loadSpec.isAutoRefresh) return;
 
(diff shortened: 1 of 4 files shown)


admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -302,4 +320,15 @@ export class AlertBannerModel extends HoistModel {
         await XH.alertBannerService.checkForBannerAsync();
         await this.refreshAsync();
     }
+
+    private async getRolesAsync() {
+        try {
+            const roles = await XH.fetchJson({
+                url: 'roleAdmin/list'
+            });
```
amcclain, line 328: This service might not always be enabled - some apps don't use Hoist's built-in role management.  In that case, this would throw.  Looks like we should have this call `roleAdmin/config` first to determine if enabled. If so, go on to load the roles. If not, might end up updating the UI, depending on how much we say about roles that's visible to users.

svc/AlertBannerService.ts
```diff
@@ -50,16 +50,18 @@ export class AlertBannerService extends HoistService {
...
+            {active, expires, publishDate, message, intent, iconName, enableClose, limitTo} = data,
             {lastDismissed, onClose} = this;
 
         if (
             !active ||
             !message ||
             (expires && expires < Date.now()) ||
             (lastDismissed && lastDismissed > publishDate) ||
-            !this.isTargetedApp(clientApps)
+            isEmpty(limitTo) ||
```
amcclain, line 61: This looks wrong - believe instead you want all the helper methods below to return true if no limits spec'd

svc/AlertBannerService.ts
```diff
@@ -109,8 +111,42 @@ export class AlertBannerService extends HoistService {
         XH.localStorageService.set('xhAlertBanner.lastDismissed', Date.now());
     };
 
-    private isTargetedApp(clientApps: string[]): boolean {
-        return isEmpty(clientApps) || clientApps.includes(XH.clientAppCode);
+    /**
+     * Returns `true` if there are no limitations on application or if the
+     * current application is included in the defined limits.
+     * Returns `false` otherwise.
```
amcclain, line 117: Nit - we can leave off the "returns false otherwise" - implied.  TBH these first two private methods don't really need comments - they are pretty clear on their own.  Could amend the first to `isTargetedClientApp` for a bit of extra clarity.  A comment on the role method to indicate that user needs to have at least one of the roles might be helpful.

admin/tabs/general/alertBanner/AlertBannerModel.ts
```diff
@@ -199,14 +202,29 @@ export class AlertBannerModel extends HoistModel {
...
+                label: `App: ${app}`,
+                value: `app-${app}`
+            }))
+            .concat([{label: `App Version: ${XH.appVersion}`, value: `appVer-${XH.appVersion}`}])
+            .concat(
+                this.roles?.map(role => ({
+                    label: `Role: ${role}`,
+                    value: `role-${role}`
+                }))
+            );
```
amcclain, line 225: Would alpha-sort these if we go this route

admin/tabs/general/alertBanner/AlertBannerPanel.ts
```diff
@@ -163,16 +163,14 @@ const formPanel = hoistCmp.factory<AlertBannerModel>(({model}) => {
...
-                                options: XH.clientApps.map(app => ({
-                                    label: app,
-                                    value: app
-                                }))
+                                noOptionsMessageFn: () =>
+                                    'Enter one or more app codes, app versions, or roles.',
+                                options: limitToOptions
                             }),
                             info: fragment(
                                 span('Specify what apps should the banner be shown to.'),
```
amcclain, line 176: This help text is now mismatched

amcclain (commented):
So I had not thought of combining these into a single property.  Appreciate the implementation here, trying to weigh pros/cons.

Pros 
* Simpler, maybe, in that there's one new (replacement) top-level property on the spec.
* and still one form control in the UI.

Cons 
* minor breaking change for any apps that had a preset saved with clientApps - I am not overly worried about it, there might be one or two, but if we added two new props for the new limits, the existing ones would still work
* would it be more clear to have dedicated controls for each limit? Thinking of how limits across dimensions are AND'd together, OR'd within.
* Limiting the current app version helpful in a way - although if you wanted to have a banner ready to show to an upcoming release you couldn't do it. If we had a dedicated text input you could put in current or next, etc.
* Don't want to save appVersion in presets - easier to selectively include if in dedicated field

I dunno - obviously making a case here for separate fields to some extent, but happy to discuss, do appreciate aspects of this approach

PR #3766, opened 2024-08-24
Mask dialog when saving role category

Ryanseanlee:
Hoist P/R Checklist
-------------------


admin/tabs/userData/roles/recategorize/RecategorizeDialog.ts
```diff
@@ -53,7 +55,7 @@ const tbar = hoistCmp.factory<RecategorizeDialogModel>(({model}) => {
             icon: Icon.check(),
             intent: 'success',
             disabled: model.categoryName == null,
-            onClick: () => model.saveAsync()
+            onClick: () => model.saveAsync().linkTo(model.savingTask)
```
lbwexler, line 58: the task should be linked to the promise chain in the model on line 58 (you can also include the refresh in that promise chain as well).  

We always want to push as much logic as this as we can into the model methods, and leave the UI as simple triggers to it -- that way if we add new ways to trigger saveAsync(), the masking will also be in place

PR #3774, opened 2024-09-04
New `ViewManager` for persisting named bundles of grid and other component state

Ryanseanlee:
Hoist P/R Checklist
-------------------

Remaining TODOs: 
- [ ] `Restore view` function when Default view is selected


svc/JsonBlobService.ts
```diff
@@ -36,20 +36,22 @@ export class JsonBlobService extends HoistService {
...
         type,
         name,
+        acl,
         value,
         meta,
         description
     }: {
         type: string;
         name: string;
+        acl?: string | PlainObject;
```
ghsolomon, line 46: Why `PlainObject`?

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+import {Icon} from '@xh/hoist/icon/Icon';
+import {menu, menuDivider, menuItem, popover} from '@xh/hoist/kit/blueprint';
+import {groupBy, keys, sortBy} from 'lodash';
+import {ReactNode} from 'react';
+import {manageDialog} from './impl/ManageDialog';
+import {saveDialog} from './impl/SaveDialog';
+import './PersistenceManager.scss';
+import {PersistenceManagerModel} from './PersistenceManagerModel';
+
+interface PersistenceManagerModelProps extends HoistProps {
```
ghsolomon, line 13: Believe our convention is to specify the model type as the generic argument to `HoistProps` (i.e. `interface PersistenceManagerModelProps extends HoistProps<PersistenceManagerModel>`). Then you don't need to specify the `model` type on line 14

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+                selectedObject,
+                isShared,
+                capitalPluralNoun,
+                manageDialogModel,
+                saveDialogModel
+            } = model;
+
+            return fragment(
+                hbox({
+                    className: 'persistence-manager',
```
ghsolomon, line 35: Believe we typically prefix our framework classNames with `xh` (i.e. `xh-persistence-manager`)

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+                    items: [
+                        popover({
+                            item: button({
+                                text: selectedObject?.name
```
ghsolomon, line 39: Nit - since `getHierarchyDisplayName` is null-safe, could simplify this slightly with `text: getHierarchyDisplayName(selectedObject?.name) ?? capitalPluralNoun`

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+                                    div({className: 'xh-popup__title', item: capitalPluralNoun}),
+                                    objMenu({minimal})
+                                ]
+                            }),
+                            placement: 'bottom-start'
+                        }),
+                        saveButton()
+                    ]
+                }),
+                manageDialogModel ? manageDialog({key: manageDialogModel.xhId}) : null,
```
ghsolomon, line 57: I don't think we need to specify the `key` to react to model changes anymore. Might be more Hoisty to use `omit` here rather than the ternary

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+//------------------------
+
+const saveButton = hoistCmp.factory<PersistenceManagerModel>({
+    render({model}) {
+        return button({
+            icon: Icon.save(),
+            tooltip: `Save changes to this ${model.noun}`,
+            intent: 'primary',
+            omit: !model.enableTopLevelSaveButton || !model.canSave,
+            onClick: () => model.saveAsync(null).linkTo(model.loadModel)
```
ghsolomon, line 74: Why pass `null` to `saveAsync`?

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+        });
+    }
+});
+
+const objMenu = hoistCmp.factory<PersistenceManagerModelProps>({
+    render({model, minimal}) {
+        const {pluralNoun, objects, loadModel} = model,
+            grouped = groupBy(objects, it => it.group),
```
ghsolomon, line 82: Nit - could use the string property name as the iteratee arg to `groupBy` (i.e. `groupBy(objects, 'group')`)

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+                        menuItem({
+                            icon: Icon.plus(),
+                            text: 'New...',
+                            onClick: () => model.createNewAsync().linkTo(loadModel)
+                        }),
+                        menuItem({
+                            icon: Icon.save(),
+                            text: 'Save',
+                            disabled: !model.canSave,
+                            onClick: () => model.saveAsync(null).linkTo(loadModel)
```
ghsolomon, line 107: Same question re: passing `null` here

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,214 @@
...
+            grouped = groupBy(objects, it => it.group),
+            sortedGroupKeys = keys(grouped).sort(),
+            items = [];
+
+        sortedGroupKeys.forEach(group => {
+            items.push(menuDivider({title: group}));
+            items.push(...hierarchicalMenus(sortBy(grouped[group], 'name')));
+        });
+
+        return menu({
```
ghsolomon, line 91: Maybe a little clearer not to use the `fragment` here?

i.e. ```
if (minimal) return items;
return [
    ...items,
    menuDivider(),
    menuItem({
...
```

File: desktop/cmp/persistenceManager/PersistenceManager.ts [hunk 10]
```diff
@@ -0,0 +1,214 @@
...
+        });
+    }
+});
+
+/**
+ * @param records
+ * @param depth  used during recursion, depth in the path string/hierarchy
+ * @returns an array of menuItem()s
+ */
+function hierarchicalMenus(records: PlainObject[], depth: number = 0): ReactNode[] {
```
ghsolomon, line 138: Interesting to think about whether a function like this should live at the component or model level.  Seems like this could be a method in the model that returns an object representing a menu structure, and that's what could be passed to a component capable of rendering it.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+ * save/update/delete their own private objects.
+ */
+export interface PersistenceManagerProps {
+    // /** Endpoint for rest controller. */
+    // url: string;
+
+    /** Key used in JsonBlob */
+    type: string;
+    /** User-facing name/label for an object managed by this model. */
+    noun: string;
```
ghsolomon, line 40: Completely get the variable name, but wonder if `noun` sounds a little funky? Maybe `entity`?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+ * mapping configs. This model loads all available views from a configured endpoint and exposes a `provider` property
+ * that can be passed to any HoistModel that takes a `persistWith` config.
+ *
+ * Objects managed by this system can be private or shared based on a user's company and/or app roles. The `acl` field
+ * on each persisted object determines if and how it is shared. Users with the configured `adminRole` can share saved
+ * objects to all app users and save changes to shared objects directly. Users with the configured `editorRole` can
+ * share objects to their own company only and modify shared objects shared only within their company. All users can
+ * save/update/delete their own private objects.
+ */
+export interface PersistenceManagerProps {
```
ghsolomon, line 33: Should we rename this `PersistenceManagerConfig`? `Props` typically refers to the props passed to a React component. Believe we typically suffix the objects passed to our model constructors with `Config`

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    // /** Endpoint for rest controller. */
+    // url: string;
```
ghsolomon, line 35: Commented out URL obsolete / good to remove?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    /** Key used in JsonBlob */
+    type: string;
+    /** User-facing name/label for an object managed by this model. */
+    noun: string;
+    /** Whether user can publish or edit globally shared objects. */
+    canManageGlobal: Thunkable<boolean>;
+    /** Async callback triggered when view changes. Should be used to recreate the affected models. */
+    onChangeAsync: (value: PlainObject) => void;
```
ghsolomon, line 44: Would be nice to get more specific typing here. Could make `PersistenceModel` generic, where the generic argument is the serialized type.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    /** Used to persist this model's selected ID. */
+    persistWith: PersistOptions;
+    /** True (default) to render a save button alongside the primary menu button when dirty. */
+    enableTopLevelSaveButton?: boolean;
+    /** Fn to produce a new, empty object - can be async. */
+    newObjectFnAsync?: () => PlainObject;
```
ghsolomon, line 50: Same comment as above - could be a closure that returns the generic type for this model

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+}
+
+export class PersistenceManagerModel extends HoistModel {
+    //------------------------
+    // Persistence Provider
+    // Pass this to models that implement `persistWith` to include their state in the view.
+    //------------------------
+    provider: PersistOptions = {
```
ghsolomon, line 58: Can make `readonly`

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+        getData: () => cloneDeep(this.pendingValue ?? this.value ?? {}),
+        setData: value => this.mergePendingValue(value)
+    };
+
+    @observable.ref @managed manageDialogModel: ManageDialogModel;
+
+    @observable.ref @managed saveDialogModel: SaveDialogModel;
+
+    enableTopLevelSaveButton: boolean = true;
```
ghsolomon, line 67: Make `readonly`?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    newObjectFn: () => PlainObject;
```
ghsolomon, line 69: Make `readonly` and make return type generic?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    /** Current state of the active object, can include not-yet-persisted changes. */
+    @observable.ref pendingValue: PlainObject = null;
```
ghsolomon, line 72: Make type generic?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    @observable.ref views: PlainObject[] = [];
```
ghsolomon, line 74: Make type generic?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    @bindable selectedId: StoreRecordId;
+
+    /** Reference Key used to query JsonBlobs */
+    type: string;
```
ghsolomon, line 79: Make readonly?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    /** Configured word to label an object persisted by this model. */
+    noun: string;
```
ghsolomon, line 82: Make readonly? In general, would group the immutable properties separate from the observable state

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    get pluralNoun(): string {
```
ghsolomon, line 84: Not sure these methods on the model are adding much value. Could add `@computeOnce` decorators if you want to cache these. Otherwise, I think I would just call the lodash methods directly?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+        return capitalize(this.pluralNoun);
+    }
+
+    private readonly canManageGlobalFn: Thunkable<boolean>;
+
+    get canManageGlobal(): boolean {
+        return executeIfFunction(this.canManageGlobalFn);
+    }
+
+    onChangeAsync?: (value: PlainObject) => void;
```
ghsolomon, line 102: Make type generic

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    get value(): PlainObject {
+        return this.selectedObject?.value;
```
ghsolomon, line 105: Make type generic

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    get objects(): PlainObject[] {
+        return this.views;
```
ghsolomon, line 109: Make type generic

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    get selectedObject(): PlainObject {
```
ghsolomon, line 112: Make type generic

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    get capitalNoun(): string {
+        return capitalize(this.noun);
+    }
+
+    get capitalPluralNoun(): string {
+        return capitalize(this.pluralNoun);
+    }
+
+    private readonly canManageGlobalFn: Thunkable<boolean>;
```
ghsolomon, line 96: Would separate the private and public members

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+        onChangeAsync,
+        persistWith,
+        canManageGlobal,
+        enableTopLevelSaveButton = true,
+        newObjectFnAsync
+    }: PersistenceManagerProps) {
+        super();
+        makeObservable(this);
+
+        throwIf(
```
ghsolomon, line 149: I would take this out. It's clear from the type that `onChangeAsync` is a required property.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    override async doLoadAsync(loadSpec: LoadSpec) {
+        const rawViews = await XH.jsonBlobService.listAsync({type: this.type, includeValue: true});
+
+        runInAction(() => (this.views = this.processRaw(rawViews)));
+
+        const {objects} = this;
+        // Auto-create an empty view if required
+        if (!objects.length) {
+            const newValue = await this.newObjectFn(),
+                newObject = await XH.jsonBlobService.createAsync({
```
ghsolomon, line 195: Wonder if we should be wrapping these in a `try` / `catch` and handling possible fetch exceptions

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+            await this.saveAsync(true);
+        }
+    }
+
+    async saveAsync(skipToast: boolean = false) {
+        const {token, id} = this.selectedObject;
+        if (this.isShared) {
+            if (!(await this.confirmShareObjSaveAsync())) return;
+        }
+        await XH.jsonBlobService.updateAsync(token, {
```
ghsolomon, line 229: Same comment as above re: `try / catch`. What if we fail to save?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+            name,
+            description,
+            value: newValue,
+            isAdd: true
+        });
+    }
+
+    @action
+    openSaveDialog(objStub: ObjStub) {
+        this.saveDialogModel = new SaveDialogModel(this, objStub);
```
ghsolomon, line 264: Interesting pattern here to create a new `SaveDialogModel` every time rather than one `readonly` model with an `isOpen` state

desktop/cmp/persistenceManager/impl/ManageDialog.ts
```diff
@@ -0,0 +1,160 @@
...
+    displayName: 'ManageDialog',
+    model: uses(ManageDialogModel),
+
+    render({model}) {
+        return dialog({
+            isOpen: true,
+            icon: Icon.gear(),
+            title: `Manage ${model.parentModel.capitalPluralNoun}`,
+            className: 'persistence-manager__manage-dialog',
+            style: {width: 800, height: 475},
```
ghsolomon, line 26: This seems like it might be a wide width for the framework to assume will be available.  Might be worth setting a `maxWidth` of `90vw` to accommodate for smaller browser windows

desktop/cmp/persistenceManager/impl/SaveDialogModel.ts
```diff
@@ -0,0 +1,86 @@
+import {FormModel} from '@xh/hoist/cmp/form';
+import {HoistModel, managed, PlainObject, TaskObserver, XH} from '@xh/hoist/core';
+import {lengthIs, required} from '@xh/hoist/data';
+import {makeObservable} from '@xh/hoist/mobx';
+import {PersistenceManagerModel} from '../PersistenceManagerModel';
+
+export class SaveDialogModel extends HoistModel {
+    parentModel: PersistenceManagerModel;
+
+    @managed saveTask = TaskObserver.trackLast();
```
ghsolomon, line 10: `TaskObserver` does not have a `destroy()` method, so no need to mark `managed`. However, can make `readonly`

desktop/cmp/persistenceManager/impl/SaveDialogModel.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    @managed formModel = new FormModel({
```
ghsolomon, line 12: Would make `readonly`. Model config seems lengthy enough to break out into its own method (i.e. `@managed formModel = this.createFormModel()`

desktop/cmp/persistenceManager/impl/SaveDialogModel.ts
```diff
@@ -0,0 +1,86 @@
...
+                            return `An entry with name "${value}" already exists`;
+                        }
+                    }
+                ]
+            },
+            {name: 'description'}
+        ]
+    });
+
+    objStub: ObjStub;
```
ghsolomon, line 30: `readonly`?

desktop/cmp/persistenceManager/impl/SaveDialogModel.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (!isValid) return;
+
+        const newObj = await XH.jsonBlobService
+            .createAsync({
+                type: this.parentModel.type,
+                name,
+                description,
+                value: objStub.value
+            })
+            .catchDefault();
```
ghsolomon, line 73: If this throws, the exception will be caught and surfaced to the user, but the dialog will still close. Seems like this isn't what we'd want.

desktop/cmp/persistenceManager/impl/SaveDialogModel.ts
```diff
@@ -0,0 +1,86 @@
...
+            description: this.isAdd ? `` : objStub.description
+        });
+    }
+
+    close() {
+        this.parentModel.closeSaveDialog();
+    }
+
+    async saveAsAsync() {
+        return this.doSaveAsAsync().linkTo(this.saveTask).catchDefault();
```
ghsolomon, line 53: Looks like this only catches exceptions thrown by `parentModel.refreshAsync()` -- not sure that's right

desktop/cmp/persistenceManager/impl/ManageDialogModel.ts
```diff
@@ -0,0 +1,227 @@
...
+import {makeObservable, observable} from '@xh/hoist/mobx';
+import {wait} from '@xh/hoist/promise';
+import {PersistenceManagerModel} from '../PersistenceManagerModel';
+
+export class ManageDialogModel extends HoistModel {
+    parentModel: PersistenceManagerModel;
+
+    @observable isOpen: boolean = false;
+
+    @managed gridModel: GridModel;
```
ghsolomon, line 15: Can make `readonly`

desktop/cmp/persistenceManager/impl/ManageDialogModel.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    @managed formModel: FormModel;
```
ghsolomon, line 17: Can make `readonly`

desktop/cmp/persistenceManager/impl/ManageDialogModel.ts
```diff
@@ -0,0 +1,227 @@
...
+        return new GridModel({
+            sortBy: 'name',
+            groupBy: 'group',
+            stripeRows: false,
+            rowBorders: true,
+            hideHeaders: true,
+            showGroupRowCounts: false,
+            store: {
+                fields: [
+                    'token',
```
ghsolomon, line 185: Why no types for `token`, `name`, `description`, `isShared`?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+    openManageDialog() {
+        this.manageDialogModel = new ManageDialogModel(this);
+    }
+
+    @action
+    closeManageDialog() {
+        const {manageDialogModel} = this;
+        this.manageDialogModel = null;
+        XH.safeDestroy(manageDialogModel);
+        this.refreshAsync();
```
ghsolomon, line 288: Closing the dialog will trigger a refresh, even if the user didn't do anything. I guess that's fine - just wondering why?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
ghsolomon: Would also love to discuss the concept of creating a "New" view in general. Wondering if it makes more sense to only support "Save as..." for creating new views.  Otherwise, what is "New" other than loading some default view and saving it, then requiring the user to edit it, and save their changes?

ghsolomon:
Just a few overall / general thoughts:

- Wondering if we really want to support "New" vs. just requiring the user to load some existing "default" view, make their changes, and save that.
- Should we indicate in the UI that we're in a view, but that we have unsaved changes? Perhaps by using an asterisk instead of a checkmark?
- In cases where we don't have any shared views, could we remove the "My ${noun}" group row? 
- It would be nice to build a context-menu version of this for cases where we're only interested in saving grid views
- Discuss ordering of popover buttons and names. Consider renaming Reset to Revert?
- We should think about ways to componentize this further and/or expose properties we expect apps will wish to customize (the popover button, manage dialog, etc)
- Allow for multi-selection manage dialog for deleting multiple views at once. Could also toggle "Shared" for multiple views at once

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,345 @@
...
+        super();
+        makeObservable(this);
+
+        throwIf(
+            !isFunction(onChangeAsync),
+            'PersistenceManagerModel requires an `onChangeAsync` callback function'
+        );
+
+        this.type = type;
+        this.canManageGlobalFn = canManageGlobal;
```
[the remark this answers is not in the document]
Ryanseanlee: we have a canManageGlobal getter. was also wondering if we should have a better name lets discuss

cmp/grid/GridModel.ts
```diff
@@ -629,8 +629,10 @@ export class GridModel extends HoistModel {
      *
      * @returns true if defaults were restored
      */
-    async restoreDefaultsAsync(): Promise<boolean> {
-        if (this.restoreDefaultsWarning) {
+    async restoreDefaultsAsync(
```
ghsolomon, line 632: Nice improvement here

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+                    omitDefaultMenuComponent
+                } = model,
+                displayName = entity.displayName;
+
+            return fragment(
+                hbox({
+                    className: 'xh-persistence-manager',
+                    items: [
+                        popover({
+                            omit: omitDefaultMenuComponent,
```
ghsolomon, line 41: Would break this out into its own component

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+                }),
+                menuItem({
+                    icon: Icon.reset(),
+                    text: 'Revert View',
+                    disabled: !model.isDirty,
+                    onClick: () => model.resetAsync().linkTo(loadModel)
+                }),
+                menuItem({
+                    icon: Icon.refresh(),
+                    text: 'Reset Defaults',
```
ghsolomon, line 117: Our `Grid` context menu item text is `Restore Grid Defaults`. Maybe this should be `Restore Default View`?

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+                        popover({
+                            omit: omitDefaultMenuComponent,
+                            item: button({
+                                text: model.getHierarchyDisplayName(selectedView?.name) ?? `-`,
+                                icon: isShared ? Icon.users() : Icon.bookmark(),
+                                rightIcon: Icon.chevronDown(),
+                                outlined: true
+                            }),
+                            content: div({
+                                items: [
```
ghsolomon, line 49: Nit - could just pass the children as multiple args to `div` rather than an `ElementSpec`

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+ */
+interface Entity {
+    /** Key used in JsonBlob */
+    name: string;
+    /** User-facing name/label for an object managed by this model. */
+    displayName?: string;
+}
+
+export interface PersistenceManagerConfig<T extends PlainObject> {
+    entity: string | Entity;
```
ghsolomon, line 52: Would add a docstring to explain what `entity` is

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+    /** True to omit the default menu component. Should be used when creating custom app-specific component */
+    omitDefaultMenuComponent?: boolean;
```
ghsolomon, line 54: I would make this a component prop since it concerns the view and not really the backing model.

desktop/cmp/persistenceManager/PersistenceManager.ts
ghsolomon: Or do we call this `Clear View` or something that more closely represents the implementation? I'm not sure what's best here.  In any case, I think we'll want a prop to enable/disable this option since I could imagine there being apps that _always_ want the user to be in _some_ view.

desktop/cmp/persistenceManager/impl/ManageDialogModel.ts
```diff
@@ -0,0 +1,243 @@
...
+    get selIsShared(): boolean {
+        return this.gridModel.selectedRecord?.data.isShared ?? false;
+    }
+
+    get userCreated(): boolean {
+        return this.gridModel.selectedRecord?.data.createdBy === XH.getUser().username;
+    }
+
+    get canDelete(): boolean {
+        return this.parentModel.views.length > 1 && (this.canManageGlobal || !this.selIsShared);
```
ghsolomon, line 32: Along the lines of `"Restore Default View"`, seems like maybe we should support deleting the last view for apps that allow that.

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+    minimal?: boolean;
+}
+
+export const [PersistenceManager, persistenceManager] =
+    hoistCmp.withFactory<PersistenceManagerProps>({
+        displayName: 'PersistenceManager',
+        model: uses(PersistenceManagerModel),
+
+        render({model, minimal = false}) {
+            const {
```
ghsolomon, line 26: Something to think about - should we mask on the initial load? (i.e. while getting the list of views before we load the selected view).  There will inevitably be some jumpiness here.  Ideally we'd want to mask the components for which we're controlling state so the user can't nudge something before the selected view is loaded.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    // TODO - Carefully review if this method needs isStale checks, and how to properly implement them.
+    override async doLoadAsync(loadSpec: LoadSpec) {
+        const {name} = this.entity,
+            rawViews = await XH.jsonBlobService.listAsync({type: name, includeValue: true});
+
+        runInAction(() => (this.views = this.processRaw(rawViews)));
+
+        // Always call selectAsync to ensure pendingValue updated and onChangeAsync callback fired if needed
+        const id = this.selectedView?.id ?? this.views[0].id;
```
ghsolomon, line 220: Worth discussing - do we want to fallback on the first view like this? What about for an app that allows us to not be in a view?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+    /** Whether user can publish or edit globally shared objects. */
+    canManageGlobal: Thunkable<boolean>;
+    /** Async callback triggered when view changes. Should be used to recreate the affected models. */
+    onChangeAsync: (value: T) => void;
+    /** Used to persist this model's selected ID. */
+    persistWith: PersistOptions;
+    /** True (default) to render a save button alongside the primary menu button when dirty. */
+    enableTopLevelSaveButton?: boolean;
+    /** Fn to produce a new, empty object - can be async. */
+    newObjectFnAsync?: () => T;
```
ghsolomon, line 64: Good to remove?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+    /** True to omit the default menu component. Should be used when creating custom app-specific component */
+    omitDefaultMenuComponent?: boolean;
+    /** Whether user can publish or edit globally shared objects. */
+    canManageGlobal: Thunkable<boolean>;
+    /** Async callback triggered when view changes. Should be used to recreate the affected models. */
+    onChangeAsync: (value: T) => void;
+    /** Used to persist this model's selected ID. */
+    persistWith: PersistOptions;
+    /** True (default) to render a save button alongside the primary menu button when dirty. */
+    enableTopLevelSaveButton?: boolean;
```
ghsolomon, line 62: I would make this a component prop since it concerns the view and not really the backing model.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    private readonly _canManageGlobal: Thunkable<boolean>;
+
+    readonly onChangeAsync?: (value: T) => void;
+
+    readonly entity: Entity;
+
+    readonly omitDefaultMenuComponent: boolean = false;
+
+    @observable.ref @managed manageDialogModel: ManageDialogModel;
```
ghsolomon, line 87: No longer needs to be `observable`. Can make `readonly`

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    @observable.ref @managed saveDialogModel: SaveDialogModel;
```
ghsolomon, line 89: No longer needs to be `observable`. Can make `readonly`

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+                                    }),
+                                    objMenu({minimal})
+                                ]
+                            }),
+                            placement: 'bottom-start'
+                        }),
+                        saveButton()
+                    ]
+                }),
+                manageDialog({omit: !manageDialogModel}),
```
ghsolomon, line 62: Remove `omit` - will always have a model

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+                saveDialog({omit: !saveDialogModel})
```
ghsolomon, line 63: Remove `omit` - will always have a model

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,400 @@
...
+    get favoritedViews(): PersistenceView<T>[] {
+        return this.views.filter(it => this.favorites.includes(it.token));
+    }
+
+    get viewTree(): TreeView[] {
+        const groupedViews = groupBy(this.views, 'group'),
+            sortedGroupKeys = keys(groupedViews).sort(),
+            ret = [];
+
+        if (this.favoritedViews.length > 0) {
```
ghsolomon, line 139: I wouldn't include the favorited views twice- the additional rendering of favorites with an accompanying divider feels more like a component-level concern.  Would just have this getter return the main tree.

desktop/cmp/persistenceManager/Types.ts
```diff
@@ -0,0 +1,28 @@
...
+    name: string;
+    owner: string;
+    token: string;
+    type: string;
+    group: string;
+    isShared: boolean;
+    value: T;
+}
+
+export interface TreeView {
```
ghsolomon, line 21: Let's think about the name for this interface. Maybe `PersistenceViewTree`? Or just `ViewTree` but I'm afraid that could get confused with Cube Views.

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -0,0 +1,152 @@
...
+            return menuDivider({title: text});
+        case 'menuFolder':
+            return menuItem({
+                text,
+                icon,
+                shouldDismissPopover: false,
+                children: items ? items.map(child => buildView(child, model)) : []
+            });
+        case 'view':
+            return menuItem({
```
ghsolomon, line 145: Maybe add a little favorite star icon to the right of the view?

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -15,67 +15,67 @@ import {pluralize} from '@xh/hoist/utils/js';
...
+        render({model, ...props}) {
+            return fragment(defaultMenu({...props}), manageDialog(), saveDialog());
         }
     });
 
 //------------------------
 // Implementation
 //------------------------
 
+const defaultMenu = hoistCmp.factory<PersistenceManagerModel>({
```
ghsolomon, line 38: To discuss - might be worth exporting this as its own component that `uses(PersistenceManagerModel)`

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -15,67 +15,67 @@ import {pluralize} from '@xh/hoist/utils/js';
 interface PersistenceManagerProps extends HoistProps<PersistenceManagerModel> {
     /** True to disable options for saving/managing items. */
     minimal?: boolean;
+    /** True (default) to render a save button alongside the primary menu button when dirty. */
+    enableTopLevelSaveButton?: boolean;
+    /** True to omit the default menu component. Should be used when creating custom app-specific component */
+    omitDefaultMenuComponent?: boolean;
```
ghsolomon, line 21: To discuss today - could make this `viewMenu: boolean | ViewMenuProps` and default it to `true` if we want something a bit more flexible. Then we could remove `enableTopLevelSaveButton` as a top-level prop here and make it part of `ViewMenuProps`

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -15,67 +15,67 @@ import {pluralize} from '@xh/hoist/utils/js';
...
-                            placement: 'bottom-start'
-                        }),
-                        saveButton()
-                    ]
-                }),
-                manageDialog({omit: !manageDialogModel}),
-                saveDialog({omit: !saveDialogModel})
-            );
+        render({model, ...props}) {
+            return fragment(defaultMenu({...props}), manageDialog(), saveDialog());
```
ghsolomon, line 30: To discuss - either exporting the dialogs, exposing more props for customizing, or making them pluggable (i.e. `PersistenceManagerProps` could accept a `manageDialogCmp` and `saveDialogCmp` that default to our Hoist implementations)

desktop/cmp/persistenceManager/PersistenceManager.ts
```diff
@@ -15,67 +15,67 @@ import {pluralize} from '@xh/hoist/utils/js';
...
 //------------------------
 
+const defaultMenu = hoistCmp.factory<PersistenceManagerModel>({
+    render({
+        model,
+        omitDefaultMenuComponent = false,
+        minimal = false,
+        enableTopLevelSaveButton = true
+    }) {
+        const {selectedView, isShared, entity} = model,
```
ghsolomon, line 45: Would early out if we’re omitting the default menu component or not render this component at all.

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -49,19 +49,16 @@ interface Entity {
...
     /** Async callback triggered when view changes. Should be used to recreate the affected models. */
     onChangeAsync: (value: T) => void;
     /** Used to persist this model's selected ID. */
     persistWith: PersistOptions;
-    /** True (default) to render a save button alongside the primary menu button when dirty. */
-    enableTopLevelSaveButton?: boolean;
-    /** Fn to produce a new, empty object - can be async. */
-    newObjectFnAsync?: () => T;
+    /** Optional flag to force selection of a view. Defaults false*/
+    allowEmpty?: boolean;
```
ghsolomon, line 61: Thanks for adding this - interested in discussing as a group this afternoon

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -127,30 +123,25 @@ export class PersistenceManagerModel<T extends PlainObject = PlainObject> extend
         return !!this.selectedView?.isShared;
     }
 
+    get isAllowEmpty(): boolean {
+        return this._allowEmpty;
```
ghsolomon, line 127: Why the getter?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -229,6 +218,7 @@ export class PersistenceManagerModel<T extends PlainObject = PlainObject> extend
 
         this.setPendingValue(value);
         await this.onChangeAsync(value);
+        this._loadedInitially = true;
```
ghsolomon, line 221: I think it would be nice if we can leverage the `LoadSpec` API to check if we’ve already completed a successful load.

ghsolomon: That way we wouldn’t need this update this observable and could just have a getter.

desktop/cmp/persistenceManager/Types.ts
```diff
@@ -18,11 +18,20 @@ export interface PersistenceView<T extends PlainObject = PlainObject> {
...
+          type: 'directory';
+          items: PersistenceViewTree[];
+      }
+    | {
+          type: 'view';
+          id: number;
+          isFavorite: boolean;
+      }
+    | {
+          type: 'divider';
```
ghsolomon, line 35: Would remove `divider`. To me, that’s a view-level concern and not a concern of the model / data.

Ryanseanlee: Without 'divider' how would we know where to put in the menuDivider?

desktop/cmp/persistenceManager/PersistenceManagerModel.ts
```diff
@@ -0,0 +1,402 @@
...
+    /** Optional flag to allow empty view selection. Defaults false*/
+    enableDefault?: boolean;
+}
+
+export class PersistenceManagerModel<T extends PlainObject = PlainObject> extends HoistModel {
+    //------------------------
+    // Persistence Provider
+    // Pass this to models that implement `persistWith` to include their state in the view.
+    //------------------------
+    readonly provider: PersistOptions = {
```
ghsolomon, line 59: Wonder if we should make a new `PersistenceManagerProvider` type so the API is consistent with `DashViewProvider`.

core/persist/persistenceManager/Types.ts
```diff
@@ -0,0 +1,34 @@
+import {PlainObject} from '@xh/hoist/core';
+import {LocalDate} from '@xh/hoist/utils/datetime';
+
+export interface PersistenceView<T extends PlainObject = PlainObject> {
+    id: number;
```
ghsolomon, line 5: I know this is from the JsonBlob interface, but wondering for my own knowledge - why the need for id and token if token is the unique ID for the blob?

amcclain:
Latest process tracked @ #3821

Issue #3778, opened 2024-09-06
Admin Console requires additional refresh after instance change


amcclain:
This might be an issue with single-instance deploys only, but is easy to replicate:

* Start local server, single instance
* Open Admin > Cluster tab, log viewer especially good demo
* Restart server, comes up as new instance
* Hoist detects instance change, updates in footer, admin cluster grid remains fixed until next auto-refresh on old instance
* With log viewer polling, server log spammed with `LogViewerAdminController [ERROR] | getFile | atm@xh.io | Unable to find cluster instance f5e9df23 [InstanceNotFoundException]`

Should admin app trigger `XH.refreshAppAsync()` on instance change? Could disrupt user in middle of editing something.   Prompt user so they know?

amcclain:
Maybe it’s just the cluster tab model that needs its own reaction to refresh its context, including the grid of instances.

ghsolomon:
I also believe the issue is that the `ClusterTabModel` does not auto-refresh on as tight of an interval as the `EnvironmentService` poll.  Seems reasonable to add a reaction in `ClusterTabModel` that triggers a refresh whenever the local instance changes.  However, that still wouldn't cover the case where some other (non-local) instance is selected in the top-grid, that instance happens to go down, the local instance has **not** changed, so the `ClusterTabModel` doesn't refresh yet, and the client is left temporarily calling the `LogViewerAdminController` for the obsolete instance's logs.

I suppose we could make the auto-refresh interval tighter, but I'm not sure it's really worth it?  Seems like we'd get a lot out of just adding the reaction to local instance changes, like you suggested, so at least there is consistency there.

PR #3782, opened 2024-09-11
Refresh `ClusterTabModel` more frequently

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/cluster/ClusterTabModel.ts
+++ b/admin/tabs/cluster/ClusterTabModel.ts
@@ -91,7 +91,7 @@ export class ClusterTabModel extends HoistModel {
             },
             {
                 track: () => XH.environmentService.serverInstance,
-                run: () => this.refreshAsync()
+                run: this.refreshAsync
             }
         );
     }
--- a/admin/tabs/monitor/MonitorTabModel.ts
+++ b/admin/tabs/monitor/MonitorTabModel.ts
@@ -68,7 +68,7 @@ export class MonitorTabModel extends HoistModel {
 
         this.addReaction({
             track: () => this.showInactive,
-            run: () => this.refreshAsync()
+            run: this.refreshAsync
         });
     }
 


ghsolomon:
Addresses issue #3778 (Admin Console requires additional refresh after instance change)
1. When `ClusterTabModel.loadAsync` call fails, the top grid is cleared
2. `ClusterTabModel` refreshes when local instance change is observed
3. `ClusterTabModel` auto-refreshes on a tighter (5s) interval - same as `MonitorTabModel`

admin/tabs/cluster/ClusterTabModel.ts
```diff
@@ -57,20 +60,28 @@ export class ClusterTabModel extends HoistModel {
...
+            gridModel.clear();
             XH.handleException(e);
         }
     }
 
     constructor() {
         super();
 
+        this.timer = Timer.create({
+            runFn: this.autoRefreshAsync,
```
amcclain, line 80: Not sure about continuing to fire this when the tab is not visible, especially since it has a modal exception handler.

Should we make this conditional upon tab visibility?

admin/tabs/cluster/ClusterTabModel.ts
```diff
@@ -57,20 +60,28 @@ export class ClusterTabModel extends HoistModel {
...
+                isLocal: row.name == XH.environmentService.serverInstance,
                 usedHeapMb: row.memory?.usedHeapMb,
                 usedPctMax: row.memory?.usedPctMax
             }));
 
             gridModel.loadData(data);
             await gridModel.preSelectFirstAsync();
         } catch (e) {
+            gridModel.clear();
             XH.handleException(e);
```
amcclain, line 72: This seems like a good candidate for an inline exception message, vs. the modal.  Fast polling + modal exceptions + admin console not an ideal combination.

A component we don't have yet but which seems ideal for this would be some kind of errorMessageBanner. I would love to keep the last-loaded data in the grid, but then show a non-modal, non-grid-replacing, but still-very-obvious alert that the data in that grid failed on last refresh.  Would be interested to know if that's a use case or idea that's come up elsewhere.

For now, we could swap out the grid with our standard errorMessage component to avoid the modal - or if the above sounds interesting, explore it as a one-off in this component to see if we like it.

admin/tabs/cluster/ClusterTabModel.ts
```diff
@@ -80,7 +91,7 @@ export class ClusterTabModel extends HoistModel {
             },
             {
                 track: () => XH.environmentService.serverInstance,
-                run: () => this.gridModel.agApi.refreshCells({force: true})
+                run: this.refreshAsync
```
amcclain, line 94: We could consider nixing this now given the overall autorefresh.  You will be getting a refresh within 5s anyway, usually within 3s. Feels like "close enough" - maybe just on the line.

admin/tabs/cluster/ClusterTabModel.ts
TomTirapani: +1 for ther "errorMessageBanner" idea

admin/tabs/cluster/ClusterTabModel.ts
ghsolomon: True - I guess just depends on whether we care about the grid and version bar getting out of sync for a couple seconds.

admin/tabs/cluster/ClusterTabModel.ts
ghsolomon: Well that's a very good point! I think the issue with keeping the last-loaded in the grid was that it preserved the grid selection, which the log viewer model was observing to poll for instance logs.  We could clear the selection on error or have the models of child components observe that this model is in an error state, although maybe cleaner to just display the error panel in place of the grid?

admin/tabs/cluster/ClusterTabModel.ts
ghsolomon: Makes sense - would probably want to do the same for `MonitorTabModel` then?

amcclain: Monitor has special toast behavior I believe - deliberately refreshing and alerting admin who is looking at app that monitor status has changed

amcclain: although arguably that should not be dependent on have you loaded tab

admin/tabs/cluster/ClusterTabModel.ts
amcclain: child comps should have their own reasonable error handling i would think

admin/tabs/cluster/ClusterTabModel.ts
amcclain: happy to leave it as-is - since it's async it's never going to be instant anyway, but doesn't seem like there's a real cost here, no need to over-think in either direction

lbwexler: great points, especially with the inline error and not refreshing when hidden

PR #3796, opened 2024-09-28
Add new Timing block to app loading track message

lbwexler:
Add new Timing block to app loading track message.
+ New state for 'Authenticating'
+ Misc cleanups

Hoist P/R Checklist
-------------------


amcclain (approved):
This is very nice, thank you!  

Would it be worth splitting up `INITIALIZING` to distinguish between `HOIST_INITIALIZING` and `APP_INITIALIZING` - transitioning to the latter right before we call `await this.appModel.initAsync();`?

Wasn't sure how much logic might be out there (if any) that keys off of that state.

lbwexler (commented):
Great catch!!

lbwexler:
Went for the suggested additional states.  Think we should go ahead and leverage all this, and I doubt anyone was observing it specifically.

Issue #3797, opened 2024-10-01
Resizable panels / splitters broken in desktop Safari


amcclain:

lbwexler:
This seems concerning -- think we should understand what's happening here sooner rather than later

TomTirapani:
I'm looking into this... it seems this test is failing:
https://github.com/xh/hoist-react/blob/develop/desktop/cmp/panel/impl/dragger/DraggerModel.ts#L256-L258

DragEvent's `buttons` is `0` in Safari, but `1` in Chrome, which causes us to exit out of the drag early here:
https://github.com/xh/hoist-react/blob/develop/desktop/cmp/panel/impl/dragger/DraggerModel.ts#L93-L96

Investigating why / how this could be the case, and what we can do about it. Seems like a low-level Safari bug to me - will post findings here

TomTirapani:
I came up with a workaround which resolves the issue:
https://github.com/xh/hoist-react/pull/3799

I'll keep looking for a better solution / more information

PR #3799, opened 2024-10-03
Panel dragger workaround for Safari

TomTirapani:
Safari is not setting `buttons` or `button` to `1` on Drag events, which is causing our Panel resizing to early out.
See: https://github.com/xh/hoist-react/issues/3797

This workaround adds a fallback check to the deprecated `MouseEvent.which` property, which *is* being set:
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which#value_for_mouseevent_non-standard

As a deprecated API we obviously should not rely on it, but I **think** its safe to use as fall back for this specific issue in lieu of a better solution.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ## 69.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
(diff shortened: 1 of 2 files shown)


lbwexler:
Thanks for the quick turnaround here Tom.  I think this is very pragmatic.

PR #3801, opened 2024-10-04
Update all log levels when a single row is changed

lbwexler:
Threw in a Refresh button to our RestGrid for good measure -- hope that makes sense?  Obviously not attached to that.

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/cluster/logs/levels/LogLevelDialog.ts
+++ b/admin/tabs/cluster/logs/levels/LogLevelDialog.ts
@@ -4,90 +4,39 @@
  *
  * Copyright © 2024 Extremely Heavy Industries Inc.
  */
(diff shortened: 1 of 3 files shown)


amcclain:
Hi @lbwexler - thanks for this - sorry for not getting to review in a more timely manner-  was on my list!

I wanted to check in re. addition of the refresh button to RestGrid, without any optionality on it.  We end up with this "double refresh button" in a few places in the admin app, and I know we have several client apps that also have full-tab rest grids that will render the same, with the new refresh button right by the app refresh:

Screenshot 2024-10-08 at 2 42 29 PM

I think I would have voted to make this opt-in, so a dev could turn it on for a case like the log levels grid while the base case didn't change. WDYY?

lbwexler:
Agree its a problem - will make it optional

PR #3811, opened 2024-10-14
Persist memory monitoring

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/cluster/memory/MemoryMonitorPanel.ts
+++ b/admin/tabs/cluster/memory/MemoryMonitorPanel.ts
@@ -8,13 +8,13 @@ import {AppModel} from '@xh/hoist/admin/AppModel';
 import {MemoryMonitorModel} from '@xh/hoist/admin/tabs/cluster/memory/MemoryMonitorModel';
 import {chart} from '@xh/hoist/cmp/chart';
 import {grid, gridCountLabel} from '@xh/hoist/cmp/grid';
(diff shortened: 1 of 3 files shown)


admin/tabs/cluster/memory/MemoryMonitorPanel.ts
```diff
@@ -82,3 +88,41 @@ const bbar = hoistCmp.factory<MemoryMonitorModel>({
...
+
+const instanceSelect = hoistCmp.factory<MemoryMonitorModel>({
+    render({model}) {
+        const {pastInstances, pastInstance, instanceName} = model,
+            options = [{value: null, label: instanceName, lastUpdated: null}] as any;
+        pastInstances?.forEach(it => {
+            options.push({value: it, label: it.name, lastUpdated: it.lastUpdated});
+        });
+        return fragment(
+            select({
```
TomTirapani, line 100: Should we disable this dropdown if you have no past instances?

TomTirapani: Also, it seems a little bit weird that this control offers you **selected** current instance + **all** past instances, but to change to a different current instance you need to change the top selection.

I understand how it ended up that way and am **not** suggesting we allow you to switch between running instances here - it just seems a bit strange.

lbwexler: totally agree and was vexed by this -- we could revisit at some point -- thought this was somewhat consistent with how we were handling the log panel

PR #3814, opened 2024-10-17
Persistable

ghsolomon:
Hoist P/R Checklist
-------------------


ghsolomon:
To Do:
- CHANGELOG entry

Discuss:
- Persisting widths of autosized columns
- Ability for Persistable to specify custom equality comparator

lbwexler:
Greg and I begin a pretty lengthy review of this on Friday -- looking forward to pick it up again in Monday.  Two thoughts I wanted to pass along:

- Really do think we should fold the exception handling of PersistenceProvider in to create.  If we ever decide we want it to throw, we could add another method or a flag, but I think we should dry this up for now.
- Would it be helpful for the PersistenceProvider to provide its initial resolved state as a *property* on itself, rather than an immediate callback to setPersistedState?  It seems like that might make it a bit easier for implementations that have a bit of a different flow during construction that later on.  Just a thought -- realize you are currently working through GroupingChooser/FilterChooser to deal with some complexities there and thought that perhaps that would help.

desktop/cmp/dash/container/DashContainerModel.ts
```diff
@@ -193,8 +193,10 @@ export class DashContainerModel
         if (persistWith) {
             try {
```
lbwexler, line 194: Remove exception handling here?

core/persist/PersistenceProvider.ts
```diff
@@ -62,57 +64,67 @@ export abstract class PersistenceProvider<S> {
      * Throws if unable to perform initial read operation.
      *
      * Note:
-     * - `destroy()` must be called when the provider is no longer needed.
+     * - `destroy()` must be called when the provider is no longer needed unless an owner is provided.
```
lbwexler, line 65: why aren't target and owner just both required for simplicity (where owner can be defaulted to Target, if its a HoistBase).  Check and throw at runtime.

desktop/cmp/panel/PanelModel.ts
```diff
@@ -107,11 +107,16 @@ export interface PanelConfig {
     xhImpl?;
 }
 
+export interface PersistablePanelState {
+    collapsed?: boolean;
+    size?: number | string;
+}
```
amcclain, line 113: Should this be `PanelPersistState` for consistency?

desktop/cmp/panel/PanelModel.ts
```diff
@@ -352,6 +333,22 @@ export class PanelModel extends HoistModel {
         this.resizeWhileDragging = v;
     }
 
+    //---------------------------------------------
+    // Persistable Interface
+    //---------------------------------------------
+    getPersistableState(): PersistableState<PersistablePanelState> {
+        const ret: PersistablePanelState = {};
+        if (this.collapsible) ret.collapsed = this.collapsed;
+        if (this.resizable) ret.size = this.size;
```
amcclain, line 342: I suppose here is where we could look to convert the current pixel size back into a percentage based size, if defaultSize spec'd as a percentage...

amcclain (commented):
Whew!  Minor comments / questions below, but no blockers I have seen so far.  Moving on to try testing this with some real world use cases within a client app.

cmp/grouping/GroupingChooserModel.ts
```diff
@@ -140,44 +146,25 @@ export class GroupingChooserModel extends HoistModel {
...
-            } catch (e) {
-                this.logError(e);
-                XH.safeDestroy(this.provider);
-                this.provider = null;
-            }
-        }
+        this.setValue(value);
+        this.setFavorites(favorites);
+
+        this.persistValue = persistWith.persistValue ?? true;
```
amcclain, line 152: Woops - we still need the `if (persistWith) ...` wrapper around all this

amcclain:
Should we consider unwinding the nested `GridPersistenceModel` and make `GridModel` itself a Persistable, directly?  I am thinking at the least we might end up adding the two public methods from the interface to GM, even if they still delegated to the inner model.  But then I wondered if - given all the things already directly on the class - we were really getting much from the extra layer.

The answer might depend in part on our plan for exposing `setPersistableState()` as a first-class public API - i.e. something that custom (pre-existing) implementations of our WIP view manager concept might call within app code. Do we want those usages to have to reach into `gridModel.persistenceModel`?  I don't think so - nicer to have it on GM directly.

amcclain:
Should we consider a way to register custom `PersistenceProvider` implementations in such a way that the `PersistenceProvider.create()` factory can create them?  

We need support custom providers - eg the features similar to our WIP `ViewManagerModel` that have already been built within client apps and can't immediately be replaced by the new Hoist system. At the same time, we have an incentive to keep developers using the factory, with its built-in target binding and exception handling.

Could do something like:
```
static registerProviderType(type: string, ProviderClass: new () => PersistenceProvider) {}
```

Simple, but would require devs to spec their custom `PersistOptions.type` anytime they wanted to create one. A bit awkward:

```
persistWith: {type:'customViewManager', customViewManager: myCustomViewManagerInstance}`
```

Maybe allow devs to make it look a bit more like the Hoist built-ins with something like:
```
static registerCustomProviderType(selector: (persistOptions: PersistOptions, ProviderClass: new () => PersistenceProvider) => boolean)
```

Then have the factory run any options it is asked to create through any custom selector(s) provided in this way?  So at the bottom of the file that defines your custom provider, you run e.g..

```
PersistenceProvider.registerCustomProviderType(
  (opts) => !!opts.customViewManager,
  CustomViewManagerModel
)
```

amcclain:
Looking ahead to our `ViewManager` impl - and more immediately looking at some of our custom client app examples - I'm wondering how best to have the manager, a `PersistenceProvider` _push_ newly selected bundles of state back into the things it has been asked to persist.

Would it be something like:
```
export class ViewManagerModel extends HoistModel {
    // Complete bundle of fully persisted state - updated when view selected or pendingValue saved.
    @observable value;
    // Current reported state from all persisted providers - updated when any provider reports a change.
    @observable pendingValue;

    // All providers bound to this view manager.
    providers: ViewManagerProvider<any>[] = [];

    // other view manager stuff

    constructor() {
        super();
        // other view manager stuff

        this.addReaction({
            track: () => this.value,
            run: () => {
                // Tell all providers to distribute the newly selected state to their targets.
                this.providers.forEach(p => p.pushStateToTarget());
            }
        });
    }
}

class ViewManagerProvider<S> extends PersistenceProvider<S> {
    viewManager: ViewManagerModel;

    constructor(cfg: PersistenceProviderConfig<S>) {
        super(cfg);
        const {viewManager} = cfg.persistOptions;
        this.viewManager = viewManager;
        viewManager.providers.push(this);
    }

    // Not part of the superclass API - should it be?
    pushStateToTarget() {
        this.target.setPersistableState(this.read());
    }

    override readRaw() {
        return this.viewManager.value;
    }

    override writeRaw(data) {
        this.viewManager.pendingValue = data;
    }

    override destroy() {
        pull(this.viewManager.providers, this);
        super.destroy();
    }
}

PersistenceProvider.registerCustomProviderType(
    pOpts => !!pOpts.customViewManager,
    ViewManagerProvider
);
```

Used like:
```
export class MyAppFeatureModel extends HoistModel {
    @managed viewManager = new ViewManagerModel({some:'config'});
    persistWith = {viewManager: this.viewManager};
    
    @persist appFlag = false;

    gridModel = new GridModel(({
        persistWith: this.persistWith,
        // ...opts....
    }))
}
```

core/persist/PersistenceProvider.ts
```diff
@@ -32,100 +39,162 @@ import {throwIf} from '@xh/hoist/utils/js';
...
     clear(path: string = this.path) {
         const obj = cloneDeep(this.readRaw());
         unset(obj, this.path);
         this.writeRaw(obj);
     }
 
-    /**
-     * Clear *all* state held by this object.
-     */
+    /** Clear *all* persisted data managed by this provider. */
```
amcclain, line 138: This API doesn't sit well with me - unclear why we have it, and I don't think the doc comment is correct.  

This method doesn't just clear the data this particular provider manages - that's `clear()` above, which is scoped to the provider's own path. Instead `clearAll()` clears.... all data held by the same backing implementation, across all paths, I think....  Is that something that we want/need to be able to call on a single provider?

cmp/grid/impl/GridPersistenceModel.ts
```diff
@@ -46,70 +40,92 @@ export class GridPersistenceModel extends HoistModel {
...
-            this.addReaction(this.sortReaction());
+        const provider = PersistenceProvider.create({
+            persistOptions: {
+                path: 'grid',
+                ...persistWith
+            },
+            target: this
+        });
+
+        if (provider) {
```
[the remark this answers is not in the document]
ghsolomon: Let's talk about this - our current rule is that we want to create our provider after we have initialized our "code" state

core/persist/DashViewProvider.ts
```diff
@@ -29,10 +31,10 @@ export class DashViewProvider extends PersistenceProvider {
     }
 
     override writeRaw(data) {
-        this.dashViewModel.setViewState(data);
+        this.dashViewModel.viewState = data;
     }
 
     override clearRaw() {
-        return this.dashViewModel.setViewState(null);
+        return (this.dashViewModel.viewState = null);
```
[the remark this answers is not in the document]
ghsolomon: You know, this is a great question... how odd... going to remove.

desktop/cmp/panel/PanelModel.ts
ghsolomon: I think we're actually ok - the issue I was seeing was actually due to the percent-specified default size not complying with a specified minSize.

amcclain:
Latest process tracked @ #3821 - although I still want to loop through my comments above as I believe that some of them still merit some review or fixes.

PR #3816, opened 2024-10-21
Fixed `genDisplayName` method to use reg-ex that works on iOS Safari …

cnrudd:
…16 (used by BB Access)

Hoist P/R Checklist
-------------------


cnrudd:
After more digging, I think that the regex that I changed is probably not the cause of the error on the clients device. 
There is probably an offending regex in some library code.  We are going to ask the client to make sure he is on ios18 before digging further.

cnrudd:
Source of issue was found in new markdown support, brought in to framework in v66.1.0.
offending reg ex is in this lib:
https://github.com/syntax-tree/mdast-util-gfm-autolink-literal
see their issue: https://github.com/syntax-tree/mdast-util-gfm-autolink-literal/issues/10

Closing PR without merging - @amcclain already put the fontawesome fix into develop.

PR #3824, opened 2024-11-13
Add public `ViewManagerModel.viewSelectionObserver`

amcclain:
+ For use by apps to mask components during view select/rebuild.

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/persist/viewmanager/ViewManagerModel.ts
+++ b/core/persist/viewmanager/ViewManagerModel.ts
@@ -7,6 +7,7 @@ import {
     PersistenceProvider,
     PersistOptions,
     PlainObject,
+    TaskObserver,
     Thunkable,
     ViewManagerProvider,
     XH
@@ -22,7 +23,7 @@ import {View, ViewTree} from './Types';
 
 export interface ViewManagerConfig {
     /**
-     * True (default) to allow user to opt-in to automatically saving changes to their private
+     * True (default) to allow user to opt in to automatically saving changes to their private
      * views - requires `persistWith`.
      */
     enableAutoSave?: boolean;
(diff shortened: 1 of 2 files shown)


amcclain:
Testing out our new stuff in an actual app that manages a fair amount of state via viewManager - on switch there's a noticeable lag/pause while the components are all given their new state.  Exposing this task observer allowed me to put a mask in place, which I think works well to let the user know what's happening.

core/persist/viewmanager/ViewManagerModel.ts
```diff
@@ -278,18 +289,19 @@ export class ViewManagerModel<T extends PlainObject = PlainObject>
...
-        this.selectedToken = token;
-
-        // Allow this model to restore its own persisted state in its ctor and note the desired
-        // selected token before views have been loaded. Once views are loaded, this method will
-        // be called again with the desired token and will proceed to set the value.
-        if (isEmpty(this.views)) return;
-
-        this.setValue(this.selectedView?.value ?? ({} as T));
+        // Introduce minimal wait and link to viewSelectionObserver to allow apps to mask.
+        await wait(100)
```
amcclain, line 293: @ghsolomon rightly pointed out that this random `100` value is unfortunate. I re-tested, and I found that I was unable to get the mask to reliably render in my app when swapping views unless I left this "high" value in place.  `await wait()` and `await wait(1)` did not do it.

(The app test case involve some heavyweight re-filtering of a large recordset + dashboard layout changes, but nothing that extreme - definitely a "real world" case.)

We discussed how ideally we could use some utility that perhaps keyed off of animation frames or otherwise was able to better detect that rendering had "settled down" - for want of a more technical term...

PR #3839, opened 2024-11-22
Fixes to DashContainerModel state handling + renders

amcclain:
+ State updates were being dropped if new state applied during re-render, when ref not available.
+ Observed in use case with ViewManager, where the parent component was being moved in the layout at same time as new dash state applied.
+ Don't ignore state updates if not rendered - store and save for later application.
+ But improve ref reaction and async state application code to avoid unintended / overlapping application of state when both state and render changing at the same time.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/dash/container/DashContainerModel.ts
+++ b/desktop/cmp/dash/container/DashContainerModel.ts
@@ -214,8 +214,12 @@ export class DashContainerModel
         // Initialize GoldenLayout with initial state once ref is ready
         this.addReaction(
             {
-                track: () => [this.containerRef.current, this.layoutLocked],
-                run: () => this.loadStateAsync(this.state)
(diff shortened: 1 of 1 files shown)


desktop/cmp/dash/container/DashContainerModel.ts
```diff
@@ -214,8 +214,12 @@ export class DashContainerModel
         // Initialize GoldenLayout with initial state once ref is ready
         this.addReaction(
             {
-                track: () => [this.containerRef.current, this.layoutLocked],
-                run: () => this.loadStateAsync(this.state)
+                track: () => [this.containerRef.current, this.layoutLocked] as const,
```
amcclain, line 217: I did not test the layoutLocked change path, but good to ensure that all working as expected

PR #3848, opened 2024-12-10
Improved accuracy of `IconProps` interface

amcclain:
- Use the `IconName` and `IconPrefix` types provided by FontAwesome.
- Define a `ResolvedIconProps` interface, for specific usages where we know `iconName` and `prefix` have been applied and will be present.

Hoist P/R Checklist
-------------------


amcclain:
I started looking at this due to warnings being thrown around icons that turned out were actually caused by https://github.com/xh/hoist-react/commit/9435122c610de16e770ffb644cce9b83b14f8b3a

But I had started tightening things up and still there is some utility in taking these changes.  My only concern is that the use of the `IconName` exported type from FA might flag false-positive warnings of some kind, or require apps that are passing icon names as vars to more precisely type them.  Not sure that would happen in practice though, or be a blocker - could be helpful...

icon/Icon.ts
```diff
@@ -59,27 +53,50 @@ export interface IconProps extends HoistProps, Partial<Omit<FontAwesomeIconProps
...
+     * your app bootstrap code, eg:
+     * ```
+     *   import {library} from '@fortawesome/fontawesome-svg-core';
+     *   import {faDreamIcon} from '@fortawesome/pro-regular-svg-icons';
+     *   library.add(faDreamIcon);
+     * ```
+     * and then pass its string name to this factory: `icon({iconName: 'dream-icon'})`
      */
-    icon(opts?: IconProps): any {
+    icon(opts: IconProps & {iconName: IconName}): any {
```
lbwexler, line 99: why is the & {iconName:IconName} needed?

lbwexler (approved):

lbwexler:
I am for getting this merged -- seems like a nice pickups to get a list of all of the available names. Very easy to cast `as IconName` for any problematic usages.

icon/Icon.ts
lbwexler: I tried without it, and it seemed to work, but perhaps am missing something?

ghsolomon (approved):
Looks good to me

PR #3849, opened 2024-12-10
Avoid no-op state updates in DashContainerModel, shorten debounce

amcclain:
- Changes to GoldenLayout state often do not result in actual diffs to processed Hoist state - don't publish a new state object if that's the case.
- Shorten debounce on `publishState()` - allows actual state change to be reported in a much more timely fashion.
- Removing debounce entirely throws - looks like an initial timing / render / model lookup issue, which could warrant further investigation to see if we can handle it better.

Hoist P/R Checklist
-------------------


lbwexler:
am all for this

lbwexler (approved):

PR #3865, opened 2024-12-19
View state storage

lbwexler:
Hoist P/R Checklist
-------------------


cmp/viewmanager/ViewManagerModel.ts
```diff
@@ -471,31 +469,45 @@ export class ViewManagerModel<T = PlainObject> extends HoistModel {
...
             track: () => [this.pendingValue, this.autoSave],
             run: () => this.maybeAutoSaveAsync(),
             debounce: 5 * SECONDS
         });
+
+        this.addReaction({
+            track: () => [this.view, this.userPinned, this.autoSave],
+            run: () => api.updateStateAsync(),
+            debounce: 5 * SECONDS,
+            fireImmediately: this.view.token != initialToken
```
ghsolomon, line 509: Curious about the `fireImmediately`. Wondering why not wait until the user has changed something?

lbwexler: wanted to catch case where existing view was being re-written because it was no longer present, or a different one.  On the fence here for sure

ghsolomon: I see- so when a user's current view is no longer available, either the "default" view or a code-specified "initial view" will be loaded, and that view will be immediately persisted as the user's current view?  I wonder if it would be better to clean (remove) their persisted current view in that case since it's no longer a view they chose.  To do that might just adding unnecessary complexity though.

Along similar lines - is it strange that pinning a view will also cause a user's current view to be persisted?  I'm not sure it's really an issue, but wanted to flag.

cmp/viewmanager/DataAccess.ts
```diff
@@ -0,0 +1,179 @@
...
+    userPinned: Record<string, boolean>;
+    autoSave: boolean;
+}
+
+/**
+ * Supporting class for accessing and updating ViewManager and View data.
+ *
+ * @internal
+ */
+export class DataAccess<T> {
```
amcclain, line 42: A general question - do we have any other classes like this, or named like this?  Just trying to relate this to any patterns we already have for breaking up functionality along these lines.

lbwexler: I couldn't find a direct analog, and as you might have noticed, I have been playing around with various names for it.

The class gets a lot thinner with the move to a server-side view service, so probably less necessary but thought it was worth it to keep it around, with the main class being so large.

lbwexler:
CR: Greg

Issue #3843, opened 2024-12-03
Allow Cube Views to opt-out of flattening redundant children


TomTirapani:
Our cube currently has hard-coded logic to flatten "redundant children"; i.e., groups with a single child with the same dimension values: https://github.com/xh/hoist-react/blob/develop/data/cube/row/BaseRow.ts#L66-L75

We should offer a flag to allow cube view queries to opt-out of this behaviour.

amcclain:
"Redundant child" - ouch!

But seriously, I have no doubt there is a specific use case behind this request, and it seems like it should be pretty straightforward to implement, assuming we don't have any strong expectations anywhere else that this processing has been applied (I don't know why there would be).

PR #3867, opened 2024-12-23
Allow Cube Query to opt-out of flattening redundant children

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/3843

Seemed like a quick win, that would allow us to clean up some code at a flagship client app. Wasn't sure about the name, its tricky to describe what this does in just a couple of words - any suggestions welcome!

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/cube/Query.ts
+++ b/data/cube/Query.ts
@@ -52,6 +52,9 @@ export interface QueryConfig {
     /** True to include leaf nodes in return.*/
     includeLeaves?: boolean;
 
+    /** True (default) to flatten single-child hierarchies into parents. */
+    flattenRedundantChildren?: boolean;
+
     /**
      * Optional function to be called for each aggregate node to determine if it should be "locked",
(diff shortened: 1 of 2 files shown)


lbwexler:
We discussed, and I like this change al ot, as it also helps document a bit of a subtle behavior.  

One unfortunate/confusing thing is that it is a lot like an `omitFn`, and I suppose we might have been able to implement it as an omitFn, although because of its recursive nature of zapping multiple nodes in a chain, I am not sure how straightforward that would be.

Perhaps we should call it `omitRedundantNodes` or `flattenRedundantNodes`?

lbwexler:
I think it really is an `omitFn`.  We could make it the default omit function on cube e.g. `Cube.omitFn = omitRedundantNodes` but that would be a breaking change for anyone overriding the `omitFn`, who would now have to also refer to the default implementation.

Maybe best to just call it `omitRedundantNodes` and have `omitFn` and `omitRedundantNodes` just mention each other in their docstrings.

PR #3869, opened 2024-12-24
Added support for `AuthZeroClientConfig.audience`

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/security/authzero/AuthZeroClient.ts
+++ b/security/authzero/AuthZeroClient.ts
@@ -4,7 +4,8 @@
  *
  * Copyright © 2024 Extremely Heavy Industries Inc.
  */
-import {Auth0Client, Auth0ClientOptions} from '@auth0/auth0-spa-js';
+import type {Auth0ClientOptions, AuthorizationParams} from '@auth0/auth0-spa-js';
+import {Auth0Client} from '@auth0/auth0-spa-js';
 import {XH} from '@xh/hoist/core';
 import {never, wait} from '@xh/hoist/promise';
 import {Token, TokenMap} from '@xh/hoist/security/Token';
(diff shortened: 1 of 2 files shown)


amcclain:
To trigger pre-existing issue (without this change) on Toolbox - 

* Disable third party cookies in your browser (Safari > Preferences > Privacy > tick "Prevent cross-site tracking"
* Attempt to load Toolbox on localhost without this change - you should see a "login_required" error due to blocked use of iframe / cookies.  Note this would effect any app attempting to use an access token while on localhost or while using an auth0 `domain` that does not match the app's deployed domain

With this change in hoist-react, update Toolbox's client-creation code to spec `audience: 'toolbox.xh.io'` like so:
Screenshot 2024-12-23 at 4 08 52 PM

Login should now proceed OK, with both ID and access tokens available.  Note also only a single call to the Auth0 `token` endpoint now needs to be made by the client.

amcclain:
Note that this change does not attempt to support a possible configuration where multiple access tokens are configured for different audiences.  I believe to support that we would need more cases around need for interactive flows from user.  Did not think that was a priority - not something we've seen yet.

security/authzero/AuthZeroClient.ts
```diff
@@ -88,14 +101,16 @@ export class AuthZeroClient extends BaseOAuthClient<AuthZeroClientConfig, AuthZe
...
+            authorizationParams: this.buildAuthParams({scope: this.loginScope})
         });
         await never();
     }
 
     protected override async doLoginPopupAsync(): Promise<void> {
         try {
-            await this.client.loginWithPopup({authorizationParams: {scope: this.loginScope}});
+            await this.client.loginWithPopup({
+                authorizationParams: this.buildAuthParams({scope: this.loginScope})
```
lbwexler, line 112: Is it not the case that it will get the default audience as now specified in `createClient`?  Just as it will get the redirect_url?    If so, I think the redundant calls to buildAuthParams is confusing.

amcclain: Fixed thanks

Issue #3833, opened 2024-11-18
Grid Filter is wrongly cleared when the `Grid` itself is cleared


ghsolomon:
See screen capture below:

https://github.com/user-attachments/assets/bad64dd2-4476-4127-af26-988ec3844f04

PR #3873, opened 2024-12-26
Make Column Filter UI lazy.

lbwexler:
Fixes mysterious bug. https://github.com/xh/hoist-react/issues/3833

Could have figured out what the problem was, but noticed only manifesting on "stale" copies of the filter that had previously been mounted.

Was surprised that these models were still being created from the get go, although I know we discussed. Making this disposable is more efficient, lightweight, and easier to reason about.

Hoist P/R Checklist
-------------------


TomTirapani:
Looks good and works well - fixed a small issue with a broken import and left over logging. 100% agree that it makes sense to recreate the state when the popover is opened, much easier to reason about 👍

PR #3883, opened 2024-12-31
Tighten/simplify implementation of HoistModel.componentProps

lbwexler:
- Fix difficult bug with re-render loops.
- Avoid mobx object/proxy object creation for models that don't use this property
- Avoid extra prop processing for linked models that don't use this property

Hoist P/R Checklist
-------------------

### DIFF
--- a/core/model/HoistModel.ts
+++ b/core/model/HoistModel.ts
@@ -4,9 +4,9 @@
  *
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
-import {action, makeObservable, observable} from '@xh/hoist/mobx';
+import {action, computed, comparer, makeObservable, observable} from '@xh/hoist/mobx';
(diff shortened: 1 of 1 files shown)


ghsolomon:
Nice cleanup!

amcclain:
I am sure this is a great cleanup, but admit I don't really understand it.

Were we previously doing the `apply` and `delete` checks on the singular `_componentProps` object to trigger reactions to the proxied observable?

To put it another way, we had/still have the comment on the getter:

> Observability is based on a shallow computation for each prop (i.e. a reference change in any particular prop will trigger observers to be notified).

Is it correct to say that the `@computed({equals: comparer.shallow})` annotation is now making that the case?  Was that the "extra processing" referenced by:
 
> Avoid extra prop processing for linked models that don't use this property

PR #3889, opened 2025-01-06
Pass token, not viewInfo and surface raw errors to make API more flexible.

cnrudd:
In my routable views implemention in which routes have view tokens,

I had to do this when the route got a new token:

```
const info = viewMngr.views.find(v => v.token === token);
if (!info) {
    try {
           await XH.jsonBlobService.getAsync(token)
    } catch (e) {
     // handle view doesn't exist & view not shared errors


    }
   
    viewMngr.selectViewAsync(info);
}

```

After this change, the flow is simplified to:

```
 try {
      await viewMngr.selectViewAsync(info);
    } catch (e) {
     // handle view doesn't exist & view not shared errors
}
```

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/viewmanager/DataAccess.ts
+++ b/cmp/viewmanager/DataAccess.ts
@@ -48,14 +48,14 @@ export class DataAccess<T> {
     }
 
     /** Fetch the latest version of a view. */
-    async fetchViewAsync(info: ViewInfo): Promise<View<T>> {
+    async fetchViewAsync(token: String): Promise<View<T>> {
         const {model} = this;
-        if (!info) return View.createDefault(model);
+        if (!token) return View.createDefault(model);
         try {
-            const raw = await XH.fetchJson({url: 'xhView/get', params: {token: info.token}});
+            const raw = await XH.fetchJson({url: 'xhView/get', params: {token}});
             return View.fromBlob(raw, model);
         } catch (e) {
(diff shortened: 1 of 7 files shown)


cmp/viewmanager/DataAccess.ts
```diff
@@ -48,14 +48,14 @@ export class DataAccess<T> {
...
 
     /** Fetch the latest version of a view. */
-    async fetchViewAsync(info: ViewInfo): Promise<View<T>> {
+    async fetchViewAsync(token: String): Promise<View<T>> {
         const {model} = this;
-        if (!info) return View.createDefault(model);
+        if (!token) return View.createDefault(model);
         try {
-            const raw = await XH.fetchJson({url: 'xhView/get', params: {token: info.token}});
+            const raw = await XH.fetchJson({url: 'xhView/get', params: {token}});
```
cnrudd, line 55: I wonder if the params here should also include `type` from ViewManagerModel, so that a token for a blob not related to this view manager cannot be fetched.  (this risk precedes this change)

lbwexler: could also just throw in View.fromBlob, if it does not match -- easier change.

ghsolomon (approved):
Looks great

Issue #3810, opened 2024-10-11
Grid Column Filter Value Tab UX improvements


cnrudd:
I discussed and reviewed this with @amcclain and we came to the conclusion that adopting our client's suggestion for our Column Filter Values tab to behave more like Excel's Column Filter would be a good improvement.

The client's suggestion was made after using a Column Filter configured to show the "Apply Filter" button (`commitOnChange: false`), so the following guidance definitely applies to that configuration, but I think this improvement could be made also to the Hoist Column Filter when it is configured to be applied automatically (`commitOnChange: true` - no Apply button showing).

Excel does not appear to offer a Column Filter configured for `commitOnChange: true` , with no "OK" button.

The better experience we see in Excel is that the user does not need to pre-uncheck all of the available rows in the Values grid before typing into the filter field.  The user can just type into the filter field, and then hit OK and the grid will be filtered by the values that appeared in the Values grid.  When a user starts typing in the filter field, the "Select All" checkbox turns into a "Select All Search Results" checkbox, and an additional "Add current selection to filter" checkbox is inserted beneath it.

In Hoist, in order to filter by all of the results that typing in the filter field produced, a user must first uncheck the 'Select All' checkbox, then type their search string into the filter field above, and then check "select all" or the individual values he wants to filter on, and then hit "Apply Filter".  It is not as intuitive, and users often need to be shown how to perform this sequence.

Please reach out to me if you would like a demo of the differences that I attempted to describe above.  These UIs and their use are not the easiest to describe and attempts to do so do not make for the most exciting reading, to be sure.

**Excel Column Filter**
<img width="444" alt="image" src="https://github.com/user-attachments/assets/b43a2436-8ac9-4dcd-b05e-cb25eed49c0f">

**Hoist Column Filter**
<img width="913" alt="image" src="https://github.com/user-attachments/assets/f0b6cf13-bf9c-4db4-8443-c79ecfd1c4b6">

PR #3894, opened 2025-01-09
Windows Excel like behavior for grid column filtering

oodarluz:
This PR addresses this issue: https://github.com/xh/hoist-react/issues/3810

Hoist P/R Checklist
-------------------


desktop/cmp/grid/impl/filter/headerfilter/values/ValuesTabModel.ts
```diff
@@ -26,6 +26,10 @@ export class ValuesTabModel extends HoistModel {
     /** Bound search term for `StoreFilterField` */
     @bindable filterText: string = null;
 
+    /** Available only when commit on change is false merge
+     * current filter with pendingValues on commit*/
```
cnrudd, line 30: Comment layout should be corrected.

desktop/cmp/grid/impl/filter/headerfilter/values/ValuesTab.ts
```diff
@@ -47,7 +48,33 @@ const tbar = hoistCmp.factory(() => {
...
+        className: 'store-filter-header',
+        items: [
+            hframe(
+                checkbox({
+                    disabled: store.empty,
+                    displayUnsetState: true,
+                    value: allVisibleRecsChecked,
+                    onChange: () => model.toggleAllRecsChecked()
+                }),
+                span(`(Select All${filterText ? ' Search Results' : ''})`)
```
cnrudd, line 67: Are the parentheses just inside the backticks really needed here?

cnrudd (changes requested):
Otito and I went over this several times and ironed out the kinks.  My current comments are just minor linting nits.

We confirmed that the new implementation works like Excel.
We tested the following flows/cases:

1. commitOnChange: true|false
2. regular grid & tree grid
3. statefulness when switching from value to custom tab
4. hide/show "add to existing filter"
5. grid font size tiny, standard, large, etc.

There is a lot to test, I don't think we missed anything, but anyone reviewing should really try to break it.

desktop/cmp/grid/impl/filter/headerfilter/values/ValuesTab.ts
oodarluz: No, I just have them to mirror the look for excel

amcclain:
Great thanks guys - I would like to have a look at this.  Will do so today and reach out with any questions.

amcclain:
Have only had a quick chance to try this out - will look more closely, and compare with Excel - but I ran into a usability issue pretty quick where I wanted to search for and select multiple values that required different searches to find.

E.g. using Toolbox's primary grid filtering example, let's say:

- I want to filter on two companies - Anderson LLC and Batz Group
- Open Company col filter
- Type "anderson" - list and grid filters down to a collection of matches
- Click header to uncheck all, then check only Anderson LLC - first company now showing as desired - leave col filter open
- Clear search within filter dialog, then type "batz" to find and select my second company
- All "batz" results are now selected, and I've lost my Anderson filter.

Seems problematic, no?  Will be interested to see how the Excel model we are looking to track handles this same "search for and filter this then that" use case.

amcclain:
Well, I see Excel (on my Mac) doing something very similar - to be honest, it feels just as confusing if not more in Excel, not sure how to feel about it but a good thing in terms of our goal of replicating Excel.

It's surprising how difficult I find it to intuit how this control works (or how I want it to work, speaking about the Excel version just as much as ours).  But wanted to chat back here quick to confirm what you say above re. this working much like Excel

amcclain:
OK - last comment for now - very helpful to try on Windows, as that's clearly (and correctly!) the version that matches the behavior we're implementing with this change.  

Note that Excel for Windows does not show the "auto apply" checkbox we see on Mac above.  I was wondering what the default `commitOnChange` value was for our control - the docs say `true` but it didn't look like that to me scanning the code, could you please confirm?

https://github.com/xh/hoist-react/blob/16d74c321d4b5111d05759c37bcba301708e9873/cmp/grid/Types.ts#L93

I am thinking it should be `false`

oodarluz:
The default commit on change is True, should I make the change to have it false and update the documentation? 
https://github.com/xh/hoist-react/blob/16d74c321d4b5111d05759c37bcba301708e9873/cmp/grid/filter/GridFilterModel.ts#L49

lbwexler (approved):
Looks good to me!  Fixes a good number of bugs.

PR #3895, opened 2025-01-10
Support to register a custom `PersistenceProvider` class

amcclain:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/persist/PersistenceProvider.ts
+++ b/core/persist/PersistenceProvider.ts
@@ -67,14 +67,15 @@ export abstract class PersistenceProvider<S> {
 
     private disposer: IReactionDisposer;
 
-    private static readonly PROVIDERS: Record<string, typeof PersistenceProvider> = {
-        pref: PrefProvider,
-        localStorage: LocalStorageProvider,
-        sessionStorage: SessionStorageProvider,
-        dashView: DashViewProvider,
-        viewManager: ViewManagerProvider,
-        custom: CustomProvider
-    };
+    private static readonly PROVIDERS: Record<PersistenceProviderType, typeof PersistenceProvider> =
+        {
+            pref: PrefProvider,
+            localStorage: LocalStorageProvider,
+            sessionStorage: SessionStorageProvider,
+            dashView: DashViewProvider,
+            viewManager: ViewManagerProvider,
+            custom: CustomProvider
+        };
 
     /**
      * Register a custom `PersistenceProvider` subclass for use by the {@link create} factory called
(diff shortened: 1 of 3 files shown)


core/persist/PersistenceProvider.ts
```diff
@@ -215,3 +220,27 @@ export abstract class PersistenceProvider<S> {
         return null;
     }
 }
+
```
amcclain, line 223: Realize that most of our interfaces are at top (including pre-existing one in this file), but this seemed like an advanced/wonky case that was more distracting up top than I wanted.  Don't feel strongly about it, but that's why I threw it in at the bottom instead.

core/persist/PersistenceProvider.ts
```diff
@@ -98,29 +122,10 @@ export abstract class PersistenceProvider<S> {
...
-                case 'custom':
-                    ret = new CustomProvider(cfg);
-                    break;
-                default:
-                    throw XH.exception(`Unknown Persistence Provider for type: ${type}`);
-            }
+            const clazz = this.PROVIDERS[type] as any;
+            throwIf(!clazz, `Unknown Persistence Provider type: ${type}`);
 
+            ret = new clazz(cfg) as PersistenceProvider<S>;
```
lbwexler, line 128: I *think* if the provider has to be a class with a constructor of this form, we might have a way to incorportate that into the type

core/persist/PersistenceProvider.ts
```diff
@@ -98,29 +122,10 @@ export abstract class PersistenceProvider<S> {
                 if (rest.getData || rest.setData) type = 'custom';
```
lbwexler, line 122: I still *wonder* if we would want to take advantage of this change to make the shortcut form above deprecated, and favor a form of {type: 'pref', key: 'MyPreferenceKey'}.  At that point, not really sure if we need pre-registration, if type could have the type of 'string| type of PrefenceProvider'.

Just thinking out loud here to get all options on the table.  Very much like this proposal as written as well

core/persist/PersistenceProvider.ts
ghsolomon: Making a PR now with that type tweak

core/persist/PersistenceProvider.ts
amcclain: Happy to hand this over to you guys to decide - this was my proposal to move things incrementally along and minimize breaking changes.

amcclain:
Replaced by improved approach in #3902

PR #3900, opened 2025-01-13
Handle edge-case where OAuth redirect opens new tab instead

### LINKED ISSUE #3899: Handle OAuth client edge case where redirect opened as new tab by browser
I use the Arc browser (based on Chromium) and sometimes experience the following flow when logging into an app that uses our `AuthZeroClient` for auth:

* App configured to process OAuth flow via redirect
* Browser decides to open a new tab instead of actually redirecting (exactly why is unclear, might be a site settings option around popup/redirect permissions, or something to do with how Arc lets you "pin" a site)
* New tab opens with Auth0 login, I can complete that and am routed back to the app OK, but still in the newly opened tab.
* Old tab (the one I started from) is left indefinitely on the spinner due to `await never()` here: https://github.com/xh/hoist-react/blob/61b298d0e8f9d264453da092faf26f97809664d0/security/authzero/AuthZeroClient.ts#L106
* Swapping back to that tab leaves me confused as to what is happening - it will spin forever, although I might stare at it for a while thinking that it's just loading for the first time and is about to complete.
* Refreshing the tab gets me into the app OK, as I am in fact generally logged in (via the other tab that was opened) - provided I go back to the "dead tab" within the lifetime of my auth session.

The idea of the `await never()` was to ensure that the app init was (permanently) halted, as we expect the site to unload in that tab and be replaced by the auth provider.  

I propose we replace it with a long wait (say, 10s) to ensure we show the mask for as long as the browser might take to process the redirect.  But after that, we could either throw or simply await (forever) a new alert with a message like "Authentication did not complete in the expected amount of time, or a new tab was opened. Please refresh this tab in your web browser to try again."

It's an edge case but seems like a reasonable fix to avoid.

amcclain:
- Fixes #3899

Hoist P/R Checklist
-------------------

### DIFF
--- a/security/BaseOAuthClient.ts
+++ b/security/BaseOAuthClient.ts
@@ -4,8 +4,11 @@
  *
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
+import {br, fragment} from '@xh/hoist/cmp/layout';
 import {HoistBase, managed, XH} from '@xh/hoist/core';
+import {Icon} from '@xh/hoist/icon';
 import {action, makeObservable} from '@xh/hoist/mobx';
+import {never, wait} from '@xh/hoist/promise';
 import {Token, TokenMap} from '@xh/hoist/security/Token';
 import {Timer} from '@xh/hoist/utils/async';
 import {MINUTES, olderThan, ONE_MINUTE, SECONDS} from '@xh/hoist/utils/datetime';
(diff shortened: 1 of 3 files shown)


amcclain:
Didn't think this warranted a changelog entry - edge case for sure, no rush on releasing either.

Issue #3903, opened 2025-01-15
Review performance of column adds/removes + `useVirtualColumns` default for grids with many columns


amcclain:
Grids with a large number of columns can have noticeably poor performance when adding or removing many columns at once via the column chooser, or by loading saved state via ViewManager.

We should:

* See if we can pin down what a "large number" is in this context - is there any threshold where performance starts to fall off steeply (30 cols, 50, 70?).
* Review the current default value of `useVirtualColumns: false`. Refresh our memory as to any downside to always enabling, consider printing a console warning for devs if not enabled and colCount > threshhold, or have the grid auto-adjust the default based on colCount.
* Review if we could be updating column state more efficiently - performance hit seems especially bad when removing a bunch of columns at once - are we applying updates a column at a time, are we using the optimal AG Grid API?

(Filed as follow-up to Slack chat in #hoist on 2025-01-15)

amcclain:
Closing this after merge of #3905 - this was the immediate cause of poor performance when showing/hiding a large number of columns at once, as col state changes were (prior to that change) processed sequentially.

This ticket also suggested a dev warning or similar to encourage use of virtual columns, but I think we accomplished the more important and actionable improvement with the PR.

PR #3905, opened 2025-01-15
Optimize column state sync

haynesjm42:
Partial fix for #3903 

Hoist P/R Checklist
-------------------


cmp/grid/Grid.ts
```diff
@@ -472,65 +474,53 @@ export class GridLocalModel extends HoistModel {
...
 
-                // 0) Insert the auto group col state if it exists, since we won't have it in our column state list
+                // Insert the auto group col state if it exists, since we won't have it in our column state list
                 const autoColState = agColState.find(c => c.colId === 'ag-Grid-AutoColumn');
                 if (autoColState) {
-                    colState.splice(agColState.indexOf(autoColState), 0, autoColState);
+                    colState.splice(
+                        agColState.indexOf(autoColState),
+                        0,
+                        autoColState as ColumnState
```
amcclain, line 483: Does this case risk spoiling the change detection below due to unfortunate hide/hidden mismatch?

I know it's an edge-case where we have this auto-column inserted by the grid - it's due to direct use of AG-grid grouping APIs, is that right?  So maybe nothing we need to worry about, just checking if there's some tweak we would make to get it 100%

cmp/grid/Grid.ts
```diff
@@ -472,65 +474,53 @@ export class GridLocalModel extends HoistModel {
...
-                            updateBodyWidths: true,
-                            fireResizedEvent: true
-                        });
-                    }
-
-                    return;
-                }
 
-                // 2) Otherwise do an (expensive) full refresh of column state
-                // Merge our state onto the ag column state to get any state which we do not yet support
```
amcclain, line 520: So this key change here - where we are now only going to pass any _different_ state, relies on the partial update behavior [documented here](https://www.ag-grid.com/javascript-data-grid/column-state/#applying-partial-state), is that correct?

I went looking for the changelog entry just to confirm when that change happened on the AG Grid side - I think I found it, but it's listed as fixed in 32.2.0 and we are on 31.x:

Screenshot 2025-01-15 at 6 33 15 PM

Does this change need us to take that update, or am I missing something else here?

amcclain (commented):
This looks like a great simplification and clean-up - and now I see why we saw performance worsen in proportion to number of columns being shown/hidden!

Eager to take it - just now wondering if it requires update to AG v32....

amcclain:
If you haven't already, we should have a look over the calls to `applyColumnState` in `AgGridModel` also, to see if anything there needs to be (or could be) adjusted to account for changes to that API.

cmp/grid/Grid.ts
haynesjm42: Good catch, only an issue when using ag-grid grouping without full-row groups, but worthwhile to be correct

cmp/grid/Grid.ts
haynesjm42: Hmm they certainly claimed that it was supported in 31.x in the way we are using it, and it seemed to work fine in my testing. Not clear what that feature request actually changed, doesn't looks like the docs for that were changed at all.

 https://www.ag-grid.com/archive/31.3.2/javascript-data-grid/column-state/#applying-partial-state

amcclain: Support ticket filed with AG for clarification, just to be careful here (although as discussed, good for us to get to 32.x soon anyway)

amcclain: Confirmed by AG that ticket related to a new flag `partialColumnState` that relates to _initial_ grid state only - https://www.ag-grid.com/react-data-grid/grid-state/#reference-GridState-partialColumnState

amcclain (approved):

amcclain:
@lbwexler ready to merge this one?  Seems like a nice win, potentially big pick-up for wide grids. Would love to get it in, start burning in.

lbwexler:
Hi -- I think we should definitely merge this now if you and John are comfortable -- all for it.
I personally have not reviewed this one -- if you think you want another set of eyes on it, John will probably need to take me through it.

PR #3907, opened 2025-01-16
Remove windows-unfriendly script from package.json

haynesjm42:
@lbwexler you had added this line a long while back (maybe when doing the initial Typescript stuff?) but this doesn't work on windows and causes the pre-commit hooks to fail (which really messes up merges in intellij). I could maybe figure out how to make this work on windows but pretty sure it shouldn't be needed.

Do you remember why this was added? There should be an executable installed in `node_modules/.bin` to run the local install of `tsc` so we should not need to alias it like this in our package.json - curious if you guys have issues with this script removed on your macs

Hoist P/R Checklist
-------------------


amcclain:
Thanks John, agreed should not be any need for this.

PR #3924, opened 2025-02-21
New param for XH.reloadApp() to remove query params

Ryanseanlee:
Hoist P/R Checklist
-------------------

### DIFF
--- a/core/XH.ts
+++ b/core/XH.ts
@@ -392,18 +392,22 @@ export class XHApi {
     /**
      * Trigger a full reload of the current application.
      *
-     * @param path - relative path to reload (e.g. 'mobile/').  Defaults to the
+     * @param path - Relative path to reload (e.g. 'mobile/').  Defaults to the
      * existing location pathname.
+     * @param removeQueryParams - To remove existing URL's parameter string.
      *
      * This method will reload the entire application document in the browser - to trigger a
      * refresh of the loadable content within the app, use {@link refreshAppAsync} instead.
      */
     @action
-    reloadApp(path?: string) {
+    reloadApp(path?: string, removeQueryParams: boolean = false) {
         never().linkTo(this.appLoadModel);
         const {location} = window,
             href = path ? `${location.origin}/${path.replace(/^\/+/, '')}` : location.href,
             url = new URL(href);
+
+        if (removeQueryParams) url.search = '';
+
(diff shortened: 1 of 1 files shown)


core/XH.ts
```diff
@@ -392,18 +392,22 @@ export class XHApi {
...
+     * @param path - Relative path to reload (e.g. 'mobile/').  Defaults to the
      * existing location pathname.
+     * @param removeQueryParams - To remove existing URL's parameter string.
      *
      * This method will reload the entire application document in the browser - to trigger a
      * refresh of the loadable content within the app, use {@link refreshAppAsync} instead.
      */
     @action
-    reloadApp(path?: string) {
+    reloadApp(path?: string, removeQueryParams: boolean = false) {
```
amcclain, line 403: We want to call this from the `AppContainerModel` code that is reloading the page when there is an exception requiring a reload - that's the main case I wanted to catch.

In that use case, we are not spec'ing a path - we want to reload at the current location.  We would need to call this as `reloadApp(null, true)` which is OK but a little sub-optimal.

How about we change the signature to accept a single arg that is either a string - the path, for backwards compatibility - or an object with optional keys `path` and `removeQueryParams`:

```
reloadApp(opts?: string | {path?: string; removeQueryParams?: boolean})
```

That way if we need to add a new option in the future, we can do so without making more breaking changes. And for the refresh after exception requiring reload, we can call as `reloadApp({removeQueryParams: true})`

core/XH.ts
```diff
@@ -392,18 +392,28 @@ export class XHApi {
     /**
      * Trigger a full reload of the current application.
      *
-     * @param path - relative path to reload (e.g. 'mobile/').  Defaults to the
-     * existing location pathname.
+     * @param opts - Options for the reload. Can be either:
```
lbwexler, line 395: This is good, but I think take it all way to export an explicit type `ReloadAppOptions`  that is  `string | {path?: string; removeQueryParams?: boolean}`.

This is just easier to document in tsdoc and looks better in the signature

lbwexler:
@Ryanseanlee -- check this out and let me know if it makes sense to you?  Idea is that we want to let the new interface handle the documentation for us.  Also want to make the code focus on the new interface -- the backward compatibility is more like "adapter" code

PR #3937, opened 2025-03-11
CodeInput/JsonInput should format automatically in readOnly mode.

RyanMG:
Linked issue: https://github.com/xh/hoist-react/issues/1923

### What

Adds an `autoFormat` prop to `CodeInput` / `JsonInput`. When set, the input formats its content for display via the configured `formatter`:

- **Readonly inputs** - defaults to `true` (pass `false` to opt out). Callers can bind raw values and let the input render them.
- **Editable inputs** - opt in with `autoFormat: true` to format content on blur (never mid-edit, so user edits and cursor position are preserved while typing). Users get clean, consistently-formatted JSON without reaching for the format button.

The `formatter` is invoked within a try/catch, so a formatter that throws (e.g. on invalid JSON) leaves the original content unchanged.

### Why

Callers no longer need to pre-format JSON before handing it to a `jsonInput`. For readonly displays they can pass the raw value (object or string) and let the input render it - since `fmtJson` delegates to `JSON.stringify`, an object's `toJSON()` is honored automatically. This makes display code more declarative (value = data, `formatter` = how to render) and removes a pile of inline `fmtJson(...)` / `JSON.stringify(..., null, 2)` calls across the Admin Console.

The editable case is an equally meaningful win: writable JSON editors stay tidy on their own. Enabled on the two library writable editors - the grid filter dialog (`GridFilterDialog`) and the REST form `json` field (`RestFormField`) - dropping the manual pretty-printing in the former.

Admin JSON panels are consolidated behind a small shared `adminJsonDisplay` component.

Companion Toolbox PR: https://github.com/xh/toolbox/pull/869

Hoist P/R Checklist
-------------------


desktop/cmp/input/JsonInput.ts
```diff
@@ -21,11 +21,18 @@ export const [JsonInput, jsonInput] = hoistCmp.withFactory<JsonInputProps>({
     displayName: 'JsonInput',
     className: 'xh-json-input',
     render(props, ref) {
+        let {value, ...jsonInputProps} = props;
+
+        if (typeof value !== 'string') {
```
RyanMG, line 26: Originally was checking if value was an object, but `null` seems to be a valid value to be passed in, and is an object in JS.

desktop/cmp/input/JsonInput.ts
```diff
@@ -21,11 +22,20 @@ export const [JsonInput, jsonInput] = hoistCmp.withFactory<JsonInputProps>({
     displayName: 'JsonInput',
     className: 'xh-json-input',
     render(props, ref) {
+        let {value, ...codeInputProps} = props;
+
+        if (isPlainObject(value)) {
+            value = JSON.stringify(value, null, 2);
```
haynesjm42, line 28: Any processing of the value should be done in the input model, in this case that would be `CodeInputModel` since this input control is just a wrapper on the `codeInput` component, might make sense to introduce a `JsonInputModel` class here for special json handling. `HoistInputModel` is the base class for all input component models and has methods for converting between the external and internal values, which is where code like this needs to live. It is important to use those methods as our input components support both the controlled mode (providing `value` and `onChange` in props) as well as binding to a property of a provided model.

`DateInput` is probably a good example to follow here, where it has a `valueType` prop to tell it whether it should be dealing with a `LocalDate` or a `Date`.

I'm personally in favor of this change overall, others might feel differently as it does open the door for potentially confusing behavior as the object will be stringified and then parsed when written back, though any usages currently would effectively be doing the same thing, just that they'd just be needing to do the stringifying and parsing themselves.

@amcclain / @lbwexler any thoughts on supporting both strings and objects in JsonInput?

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@
 * Modify `TabContainerModel` to make more methods `protected`, improving extensibility for advanced
 use-cases.
 
+### ⚙️ Technical
+
+* Updated the `jsonInput` component to auto-format its provided value for display.
+It can accept JSON, stringified JSON or plain objects and output proper pretty-printed JSON in the
+UI - without the need for manual formatting in the call.
+
```
amcclain, line 15: We had a release underneath this PR, sorry :) - will need to move this up

desktop/cmp/input/JsonInput.ts
amcclain: It's a good question - I had not considered it.

My concern was that it could muddy the waters a bit in terms of what value types this component supports. Was thinking we could use a type to represent valid JSON values - typefest has `JsonValue` which I believe would be it.

However for that to be a valid concern at all, we'd need to actually spec the type of values that inputs support, and as David points out above, `HoistInputProps.value` is stuck on `any`.  Can we add a generic to support setting the value type on that props interface?  

That would be a very different change from pretty-printing - I think that still need to be addressed, see comment on the linked issue, but adding proper value typing _seems_ like a good idea.  My only concern is that if we did so we would create work for devs who were unwittingly benefitting from looser typing with `any`, but that's kinda the idea...

admin/tabs/activity/tracking/detail/ActivityDetailModel.ts
```diff
@@ -9,15 +9,14 @@ import * as Col from '@xh/hoist/admin/columns';
...
-import {fmtJson} from '@xh/hoist/format';
 import {action, computed, makeObservable, observable} from '@xh/hoist/mobx';
 import {ActivityTrackingModel} from '../ActivityTrackingModel';
 
 export class ActivityDetailModel extends HoistModel {
     @lookup(ActivityTrackingModel) activityTrackingModel: ActivityTrackingModel;
     @managed gridModel: GridModel;
     @managed formModel: FormModel;
-    @observable formattedData;
+    @observable displayData;
```
RyanMG, line 19: This model is no longer doing the formatting. Update the name to reflect. See https://github.com/xh/hoist-react/pull/3937/files#diff-2a82d78cc59aadd90860fcd0383be76d3b7491d7b19736a18857b6b93d30cd66L117-L125

desktop/cmp/input/CodeInput.ts
```diff
@@ -326,22 +334,33 @@ class CodeInputModel extends HoistInputModel {
...
-            return this.componentProps.formatter(str);
-        } catch (e) {
-            return str;
+            return this.componentProps.formatter(val);
+        } catch (ignored) {
+            return val;
         }
     }
 
+    private formatValue() {
```
RyanMG, line 353: @amcclain had some naming / formatting updates he wanted done here.

1. onAutoFormat -> formatValue. We are choosing to use autoFormat as the new prop (on interface) to format readonly data when loading into the UI. Anselm felt onAutoFormat was not the best name of the reaction happening here (to a user requesting the use of the formatter to update their UI)
2.  tryPrettyPrint was use only once, in this method, to run the formatter. Inlining that call to the method itself.

desktop/cmp/input/CodeInput.ts
```diff
@@ -326,22 +334,33 @@ class CodeInputModel extends HoistInputModel {
         if (this.cursor) this.clearSearchResults();
     };
 
-    onAutoFormat = () => {
-        if (!isFunction(this.componentProps.formatter)) return;
+    override toInternal(val: any) {
```
RyanMG, line 337: Override fo the toInternal used by the `HoistInputModel`. Now will run the formatter if requested and properly configured.

admin/tabs/activity/tracking/detail/ActivityDetailView.ts
```diff
@@ -134,10 +134,11 @@ const detailRecForm = hoistCmp.factory<ActivityDetailModel>(({model}) => {
 const additionalDataJsonInput = hoistCmp.factory<ActivityDetailModel>(({model}) => {
     return jsonInput({
         readonly: true,
+        autoFormat: true,
         width: '100%',
         height: '100%',
         showCopyButton: true,
-        value: model.formattedData
+        value: model.displayData
```
RyanMG, line 141: See https://github.com/xh/hoist-react/pull/3937/files#r2001715715

desktop/cmp/input/CodeInput.ts
```diff
@@ -326,22 +334,33 @@ class CodeInputModel extends HoistInputModel {
...
-            return this.componentProps.formatter(str);
-        } catch (e) {
-            return str;
+            return this.componentProps.formatter(val);
+        } catch (ignored) {
+            return val;
         }
     }
 
+    private formatAndSetEditorValue() {
```
RyanMG, line 353: @amcclain had some naming / formatting updates he wanted done here.

onAutoFormat -> formatValue was the first pass at the name. We are choosing to use autoFormat as the new prop (on interface) to format readonly data when loading into the UI. Anselm felt onAutoFormat was not the best name of the reaction happening here (to a user requesting the use of the formatter to update their UI)
tryPrettyPrint was use only once, in this method, to run the formatter. Inlining that call to the method itself.
However, we subsequently added safeFormatValue, which is much leaner - it just takes a val, tries to format it, and returns it one way or the other. This existing method was already safe (try/catch block), so to a future developer it may be configuring why the similar names. So I opted to update this method to denote that it would be sending the value to the editor (code mirror) instead of returning said value.

desktop/cmp/input/JsonInput.ts
```diff
@@ -22,8 +22,8 @@ export const [JsonInput, jsonInput] = hoistCmp.withFactory<JsonInputProps>({
     className: 'xh-json-input',
     render(props, ref) {
         return codeInput({
-            linter: linter,
-            formatter: fmtJson,
+            linter,
+            formatter: safeFmtJson,
```
RyanMG, line 26: Unsure if we want this? We create a "safe" version of `fmtJson` (https://github.com/xh/hoist-react/pull/3937/files#diff-1bd2a34b6889e1254a813e22f3fecee2169724007549f9c6d2396658c54ebc7bR76). I made this update to use that for jsonInput, but that may create issues I am not thinking of?

RyanMG:
Had a meeting this morning with @amcclain, and in that we reviewed this issue broadly. Updates done in that meeting applied to this PR. CC:// @haynesjm42

desktop/cmp/input/CodeInput.ts
```diff
@@ -43,6 +43,14 @@ export interface CodeInputProps extends HoistProps, HoistInputProps, LayoutProps
     /** True to focus the control on render. */
```
RyanMG, line 43: This is a `desktop` component. I did not see a mobile version, but let me know if I just overlooked something.

haynesjm42 (approved):

amcclain:
See - autoformatting dreams do (eventually) come true! :D

Issue #3943, opened 2025-03-14
Tabbing into a `BooleanEditor` with `quickToggle: true` toggles the value


TomTirapani:
If a user tabs into an editable column that uses a `BooleanEditor` with `quickToggle: true`, the value is toggled. I would expect the user to have to take an additional action (e.g. click, hit enter) to toggle the value - using the tab key to navigate grids shouldn't change values.

To reproduce, move the checkbox column to the right and tab into it from a previous column:
https://toolbox-dev.xh.io/app/grids/inlineEditing

<img width="342" alt="Image" src="https://github.com/user-attachments/assets/4cf5530c-911d-416d-b4bb-cadf338360b7" />

<img width="315" alt="Image" src="https://github.com/user-attachments/assets/471ab863-98ef-49de-a82e-db58ec5f2aa2" />

TomTirapani:
See PR for possible workaround: https://github.com/xh/hoist-react/pull/3944

PR #3944, opened 2025-03-14
Workaround for `BooleanEditor.quickToggle` and keyboard tab interaction

TomTirapani:
Fixes https://github.com/xh/hoist-react/issues/3943

I'm not entirely happy with this workaround - however, I spent far too long trying different approaches to no avail. Its still a net positive over the previous behaviour. If anybody has any better solutions I'm all ears :)

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/editors/BooleanEditor.ts
+++ b/desktop/cmp/grid/editors/BooleanEditor.ts
@@ -49,11 +49,23 @@ export const [BooleanEditor, booleanEditor] = hoistCmp.withFactory<BooleanEditor
     }
 });
 
-function useInstantEditor({onValueChange, initialValue, stopEditing}: CustomCellEditorProps, ref) {
+function useInstantEditor(
+    {onValueChange, initialValue, stopEditing, eventKey, eGridCell}: CustomCellEditorProps,
+    ref
+) {
+    // Don't toggle if the user has tabbed into the editor. See https://github.com/xh/hoist-react/issues/3943.
+    // Fortunately, `eventKey` is null for tab, so we can use that to accept other keyboard events.
+    // Unfortunately, it is also null for mouse events, so we check if the grid cell is currently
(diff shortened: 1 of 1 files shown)


desktop/cmp/grid/editors/BooleanEditor.ts
```diff
@@ -49,11 +49,23 @@ export const [BooleanEditor, booleanEditor] = hoistCmp.withFactory<BooleanEditor
...
+    {onValueChange, initialValue, stopEditing, eventKey, eGridCell}: CustomCellEditorProps,
+    ref
+) {
+    // Don't toggle if the user has tabbed into the editor. See https://github.com/xh/hoist-react/issues/3943.
+    // Fortunately, `eventKey` is null for tab, so we can use that to accept other keyboard events.
+    // Unfortunately, it is also null for mouse events, so we check if the grid cell is currently
+    // underneath the mouse position via `:hover` selector.
     useEffect(() => {
-        onValueChange(!initialValue);
+        const els = document.querySelectorAll(':hover'),
```
RyanMG, line 61: I can also not find a better solution. It seems like AG grid swallows the important information here (as you note - the source of the event being tab). Another solution I looked into is not setting the checkbox for "edit mode" when navigating to it (a re-render as the checkbox enters edit mode). This was also stymied by AG grid's provided details - again not being told enough to determine if we are being asked to go into edit mode because the field is being focused to via keyboard navigation or a mouse click.

As for this fix - its very edge case, but it looks like I can still get the error by having my mouse cursor over the checkbox while following the steps and tabbing in. In that case, the hover source IS the element in question. However, this is a much rarer occurrence, so this change looks like a net win.

lbwexler:
This is a really horrible bug -- thanks for addressing, and not letting the perfect be the enemy of the good!

PR #3946, opened 2025-03-14
Allow for form value persistence with integrated config options

RyanMG:
This is a off hand request that came up during a PR review session this week. Forms are not generally persisted as state - they are usually filled out, saved, and their output is persisted for use elsewhere. However, there are isntances in Hoist based applications where forms are being used in a more stateful way, such as storing filtering options for other data.
At least one such use case was seen in the PR review, and persisting the form state was implemented manually. It was suggested that this would be a good optional set of options to provide in the hoist react form model proper, to allow this persistance without additional boilerplate.

This PR / change uses as a baseline the configuration and execution of configurable persistence from the GroupingChooserModel, updated to work with the added complexity of a form.

Hoist P/R Checklist
-------------------


cmp/form/FormModel.ts
```diff
@@ -22,13 +30,21 @@ export interface FormConfig {
...
     disabled?: boolean;
     readonly?: boolean;
 
     /** @internal */
     xhImpl?: boolean;
 }
 
+export interface FormPersistOptions extends PersistOptions {
+    /** If persisting only a subset of all fields, provide an array of field names. */
+    fieldsToPersist?: string[];
```
lbwexler, line 45: could consider using two properties -- fieldsToInclude and fieldsToExclude.  We have that pattern elsewhere

cmp/form/FormModel.ts
```diff
@@ -274,4 +293,37 @@ export class FormModel extends HoistModel {
...
+                    path: `${path}.fields.${name}.value`,
+                    ...rootPersistWith
+                },
+                target: {
+                    getPersistableState: () => {
+                        return new PersistableState(this.fields[name].value);
+                    },
+                    setPersistableState: ({value}) => {
+                        // There is no metadata on a field to denote it is a date.
+                        // Use a regex matcher to tests for dates and format matches accurately.
```
lbwexler, line 316: can/should we generalize this?

RyanMG: Are you referring to updating the form field to denote its 'type" (e.g. date, number, etc)? to use here, rather than a regex matcher for date-like strings?

cmp/form/FormModel.ts
```diff
@@ -274,4 +293,37 @@ export class FormModel extends HoistModel {
...
+                        if (isString(value) && value.match(/^\d{4}-\d{2}-\d{2}$/)) {
+                            this.setValues({[name]: LocalDate.from(value)});
+                            return;
+                        }
+
+                        this.setValues({[name]: value});
```
RyanMG, line 322: The internal means to set a value is this setValues (plural). A form has multiple child models - one for each form field. Hence the form parent's setValues implementation iterates through all child models, comparing names to the provided name in the argument object passed into setValues.

RyanMG:
@ghsolomon thanks for the review. When I get caught up on tasks and can return to this, I will make the requested change(es) and / or request some follow up!

RyanMG:
I was able to get this working with a single persistence provider for the whole form.

`I think we probably need to include the "valueType" in the serialized interface to know how to properly deserialize.`
Where can I get access to this? From the point where I am doing the work, I can get access to a `fieldModel`, but I don't see where on this model it tells me its data type (or value type).
Thanks.

ghsolomon:
> Where can I get access to this? From the point where I am doing the work, I can get access to a fieldModel, but I don't see where on this model it tells me its data type (or value type).

You're right -- unlike Store Fields, FormFields don't have a `type` property.  I think the best we could probably do is something like what we do in `GridExportService` for determining a cell-specific type:
https://github.com/xh/hoist-react/blob/develop/svc/GridExportService.ts#L417

We could recurse into the values of arrays or objects to handle nested Dates + LocalDates, but I think that's about the best we could do. @lbwexler's suggestion to add a hook sounds like a good way for apps with more complex objects (could imagine custom fields with values that are class instances) to specify their deserialization methods.

RyanMG:
@ghsolomon I have a "fix". I dont love it, so I am posting it in its base working state before I do any cleanup  or other work to finish it off.
All in this last commit:
https://github.com/xh/hoist-react/pull/3946/commits/bd1e4c5fc1dad52bcad336360cac9af8f17c9084

This feels like bending over backwards to accommodate one narrow use case. In theory it keeps things clean since we have one source of truth on the form field's value. However, it would feel cleaner just to require the user to provide a value type in the form model field object IF they want to use persistence. Scopes the change down to only as needed.

I tried making a little helper like you linked for the grid. It may be I am missing something, but it was not proving useful to me. The main issue is that a date out of persistence is a string, and the only want to verify that is to attempt to turn it into a date and test the result. But then technically we can run into issues if a user enters a date like string into a regular text field. Moreover, the date test won't know about local date vs any other internal date types.
But please let me know if I mis understood your request there.

cmp/form/FormModel.ts
```diff
@@ -316,7 +316,13 @@ export class FormModel extends HoistModel {
                 getPersistableState: () => {
                     return new PersistableState(
                         fieldNameMap.reduce(
-                            (acc, name) => ({...acc, [name]: this.fields[name].value}),
+                            (acc, name) => ({
+                                ...acc,
+                                [name]: {
+                                    value: this.fields[name].value,
+                                    valueType: this.fields[name].valueType
```
RyanMG, line 323: See new property https://github.com/xh/hoist-react/pull/3946/commits/bd1e4c5fc1dad52bcad336360cac9af8f17c9084#diff-1810981277d792338bd36c1316e1a8d4a7e7c5a78e652f58588b8b1c607f7337R191

cmp/form/FormModel.ts
```diff
@@ -326,20 +332,29 @@ export class FormModel extends HoistModel {
...
             },
             owner: this
         });
     }
+
+    /**
+     * TODO
+     */
+    private processFormValue(persistedValue: PlainObject) {
+        if (isNil(persistedValue.value) || persistedValue.value === '') return null;
```
RyanMG, line 348: Is a user enters a value then clears it, we dont want to populate the non-result.

cmp/form/field/BaseFieldModel.ts
```diff
@@ -183,6 +188,22 @@ export abstract class BaseFieldModel extends HoistModel {
...
 
+    setValueType(elementName: string, valueType?: string) {
+        switch (elementName) {
+            case 'Checkbox':
+                this.valueType = 'boolean';
+                break;
+            case 'NumberInput':
+                this.valueType = 'number';
+                break;
+            case 'DateInput':
```
RyanMG, line 199: I assume there are more of these I need to track down an support.

desktop/cmp/form/FormField.ts
```diff
@@ -171,6 +171,10 @@ export const [FormField, formField] = hoistCmp.withFactory<FormFieldProps>({
         if (disabled) classes.push('xh-form-field-disabled');
         if (displayNotValid) classes.push('xh-form-field-invalid');
 
+        // The form model does not know the value type of any given field.]
+        // Use the UI element to set the value type on the form field model.
+        model.setValueType(childElementName, child?.props?.valueType);
```
RyanMG, line 176: I hate it. But this does let me derive the value type from its associated input.

RyanMG:
@ghsolomon this works. My date testing was not working when I attempted something similar - obviously I had some outside error, and walked away from it too quickly. 
Thanks.

cmp/form/FormModel.ts
```diff
@@ -22,13 +43,23 @@ export interface FormConfig {
...
+    persistWith?: FormPersistOptions;
+
     disabled?: boolean;
     readonly?: boolean;
 
     /** @internal */
     xhImpl?: boolean;
 }
 
+export interface FormPersistOptions extends PersistOptions {
```
ghsolomon, line 56: Should this be a waterfall where only one of these two properties is respected?

i.e. 
`persistedFields = fieldsToExclude ? without(allFields, ...fieldsToExclude) : fieldsToInclude ?? allFields`

Just thinking it would be strange to specify both since it would be clearer to just not include the "excluded" fields in your list of "included" fields.

cmp/form/FormModel.ts
```diff
@@ -274,4 +308,83 @@ export class FormModel extends HoistModel {
...
         );
     }
+
+    private initPersist({
+        fieldsToInclude = null,
+        fieldsToExclude = null,
+        path = 'formValues',
+        ...rootPersistWith
+    }: FormPersistOptions) {
+        const fieldNameMap = this.createFormFieldsToPersistListing(
```
ghsolomon, line 318: Might just be me mixing Java with JS, but to me a Map suggests an object with key/value pairs. I might just call this "fieldsToPersist" or "fieldNamesToPersist".

cmp/form/FormModel.ts
```diff
@@ -274,4 +308,83 @@ export class FormModel extends HoistModel {
...
+        );
+
+        PersistenceProvider.create({
+            persistOptions: {
+                path,
+                ...rootPersistWith
+            },
+            target: {
+                getPersistableState: () => {
+                    return new PersistableState(
```
ghsolomon, line 330: Since the `serialize` and `deserialize` methods are recursive, could simplify this to just:


```
getPersistableState: () =>
                    new PersistableState(this.serialize(pick(this.getData(), fieldNameMap))),


setPersistableState: ({value: formValues}) =>
                    this.setValues(this.deserialize(pick(formValues, fieldNameMap)))
```

cmp/form/FormModel.ts
```diff
@@ -274,4 +308,83 @@ export class FormModel extends HoistModel {
...
+                        }, {} as PlainObject)
+                    );
+                }
+            },
+            owner: this
+        });
+    }
+
+    private serialize(formValue: unknown) {
+        if (isArray(formValue)) return formValue.map(this.serialize);
```
ghsolomon, line 355: Does passing the function reference like this work? I think because `serialize` is not an arrow function, it will lose its "this" context when passed this way, but maybe it doesn't matter since the method implementation isn't referencing any other members on the model besides the method itself.

cmp/form/FormModel.ts
```diff
@@ -274,4 +308,83 @@ export class FormModel extends HoistModel {
...
+        if (isDate(formValue)) return {__type: 'date', value: formValue.toJSON()};
+        if (isLocalDate(formValue)) return {__type: 'localDate', value: formValue.toJSON()};
+        if (isObject(formValue)) return mapValues(formValue, this.serialize);
+        return formValue;
+    }
+
+    private deserialize(formValue: unknown) {
+        if (isArray(formValue)) return formValue.map(this.deserialize);
```
ghsolomon, line 363: Same question as above re passing a reference to the method itself when not using an arrow function

ghsolomon:
I think this is looking good - had one last thought. If there is persisted state, the form will begin in a dirty state. That seems right and in keeping with the rest of our persistence story, but wondering if there would ever be a use case where you'd want the form to *initialize* to the persisted state, such that resetting the form restores the persisted state rather than the in-code default state.

RyanMG:
@ghsolomon Im not sure I have the answer for that. The use case which brought this up was for places where forms are used as filters for datasets, but even there your use case is likely valid.

Im also not sure how much that change entails to what is here now. But if this is what we need, I can look into it.

cmp/form/FormModel.ts
```diff
@@ -274,4 +310,50 @@ export class FormModel extends HoistModel {
...
+                    this.setValues(this.deserialize(pick(formValues, fieldNamesToPersist)))
+            },
+            owner: this
+        });
+    }
+
+    private serialize = (formValue: unknown) => {
+        if (isArray(formValue)) return formValue.map(this.serialize);
+        if (isDate(formValue)) return {__type: 'date', value: formValue.toJSON()};
+        if (isLocalDate(formValue)) return {__type: 'localDate', value: formValue.toJSON()};
```
amcclain, line 343: I'm down with this use of `__type` - seems very standard and reasonable to me. Anyone inspecting the persisted contents should be able to suss out what's going on.

In a few other places we use `_xh` as a prefix - e.g. `_xhType` or `_xhFieldType`.  Mainly thinking of possible future cases where we are serializing type/value pairs in other contexts and this convention gets spread around, would there be any advantage to using that `xh` to signal "this is something that Hoist is processing"?  

Don't feel strongly about it, just throwing it out there. Am 100% fine if `__type` is a more general convention, or feels like one.

Great that we are pushing past this trickiness.

amcclain:
>  wondering if there would ever be a use case where you'd want the form to initialize to the persisted state

I had the same question, and think there's a decent chance that this will come up...but also might not...  If we needed to support, I think we could handle with a custom persist option (something like `initPersistedValues`, or something that's more intelligible.  Should then be straightforward to switch on that to call either `init` or `setValues`.

BUT unless we know we need it, let's not add it yet and wait and see. The use cases we're anticipating for this are typically _not_ ones where we look to see if the form model is dirty, or if they are (e.g. a query form with a reset option), the current behavior is almost certainly the one you would want.

So all that said, this looks great to me and - assuming it's all working - I'd love to get it merged.  @minusleaf and I just today came across a case in a client app where we could use this.

amcclain (approved):

amcclain:
Arg sorry - didn't realize that I had written up but then never submitted a set of earlier comments on this PR.  Just resolved them now - happily, all of them had already been discussed and addressed, best case scenario :)

cmp/form/FormModel.ts
```diff
@@ -22,13 +45,23 @@ export interface FormConfig {
...
 
     /** @internal */
     xhImpl?: boolean;
 }
 
+export interface FormPersistOptions extends PersistOptions {
+    /** If persisting only a subset of all fields, provide an array of field names. */
+    fieldsToInclude?: string[];
+    /** If excluding a subset of all fields, provide an array of field names. */
+    fieldsToExclude?: string[];
```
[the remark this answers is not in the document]
RyanMG: Updated.

cmp/form/FormModel.ts
ghsolomon: I had the same thought last night and agree -- probably better to use the xh prefix so we could do this elsewhere, be consistent, and not worry about a name collision.

Was thinking it would be nice if our `markPersist` and `@persist` decorators also handled serializing / deserializing `Dates` and `LocalDates` and collections that contain them. (Ex. `bindable` date input that you want to persist / don't need a whole `Form` and `FormModel`).

ghsolomon: Just to clarify -- not saying we need to add that for this PR, but using the `xh` prefix convention gives us room to implement this in more places where we might not be able to make as many assumptions about the underlying data.

amcclain: Great agreed 100% good point about persisting (Local)Date properties directly - as a follow-on enhancement

cmp/form/FormModel.ts
RyanMG: Fixed

cmp/form/FormModel.ts
RyanMG: Fixed

RyanMG:
@amcclain @ghsolomon Updated with the last requested change + merge up develop

amcclain:
Excellent. 

@lbwexler / @ghsolomon this looks ready to go to me - would love to merge today as I have accumulated 1-2 use cases for it.  Thanks for having another look.

ghsolomon:
Looks good to me!

lbwexler:
Looking now and will aim to merge

Issue #3947, opened 2025-03-18
WebSocket heartbeat should report client app version, show in admin grid


amcclain:
Each client with a connected websocket sends a heartbeat message back on that channel to the server. Their status is viewable on the Admin Cluster > Websockets tab.

It would be helpful for the heartbeat response from each client to include the current version number of the client app loaded in the browser, in case it ever differs from the released version (eg after a release, for a user who might have dismissed or ignored the update banner).  This should be visible as a new column in the admin grid.

RyanMG:
@amcclain PRs are up for review.

In app:
<img width="1366" alt="Image" src="https://github.com/user-attachments/assets/fcbbc308-1dc5-468b-bf45-10a59954244d" />

RyanMG:
I received comments today from @lbwexler . I will loop all stakeholders in when those changes are ready for re-review.

lbwexler:
Namely, we discussed that it would be perhaps cleaner to keep the channel immutable (and the heartbeat simple) and have the app version (or any other meta data we wish to send), set when the channel is created.  See the following --header or query params are fine with me.  Perhaps header with json content is most scalable.

https://stackoverflow.com/questions/24933593/is-there-a-way-to-send-meta-data-when-opening-a-websocket

PR #3948, opened 2025-03-18
Log client app version in websocket to be displayed for admins

RyanMG:
Ticket https://github.com/xh/hoist-react/issues/3947

Server update [here](https://github.com/xh/hoist-core/pull/451)

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/cluster/instances/websocket/WebSocketColumns.ts
+++ b/admin/tabs/cluster/instances/websocket/WebSocketColumns.ts
@@ -77,7 +77,7 @@ export const clientAppVersion: ColumnSpec = {
     field: {
         name: 'clientAppVersion',
         type: 'string',
-        displayName: 'Client App Version'
+        displayName: 'Client Version'
     },
     width: 120
 };
--- a/admin/tabs/cluster/instances/websocket/WebSocketColumns.ts
+++ b/admin/tabs/cluster/instances/websocket/WebSocketColumns.ts
@@ -72,3 +72,12 @@ export const lastReceivedTime: ColumnSpec = {
     ...Col.compactDate,
     width: 140
 };
+
+export const clientAppVersion: ColumnSpec = {
+    field: {
+        name: 'clientAppVersion',
+        type: 'string',
+        displayName: 'Client App Version'
+    },
+    width: 120
+};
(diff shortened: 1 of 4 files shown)


amcclain:
Great, thanks for jumping on this!  Just had a chance to update the ticket with a proper description.

Goal is to get this version info into the Hoist admin console grid.  We'll need to look at hoist-core to see how these messages are handled and thread that through into the `HoistWebSocketChannel` object, which is what is holding the state reported by that grid.

admin/tabs/cluster/instances/websocket/WebSocketColumns.ts
```diff
@@ -72,3 +72,12 @@ export const lastReceivedTime: ColumnSpec = {
...
     width: 140
 };
+
+export const clientAppVersion: ColumnSpec = {
+    field: {
+        name: 'clientAppVersion',
+        type: 'string',
+        displayName: 'Client App Version'
+    },
+    width: 120
```
RyanMG, line 82: 120 is holding "7.0 SNAPSHOT" with no issue, but unsure how much we expect this value could grow.

svc/WebSocketService.ts
```diff
@@ -173,7 +173,11 @@ export class WebSocketService extends HoistService {
     heartbeatOrReconnect() {
         this.updateConnectedStatus();
         if (this.connected) {
-            this.sendMessage({topic: this.HEARTBEAT_TOPIC, data: 'ping'});
+            this.sendMessage({
+                topic: this.HEARTBEAT_TOPIC,
+                data: 'ping',
+                clientAppVersion: XH.appVersion
```
RyanMG, line 179: I initially made `data` an object and sent `clientAppVersion` as a property in said object.

I am unsure though how clients may be using this service. If they are extending and overriding `data` in `sendMessage` we would lose this value. However, there is also nothing stopping them from overriding `heartbeatOrReconnect` either.
Open to suggestions or notes to help better understand real world use cases.

RyanMG:
@lbwexler updated.

1. Update the column name from Client App Version -> Client Version
2. Send client version with the initial handshake, rather than with each heartbeat

lbwexler (approved):

Issue #3950, opened 2025-04-01
`Store.summaryRecord` should not throw when `summaryRecords` is null


ghsolomon:
`Store.summaryRecord` throws when there are multiple `summaryRecords`:
https://github.com/xh/hoist-react/blob/develop/data/Store.ts#L696

This check makes some sense, but because `summaryRecords` can be null, the length check itself can also (unintentionally) throw.  In the case of `summaryRecords === null`, `get summaryRecord()` should probably also return `null`.

RyanMG:
Is this all you were looking for? https://github.com/xh/hoist-react/compare/summary-records-null-check?expand=1
Or was there more needed I am missing?

RyanMG:
@ghsolomon https://github.com/xh/hoist-react/pull/3952

PR #3952, opened 2025-04-02
Null check on summaryRecords before accessing / using value

RyanMG:
For https://github.com/xh/hoist-react/issues/3950

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.ts
+++ b/data/Store.ts
@@ -18,6 +18,7 @@ import {
     isEmpty,
     isFunction,
     isNil,
+    isNull,
     isString,
     values,
     remove as lodashRemove,
@@ -692,6 +693,8 @@ export class Store extends HoistBase {
      * for backwards compat with app code predating support for multiple {@link summaryRecords}.
      */
     get summaryRecord(): StoreRecord {
+        if (isNull(this.summaryRecords)) return null;
+
         throwIf(
             this.summaryRecords.length > 1,
             'Store has multiple summary records - must access via Store.summaryRecords.'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v73.0.0-SNAPSHOT - unreleased
 
 * Support for reporting Client Version in Admin WebSockets tab.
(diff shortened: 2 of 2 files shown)


data/Store.ts
```diff
@@ -692,6 +693,8 @@ export class Store extends HoistBase {
      * for backwards compat with app code predating support for multiple {@link summaryRecords}.
      */
     get summaryRecord(): StoreRecord {
```
RyanMG, line 695: I was going to update this to `get summaryRecord(): StoreRecord | null {` to reflect this update. However, doing so will make every call to `summaryRecord` require a null check before using the returned records. I didn't want to introduce that level of added complexity before discussing with the team.

ghsolomon: Yeah - you’ll find a lot of methods / properties in Hoist that can return null. Our base tsconfig does not have strictNullChecks enabled. I think what you have is appropriate since it’s consistent with the rest of the framework.

data/Store.ts
```diff
@@ -692,6 +693,8 @@ export class Store extends HoistBase {
...
+        if (isNull(this.summaryRecords)) return null;
```
ghsolomon, line 696: Nit - would be fine to just check if `summaryRecords` is truthy if you wanted to just write `if (!this.summaryRecords)`, but I believe what you have is correct since `summaryRecords` cannot be `undefined`.

ghsolomon (approved):

data/Store.ts
amcclain: Re tightening down our tsconfig - my goal is for us to (one day) get to setting `noImplicitAny: true`, then we could turn our sights on `strictNullChecks`.  

I do believe that it would be a net positive, but the cost/benefit calculation now is....unclear... given the amount of change it would entail.

data/Store.ts
amcclain: I'm also a frequent user of lodash `isEmpty` - would be usable here, as we want to return `null` if `summaryRecords` is nullish or empty.

Issue #3939, opened 2025-03-12
Enable `MsalClient` to opt-in to performance telemetry


amcclain:
I would like to have our MSAL client class import the `BrowserPerformanceClient` class and support a config-driven flag to enable it.

See https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/performance.md

Note too option to have performance tracing stopped, which might be something we wish to have happen automatically, eg after full app init (as a further config option around this feature).

lbwexler:
Am going to try this out in application world first -- we don't need to make any changes in Hoist.  After that, we can see what, if anything makes sense for the toolkit to implement.  This may dovetail with some potential work around client-side monitoring.

amcclain:
From the linked doc, it looks like we're supposed to pass the entire overall config into the telemetry client:

```
const msalInstance = new PublicClientApplication({
    ...msalConfig,
    telemetry: {
        client: new BrowserPerformanceClient(msalConfig)
    }
});
```

That merged `msalConfig` is constructed only within `MsalClient`, so we can't accurately do that in app world alone. Unclear what possible impact would be of passing telemetry client partial/incomplete config.

PR #3954, opened 2025-04-04
MSAL v4.8 w/telemetry + include headers cb in `FetchService` track timing

lbwexler:
Hoist P/R Checklist
-------------------

### DIFF
--- a/security/msal/MsalClient.ts
+++ b/security/msal/MsalClient.ts
@@ -289,18 +289,14 @@ export class MsalClient extends BaseOAuthClient<MsalClientConfig, MsalTokenSpec>
                         {name, startTimeMs, durationMs, success, errorName, errorCode} = e,
                         eTime = startTimeMs ? new Date(startTimeMs) : new Date();
 
-                    if (!events[name]) {
-                        events[name] = {
-                            firstTime: eTime,
-                            lastTime: eTime,
-                            successCount: 0,
-                            failureCount: 0,
-                            duration: {count: 0, total: 0, average: 0, worst: 0},
-                            lastFailure: null
-                        };
-                    }
-
-                    const eResult = events[name];
+                    const eResult = (events[name] ??= {
+                        firstTime: eTime,
+                        lastTime: eTime,
+                        successCount: 0,
+                        failureCount: 0,
+                        duration: {count: 0, total: 0, average: 0, worst: 0},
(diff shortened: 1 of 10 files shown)


amcclain (approved):

amcclain:
Pls see https://github.com/xh/hoist-react/issues/3939#issuecomment-2779011907 in case we want to make what I think is still a required change in Hoist `MsalClient` to support enabling built-in MSAL telemetry

security/msal/MsalClient.ts
```diff
@@ -247,6 +267,82 @@ export class MsalClient extends BaseOAuthClient<MsalClientConfig, MsalTokenSpec>
...
+                        eResult.failureCount++;
+                        eResult.lastFailure = {
+                            time: eTime,
+                            code: errorCode,
+                            name: errorName,
+                            raw: e
+                        };
+                    }
+
+                    const {duration} = eResult;
```
lbwexler, line 318: 318 can be nested in if block. (super nit)

security/msal/MsalClient.ts
```diff
@@ -247,6 +267,82 @@ export class MsalClient extends BaseOAuthClient<MsalClientConfig, MsalTokenSpec>
...
+                    if (durationMs) {
```
lbwexler, line 319: is zero milliseconds ever a possibility for a no-op cache lookup?   safer would be isFinite()

security/msal/MsalClient.ts
```diff
@@ -247,6 +267,82 @@ export class MsalClient extends BaseOAuthClient<MsalClientConfig, MsalTokenSpec>
...
+                } catch (e) {
+                    this.logError(`Error processing telemetry event`, e);
+                }
+            });
+        });
+
+        this.logInfo('Telemetry enabled');
+    }
+
+    disableTelemetry(): void {
```
lbwexler, line 334: Do we envision ever doing this?  Don't feel strongly about it, am curious if its worth it

security/msal/MsalClient.ts
amcclain: I imagined it could be, but was also deliberately skipping those cases on the theory that it might be useful to track events-with-non-zero-duration only in this nested block

amcclain (approved):
I'm good to go on this when you are, @lbwexler

PR #3958, opened 2025-04-09
Additional enhancement/simplifications to Client Health

lbwexler:
+ ship out less health related display info back to opening message only.
+ simpler date formatting for MSAL health info + better timing for registration to avoid race condition
+ Added new ids for client and session.
+ Added PageState, idle, and web socket information to health report.
+ Remove concept of tracking on WARN level.  We are overpromising here -- no concrete plan for actually evaluating this client side state so better to hold off 

Hoist P/R Checklist
-------------------

### DIFF
--- a/appcontainer/AppStateModel.ts
+++ b/appcontainer/AppStateModel.ts
@@ -4,10 +4,10 @@
  *
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
-import {AppState, AppSuspendData, HoistModel, XH} from '@xh/hoist/core';
+import {AppState, AppSuspendData, HoistModel, PlainObject, XH} from '@xh/hoist/core';
 import {action, makeObservable, observable} from '@xh/hoist/mobx';
 import {Timer} from '@xh/hoist/utils/async';
-import {camelCase, isBoolean, isString, mapKeys} from 'lodash';
+import {camelCase, isBoolean, isString, mapKeys, pick} from 'lodash';
 
 /**
  * Support for Core Hoist Application state and loading.
@@ -94,9 +94,13 @@ export class AppStateModel extends HoistModel {
                     data: {
                         appVersion: XH.appVersion,
                         appBuild: XH.appBuild,
+                        clientId: XH.clientId,
+                        sessionId: XH.sessionId,
                         locationHref: window.location.href,
                         timings: mapKeys(timings, (v, k) => camelCase(k)),
-                        clientHealth: XH.clientHealthService.getReport()
+                        clientHealth: XH.clientHealthService.getReport(),
+                        window: this.getWindowData(),
+                        screen: this.getScreenData()
                     },
                     logData: ['appVersion', 'appBuild'],
                     omit: !XH.appSpec.trackAppLoad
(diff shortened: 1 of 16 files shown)


security/msal/MsalClient.ts
```diff
@@ -279,7 +280,9 @@ export class MsalClient extends BaseOAuthClient<MsalClientConfig, MsalTokenSpec>
                 try {
                     const {events} = this.telemetryResults,
                         {name, startTimeMs, durationMs, success, errorName, errorCode} = e,
-                        eTime = startTimeMs ? new Date(startTimeMs) : new Date();
+                        eTime = fmtDate(startTimeMs ?? Date.now(), {
+                            fmt: 'MMM D h:mm:ssa'
```
amcclain, line 284: Why not just use millis?  Don't think we would want to bake in pre-formatted dates (in browser's timezone)

svc/ClientHealthService.ts
```diff
@@ -144,31 +115,27 @@ export class ClientHealthService extends HoistService {
...
+                clientId: XH.clientId,
+                sessionId: XH.sessionId,
+                report: this.getReport()
+            }
         });
     }
 }
 
-export interface SessionData {
+export interface GeneralData {
```
amcclain, line 133: `SessionData`?

svc/ClientHealthService.ts
```diff
@@ -144,31 +115,27 @@ export class ClientHealthService extends HoistService {
...
-            intervalMins,
-            severity: defaultSeverity,
-            ...rest
-        } = XH.trackService.conf.clientHealthReport ?? {};
-
-        const rpt = this.getReport();
-        let severity = defaultSeverity ?? 'INFO';
-        if (find(rpt, (v: any) => isPlainObject(v) && v.severity === 'WARN')) {
-            severity = 'WARN';
-        }
```
amcclain, line 157: Decided we didn't need this?  OK with me, just checking

appcontainer/AppStateModel.ts
```diff
@@ -94,9 +94,13 @@ export class AppStateModel extends HoistModel {
                     data: {
                         appVersion: XH.appVersion,
                         appBuild: XH.appBuild,
+                        clientId: XH.clientId,
+                        sessionId: XH.sessionId,
                         locationHref: window.location.href,
```
amcclain, line 99: Minor thing, but looks like `locationHref` might be legacy, from before we capture this by default on the top-level track entry (outside of data)

amcclain (commented):
Looks like a great further evolution of this feature. MSAL datetimes only thing I didn't get and believe we should change.  Thanks!

svc/ClientHealthService.ts
lbwexler: right -- was ceding the term "session" to the "sessionStorage" scope meaning, especially since we now have a 'sessionId'

svc/ClientHealthService.ts
```diff
@@ -144,31 +115,27 @@ export class ClientHealthService extends HoistService {
...
         XH.track({
             category: 'App',
             message: 'Submitted health report',
-            severity,
             ...rest,
-            data: rpt
+            data: {
+                clientId: XH.clientId,
+                sessionId: XH.sessionId,
+                report: this.getReport()
```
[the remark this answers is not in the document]
lbwexler: sure -- was on the fence about it, will do

PR #3959, opened 2025-04-10
Option in About dialog to send client health report on-demand

amcclain:
- Fix to ensure mobile toasts are shown above fullscreen dialogs.
- Make `XH.clientHealthService.sendReport()` public, although comment as internal.
- Same with `XH.trackService.pushPendingAsync()`.
- Tweak to about table alignment on mobile.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/appcontainer/AboutDialog.ts
+++ b/desktop/appcontainer/AboutDialog.ts
@@ -45,11 +45,10 @@ export const aboutDialog = hoistCmp.factory({
                     button({
                         text: 'Send Client Health Report',
                         icon: Icon.health(),
-                        omit: !XH.trackService.enabled,
+                        omit: !XH.clientHealthService.enabled,
                         onClick: async () => {
                             try {
-                                XH.clientHealthService.sendReport({trackOpts: {severity: 'INFO'}});
-                                await XH.trackService.pushPendingAsync();
+                                await XH.clientHealthService.sendReportAsync();
                                 XH.successToast('Client health report successfully submitted.');
                             } catch (e) {
                                 XH.handleException('Error sending client health report', e);
--- a/desktop/appcontainer/AboutDialog.ts
+++ b/desktop/appcontainer/AboutDialog.ts
@@ -48,7 +48,7 @@ export const aboutDialog = hoistCmp.factory({
                         omit: !XH.trackService.enabled,
                         onClick: async () => {
                             try {
-                                XH.clientHealthService.sendReport();
+                                XH.clientHealthService.sendReport({trackOpts: {severity: 'INFO'}});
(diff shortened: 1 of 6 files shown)


desktop/appcontainer/AboutDialog.ts
```diff
@@ -42,10 +42,25 @@ export const aboutDialog = hoistCmp.factory({
...
+                        icon: Icon.health(),
+                        omit: !XH.trackService.enabled,
+                        onClick: async () => {
+                            try {
+                                XH.clientHealthService.sendReport();
+                                await XH.trackService.pushPendingAsync();
+                                XH.successToast('Client health report successfully submitted.');
+                            } catch (e) {
+                                XH.handleException('Error sending client health report', e);
+                            }
```
amcclain, line 56: Thought about bundling toast, push, error handling into options on `sendReport` but didn't really seem worth it. 

I did think it was worth the extra complexity to do both of those things, however, to ensure the report gets submitted right away, assuming some kind of interactive troubleshooting is likely

lbwexler: 👍🏻

svc/ClientHealthService.ts
```diff
@@ -64,6 +61,25 @@ export class ClientHealthService extends HoistService {
...
+     * Generate and submit a report to the server, via TrackService.
+     * @internal - apps should enable via config and allow this service to submit on timer.
+     */
+    sendReport() {
+        const {intervalMins, ...rest} = XH.trackService.conf.clientHealthReport ?? {};
+
+        XH.track({
+            category: 'App',
+            message: 'Submitted health report',
+            ...rest,
```
lbwexler, line 74: our strategy of having the `..rest` coming from config including "debug" not going to be compatible with this. :-(

amcclain: Good catch - worked around it with...yet more opts....

lbwexler: maybe their is a public async version of this method that sets the severity to 'info', and awaits the pending message push.  And a private version called by that and the timer....

amcclain: FWIW now that I've added opts, we could bring in the `sendImmediately` and `showConfirmToast` etc.  Could also support passing in custom data for that one-off call.  

But still didn't think any of those were quite compelling enough to add up-front, and trying to keep as minimal as possible.  

The entire feature is clearly a bonus - was triggered from desire to see report from mobile client without waiting 15 mins, and without having dev console to trigger manually

lbwexler:
Tweaked the structure to allow the ad-hoc sending to diverge in additional ways, and to simplify the API and DRY.

Great change!

PR #3963, opened 2025-04-10
Ensure `FilterFieldSpec.values` compacted to remove nulls

amcclain:
- Nulls within provided values appear to break `FilterChooser` - suggestions don't show, and general typeahead support is compromised.
- Compact provided values in ctor processing.
- Repeat in admin activity tracking, where we mutate these directly (in an unsupported way).

Hoist P/R Checklist
-------------------

### DIFF
--- a/admin/tabs/activity/tracking/ActivityTrackingModel.ts
+++ b/admin/tabs/activity/tracking/ActivityTrackingModel.ts
@@ -5,17 +5,17 @@
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
 import {exportFilename} from '@xh/hoist/admin/AdminUtils';
-import {GroupingChooserModel} from '@xh/hoist/cmp/grouping';
+import * as Col from '@xh/hoist/admin/columns';
 import {FilterChooserModel} from '@xh/hoist/cmp/filter';
 import {FormModel} from '@xh/hoist/cmp/form';
 import {GridModel, TreeStyle} from '@xh/hoist/cmp/grid';
+import {GroupingChooserModel} from '@xh/hoist/cmp/grouping';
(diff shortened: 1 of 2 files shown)


amcclain:
I have been noticing that auto-complete filtering in activity viewer was not working reliably for one or two of our instances, but wasn't sure what could be going on.  I'm pretty sure that I've tracked it back to filter field specs getting a `null` entry in a provided values array, which can happen with lookups loaded from server.

This is *not* happening currently on toolbox, but you can recreate by manually setting values incl null on an instance - you will see it breaks quite badly, although weirdly without throwing errors.  It's confusing.

I wanted to get deeper into the component / query engine to understand what's happening, but the waters get deep quickly, hence this more tactical fix.

I am working on refactor to activity viewer in another branch - will look to remove the unsupported direct setting of these values in that branch if i can

admin/tabs/activity/tracking/ActivityTrackingModel.ts
```diff
@@ -330,12 +329,21 @@ export class ActivityTrackingModel extends HoistModel {
...
+        try {
+            const lookups = await XH.fetchJson({url: 'trackLogAdmin/lookups'});
+            this.filterChooserModel.fieldSpecs.forEach(spec => {
+                const {field} = spec,
+                    lookup = lookups[field] ? compact(lookups[field]) : null;
+
+                if (!isEmpty(lookup)) {
+                    spec.values = lookup;
+                    spec.enableValues = true;
+                    spec.hasExplicitValues = true;
```
TomTirapani, line 341: Great that we wait until we've confirmed we have non-null values before setting these flags 👍

TomTirapani (approved):
Was able to recreate the issue locally, and this confirmed this branch fixed it 👍  Looks good and safer all round

PR #3967, opened 2025-04-14
Update Admin log viewer handling for long lines

amcclain:
- Remove previous heuristic around setting width of main "content" column within the log viewer based on char count. This was not accurately ensuring all content was shown, leading to unwanted truncation of longer lines.
- Replace with hard-coded 5000px width, disable autosize completely to continue avoiding overhead there, then also hard-code width and pin line number col.

Hoist P/R Checklist
-------------------


amcclain:
The current handling around long lines threw me off here - was seeing truncated lines and took a bit to trace down where that was happening.

Alternative approaches would be:
* Increase multiplier used by heuristic to a value that would more reliably avoid truncation
* Allow autosize to handle - it was fast with medium sized (several thousand) lines - we could test with many and it would likely be fine, but it's obvs not entirely without cost.

In the end, I decided that just picking some "very wide" value seemed like a reasonable compromise and kept things extremely simple....

PR #3969, opened 2025-04-15
Use baked-in client `XH.appVersion|Build` for upgrade check

amcclain:
* Previously the versions loaded from the server on init were used as the baseline.
* The two versions *should* be the same, but in cases where a browser "restores" a tab and re-inits an app without reloading the code itself, the upgrade check would miss the fact that the client remained on an older version.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## v73.0.0-SNAPSHOT - unreleased
 
+### ⚙️ Technical
+
+* Updated the background version checking performed by `EnvironmentService` to use the app version
+  and build information baked into the client build when comparing against the latest values from
+  the server. Previously the versions loaded from the server on init were used as the baseline.
+  * The two versions *should* be the same, but in cases where a browser "restores" a tab and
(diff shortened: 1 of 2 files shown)


amcclain:
Not sure why I waited so long to do this - it's been an issue for ages, ever since browsers started playing new tricks with tabs getting restored / synced across sessions. 

I use the Arc browser which perhaps does this more than vanilla Chrome.  It can result in a tab having a version of the app's index.html page and JS/CSS source loaded/cached and "re-run" or re-initialized without checking to see if there is a newer source available.  This can result in a stale version of the client starting up fresh, which will load the current _server_ version into `EnvironmentService`.

The only risk I can foresee here - as called out in the changelog - is that some project has a misconfigured build where the client/webpack build is given a different version or build tag than the server/gradle build. Previously I don't think that would have actually mattered, now it would result in a false positive and persistent upgrade prompt.

The solution in this case would be to fix the build. If we realized that too late and had already released a version, the upgrade prompt could be set to `silent` to stop users from getting an alert they could not resolve.

RyanMG (approved):

PR #3971, opened 2025-04-17
New Clients Tab in Admin

lbwexler:
Hoist P/R Checklist
-------------------


admin/tabs/client/ClientTab.ts
```diff
@@ -7,18 +7,17 @@
...
+export const clientTab = hoistCmp.factory(() =>
     tabContainer({
         modelConfig: {
-            route: 'default.activity',
-            switcher: {orientation: 'left', testId: 'activity-tab-switcher'},
+            route: 'default.clients',
+            switcher: {orientation: 'left', testId: 'client-tab-switcher'},
             tabs: [
-                {id: 'tracking', icon: Icon.analytics(), content: activityTrackingPanel},
+                {id: 'clients', icon: Icon.desktop(), content: clientsPanel},
```
amcclain, line 20: `Connections` or something like that?

amcclain: I went ahead and updated the routes / display titles - I prefer them, happy to rename the components to match if you like them too (or leave those alone, but might as well be consistent)

admin/tabs/monitor/MonitorTabModel.ts
```diff
@@ -73,7 +74,7 @@ export class MonitorTabModel extends HoistModel {
     }
 
     override async doLoadAsync(loadSpec: LoadSpec) {
-        if (!XH.pageIsVisible) return;
```
amcclain, line 76: FWIW we are showing a toast out of this model when the status changes - with the idea that you might not have this tab visible (i.e. are on new clients tab, or configs) but would still appreciate a heads up that a monitor just ticked over. Might want to keep prior viz check if we thought that worthwhile

amcclain:
Did not attempt multi-instance to see what you mean re. possible serialization issue, but in terms of the overall change and everything here, am all for it!

PR #3990, opened 2025-05-08
Improved client `WebSocketService` heartbeat

amcclain:
- Check that client is receiving inbound messages from the server, not just successfully sending outbound heartbeats.
- Attempt reconnect if not, although with a buffer on retries.
- Add dedicated telemetry counts for heartbeat sent/ack - these should always equal or be within 1 of each other, regardless of how many other messages might be flowing in or out. Seems like a useful thing to be able to check.

Hoist P/R Checklist
-------------------

### DIFF
--- a/svc/WebSocketService.ts
+++ b/svc/WebSocketService.ts
@@ -8,7 +8,7 @@ import {HoistService, PlainObject, XH} from '@xh/hoist/core';
 import {withFormattedTimestamps} from '@xh/hoist/format';
 import {action, makeObservable, observable} from '@xh/hoist/mobx';
 import {Timer} from '@xh/hoist/utils/async';
-import {SECONDS} from '@xh/hoist/utils/datetime';
+import {olderThan, SECONDS} from '@xh/hoist/utils/datetime';
 import {throwIf} from '@xh/hoist/utils/js';
 import {find, pull} from 'lodash';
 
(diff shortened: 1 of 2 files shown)


svc/WebSocketService.ts
```diff
@@ -39,6 +39,13 @@ export class WebSocketService extends HoistService {
     static instance: WebSocketService;
 
     readonly HEARTBEAT_TOPIC = 'xhHeartbeat';
+    /** Check connection and send a new heartbeat (which should be promptly ack'd) every 10s. */
+    readonly HEARTBEAT_INTERVAL_MS = 10 * SECONDS;
+    /**  If no heartbeat ack (or other msg) received for past 30s, force a reconnect attempt. */
+    readonly HEARTBEAT_ACK_TIMEOUT_MS = 30 * SECONDS;
+    /** But wait at least 30s after a heartbeat-driven reconnect attempt before trying again. */
+    readonly HEARTBEAT_MIN_RECONNECT_INTERVAL_MS = 30 * SECONDS;
```
amcclain, line 47: I added this out of an abundance of caution - we want to detect connectivity issues quickly with 10s heartbeat interval, and if there's an issue found we'll fire off a reconnect attempt.  But if 10s later we still have an issue, wait a bit longer before trying again - something is up...

Realizing this might be better as, say, 25s, if we wanted it to skip only two timer cycles then try again on the third.

svc/WebSocketService.ts
```diff
@@ -193,15 +200,44 @@ export class WebSocketService extends HoistService {
     private heartbeatOrReconnect() {
         this.updateConnectedStatus();
         if (this.connected) {
-            this.sendMessage({topic: this.HEARTBEAT_TOPIC, data: 'ping'});
+            // We have a channel key, so we successfully registered with the server and have not
+            // received a disconnect message. We should be receiving at least heartbeat messages,
+            // but have observed cases where "something" interrupted connectivity in a surprising
+            // way and no new inbound messages were arriving, even with the socket reporting open
+            // and accepting outbound messages to send. Detect that case here.
```
amcclain, line 207: My hope is that we might learn more and could remove or edit this 🤷 comment, but for now I wanted to document/acknowledge the weirdness in-place here in the code, at least to some degree....

svc/WebSocketService.ts
```diff
@@ -361,6 +400,8 @@ export interface WebSocketTelemetry {
         connError?: WebSocketEventTelemetry;
         msgReceived?: WebSocketEventTelemetry;
         msgSent?: WebSocketEventTelemetry;
+        heartbeatReceived?: WebSocketEventTelemetry;
+        heartbeatSent?: WebSocketEventTelemetry;
```
amcclain, line 404: As per commit msg comment - seemed useful and cheap to give these dedicated counters.  Heartbeats are included in the general msgSent/Received counters, but so are app-level messages of an unpredictable volume.  I want to see how big the diff between these two numbers is in practice for clients - they should be equal or (very briefly!) within 1 of each other.

lbwexler:
This change looks fine for me, although I worry its a bit less general as to connection problems, and avoiding hyperactive reconnect attempts.  We don't want to retry the connection often whether or not we have failed due to a lack of getting pong responses, or any other reason the WS has been terminated.  The "MIN_RECONNECT" interval really should apply generally, not just to the heartbeat related disconnect.

+ Why not just declare the WS dead when pong received is too out of date according to interval and disconnect()
+ Apply the MIN_RECONNECT interval generally to the attempt to reconnect.

amcclain:
Superseded by #3992

PR #3995, opened 2025-05-15
Add new check for makeObservable on HoistBase

lbwexler:
Hoist P/R Checklist
-------------------


ghsolomon (commented):
Looks good. I wonder if we should only do this when `XH.isDevelopmentMode`?

lbwexler:
On the fence but happy to do that!

ghsolomon:
> On the fence but happy to do that!

Yeah I could see wanting this check if troubleshooting a production issue, but also would be nice not to _always_ queue up one of these checks every time a model is instantiated.  Another idea could be to have a "debugMode" soft config?

ghsolomon:
`isDevelopmentMode` is nice though since we already have it, so if no objections, I think this looks good

PR #4003, opened 2025-05-22
`ViewManagerModel.settleTime` improvements and `preserveUnsavedChanges` flag

TomTirapani:
+ Improve `ViewManagerModel.settleTime` by updating `lastPushed` when changes are pushed rather than when view is loaded.
+ Add ViewManagerModel.preserveUnsavedChanges flag to opt-out of that behaviour.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 74.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
++ Add `ViewManagerModel.preserveUnsavedChanges` flag to opt-out of that behaviour.
+
+### 🐞 Bug Fixes
++ Improves `ViewManagerModel.settleTime` by comparing to when the view is pushed to components
+  rather than when view is loaded from the server.
+
 ## v73.0.1 - 2025-05-19
 
 ### 🐞 Bug Fixes
--- a/cmp/viewmanager/View.ts
+++ b/cmp/viewmanager/View.ts
@@ -27,12 +27,6 @@ export class View<T extends PlainObject = PlainObject> {
      */
     readonly value: Partial<T> = null;
 
-    /**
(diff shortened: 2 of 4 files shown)


amcclain (commented):
To ensure I am thinking correctly about this change - will it affect/fix the settleTime handling on the _first_ actual load/push of a view only?  

That is, it looks to me like later/subsequent view changes/pushes are already getting the intended behavior, where any state that flows back from the components right after a push is effectively discounted. 

This change gets that _same_ intended behavior working for the first view as well, since now `lastPushed` is updated when the targets actually bind as opposed to when the viewManager is first created and loads its first view internally.

Which is good - good that it was already working for later view changes, and good that this change is quite targeted.

amcclain:
@lbwexler and @ghsolomon would be great to get your eyes on this - looks like a key fix in our forever war against inaccurate dirty view detection. 😉

TomTirapani:
> Also - I like the new `preserveUnsavedChanges` flag and am in favor of adding it here, but does that change directly relate to the update to `settleTime` handling?
> 
> I.e. I don't think that we would need to turn that _off_ to get the dirty detection working accurately, do we? What guidance might we offer a developer who is wondering if/when they should use that flag to disable?

No, the new flag is orthogonal and not required for the main `settleTime` handling changes to work. We could move it to its own PR and merge separately if preferred.

ghsolomon:
Looked at this with @TomTirapani, and we agree that it's already been tested in a client-app, solves that app's immediate issue and should be released.

It introduces a new public method `ViewManagerModel.noteStatePushed()` which we plan to deprecate with a future implementation, but since this has been tested, we feel it is prudent to release with a comment noting the method internal.

ghsolomon:
Closing due to overlap with https://github.com/xh/hoist-react/pull/4008 (pending testing in client apps)

PR #4004, opened 2025-05-27
Add grid "Open To Depth" functionality

febbraiod:
See Toolbox release

Hoist P/R Checklist
-------------------


febbraiod:
Corresponding TB PR:

https://github.com/xh/toolbox/pull/764

cmp/grid/GridModel.ts
```diff
@@ -1533,6 +1555,24 @@ export class GridModel extends HoistModel {
...
+
+        const {agApi} = this;
+        if (this.expandToLevel === 0) {
+            agApi.collapseAll();
+        } else if (this.expandToLevel === this.maxDepth) {
+            agApi.expandAll();
+        } else {
+            agApi.forEachNode(node => {
+                node.setExpanded(node.level < this.expandToLevel);
+            });
```
amcclain, line 1569: I was wondering how this iteration over every node holds up with very large grids.  Tried our Toolbox Admin > Tests > Grid page.  Allowed it to load its default (if intense) 200k records, the swapped to tree mode.  All OK, but then tried expanding to level three and the tab locked up with 100% CPU for...well, I gave up waiting.

Note that this does not occur with pre-existing expand/collapse all - I can use the toggle in the tree column header and the app is able to execute that operation OK and remain responsive.

Checking first to see how you were looking at / thinking about performance here, happy to discuss what we could do....

haynesjm42: In similar code in a client app we are setting the `expanded` property on the row nodes directly and then calling `agApi.onGroupExpandedOrCollapsed()` after updating all the nodes, might be worth trying to see if that performs better

data/RecordAction.ts
```diff
@@ -81,6 +81,11 @@ export interface ActionFnData {
     [x: string]: any;
 }
 
+export interface DisplayFnData extends ActionFnData {
+    /** Default display config for the action */
+    defaultConfig?: PlainObject;
```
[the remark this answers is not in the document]
febbraiod: I'll take a look

cmp/grid/GridContextMenu.ts
```diff
@@ -33,7 +33,7 @@ export type GridContextMenuToken =
     | 'autosizeColumns'
     | 'copyCell'
     | 'colChooser'
-    | 'expandCollapseAll'
+    | 'expandCollapse'
```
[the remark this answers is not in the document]
febbraiod: ok

cmp/grid/GridModel.ts
```diff
@@ -167,6 +168,9 @@ export interface GridConfig {
     /** Column ID(s) by which to do full-width grouping. */
     groupBy?: Some<string>;
 
+    /** Group level to expand to on initial load. 0 = all collapsed, 1 = only top level expanded. */
+    expandToLevel?: number;
```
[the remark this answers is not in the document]
febbraiod: Yes I think so.

cmp/grid/GridModel.ts
febbraiod: Will give it a try

data/RecordAction.ts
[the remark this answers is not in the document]
febbraiod: I think this might be cruft. Don't think this defaultConfig is set or used anywhere.

cmp/grid/GridModel.ts
febbraiod: This appears to improve things but a lot.

cmp/grid/GridContextMenu.ts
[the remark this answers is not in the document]
febbraiod: Actually I don't think I intended to make this change. I had left support for this in the MenuSupport file. Lee and I both thought it would be good to support this for backward compatibility.

cmp/grid/GridModel.ts
lbwexler: yes, performance improved, but actually not doing anything

cmp/grid/impl/MenuSupport.ts
```diff
@@ -270,3 +259,53 @@ function replaceHoistToken(token: string, gridModel: GridModel): Some<RecordActi
...
+        text: 'Expand to ...',
+        displayFn: () => {
+            // Don't show for degenerate shallow grid models, or if we don't have labels
+            const {maxDepth, expandToLevel} = gridModel;
+            if (maxDepth <= 1) return {hidden: true};
+
+            const levelLabels = executeIfFunction(gridModel.levelLabels);
+            if (!levelLabels || levelLabels.length < maxDepth + 1) {
+                gridModel.logDebug(
+                    '"levelLabels" not provided or of insufficient length. No menu items shown.'
```
febbraiod, line 292: we're sure we don't want to give Level 1, Level 2, ect generic labels by default?

febbraiod: Spoke to Lee.

Issue #3813, opened 2024-10-15
ChartModel context menu should be customizable


febbraiod:
Also, there are some context menu-related configurations in the default chart config that seem to do nothing.

The context is that we have a flagship app where not all of the context menu items work. There seems to be an issue in Highcharts with complex area chart color configuration and its export functionality. This is acceptable in the app, as it seems the buttons have always been broken but no one has reported it. We do not believe they are utilized at all (in particular: the export SVG and export PNG items).

I attempted to remove these items via the Highcharts config, but it appears those settings are ignored. There is a context menu applied to the container, but it is not customizable.

amcclain:
Dupe of #2753

Issue #3388, opened 2023-06-07
Support custom button(s) in chart panel


febbraiod:
We had a case where we wanted to have a button drawn within the chart itself to offer a popup using Hoist's modal support (without have to include a toolbar or a panel header).

Was unable to render said button with our icons (even when converted to an svg using asHtml: true).

Not 100% sure whether or not this idea is supported by highcharts but I was unable to crack it in what seemed like a reasonable amount of time for the ask, but Lee and I didn't want us to lose the idea.

amcclain:
Also linking to #2753 in case that ticket gets done first and provides another viable option here.

Issue #2753, opened 2021-10-25
Allow  Chart context menu to be customized


cnrudd:
As per client request, Chart context menu should have an API that allows the menu items to be customized.

amcclain:
GridModel context menu API (with string tokens for default actions) should be the model.

Issue #2389, opened 2021-04-06
Charts should provide extended functionality via context menu


febbraiod:
We may have need for this in a client app, previously implemented in Ext.

The context menu provides the ability to:
 
1. Copy the chart to the clipboard in varying sizes.
2. Export the chart as a png file
3. Export the series data as an Excel spreadsheet
4. Print the chart

PR #4005, opened 2025-05-28
Chart: custom context menu

cnrudd:
This addresses the following tickets:

https://github.com/xh/hoist-react/issues/2389
https://github.com/xh/hoist-react/issues/3813
https://github.com/xh/hoist-react/issues/3388
https://github.com/xh/hoist-react/issues/2753

Yes, you get 4 for 1 by merging this PR.  :)

Corresponding Toolbox PR with new custom context menu on Line Chart: https://github.com/xh/toolbox/pull/765

The scope of this PR has expanded.  The PR now proposes these changes:

1. New ability to customize chart context menus with a preset list of chart menu tokens or custom menu items.
2. Custom chart menu items will receive the active chart series points and the original right click event as args.
3. The framework's standard context menu's menu item actionFns will also receive the original right click event as a 2nd arg

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes: minor breaking change, noted in changelog.
- [x] Reviewed and tested on Mobile: not required.  no context menu on mobile
- [x] Created Toolbox branch: https://github.com/xh/toolbox/pull/765


amcclain:
Thanks for this @cnrudd  - I will review this very shortly and reach out with any questions

desktop/cmp/contextmenu/ChartContextMenu.ts
```diff
@@ -0,0 +1,194 @@
...
+    chartModel: ChartModel = null;
+    contextMenuClickEvt;
+    point;
+
+    constructor({items, chartModel, contextMenuClickEvt, point}) {
+        this.chartModel = chartModel;
+        this.contextMenuClickEvt = contextMenuClickEvt;
+        this.point = point;
+        this.items = cloneDeep(items).map(it => this.buildMenuItemConfig(it));
+    }
```
amcclain, line 68: It took me a little while reading this PR to grok, but realizing that this is a funny class / object to have.  We create it and it captures this state in its ctor, but then all we do is read its `items` property and then never reference it again (with this all happening in the `ChartLocalModel`.

Give that, it seems to me that this would be better implemented as a pure function - that's really what we need to deliver - a function that takes an event from HC and returns the items in the form that HC is ready to consume.  

Having the object right now made me immediately think it was more stateful than it really is. It also gave us another thing to keep track of in our minds when we say "contextMenu" - we have this object, the spec on `ChartModel`, and the function to always produce a new one of these objects on `ChartLocalModel`.

amcclain (commented):
This was pre-existing code, but since we're looking at the context menu here, in `Chart.ts` render function we have

`return !XH.isMobileApp ? useContextMenu(coreContents, impl.contextMenu) : coreContents;`

Seems like we should check `impl.contextMenu` instead and return `coreContents` directly if null.  That way the "not supported on mobile" fact is expressed once, and we also aren't wrapping the component in the case where the user has disabled the context menu.  

------

Thanks for making this change - I think we have all the parts we need, just wondering in my review comments here about the "packaging" - specifically the roles of `ChartModel` vs. `ChartLocalModel` and the need for the new `ChartContextMenu` class vs a functional implementation.  Have a look and LMK when you think when you have a chance.

cnrudd:
@amcclain @lbwexler This is ready for a 2nd look.

I've implemented these changes:

1. reworked as @amcclain suggestion so that menu item building is no longer in a class (and a lot of the smaller suggestions)
2. added contextMenuEvent to standard context menu (Lee's suggestion)
3. chart actionFn: grouped args in 'params' arg (Lee's suggestion)
4. chart actionFn: support providing points (for when more than one series is in chart and tooltip is shared)
5. also fixed chart mouse right to left zoom out bug
6. added/improved toolbox demos: multi series chart, new context menu item on basicPanel demo

cmp/chart/impl/ChartContextMenuItems.ts
```diff
@@ -0,0 +1,201 @@
...
+             * there are multiple series and tooltip.shared = true, points.length > 1.
+             */
+            points: any[];
+        }
+    ) => void;
+}
+
+/**
+ * If a String, value can be '-' for a separator, or a token supported by HighCharts
+ * for its native menu items, or a Hoist specific token.
```
amcclain, line 51: Can it also be any arbitrary string to insert a header?  If not, think we should spec `'-'` vs `string` here

cmp/chart/impl/ChartContextMenuItems.ts
```diff
@@ -0,0 +1,201 @@
...
+    return item as MenuItem;
+}
+
+function parseToken(token: string, chartModel: ChartModel): MenuItem | string {
+    switch (token) {
+        case 'viewFullscreen':
+            return {
+                text: 'View in full screen',
+                icon: Icon.expand(),
+                hidden: !chartModel,
```
amcclain, line 117: I notice that all of these have this `hidden: !chartModel` - are we ever rendering this menu without a `chartModel`?  Should there be some early out before we reach this point?

cmp/chart/impl/ChartContextMenuItems.ts
```diff
@@ -0,0 +1,201 @@
...
+    | 'printChart'
+    | 'downloadJPEG'
+    | 'downloadPNG'
+    | 'downloadSVG'
+    | 'downloadCSV'
+    | 'downloadXLS'
+    | 'downloadPDF'
+    | 'copyToClipboard';
+
+export interface ChartMenuItem extends Omit<MenuItem, 'actionFn' | 'items'> {
```
amcclain, line 29: Worth making the naming consistent - we could drop `Context` from the other exported members here (there's no other kind of menu), or make this `ChartContextMenuItem`

cmp/chart/impl/ChartContextMenuItems.ts
```diff
@@ -0,0 +1,201 @@
...
+                hidden: !chartModel,
+                actionFn: () => chartModel.highchart.downloadXLS()
+            };
+        default:
+            return token;
+    }
+}
+
+function isMenuItem(item: ChartMenuItemLike): item is MenuItem {
+    return !isString(item) && !isValidElement(item);
```
amcclain, line 191: This is called in just one place here, and we've already checked if the value is a string and won't reach here if so - could inline the validElement check and drop this function which I think just adds a bit more noise to the file at the moment

cmp/chart/impl/zoomout.ts
```diff
@@ -30,7 +30,8 @@ export function installZoomoutGesture(Highcharts) {
 
         Highcharts.addEvent(this, 'selection', e => {
             if (pixelDiff < 0) {
-                this.zoom(); // call w/o arguments resets zoom
+                // `undefined`'s preserve defaults, last arg `false` disables animation
```
amcclain, line 33: What's this change doing? Doesn't seem related to the others, or is it?

core/types/Interfaces.ts
```diff
@@ -284,10 +284,10 @@ export interface MenuItem {
     className?: string;
 
     /** Executed when the user clicks the menu item. */
-    actionFn?: (e: MouseEvent | PointerEvent) => void;
+    actionFn?: (e: MouseEvent | PointerEvent, contextMenuEvent?: MouseEvent | PointerEvent) => void;
```
amcclain, line 288: So the first event `e` is the click on a specific item within the menu, and the new second event `contextMenuEvent` is the right-click that caused the menu to appear in the first place?

desktop/cmp/contextmenu/ContextMenu.ts
```diff
@@ -20,6 +20,12 @@ export type ContextMenuSpec = MenuItemLike[] | ((e: MouseEvent) => MenuItemLike[
 
 export interface ContextMenuProps extends HoistProps {
     menuItems: MenuItemLike[];
+    /**
+     * Optional event because this component is sometimes used as the content in popovers, to
+     * create menu buttons.  In those cases, it is not expected that the click event contains any
```
amcclain, line 25: > because this component is sometimes used as the content in popovers, to create menu buttons

Is this not *always* how it's used?  What's an example of the "contextual information" here, or when would you expect to find something in the first event `e`?  Just as a developer I'm left a little uncertain as to what I would do with one event or the other.

cmp/chart/impl/ChartContextMenuItems.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Specification for a ChartContextMenu.  Either a list of items or a function to produce one.
+ */
+export type ChartContextMenuSpec =
+    | boolean
+    | ChartContextMenuItemLike[]
+    | ((chartModel: ChartModel) => ChartContextMenuItemLike[]);
+
+export function getChartContextMenuItems(
```
amcclain, line 63: Should we mark as `@internal`?

cmp/chart/ChartModel.ts
```diff
@@ -95,4 +111,13 @@ export class ChartModel extends HoistModel {
...
     }
+
+    private parseContextMenu(
+        spec: ChartContextMenuSpec
+    ): ChartContextMenuItemLike[] | ((chartModel: ChartModel) => ChartContextMenuItemLike[]) {
+        if (spec === false || !XH.isDesktop) return null;
+        if (isNil(spec) || spec === true) return ChartModel.defaultContextMenu;
+
+        return spec;
+    }
```
amcclain, line 122: This looks great / very clear

cmp/chart/Chart.ts
```diff
@@ -106,7 +108,7 @@ class ChartLocalModel extends HoistModel {
     model: ChartModel;
 
     chartRef = createObservableRef<HTMLElement>();
-    contextMenu: any;
+    contextMenu: (e) => MenuItemLike[];
```
amcclain, line 111: Sorry we can leave this as is, but still feels like there's some way to streamline a bit more.

Currently we have
* `ChartLocalModel.contextMenu` - function created/captured as soon as linked
* `ChartLocalModel.getContextMenu()` - function to create the above function
* `ChartModel.contextMenu` - partially processed/parsed spec, might be a function that this local model calls, passing back reference to `ChartModel` itself
* `getChartContextMenuItems()` exported util

Still wondering what we're getting from having the context menu stuff in this local model - I am probably missing some subtlety as to why we need this layer involved, but whatever that subtlety is, if we could make more clear might help whenever we revisit this code in the future.

Or maybe we could drop contextMenu from this local model altogether and have the primary `ChartModel` process the specs/tokens it's given "all the way" into the function that `ChartLocalModel.getContextMenu()` currently produces.  

So we'd have context menu stuff in two places - the main `ChartModel` and the helper util - and `ChartModel` in its ctor would process the specs it's passed all the way through to the form that the rendering within `Chart.ts` consumes and passes on to the `useContextMenu` hook.

Don't mean to be a pain with all of this, but am having a hard time keeping track of the current flow and if it could be simpler, great - if not, no worries :)

cmp/chart/Chart.ts
```diff
@@ -95,7 +94,8 @@ export const [Chart, chart] = hoistCmp.withFactory<ChartProps>({
             })
         });
 
-        return !XH.isMobileApp ? useContextMenu(coreContents, impl.contextMenu) : coreContents;
+        // Must check isMobileApp here because `useContextMenu` is not defined in mobile
+        return !XH.isMobileApp ? useContextMenu(coreContents, model.contextMenu) : coreContents;
```
lbwexler, line 98: 👍🏻

lbwexler (commented):
Thanks for taking this one step further Colin.  Just a couple of small tweaks and hope we can merge this AM

CHANGELOG.md
```diff
@@ -10,12 +10,16 @@
 * Added `ViewManagerModel.preserveUnsavedChanges` flag to opt-out of that behaviour.
 * Added `PersistOptions.settleTime` to configure time to wait for state to settle before persisting.
 * Support for gridcolumn level `onCellClicked` events.
+* Chart context menu is now fully customizable.
```
lbwexler, line 13: We could mention in the context of "general improvements to menu api"  and also mention new MenuContext object now sent generally to both handlers as second argument.

Issue #3928, opened 2025-02-25
Add computed `StoreRecord.modifiedFields` getter


amcclain:
It would be useful to have an easy way to report on which fields in a `StoreRecord` have been modified, by name.

jskupsik:
```ts
    get modifiedValues(): PlainObject {
        const ret = {id: this.id};
        this.fields.forEach(({name}) => {
            if (this.data[name] !== this.committedData[name]) {
                ret[name] = this.data[name];
            }
        });
        return ret;
    }

Issue #3927, opened 2025-02-25
`StoreRecord.isDirty` remains true even if all values edited back to unmodified state


amcclain:
See https://github.com/xh/hoist-react/blob/3835abde3ff7c59b07fa5dc4fd4613f970e94c3e/data/StoreRecord.ts#L75 where we are using strict equality to compare committed vs. current data objects.

If a user changes a value on a record, then changes it back again to exactly what it was before, I would expect the record to report clean, as `FormModel` will do.

You can try this out on https://toolbox-dev.xh.io/app/grids/inlineEditing - change the Name of a record, note that Commit All button lights up in toolbar.  Change it back to what it was - button still enabled as rec still reporting dirty.

PR #4011, opened 2025-05-31
StoreRecord includes a getter to return a list of names for all modified fields

RyanMG:
https://github.com/xh/hoist-react/issues/3928

### DIFF
--- a/data/StoreRecord.ts
+++ b/data/StoreRecord.ts
@@ -6,7 +6,7 @@
  */
 import {PlainObject} from '@xh/hoist/core';
 import {throwIf} from '@xh/hoist/utils/js';
-import {isNil, flatMap, isMatch} from 'lodash';
+import {isNil, flatMap, isMatch, isEmpty} from 'lodash';
 import {Store} from './Store';
 import {ValidationState} from './validation/ValidationState';
 import {RecordValidator} from './impl/RecordValidator';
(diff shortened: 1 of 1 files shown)


amcclain:
I should add - re the fix for #3927  - I expect we'll need to look at where the store is applying these modifications, and have it do a new comparison and potentially "reset" `data = committedData` if it detects they are again the same.  Hopefully we have a straightforward (I won't say "easy") path to doing that.  If we get held up, we can of course revisit.

RyanMG:
@amcclain Quick update to review. I see this working in the toolbox example, but it may be simpler than what is actually needed here.

data/Store.ts
```diff
@@ -705,7 +705,7 @@ export class Store extends HoistBase {
     /** True if the store has changes which need to be committed. */
     @computed
     get isDirty(): boolean {
-        return this._current !== this._committed;
+        return !equal(this._current, this._committed);
     }
```
lbwexler, line 709: So there is a whole "normalization" step (see RecordSet.normalize)  that is supposed to do this evaluation after every update to `_current` and `_committed` and keep them pointer equivalent.    the fact that the getter is computed, might mean that this is an efficient way to have the same outcome, although I will say that one benefit of the normalization step is that it just frees up memory after an edit gets committed.

data/StoreRecord.ts
```diff
@@ -147,6 +148,24 @@ export class StoreRecord {
         return this.validator?.validationState ?? 'Unknown';
     }
 
+    get modifiedData(): PlainObject {
+        if (!this.isModified) return {};
```
lbwexler: Returning null would be more efficient, if we are potentially collecting this for giant recordsets.  Most records are clean.

data/StoreRecord.ts
```diff
@@ -147,6 +148,24 @@ export class StoreRecord {
...
+
+        const ret = this.fields.reduce((acc, {name}) => {
+            if (isNil(this.committedData) || this.data[name] !== this.committedData[name]) {
+                acc[name] = this.data[name];
+            }
+            return acc;
+        }, {});
+
+        return isEmpty(ret)
```
lbwexler: normalize empty here to null.  Still on the fence about including the id, but I guess that reduces the burden on the calling code.

PR #4012, opened 2025-06-02
Allow custom icon for grouping chooser favorites

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grouping/GroupingChooser.ts
+++ b/desktop/cmp/grouping/GroupingChooser.ts
@@ -62,6 +62,7 @@ export const [GroupingChooser, groupingChooser] = hoistCmp.withFactory<GroupingC
             popoverPosition = 'bottom',
             styleButtonAsInput = true,
             testId,
+            icon,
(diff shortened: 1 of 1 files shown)


lbwexler (commented):
Don and I reviewed -- we'll add an explicit `favoritesIcon` property, so as to allow the app to add an icon to the main button.

amcclain:
And - realize this is not in scope for this change, and shouldn't be - but interesting to know if we had an ability to re-wire our icons we would use that instead - so if we could in one place in the app or shared client code repoint `Icon.bookmark()` at another FA icon, would that be better?

amcclain:
> selectDropdown (chevronDown) was requested

Ah, that's....not something I would have guessed.  It doesn't really seem quite right, in that the optional favorites add-on functionality of groupingChooser really isn't the same as "the options available in this control".  I guess it's fine...just strikes me as one of the more confusing choices that could have been made when it comes to a replacement icon

PR #4018, opened 2025-06-16
Added props to `ViewManager` to customize icons used for different view types

TomTirapani:
+ Global views use `globe` icon by default.
+ Shared views user `users` icon by default.

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/viewmanager/ViewManager.ts
+++ b/desktop/cmp/viewmanager/ViewManager.ts
@@ -22,36 +22,36 @@ import {saveAsDialog} from './dialog/SaveAsDialog';
 
 import './ViewManager.scss';
 
-/**
- * Visibility options for save/revert button.
- *
- * 'never' to hide button.
- * 'whenDirty' to only show when persistence state is dirty and button is therefore enabled.
- * 'always' will always show button.
- */
-export type ViewManagerStateButtonMode = 'whenDirty' | 'always' | 'never';
-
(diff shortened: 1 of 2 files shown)


desktop/cmp/viewmanager/ViewManager.ts
```diff
@@ -115,6 +139,16 @@ const menuButton = hoistCmp.factory<ViewManagerLocalModel>({
     }
 });
 
+const buttonIcon = hoistCmp.factory<ViewManagerLocalModel>({
+    render({model, ownedViewIcon, sharedViewIcon, globalViewIcon, defaultViewIcon}) {
+        const {view} = model.parent;
+        if (view.isOwned) return ownedViewIcon;
+        if (view.isShared) return sharedViewIcon;
```
amcclain, line 146: When a user is looking at one of their _own_ views which they have also shared with other users, best to show them the "owned" icon or the "shared" one?

Right now we would show the owned icon - both `isOwned` and `isShared` would be true, but I'm wondering if it would be more useful to clue the user into the fact that they've shared this view and others might notice changes.  

Downside is that "what I can do with this view" would not then directly track with the icon displayed. For that reason I think it might well be best to leave as is - show the owned icon - but not sure. Don't think there's a clear right answer but wanted to ask the question.

amcclain (approved):
Looks great.  I tweaked/fixed a few comments and one question for the group below, but all for merging.

Issue #2400, opened 2021-04-14
Desktop Menu component(s) - replaces direct use of Blueprint


amcclain:
This is a weird gap that came up in our discussion on PR #1800.

We provide Hoist-level support for menus in a few places:
- Desktop `contextmenu` package  - https://github.com/xh/hoist-react/tree/develop/desktop/cmp/contextmenu
- Mobile `menu` package - https://github.com/xh/hoist-react/tree/develop/mobile/cmp/menu
- Desktop/Mobile `AppMenuButton` comp

But we don't provide a first-class way to setup e.g. a button on a toolbar that shows a menu - you need to use BP kit components. 

We should review and reconcile the existing approaches and come up with a first iteration (however minimal) of a Hoist menu package for desktop.

lbwexler:
Raising the priority of this.   The ContextMenu API was actually substantially changed in Blueprint 4.0, and this is actually the main blocker to upgrading to Blueprint 4.0.

The new api does not provide an obvious way to respond to the actual mouse event, and dynamically construct the menu at that time.  Rather than invest more time and contortions into targeting the new blueprint context menu api we should just promote some verison of the mobile menu as a desktop (or cross-platform) native Hoist menu component.

amcclain:
https://github.com/szhsin/react-menu

PR #4019, opened 2025-06-24
Added `extraMenuItems` to `ViewManager`

TomTirapani:
+ Moved `parseMenuItems` to reusable util

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/impl/MenuItems.ts
+++ b/utils/impl/MenuItems.ts
@@ -1,0 +1,51 @@
+import {isMenuItem, MenuItemLike} from '@xh/hoist/core';
+import {menuDivider, menuItem} from '@xh/hoist/kit/blueprint';
+import {MenuItemProps} from '@blueprintjs/core';
+import {wait} from '@xh/hoist/promise';
+import {isOmitted} from '@xh/hoist/utils/impl/IsOmitted';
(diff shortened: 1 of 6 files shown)


utils/impl/MenuItems.ts
cnrudd, line 1: Nice to get this into a utils folder.

amcclain: Agreed!

I don't think it benefits us to hold this change back, but I was left wondering if/how we should be using this same util in any other places across our rather scattershot approach to menus.  

E.g. very similar code appears in desktop `ContextMenu.ts` non-exported function `parseItems` - would it be worth it to extend the new util so it could satisfy that use case also?  It's just a bit surprising that we have this pretty general sounding util for handling menu item parsing, but then it's called in only two specific use cases (app button menu and now viewmanager).

Although the real weirdness is that we don't have a desktop `menu` component with any custom processing like this - we only have the re-exported BP comp.  That's covered by this ancient ticket - https://github.com/xh/hoist-react/issues/2400 - and ultimately if we wanted to holistically review menu support we'd need to tackle that, at which point the new util could come into its own or be baked into a new component.

PR #4024, opened 2025-07-02
Further refinements to the `GroupingChooser` desktop UI

amcclain:
* Added new props `favoritesSide` and `favoritesTitle`.
* Deprecated `popoverTitle` prop - use `editorTitle` instead.
* Moved "Save as Favorite" button to a new compact toolbar within the popover.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
-## v75.0.0-SNAPSHOT - unreleased
+## v74.1.1 - 2025-07-02
+
+### 🎁 New Features
+
+* Further refinements to the `GroupingChooser` desktop UI.
+  * Added new props `favoritesSide` and `favoritesTitle`.
+  * Deprecated `popoverTitle` prop - use `editorTitle` instead.
+  * Moved "Save as Favorite" button to a new compact toolbar within the popover.
 
 ## v74.1.0 - 2025-06-30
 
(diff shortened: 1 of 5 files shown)


amcclain:
After discussion with senior folks at our client today and some more reflection, decided to go ahead and add a fuller set of configuration options for desktop `GroupingChooser`.

New options would allow for a configuration like the below, which is what I would recommend for the client based on where we landed with that discussion:

```
groupingChooser({
  favoritesSide: 'top',
  favoritesTitle: null,
  editorTitle: 'Custom Grouping'
})
```

CleanShot 2025-07-01 at 19 37 14@2x

But it would also allow for favorites on the left as originally requested:

CleanShot 2025-07-01 at 19 35 51@2x

Although I am not inclined to recommend given the sub-optimal empty state:

CleanShot 2025-07-01 at 20 07 40@2x

Note that in top/bottom orientation, there is special handling to drop out the favorites section completely (you can see that the button to add your first one remains):

CleanShot 2025-07-01 at 20 09 31@2x

amcclain:
The only caveat I know of is that there's not a great way to manage the overall max size of the thing.  An app could add some specific constraints in CSS - e.g. based on `vh` - might be recommended in a top/bottom orientation if users might have many faves.  I did not add these by default as we don't know where on the page the chooser might live.

In general I think we have to hope that users keep this list somewhat reasonable, or we deal with it when we have a concrete case....  CSS should give apps what they need though, didn't want to add too many more props.

PR #4031, opened 2025-07-10
Don't cache in useContextModel

lbwexler:
This change undoes the assumption we built in the toolkit that useContextModel is static and is not reactive to the model reference changing.

Definitely has performance implications (full lookup re-done on every render) as well as potential behavior implications -- do we have code that is somehow taking advantage of the fact that these models are not being swapped out from underneath the application.  But Our lookup of the "primary" model is not currently cached (as far as I can tell reading `HoistCmp`) so perhaps the performance is not a problem?  Note that useContextModel is used in the following:

+ Misc Grid helpers and associated buttons (the use case in the associated ticket)
+ RefreshContextView (used by Tab, Page, and AppContainer)
+ Panel

The broad performance and runtime implications of this change for the latter two in the list worry me the most with this change.

**UPDATE -- I think this change is worth it and ready for review.  If we can afford to be dynamic w.r.t to the main context model, we can afford to for auxillary gridModels**

**NOTE: -- I also cleaned up some tangentially related strangeness in the way we manage state for GridFindField.** 

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 Previously this value was cached on first render.
 
 * Fixes to framework components that bind to grids (e.g. `ColChooserButton`, `ColAutosizeButton`,
-  `GridFindField`) to rebind to a change to the observable GridModel available via context.
+  `GridFindField`) to rebind to a new observable GridModel available via context.
 
 ## v74.1.2 - 2025-07-03
 


lbwexler:
CR: Colin.  Lee to double check case where GridModel is initially null, only set fairly late.

PR #4044, opened 2025-07-23
Normalize `parseFilter` return to `null` w/false/empty value

amcclain:
- Previous `parseFilter` handling would allow an `undefined` input to pass through, which would then throw within `FilterChooserModel.setValue()` which does a check on `f === null` within its `validateFilter()`.
- The parser now normalizes falsy inputs and also empty arrays to `null`.

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/filter/Utils.ts
+++ b/data/filter/Utils.ts
@@ -5,24 +5,21 @@
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
 
-import {castArray, flatMap, groupBy, isArray, isFunction} from 'lodash';
-
-import {FieldFilterSpec, FilterLike} from './Types';
-import {Filter} from './Filter';
-import {FieldFilter} from './FieldFilter';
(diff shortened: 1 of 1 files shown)


data/filter/Utils.ts
```diff
@@ -32,7 +29,8 @@ export function parseFilter(spec: FilterLike): Filter {
     let s = spec as any;
 
     // Degenerate cases
-    if (!s || s instanceof Filter) return s as Filter;
+    if (s instanceof Filter) return s as Filter;
+    if (!s || (isArray(s) && s.length === 0)) return null;
```
amcclain, line 33: We could use `isNil(s)` here instead of the falsey check, but given that any other falsey value would be invalid I wasn't sure that would be desirable.  

Also could unwind the empty array handling, but I didn't really see `[] -> empty CompoundFilter` as a feature - seemed more confusing than anything.

PR #4052, opened 2025-07-26
Add new `ClipboardButton.errorMessage` prop

### LINKED ISSUE #4049: `ClipboardButton` should have configurable alerting if copy operation fails
Takes successMessage prop, but there is no built in handler for a failure / error state.

Should accept an errorMessage prop

amcclain:
- Fixes #4049
- Enable toast alert via `handleException` by default, if not suppressed via the new prop. This is a change in behavior but seems like a better default.

Hoist P/R Checklist
-------------------


amcclain:
We are using this button in a client app and taking advantage of async support to call an endpoint that generates some text to copy. It should be reliable, but could fail - seems prudent all around to add some user-visible error handling.

PR #4056, opened 2025-07-31
Add `IColChooserModel` cross-platform interface

amcclain:
Introduces a cross-platform `IColChooserModel` interface so `GridModel.colChooserModel` is properly typed across desktop and mobile, eliminating the `as any` casts previously needed to call `open()`.

This mirrors the established `IDynamicTabSwitcherModel` pattern: a shared interface in `cmp/grid/Types.ts`, implemented by both the desktop and mobile `ColChooserModel` classes, used to type the `GridModel.colChooserModel` field and the `parseChooserModel()` return.

**Changes**
- Added `IColChooserModel` (extends `HoistModel`) capturing the cross-platform control surface: `gridModel`, `isOpen`, `open()`, `close()`.
- Typed `GridModel.colChooserModel` and `parseChooserModel()` as `IColChooserModel`, dropping the `(... as any)?.open()` casts in `GridModel.showColChooser()` and `MenuSupport`.
- Both `ColChooserModel` classes now `implements IColChooserModel`.
- Added a `### ⚙️ Typescript API Adjustments` CHANGELOG entry.

Internal typing improvement only - no runtime behavior change. `extends HoistModel` keeps the public field's base-model surface intact, so it is non-breaking for consumers.

Hoist P/R Checklist
-------------------


amcclain:
@ghsolomon updated - WDYT?  Improvement?

ghsolomon:
> @ghsolomon updated - WDYT? Improvement?

I like this better.  Should `IColChooserModel` extend `HoistModel` in case someone wanted to access members of `HoistModel`?

As far as naming conventions, have we used the "I" prefix before? I understand it here to avoid colliding with the concrete implementations, but from what I gather the general Typescript community tends to discourage it as opposed to in Java, C#, etc.  Here's Microsoft's internal Typescript coding guidelines, which they emphasize only apply to TS contributors, but still could be worth noting: https://github.com/microsoft/TypeScript/wiki/Coding-guidelines.  See (2) under Names (Do not use I as a prefix for interface names.)

That said, I'm not sure what else we'd call this? `BaseColChooserModel`? `CrossPlatformColChooserModel`? Maybe in this case best to stick with the prefix, but wanted to throw this out.

amcclain:
> That said, I'm not sure what else we'd call this? `BaseColChooserModel`? `CrossPlatformColChooserModel`? Maybe in this case best to stick with the prefix, but wanted to throw this out.

Yes, I also am not a fan of the `I` prefix, but this was exactly my dilemma. I figured if we were ever going to use it, perhaps these cross platform interfaces would be the place. 

In hindsight, perhaps we should have gone with `[Desktop|Mobile]ColChooserModel` for the implementations...  That would also have reduced likelihood of grabbing wrong import / IDE confusion. Then the interface simply `ColChooserModel`.  Too late to consider that change?

amcclain:
@lbwexler I am happy to let this miss the boat on today's 86.2.0 release, but I think we should merge immediately after. The biggest point of contention - the interface name - was in a real way settled when we merged in `IDynamicTabSwitcherModel` with the dynamic tabs feature.

Note we'll need to move the CHANGELOG entry to the next release prior to merge, otherwise this has been caught up and reviewed.

amcclain:
Thanks!

PR #4058, opened 2025-08-01
Add new `Cube.modifyRecordsAsync`

ghsolomon:
Hoist P/R Checklist
-------------------

### DIFF
--- a/data/Store.ts
+++ b/data/Store.ts
@@ -624,9 +624,8 @@ export class Store extends HoistBase {
         if (!isEmpty(updateRecs)) {
             this._current = this._current.withTransaction({update: updateRecs});
             changeLog.update = updateRecs;
+            this.rebuildFiltered();
         }
-
-        this.rebuildFiltered();
 
         return changeLog;
     }


ghsolomon:
TODO:

- [ ] Do we need special handling if summaryRecord(s) were modified? `Store.updateData` has special handling, but `Store.modifyRecords` does not.
- [ ] When modifying a `Record` such that it no longer passes a `View.query.filter`, it seems the `Record` remains visible. Perhaps the filter is not being re-evaluated in this case? Need to investigate whether this is an existing issue.

ghsolomon:
> Seems very promising -- not a lot of code. Not sure what the right thing to do in terms of naming the new types is, but do think Changelog is a bit too susceptible for confusion for my tastes.

Makes sense. I had the same thought but justified it to myself since it's scoped to the `data` package.  I'll go ahead and add the `Store` prefixes.

data/Store.ts
```diff
@@ -577,20 +594,41 @@ export class Store extends HoistBase {
...
+            changeLog.summaryRecords = this.summaryRecords;
+        }
+
+        // 3) Apply changes
+        if (!isEmpty(updateRecs)) {
+            this._current = this._current.withTransaction({update: updateRecs});
+            changeLog.update = updateRecs;
+        }
 
         this.rebuildFiltered();
```
lbwexler, line 629: nit:  does rebuildFiltered need to be done if there are no changes to the updateRecs?

PR #4065, opened 2025-08-11
Commit grid column filter changes on `enter` + cancel on `esc`

ghsolomon:
… "esc" keydown

Hoist P/R Checklist
-------------------

### DIFF
--- a/desktop/cmp/grid/impl/filter/headerfilter/HeaderFilter.ts
+++ b/desktop/cmp/grid/impl/filter/headerfilter/HeaderFilter.ts
@@ -41,16 +41,9 @@ export const headerFilter = hoistCmp.factory({
                     group: 'Column Filter',
                     onKeyDown: () =>
                         // Wait for debounced reaction in `ValuesTabModel` to run before committing
-                        wait(400).then(
-                            () => (model.hasFilter || model.hasPendingFilter) && model.commit()
-                        )
-                },
-                {
-                    allowInInput: true,
-                    combo: 'escape',
-                    label: 'Cancel',
-                    group: 'Column Filter',
-                    onKeyDown: () => model.parent.close()
+                        wait(400).then(() => {
+                            if (model.hasFilter || model.hasPendingFilter) model.commit();
+                        })
                 }
             ]
         });
@@ -67,7 +60,7 @@ const bbar = hoistCmp.factory<HeaderFilterModel>({
                     icon: Icon.delete(),
                     text: 'Clear',
                     intent: 'danger',
-                    disabled: !model.hasFilter,
+                    disabled: !hasFilter,
                     onClick: () => model.clear()
                 }),
                 filler(),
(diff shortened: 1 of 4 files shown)


amcclain:
Thanks for this, Greg - just scanned through your own comments, seems like this is a pretty in-depth change that will require a careful look.

amcclain (commented):
Thank you for taking this on!

Some comments (and one clear bug report) below. 

Zooming out from the implementation specifics, I'm wavering about the "apply filter when clicking outside popup" behavior. My main concern is that it doesn't give you any obvious way to cancel out of a filtering change if you've opened the dialog and typed something into the quick filter.  Is that problematic / surprising / user-unfriendly?

Ideally we would be mind-readers and know if the user is clicking outside to say "yes that's it, let's go!" or clicking outside to say "nevermind get me outta here, don't change anything".... 🤔 

Perhaps if we a) added the enter key handling and b) lit up the apply button when there was a real change to apply we could continue to treat "click outside" as a cancel.  We could also put it in the hands of the developer, but as always I'm wary of allowing a standardized built-in control like this to behave differently across apps, or grids in the same app....

edit - another option would be to make the "click outside -> apply" change but also add a "Cancel" option to the bottom toolbar.  Could remove the word "Filter" from the clear/apply buttons to avoid overcrowding....

ghsolomon:
Thanks for the feedback here. I played around a bit more in Excel, where filters are NOT committed on close but are committed on enter keydown.  Excel emphasizes the "OK" button once the filter is changed, even if the user then un-dirties the filter, which is interesting. Would be inclined to do something similar (but smarter?). Enable OK/Apply when there is a filter or pending filter, but emphasize it (thicker border? fill variation?) only when the filter is dirty.

If we did the above, I agree we could reasonably cancel on-close and require the user to hit "Enter" or click apply to commit their changes. Will update the PR


https://github.com/user-attachments/assets/a9750d24-530e-4e6a-9149-bb41d984ed47

desktop/cmp/grid/impl/filter/headerfilter/HeaderFilter.ts
```diff
@@ -31,32 +32,59 @@ export const headerFilter = hoistCmp.factory({
...
+                    intent: isDirty ? 'primary' : null,
+                    minimal: !isDirty,
+                    outlined: !isDirty,
                     onClick: () => model.commit()
+                }),
+                button({
+                    omit: commitOnChange,
+                    text: 'Cancel',
+                    outlined: true,
+                    onClick: () => model.parent.close()
```
amcclain, line 87: We put the OK/save/apply button to the right of cancel quite consistently across Hoist - would like to keep that ordering here.

amcclain: Would also drop outline on cancel - looks a bit weird in compact toolbar, clear filter not outlined

desktop/cmp/grid/impl/filter/headerfilter/HeaderFilterModel.ts
```diff
@@ -29,43 +39,52 @@ export class HeaderFilterModel extends HoistModel {
...
     }
 
-    get hasPendingFilter() {
+    get hasPendingFilter(): boolean {
+        return !!this.pendingFilter;
+    }
+
+    @computed
+    get isDirty(): boolean {
+        return !isEqual(this.columnFilters, flattenFilter(parseFilter(this.pendingFilter)));
```
amcclain, line 80: I'm actually a bit surprised that `isEqual` works with `Filter` instances, but it seems to.  We do have `Filter.equals()` - should / could we be using that?

amcclain (approved):

PR #4067, opened 2025-08-13
`GridCountLabel` supports a new `excludeParents` prop

cnrudd:
I found I need this feature on a tree grid where counting the grouping rows confuses things.

If we wanted to go further, seems like the store could also support some more getters, and then the logic inside `GridCountLabel` would be simplified.

```
get children(): StoreRecord[] {
     return this._filtered.filter(it => !it.children.length);
}

get allChildren(): StoreRecord[] {
     return this._current.filter(it => !it.children.length);
}

get childrenCount(): number {
      return this.children.length;
}

get allChildrenCount(): number {
      return this.allChildren.length;
}
```

Also, if this feature has legs, I would apply the same changes to `StoreCountLabel`.

Hoist P/R Checklist
-------------------


cnrudd:
> This proposal defines a "parent" as "a record with non-zero amount of children". OK for parent rows where all children have been filtered out to "become" children? Maybe that's desired, or just doesn't matter?

In my experience testing this solution, when all children are filtered out, the parent rows are filtered out also.  Store.records (from which the count is derived) is the list of filtered records, and I don't see parent records in that array that don't have children.  Maybe there is a difference here between how the list shakes out for tree grids vs grouped grids.  I've only tested on tree grids.

> Copilot flags fact that we are allocating an array and filtering across all records to get this count, vs other cases which use pre-computed counts on RecordSet. Is that a practical concern?

I am testing on a grid with 16k rows and I don't see any issues.   Maybe on very large grids this would be a reason not to pre-compute `children`, `allChidren` in `store`, as suggested above, and just keep this logic on this component.

> It's only 8:30am and includeChildren + excludeParents making my head hurt a bit :) New param true ignored if old param false, should we just return nothing instead?

Certainly could do that.  Could also warn in console to alert developer that that configuration doesn't make sense.

> What about StoreCountLabel?

I mentioned in description at top that whatever was done to `GridCountLabel` could be applied to `StoreCountLabel`.

cnrudd:
I just tested on our toolbox companies grid, which is not a tree grid, but grid with grouping.   The problem of excess row counts (or counting groups and children when I only want children counted) doesn't happen there because the grouping is done at the agGrid level, not at the store level, so there are no parent/child relationships in the store.  Every record is a root record.   

So, just to help the discussion and limit the scope of problem, it only applies to when `gridModel.treeMode = true`.   
This is something we could to the docs for includeChildren right now, only applies to tree grids, not grids with groupBy.

lbwexler:
Implementation is correct and cheap — `children` is an O(1) map lookup, so the per-render leaf filter is a single O(N) scan (fine at 16k rows). No behavior concerns.

The blocker is API shape. The two opposite-sense booleans (`excludeParents` is silently ignored when `includeChildren` is false) make an illegal state representable. Collapsing to a single `includeMode: 'roots' | 'all' | 'leaves'` maps exactly onto the three meaningful states and removes the foot-gun — keep `includeChildren` working but deprecated, drop next major.

Same treatment should apply to `StoreCountLabel`, and a `StoreRecord.isLeaf`/`isParent` getter would let us express "parent = has visible children" once instead of inlining `!it.children.length`.

(Branch was 544 commits behind develop — caught up locally.)

PR #4092, opened 2025-09-12
validEmails constraint implemented

cnrudd:
For forms that let users send emails, this seems necessary for the subject, cc, and bcc input fields.

Hoist P/R Checklist
-------------------

- [x] Reviewed and tested on Mobile: not required
- [x] Created Toolbox branch / PR: not required

### DIFF
--- a/data/validation/constraints.ts
+++ b/data/validation/constraints.ts
@@ -38,26 +38,21 @@ export const validEmail: Constraint<string> = ({value, displayName}) => {
 
 /**
  * Validate all email addresses in a field that allows multiple email addresses
- * separated by a semicolon - separator used by Outlook for multiple email addresses.
+ * separated by a semicolon - the separator used by Outlook for multiple email addresses.
  * All email addresses must be unique.
  */
 export function validEmails(c?: ValidEmailsOptions): Constraint<string> {
     return ({value, displayName}) => {
         if (isNil(value)) return null;
 
-        let emails = value.split(';');
-        if (c?.strict) {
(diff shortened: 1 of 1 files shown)


data/validation/constraints.ts
```diff
@@ -27,19 +27,65 @@ export const required: Constraint = ({value, displayName}) => {
...
+    };
+}
+export interface ValidEmailsOptions {
+    /**
+     * True to not allow whitespace between addresses and not allow a trailing semicolon.
+     * Defaults to false.
+     * False is the more user-friendly option, the user can enter spaces after semicolons, and a trailing semicolon
+     * and the user will not be bothered with a validation error, but downstream code will need to handle trimming.
+     */
+    strict?: boolean;
```
lbwexler, line 80: I think these options are overkill.  Not hard for app to do the split/count if it needs to validate the number.  And strict can just be fixed at false.

cnrudd: I'm ok with removing the `strict`, and hardcoding the `strict:false` code, but can we keep the `min/maxCount`?  Wouldn't it be annoying to a user if their form looked valid and then when they hit submit they got a second surprise message saying they have an error in their email field?  The situation erodes confidence.
Also OK with removing `minCount`, and keeping `maxCount`, but setting `minCount=1` would eliminate the need to import the `required` rule.

cnrudd:
Done.  `strict` flag removed.  Also made another small cleanup.

> OK. Let's just remove the strict flag. Sounds good.

PR #4097, opened 2025-09-18
Fixes Circular Dependency btwn LogUtils and LangUtils and reduces risk of future occurence

cnrudd:
Hoist P/R Checklist
-------------------


cmp/treemap/TreeMap.ts
```diff
@@ -21,7 +21,8 @@ import {errorMessage} from '@xh/hoist/cmp/error';
 import {mask} from '@xh/hoist/cmp/mask';
 import {Highcharts} from '@xh/hoist/kit/highcharts';
 import {wait} from '@xh/hoist/promise';
-import {logError, logWithDebug} from '@xh/hoist/utils/js';
+import {logWithDebug} from '@xh/hoist/utils/js';
```
amcclain, line 24: This was surprising - looked like a mistake in this context, then realized that its in our decorators package which is aliased in via `/js`.  

Seems like we might have a choice - either keep this where it is (all decorators come from Decorators.ts) or move it into the new log utils (all log methods come from LogUtils.ts).

utils/js/Decorators.ts
```diff
@@ -7,7 +7,7 @@
 import {XH} from '@xh/hoist/core';
 import {debounce, isFunction} from 'lodash';
 import {getOrCreate, throwIf, warnIf} from './LangUtils';
-import {withDebug, withInfo} from './LogUtils';
+import {withDebug, withInfo} from '../log';
```
amcclain, line 10: Should we be using the relative imports here, or avoid those and consistently import from the full path?

utils/js/LangUtils.ts
```diff
@@ -4,9 +4,9 @@
  *
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
-import {PlainObject, Thunkable} from '@xh/hoist/core';
+import type {PlainObject, Thunkable, LogSource} from '@xh/hoist/core';
```
amcclain, line 7: Is the `import type` option something that we should be using more consistently and rigorously across our Hoist codebase?

Also - was this particular change just a matter of style/best practice, or was this actually a key to solving the problem?

CHANGELOG.md
```diff
@@ -3,8 +3,10 @@
 ## 76.0.0-SNAPSHOT - unreleased
 
 ### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - upgrade to Hoist Core, change to Tab constructor)
-*  The constructor for `TabModel` has changed to take its owning container as a second argument.
+* The constructor for `TabModel` has changed to take its owning container as a second argument.
    (Most applications do not create `TabModels` directly, but it is possible.)
+* `checkVersion` and related methods' import path has changed. Update imports from
+  `@xh/hoist/utils/js/VersionUtils` to `@xh/hoist/utils/version`;  (Unlikely to affect most apps.)
```
amcclain, line 9: Need to call out the logging path change, just as much / more so than the version path change, right?

amcclain (commented):
Thank you for continuing to work through this!

* Could we summarize - for this PR + resulting squash commit message - what was going wrong here?  So we have some record of what we've learned and how this resolves?
* How could we document that in the code in any way?
* Would be answered by above, but why did VersionUtils also need to move?

CHANGELOG.md
cnrudd: yes.  done.

cmp/treemap/TreeMap.ts
cnrudd: See PR intro above. I don't think moving any log related decorators into the /log/ packages is a good idea.

utils/js/Decorators.ts
cnrudd: changed to use full path.  My thinking is: imports outside of package should use full path.

utils/js/LangUtils.ts
cnrudd: > Is the import type option something that we should be using more consistently and rigorously across our Hoist codebase?

I think so, based on what I read.

I don't think this was the root circ dep cause.  This change is more about best practices.

amcclain (approved):

utils/js/LangUtils.ts
lbwexler: I have seen where it resolves circular dependencies and am a big fan.  Not sure if it was needed here.

cmp/treemap/TreeMap.ts
lbwexler: I would have preferred to leave the logging stuff together, but understand colin's argument! so am good with that.

PR #4121, opened 2025-10-23
Added configurability for PersistenceProvider clearing of default state

### LINKED ISSUE #3945: `PersistenceProvider` clearing of default state should be an explicit option on target
Within `PersistenceProvider.bindToTarget`, we added a reaction that will actively clear the persisted state if it comes to equal the default state:

https://github.com/xh/hoist-react/blob/59c78e8f4e97f03b6cb040fa4ce07544b8bd2e35/core/persist/PersistenceProvider.ts#L187

The intention of this was to avoid eagerly capturing or "snapshotting" default state in such a way that code changes establishing a new default would not be respected.  We were very focused on the case of grid columns, especially with preferences, which are "auto saving" by nature. 

I.e. we wanted to avoid the following, which we saw as an anti-pattern:
* Grid persisted to pref
* User visits grid, changes nothing about the columns, has no opinion on the columns, but snapshot of default state is saved to pref
* App code is changed to introduce a new column that is visible by default 
* User is told about new column, looks for it in grid, but does not see it.

That change was made with the big updates to persistence several months ago.

Now, working on a client app with a custom viewManager-ish implementation, we observed the following:

* The custom vm impl had a "new view" option that we did not want to remove.  We went to implement this by creating a combined default state object for all persisted components - we asked each component's provider for its default state and installed it at the right path. Our expectation was that this would correctly represent the default state of the components.
* We saved that state in our persisted view record and then pushed it onto the components 
* The components naturally round-tripped the state back through the providers.
* We expected our view to remain clean - ie the state we pushed onto components was their default state, so the state they then report should match.
* For most components that was the case, but for a grid, the columns came back from the round-trip erased. So our newly saved view was immediately dirty in a surprising way.

We tracked this down to the erase-default-state behavior, but noticed it wasn't happening to all state - ie grid erased its columns, but not its default sortBy

Issue 1 with current impl:
* It is handled by a reaction that might or might not be triggered by simply instantiating the component. Some underlying rendered state might change (eg columns autosized) that can trigger this reaction in a way that erases the persisted state, but other bits of state (eg sortBy string) won't trigger this reaction and won't cause default state to be erased. So if and when this erasing happens is a bit unpredictable.

Issue 2:
* While the rationale above still makes sense for grid columns, I don't think it generalizes very well.  
* E.g. we have saved state that includes a boolean `showActiveItemsOnly`. Say the default in-code value is true.  
* User sets up a view, flips that switch off then on again, then adds a filter on a customer and saves it as "Active items for Customer X" - which is exactly what they are seeing at that time. That activeOnly setting is essential to their report, but happens to match the default and is not persisted.  
* In an unrelated decision, the default in the code is changed to false and a new version of the app is released.
* User returns to their saved view, which now unexpectedly (and without them doing anything) shows all items, not just active ones.

Proposal:
* Add a new optional getter to the `Persistable` interface - something like `persistDefaultValue: true`. 
* Check that flag when setting up the reaction - only do so if `false`. 
* Set to false for grid columns here - https://github.com/xh/hoist-react/blob/59c78e8f4e97f03b6cb040fa4ce07544b8bd2e35/cmp/grid/impl/InitPersist.ts#L35
* Consider also allowing `PersistenceProvider` instance to have a say - ie work into `PersistOptions`

jskupsik:
Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/grid/impl/InitPersist.ts
+++ b/cmp/grid/impl/InitPersist.ts
@@ -32,6 +32,7 @@ export function initPersist(
         PersistenceProvider.create({
             persistOptions: {
                 path: `${path}.columns`,
+                persistDefaultValue: true,
                 ...persistWith
             },
             target: {
@@ -61,7 +62,6 @@ export function initPersist(
         PersistenceProvider.create({
             persistOptions: {
                 path: `${path}.sortBy`,
-                persistDefaultValue: true,
                 ...persistWith
             },
             target: {
--- a/cmp/grid/impl/InitPersist.ts
+++ b/cmp/grid/impl/InitPersist.ts
@@ -61,6 +61,7 @@ export function initPersist(
         PersistenceProvider.create({
             persistOptions: {
                 path: `${path}.sortBy`,
+                persistDefaultValue: true,
                 ...persistWith
             },
             target: {
--- a/core/persist/Persistable.ts
+++ b/core/persist/Persistable.ts
@@ -8,11 +8,14 @@ import {isEqual} from 'lodash';
 
 /**
  * Interface for objects that can be bound to a {@link PersistenceProvider}.
- * @typeParam S - must be serializable to JSON and PersistableState<S> must be observable
+ * @typeParam S - must be serializable to JSON and PersistableState<S> must be observable.
+ * @param defaultPersistableState - (optional) sets the default persistable state. If not set, the
+ * PersistenceProvider will save an initial snapshot of `getPersistableState()` as the default state.
  */
 export interface Persistable<S> {
     getPersistableState(): PersistableState<S>;
     setPersistableState(state: PersistableState<S>): void;
+    defaultPersistableState?: PersistableState<S>;
 }
 
 /** Wrapper for a serializable Persistable state object. */
--- a/core/persist/PersistOptions.ts
+++ b/core/persist/PersistOptions.ts
@@ -34,6 +34,12 @@ export interface PersistOptions {
     settleTime?: number;
 
     /**
+     * If true, the default state will be persisted just like any other state, by writing its value.
+     * Otherwise, the default state is represented by clearing/emptying the persisted value.
+     */
+    persistDefaultValue?: boolean;
+
+    /**
      * Type of PersistenceProvider to create. Specify as one of the built-in string types,
      * or a subclass of PersistenceProvider.
      *
--- a/core/persist/PersistenceProvider.ts
+++ b/core/persist/PersistenceProvider.ts
@@ -61,8 +61,9 @@ export type PersistenceProviderConfig<S = any> = {
 export abstract class PersistenceProvider<S = any> {
     readonly path: string;
     readonly debounce: DebounceSpec;
-    readonly owner: HoistBase;
     readonly settleTime: number;
+    readonly persistDefaultValue: boolean;
+    readonly owner: HoistBase;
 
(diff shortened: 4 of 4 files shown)


cmp/grid/impl/InitPersist.ts
```diff
@@ -32,6 +32,7 @@ export function initPersist(
         PersistenceProvider.create({
             persistOptions: {
                 path: `${path}.columns`,
+                persistDefaultValue: true,
```
jskupsik: Implementation of Anslem's proposal:

> Set to false for grid columns here -

Note that this can be easily overriden by users using the PersistOption for persistColumns:
```
persistColumns: { persistDefaultValue: false }
```

core/persist/Persistable.ts
```diff
@@ -8,11 +8,14 @@ import {isEqual} from 'lodash';
...
  * Interface for objects that can be bound to a {@link PersistenceProvider}.
- * @typeParam S - must be serializable to JSON and PersistableState<S> must be observable
+ * @typeParam S - must be serializable to JSON and PersistableState<S> must be observable.
+ * @param defaultPersistableState - (optional) sets the default persistable state. If not set, the
+ * PersistenceProvider will save an initial snapshot of `getPersistableState()` as the default state.
  */
 export interface Persistable<S> {
     getPersistableState(): PersistableState<S>;
     setPersistableState(state: PersistableState<S>): void;
+    defaultPersistableState?: PersistableState<S>;
```
jskupsik: This gives the user full control of the `defaultPersistableState` without the need to create an exotic `getPersistableState()` method.

lbwexler (commented):
The flag to short-circuit the erasure looks pretty straightforward and safe and I approve -- the more simple hooks we have like this, the better.

Not sure I understand how the defaultPersistableState would be used?  Would we change our implementations of getPersistableState() in hoist itself?

lbwexler:
_(Posted by Claude Code on Lee's behalf — analysis below is the assistant's, for the team to sanity-check.)_

Picking up the thread on Lee's question about \`defaultPersistableState\`.

Two separate mechanisms here:

- **\`persistDefaultValue\` (PersistOptions)** — gates the erase-to-default branch. Clean, the approvable part.
- **\`defaultPersistableState\` (Persistable)** — nothing sets it anywhere on the branch, so the \`??\` always falls through to \`getPersistableState()\`. It's an inert hook today.

The columns case is the catch: \`InitPersist\` sets \`persistDefaultValue: true\`, which fixes the round-trip erasure but revives the original anti-pattern (user's current columns get snapshotted to pref → new default-visible column never shows). \`defaultPersistableState\` is the actual fix for that — keep erasing-to-default, but compare against the *declared* default (\`GridModel._defaultState\`) instead of a bind-time snapshot. That's the piece left unwired.

Proposed path:
1. Keep \`persistDefaultValue\`.
2. Either drop \`defaultPersistableState\` from this PR, or finish it by wiring grid \`InitPersist\` to set it from \`_defaultState\`.
3. Reconsider \`persistDefaultValue: true\` on columns on its own.
4. Needs a merge from \`develop\` (conflicting) + CHANGELOG entry.

Lean: do (2)-wired since that's what closes "Issue 1"; otherwise split it out and revert the columns flag for a minimal landing.

lbwexler:
discussed with Jakub -- use case no longer exists at client, and this PR has grown stale.  We will revisit it if it ever comes up again

Issue #4141, opened 2025-11-18
Reloading GridModel records with same data but with children breaks the grid


minusleaf:
In a client app, we have a toggle that on change sets the GridModel store's `loadTreeData`. The data the grid loads contains children, and by default, `loadTreeData` is false so we don't show the relationship tree between data entries in the grid.

Problem is if we set the toggle on to show the children for the same data, we get double the top-level rows displayed where half are basically empty parent rows but have children they were supposed to show and the other half are just the parent rows by themselves. In GridModel, the `columns` -> `renderer` function doesn't seem to pass the record value at all for these new rows.

Apparently, this happens whenever the records passed into the `GridModel` are numbers. Swapping back and forth a toggle I added to show/hide child items (modifying `gridModel.store.loadTreeData`), I can consistently reproduce this behavior in Toolbox.

<img width="1241" height="592" alt="Image" src="https://github.com/user-attachments/assets/6201f771-65b9-4dc2-93f2-23952d3aa414" />

PR #4142, opened 2025-11-20
Ensure `StoreRecord.treePath` populated with stringified IDs, fixes tree grid issues

minusleaf:
Fixes: https://github.com/xh/hoist-react/issues/4141

@amcclain and I combed through `GridModel`, `AgGridModel`, and `StoreRecord` and noted that a GridModel's `expandState` is a JSON (string only) rather than a Map where key is the ID of the record. The docs don't really explain why path IDs have to be strings, but we think that if the record ID is a number, then checking for equivalence becomes a string to number comparison.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@
 
 ### 🐞 Bug Fixes
 
-* Fixed `GridModel` not appending children to the parents correctly when loaded data uses a numerical ID.
+* Fixed `GridModel` not appending children to the parents correctly when loaded data uses a
+  numerical ID.
 
 ### ⚙️ Technical
 


amcclain:
Note that https://www.ag-grid.com/javascript-data-grid/tree-data-paths/ leads off with:
> Each row's position in the hierarchy must be provided to the grid as an array of strings

We also take care to construct `StoreRecord.agId` as a string, although we did not find it necessary for those to match the contents of `treePath` - if they did, we would have breaks on string IDs as well as numeric IDs as we prefix the `agId` property with `ag_` (for some reason I don't recall and we didn't comment on).

Finally - and this _might_ be related - note that the `expandState` object produced by the grid claims to by keyed by StoreRecord IDs:

https://github.com/xh/hoist-react/blob/7729a40fd6e5d53696d11d93179b09376cb0c8a4/cmp/ag-grid/AgGridModel.ts#L411C13-L411C87

but with numeric IDs this is not possible.

PR #4157, opened 2025-12-03
Sort displayed columns in column chooser to match grid order

### LINKED ISSUE #4153: Column chooser displayed columns order should match column order in the grid

haynesjm42:
Fixes #4153 

Required minor enhancement to `LeftRightChooserModel` to provide more control over sorting in the grids.

With this change the displayed columns will be sorted to match their order in the grid, while still sorting by name for the available columns.

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## 79.0.0-SNAPSHOT - unreleased
 
 ### 🐞 Bug Fixes
-* The column chooser now displays columns in the same order as they appear in the grid.
+
+* Fixed column chooser to display columns in the same order as they appear in the grid.
+* Defaulted Highcharts font to Hoist default (--xh-font-family)
 
 ## 78.1.3 - 2025-12-04
 
--- a/cmp/chart/Chart.scss
+++ b/cmp/chart/Chart.scss
@@ -15,6 +15,11 @@
   z-index: 9999 !important;
 }
 
+// Default Highcharts font to match Hoist standard
+svg.highcharts-root {
+  font-family: var(--xh-chart-font-family) !important;
+}
+
 // Convenience class for optional use when specifying custom chart tooltips.
 // See Toolbox for example usage.
 .xh-chart-tooltip {
--- a/styles/vars.scss
+++ b/styles/vars.scss
@@ -358,6 +358,7 @@ body {
   // Charts
   //------------------------
   --xh-chart-bg: var(--chart-bg, var(--xh-bg));
+  --xh-chart-font-family: var(--chart-font-family, var(--xh-font-family));
 
 
   //------------------------


desktop/cmp/grid/impl/colchooser/ColChooserModel.ts
```diff
@@ -111,19 +114,33 @@ export class ColChooserModel extends HoistModel {
...
+                if (pinned === 'left') {
+                    return 0;
+                }
+
+                if (pinned === 'right') {
+                    return 2;
+                }
+
+                return 1;
+            });
```
amcclain: This sort wasn't super clear to me - is the idea that our `columnState` is already in the correct order unless there are pinned cols, which aren't accounted for in that ordering?

Could comment that to clarify.  (Also can't help but make me wonder if this should be handled somehow at a deeper level - seems "funny" to have to deal with this concern here, but not clear to me we had any particular contract around sorting out of `columnState`)

amcclain (commented):
You can see from my comments that I found the new LR configs a bit confusing, but  I don't have any workable suggestions to improve on where you landed with it.  Yes, the API seems a bit sub-optimal somehow, we might do it differently with a clean sheet approach, but I'm all for taking this as it gets us to the capabilities we need.  Thanks for working it out.

Approving but leaving open for @lbwexler to have a look if he has a chance

desktop/cmp/grid/impl/colchooser/ColChooserModel.ts
haynesjm42: Yeah when a column is pinned it's order in column state remains where it was before becoming pinned, and the relative order of pinned columns on the same side is correct in column state but where exactly they are relative to other columns not pinned or not on the same side is somewhat undefined, this is just how ag-grid handles things, I assume to keep the column state as a simple list, and to get the behavior of columns going back to their original position when the user unpins them (which I guess makes sense)

Will add a comment to clarify for the next person to encounter that code

desktop/cmp/grid/impl/colchooser/ColChooserModel.ts
```diff
@@ -55,6 +56,8 @@ export class ColChooserModel extends HoistModel {
             rightTitle: 'Displayed Columns',
             rightEmptyText: 'No columns will be visible.',
             leftSorted: true,
+            leftSortBy: 'text',
+            rightSorted: true,
```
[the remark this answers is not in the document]
amcclain: Agree 100% re avoiding rabbit holes on this one - if we take the opportunity to revisit column chooser more fully we'll likely move away from LR chooser anyway, and it's not otherwise some absolutely central part of the toolkit. I'm happy to have these API updates that get us what we need even if they are (unavoidably) a bit confusing/overlapping.

amcclain (approved):

PR #4184, opened 2025-12-17
Tighten `LocalDate` unit types

febbraiod:
Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/datetime/LocalDate.ts
+++ b/utils/datetime/LocalDate.ts
@@ -4,7 +4,7 @@
  *
  * Copyright © 2025 Extremely Heavy Industries Inc.
  */
-import {XH} from '@xh/hoist/core';
+import {LocalDateUnit, XH} from '@xh/hoist/core';
 import {computeOnce, throwIf} from '@xh/hoist/utils/js';
 import {isNil, isString} from 'lodash';
 import moment, {Moment, MomentInput} from 'moment';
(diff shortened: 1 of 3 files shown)


utils/datetime/LocalDate.ts
```diff
@@ -21,9 +21,33 @@ import moment, {Moment, MomentInput} from 'moment';
...
+        'years',
+        'quarter',
+        'quarters',
+        'month',
+        'months',
+        'week',
+        'weeks',
+        'day',
+        'days'
+    ];
```
febbraiod: Intentionally removing 'date' here. The 'diff' method in moment actually doesn't accept it (even though other moment methods do). Oddly, it doesn't throw, but TS complains and it seems to produce nonsense.

Right now:

moment().diff(moment('20250101'), 'day') === 350
moment().diff(moment('20250101'), 'date') === 30297322410
moment().diff(moment('20250101'), 'foo') === 30297345155

Obviously these appear to be timestamps, but in 1970?

Previously, our ensureUnitValid method normalized inputs via Moment, but 'date' normalizes to 'date', allowing an invalid unit to slip through and be passed to diff.

Anyway I think slimming down what we accept here is the right move. We are insisting on semantically accurate values. (Even if 'd' might obviously mean 'day'). And also smoothing over this inconsistency in the moment API.

PR #4192, opened 2025-12-29
Optimize `GridLocalModel.syncData` 

### LINKED ISSUE #4169: `GridLocalModel.syncData` inefficiencies
There are some inefficiencies in `GridLocalModel.syncData()` that appear to be O(n) operations that we call on every sync, even if they are irrelevant to the grid configuration.

Timings mentioned were observed in a flat grid with ~60k rows and ~80 columns (15 or so visible)

The most time is spent applying the ag-grid transaction. Applying around 2k updates took around 600ms. After more digging it appears this time is pretty much entirely spent updating the client side row model, specifically updating the sorting of the rows. I noticed that we are no longer enabling `deltaSort` in the grid options (we used to have it as experimental I think) and in this case it dropped the time spent applying the transaction to ~175ms. Need to better understand the caveats with deltaSort but seems like something we might want to enable by default.

* `updatePinnedSummaryRowData()` - this results in a call to `updateGridOptions` on the ag-grid api which appears to have an O(n) ties to the nubmer of rows. Even if the pinned rows were empty, and are being set to empty, simply calling `updateGridOptions` is taking ~20ms. Checking if the pinned row data has actually changed from what is already set so we can early-out resolves this.
* `noteAgExpandStateChange()` - this is called even if the grid is not in `treeMode` and has no `groupBy` - meaning it would never have expandable rows. Takes up around 10ms 
* `applyScrollOptimization()` - by default this is always called on every sync but I observed it taking between 80ms and 120ms. Not clear to me under which circumstances this needs to be called, but we might consider having an easy way to turn this off that doesn't require changing the grid options and setting rows heights in the app (at least I think that is how one would turn it off?)

haynesjm42:
Fixed #4169 

* Avoid doing any unnecessary work when syncing data to ag-grid
* Introduce new GridModel properties to enable/disable certain features that can impact performance when dealing with large data sets

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,6 @@
 * Improved the efficiency and logging of MsalClient.
 
 ### 📚 Libraries
-
 * react-grid-layout `1.5.0 → 2.1.1`
 * hoist-core `35.0.0`
 


cmp/grid/GridModel.ts
```diff
@@ -353,6 +353,21 @@ export interface GridConfig {
      */
     enableFullWidthScroll?: boolean;
 
+    /**
+     * Set to true to disable scroll optimization for large grids, where we proactively update the
+     * row heights in ag-grid whenever the data changes to avoid hitching while quickly scrolling
+     * through large grids.
```
lbwexler: I would be inclined to just put this quietly in "experimental" if we think its farily likely that we eventually just remove it.

PR #4196, opened 2026-01-06
Fix for a react error thrown when Blueprint menus include nested child items

RyanMG:
I found this error when implementing a new nested navigation menu in Dash. As soon as the pop over menu was shown, a console error would appear complaining about missing keys on menu items.
Screenshot 2026-01-06 at 2 38 34 PM

The error would not appear unless menu items were present which themselves were parent items to show flyout child items.
I tried adding a key to each item (parent and child), but this did not fix the issue, and when inspecting the elements before render, I could see that no key was present.

I found the issue in our Hoist "parseMenuItems" method - which takes a list of menu option config objects, and converts them into Blueprint MenuItem components. This did not have "key" in its list of props it would accept then pass into the MenuItem from blueprint.

I found the same error in Toolbox, and confirmed the error was present even before the recent Blueprint 6.0 update.
Screenshot 2026-01-06 at 2 50 15 PM

In the above case, it was the same error text output, but an easier fix.
https://github.com/xh/toolbox/blob/develop/client-app/src/desktop/tabs/panels/BasicPanel.tsx#L94-L96
Here we are not using the "parseMenuItems" method, but we ARE neglecting to provide a key to the "menuItem" listings for these nested children. So this toolbox error could be fixed with just the added key.

However, I wanted a recreate the error from Dash, so I create https://github.com/xh/toolbox/compare/bp-menu-child-item-parsing?expand=1 to update this menu code to use "parseMenuItems", and confirm that the added key would work when using this method to generate a menu.

Hoist P/R Checklist
-------------------

### DIFF
--- a/utils/impl/MenuItems.ts
+++ b/utils/impl/MenuItems.ts
@@ -4,7 +4,7 @@
  *
  * Copyright © 2026 Extremely Heavy Industries Inc.
  */
-import {isMenuItem, MenuItemLike, XH} from '@xh/hoist/core';
+import {ElementSpec, isMenuItem, MenuItemLike} from '@xh/hoist/core';
 import {menuDivider, menuItem} from '@xh/hoist/kit/blueprint';
 import {MenuItemProps} from '@blueprintjs/core';
 import {wait} from '@xh/hoist/promise';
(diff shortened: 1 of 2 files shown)


utils/impl/MenuItems.ts
```diff
@@ -43,8 +43,9 @@ export function parseMenuItems(items: MenuItemLike[]): ReactNode[] {
             const {actionFn} = item;
 
             // Create menuItem from config
-            const cfg: MenuItemProps = {
+            const cfg: MenuItemProps & {key?: string} = {
                 text: item.text,
+                key: item.key ?? XH.genId(),
```
RyanMG: Unsure about this. Would it be better to let it error, and force the developer to provide a key?

PR #4212, opened 2026-01-27
Tighten auth during load sequence

lbwexler:
This requires the related Hoist Core PR. with the same name.

Main Idea is to remove an additional step in the auth process by returning identity information straight away with /authStatus and /login, rather than requiring an extra round trip.

The related refactor should also allow Apps to more easily create custom HoistUsers out of their custom AppModels.

I considered having AuthModel also provide the aquired  IdentityInfo as a getter, rather than returning it from the framework method.   On the fence about that, but if it made more sense to you guys we could go in that direction

Hoist P/R Checklist
-------------------


CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 ## 81.0.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes
+* Change to the signatures of some `HoistAuthModel` methods to return `IdentityInfo` rather than
```
jskupsik: Reminder to add the proper `(Requires hoist-core X.Y.Z)` line to the changelog before merging, once the non-breaking server-side version of this is released.

jskupsik (approved):
Good to go out **next time we require the server version to be updated**. 

Note that it will error otherwise:
```
installHook.js:1 [ExceptionHandler] | Cannot read properties of undefined (reading 'user') | TypeError: Cannot read properties of undefined (reading 'user')
    at AuthModel.parseIdentityInfo (HoistAuthModel.ts:127:18)
    at AuthModel.getAuthStatusFromServerAsync (HoistAuthModel.ts:52:41)
    at async AuthModel.completeAuthAsync (AuthModel.ts:66:21)
    at async AppContainerModel.initAsync (AppContainerModel.ts:201:30)
```

Also do not forget about the very tiny type-only change in toolbox: https://github.com/xh/toolbox/pull/801

appcontainer/AppContainerModel.ts
```diff
@@ -197,41 +198,40 @@ export class AppContainerModel extends HoistModel {
...
                     !appSpec.enableLoginForm,
                     'Unable to complete required authentication (SSO/Auth failure).'
                 );
                 this.setAppState('LOGIN_REQUIRED');
-                return;
             }
         } catch (e) {
             this.setAppState('LOAD_FAILED');
             XH.handleException(e, {requireReload: true});
             return;
```
amcclain: Could clean up this `return` now as well if you wanted

PR #4213, opened 2026-01-28
Fix View connection logic

haynesjm42:
Don't depend on the Cube info for determining when a View update is required upon connecting to a Cube, as it is intended for application use only, and may not be used/updated at all.

Hoist P/R Checklist
-------------------

### DIFF
--- a/data/cube/Cube.ts
+++ b/data/cube/Cube.ts
@@ -155,6 +155,11 @@ export class Cube extends HoistBase {
         return this.store.empty;
     }
 
+    /** Timestamp (ms) of when the Cube data was last updated */
+    get lastUpdated(): number {
+        return this.store.lastUpdated;
+    }
+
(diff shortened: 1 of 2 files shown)


data/cube/Cube.ts
```diff
@@ -234,7 +239,7 @@ export class Cube extends HoistBase {
         this._connectedViews.add(view);
 
         // If the view is not up-to-date with the current cube data, then reload the view
-        if (view.info !== this.info) {
+        if (view.cubeDataUpdated !== this.lastUpdated) {
```
amcclain: I had wondered a bit about the previous use of `info` - this def makes more sense.  

That timestamp is really the time that the data was actually *loaded* into the store, right, as opposed to a possible business timestamp (e.g. values as-of)?  Given that, it seems like this should be a reliable indicator - i.e. no chance of accidental collision.

PR #4225, opened 2026-02-12
Add contentBoxProps to mobile Panel + scrollable to desktop Panel

lbwexler:
## Summary

- Adds `contentBoxProps` to mobile `Panel`, matching the desktop `Panel` API. Children are now
  wrapped in a `frame()` with the `xh-panel__content` CSS class, consistent with the desktop
  implementation.
- Adds `scrollable` prop to desktop `Panel`, matching the existing mobile `Panel` API. Sets
  `overflowY: 'auto'` on the content area as a convenience shorthand.
- Both Panels now have a consistent API surface with both `contentBoxProps` and `scrollable`.
- Updates CHANGELOG, Panel README, and Form README.

## Test plan

- [ ] Verify mobile Panel renders content correctly with default props
- [ ] Test mobile `contentBoxProps` with padding, flexDirection, and overflow
- [ ] Test mobile `scrollable` still works as before
- [ ] Test desktop `scrollable: true` enables vertical scrolling while keeping toolbars fixed
- [ ] Verify no regressions in desktop Panel contentBoxProps behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@
   header and collapsible content.
 * Added `FormFieldSet` component for grouping `FormFields` and displaying their aggregate validation
   state.
-* Added `contentBoxProps` to `Panel`, providing direct control over the inner frame wrapping content
-  items. Use to apply padding, change flex direction, enable scrolling, or add custom classes without
-  extra wrapper elements. Matches the existing `contentBoxProps` API on `Card`.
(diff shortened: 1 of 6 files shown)


lbwexler:
@amcclain -- thought the scrollable property would be nice to keep, and make symmetrical.   Was on the fence about the vertical assumption that mobile has.  Could be amended to, say,  scrollable: 'horizontal' | 'vertical' | 'both' | true

PR #4245, opened 2026-02-19
Add `bind` config to GroupingChooserModel

### LINKED ISSUE #4099: `GroupingChooserModel` should support binding to a `View` or `GridModel`
`GroupingChooserModel` is currently the odd one out in the family of models and components for "slicing and dicing" tabular data (`GridModel`, `View`, `FilterChooserModel`, etc.) - it should also support the `bind` api that we use elsewhere to avoid the need for setting up all the boilerplate to wire it up with the `View` or `GridModel`

haynesjm42:
## Summary
- Adds a `bind` config to `GroupingChooserModel` that accepts a `GridModel` or Cube `View`
- Automatically syncs the chooser's value to the target as it changes — calls `setGroupBy()` for GridModel, `updateQuery({dimensions})` for View
- One-way binding: the GroupingChooser is the source of truth

Resolves #4099

## See also
- xh/toolbox#812 — exercises the new `bind` property in the standard grid example

### DIFF
--- a/data/Field.ts
+++ b/data/Field.ts
@@ -36,7 +36,7 @@ export interface FieldSpec {
     /** Value to be used for records with a null, or non-existent value. */
     defaultValue?: any;
 
-    /** True to allow this field to be used for grouping. Defaults to false. */
+    /** True if this field is intended to be used for grouping.  Defaults to false. */
     isDimension?: boolean;
 
     /** Rules to apply to this field. */


haynesjm42:
I was thinking it made sense when using a GroupingChooser to give it complete control over the state since GroupingChooserModel may be persisting it, but in hindsight I don't think there is any reason not to just let it go both ways, just in case someone decides to update the state directly instead of going through the chooser.

PR #4257, opened 2026-02-24
Widen FetchOptions/TrackOptions loadSpec typing to accept LoadSpecConfig — Closes #4253

### LINKED ISSUE #4253: FetchOptions.loadSpec should accept Partial<LoadSpec>
FetchOptions.loadSpec is typed as LoadSpec, but callers that don't have a full LoadSpec instance (e.g. when only isAutoRefresh is known) must use @ts-ignore to pass a partial object.

Suggestion: Change the type to `LoadSpec | Partial<LoadSpec>` (or just `Partial<LoadSpec>`). FetchService only reads the boolean flags (`isAutoRefresh`, `isRefresh`) from it for tracking and exception handling, so a partial object is sufficient.

febbraiod:
## Summary
- Widens `FetchOptions.loadSpec` and `TrackOptions.loadSpec` to accept `LoadSpec | LoadSpecConfig`, allowing app code to pass a plain `{isAutoRefresh: true}` for direct fetch calls outside the `doLoadAsync` lifecycle.
- Adds an `instanceof LoadSpec` guard in `ExceptionHandler` so that `typeDisplay` and `loadNumber` are only extracted from full `LoadSpec` instances (a plain config would have produced `undefined` for those fields).

## Test plan
- [x] `yarn lint` passes
- [x] `tsc` passes (pre-commit hook)
- [ ] Verify exception handling still populates `loadType`/`loadNumber` for standard `doLoadAsync` flows
- [ ] Verify `{isAutoRefresh: true}` on a direct `fetchJson()` call suppresses alerts and skips tracking


core/ExceptionHandler.ts
```diff
@@ -261,11 +261,15 @@ export class ExceptionHandler {
                 delete serverDetails.lineNumber;
             }
 
-            // Remove verbose loadSpec from fetchOptions
+            // Remove verbose loadSpec from fetchOptions, extracting summary fields
+            // if a full LoadSpec instance (vs. a plain LoadSpecConfig) was provided.
             const {fetchOptions} = ret;
             if (fetchOptions?.loadSpec) {
```
lbwexler: if (fetchOptions?.loadSpec instanceof LoadSpec) removes the need for the nested if

lbwexler: duh

PR #4297, opened 2026-03-11
Add desktop SegmentedControl input

### LINKED ISSUE #4175: Add SegmentedControl wrapper
Our current `ButtonGroupInput` has the following disadvantages:

- Active selection is unclear when there are only 2 values.
- When in a "disabled" state, active selection can be even more ambiguous.
- There is no good way to vary intents across buttons because adjoining borders are effectively collapsed, leading to buttons where 3 sides are styled one color and the 4th another.
- In cases where only a single Button is rendered within a ButtonGroup, one side will have square borders, and the other side's borders will be rounded. It could be considered an anti-pattern to render a ButtonGroup with a single Button, but it can happen in dynamic forms where options are constrained to a single value based on other inputs.

Blueprint has a `SegmentedControl` component that solves these issues:
https://blueprintjs.com/docs/#core/components/segmented-control

We could consider making this a drop-in replacement for `ButtonGroupInput`, although `ButtonGroup` (the non-input component) does still seem useful for action bars that are not intended to work as radio buttons.  Having `ButtonGroup` and `ButtonGroupInput` use different underlying components breaks our current convention, so maybe we create a new input entirely -- `SelectBar`? `ButtonSelect`? `RadioStrip`? `RadioButtons`?

amcclain:
Closes #4175

## Summary

- New `SegmentedControl` HoistInput wrapping Blueprint's SegmentedControl component, providing clear visual indication of the active selection within a small set of mutually exclusive options.
- Driven by an `options` prop (like RadioInput/Select) supporting `value`/`label`/`icon`/`disabled` per option, or bare primitive values.
- Exports `BpSegmentedControl` and `bpSegmentedControl` factory from the Blueprint kit wrappers, following the `BpCard` naming pattern.
- Includes themeable CSS custom properties (`--xh-segmented-control-*`) for background, selected state, text colors, border-radius, and padding — with light/dark mode defaults derived from existing Hoist theme variables.
- Supports `intent` prop (`'none' | 'primary'`) with Blueprint's intent colors overridden to use Hoist's `--xh-intent-primary` theme variables for consistent theming.
- Supports `compact` prop for reduced sizing in compact toolbars and other space-constrained contexts, using Blueprint's `size: 'small'` and Hoist's `--xh-tbar-compact-*` dimensions.
- Fixes `RadioInput` vertical spacing — restores `margin-bottom` on non-inline radio options that was lost by Hoist's global `bp6-control` margin reset.

## Props

| Prop | Type | Default | Description |
|------|------|---------|-------------|
| `options` | `(SegmentedControlOption \| any)[]` | required | Available options |
| `fill` | `boolean` | `true` | Stretch to fill available width |
| `intent` | `'none' \| 'primary'` | `'none'` | Visual intent for the selected option |
| `compact` | `boolean` | `false` | Reduced sizing for compact toolbars |
| + all standard `HoistInputProps` | | | `bind`, `value`, `disabled`, `onChange`, `onCommit`, etc. |

## Test plan

- [ ] Verify selection tracking and value binding in Toolbox InputsPanel demo (see companion Toolbox PR)
- [ ] Verify light and dark theme rendering
- [ ] Verify CSS custom property overrides work for app-level theming
- [ ] Verify FormField integration (validation borders)
- [ ] Verify compact mode in a compact toolbar
- [ ] Verify primary intent renders with Hoist theme colors
- [ ] Verify RadioInput vertical spacing improvement for non-inline options

Generated with Claude Opus 4.6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,13 @@
 ### 🎁 New Features
 
 * Added publish controls to the Admin Metrics tab, supporting the new opt-in metrics export
-  feature in hoist-core 36.4.
+  feature in `hoist-core >= 36.4`.
 * Added `CheckboxButton` desktop input component — a button-based boolean toggle matching the
   existing mobile component. Added `checkedIcon` and `uncheckedIcon` props to both desktop and
   mobile versions for custom icon support.
+* Added `SegmentedControl` desktop input component — a toggle group for mutually exclusive options
+  with strong visual differentiation of the active selection, an improvement over `ButtonGroupInput`
+  for small option sets.
 
 ### ⚙️ Technical
 
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -100,6 +100,7 @@ Desktop form inputs with Blueprint styling:
 | `RadioInput` | Radio button group |
 | `Slider` | Range slider |
 | `ButtonGroupInput` | Segmented button selection |
+| `SegmentedControl` | Toggle group for mutually exclusive options with strong visual differentiation of the active selection |
 | `CodeInput` | Code editor with syntax highlighting |
 | `JsonInput` | JSON editor with validation |
 


amcclain:
Note on naming: "Segmented Control" is a well-established term of art across UI toolkits — it's not a Blueprint-specific concept. Apple introduced the pattern in UIKit/AppKit (UISegmentedControl, NSSegmentedControl), and it's since been adopted broadly: Android (MaterialSegmentedButton), Ant Design (Segmented), Chakra UI (SegmentedControl), and Blueprint all use the same name. The pattern describes a horizontally arranged group of mutually exclusive options where the selected option is visually elevated — distinct from a button group or radio group in its visual treatment of the active state.

Issue #3893, opened 2025-01-08
Enhanced Select to handle "selecting objects"


jskupsik:
A not-too-uncommon use case of the select prop is to let the user select (conceptually) an object.

Our select component almost supports this, but not really well and not all the time.

There are two big concerns here: 
1. We don't make it easy to *always* render a "value" using a "label" - it is just very finicky to get it right, and there are some edge cases/limitations that results in bugs where we render the value.
2. We have a strong assumption that "value" is a primitive (string or number). This leads to complexity / errors in cases where we essentially want the object to be the value, but have to work around this select limitation.

Usually, we implement "object selection" by making our `SelectOption` have a value of the object's id and a label be the object's name. This works nicely most of the time, rendering the name rather than the id, but there are a lot of edge cases/pitfalls where we render the id instead:
- The currently selected object is not currently selectable - dynamic options have to be carefully curated to never exclude the current object.
- We are using QueryFn and we are rendering a saved value not in cache - we have to manually use options to prime the initial value, or give up on using the id and instead use the name as the key.
- The form that wraps the select is readonly - never make a form that wraps a select readonly, instead make it disabled. For selectEditors, we need to remember give the column the proper renderer.

This makes using the select correctly for this purpose quite difficult.

If we do not have all of the objects readily available, it is quite difficult to use queryFn with `SelectObject`. One is tempted to make the `value` be the object itself rather than id, it almost works, but there are issues with how it is handled and occasionally is rendered as [Object object]. We should consider supporting this.

We have a  `SelectOption` - it is typed and assumed to be very strictly:
```ts
export interface SelectOption {
    value?: any;
    label?: string;
    options?: (SelectOption | any)[];
}
```
but in some places we instruct the user to place other fields on it - fields which then aren't always present if rendering a saved value not defined in `options`:
```ts
    /**
     * Function to render options in the dropdown list. Called for each option object (which
     * will contain at minimum a value and label field, as well as any other fields present in
     * the source objects).
     */
    optionRenderer?: (opt: SelectOption) => ReactNode;
```
[...]

PR #4303, opened 2026-03-16
Enhance Select for robust object value support (#3893)

jskupsik:
Fixes bugs preventing object (non-primitive) values from working correctly, and adds a `generateOptionFn` prop for resolving values not present in the current options list.

**Bug fixes for object values**
- `doQueryAsync`: dedupe via `isEqual`/`unionWith` instead of `keyBy` (string-key collision on objects).
- `optionRenderer` / selected-option matching: use deep equality so object values match and no longer render as `[object Object]` or collide with one another.

**New `generateOptionFn` prop**
- Synchronous `(value) => SelectOption` called when a selected value is not found among the current options (e.g. `queryFn`-based selects, readonly forms). Return `null` to fall back to the default value-as-label behavior.
- Resolves the label synchronously in `findOption`, so the value renders with its proper label on first paint - no async cache, reaction, or race conditions.

This supersedes the async `lookupFn` + `_lookupCache` approach earlier explored on this branch; the synchronous design is a much smaller, simpler footprint (just the prop plus the `findOption` fallback) and avoids the stale-cache / reaction-ordering bugs that approach was prone to.

Tested on desktop and mobile via the Toolbox companion PR (employee-by-id example renders the label, not the raw id, on mount, for a value not in the options list).

Hoist P/R Checklist
-------------------


lbwexler:
Could we make the `lookupFunction` synchronous?  Is is more of a `optionGenerator`?   This would simplify things greatly and assuage some concerns I am uncovering with timing and loops, and alot of the complexity.

PR #4308, opened 2026-03-20
Fixed FilterChooser / QueryEngine to better handle null values and errors

jskupsik:
## Summary

* Fixed `FilterChooser` `QueryEngine` to handle null values from `Store.getValuesForFieldFilter()` without throwing — null values in `spec.values` previously crashed `renderValue()` → `formattedValue.match()`, silently killing the suggestion dropdown.
* Wrapped `QueryEngine.queryAsync()` in a try-catch that logs errors via `model.logError()` rather than letting them vanish as unhandled promise rejections.
* The 'is' pseudo-operator is now shown in the e.g. operator hints, and 'is blank' / 'is not blank' options are offered in value suggestions when a field has null values using new `getBlankOptionsForField`.
* Consolidated `getValueMatchesForField` and `getBlankOptionsForField` into a single `getMatchesForField` method. 
* Simplified `openSearching` from a two-pass approach into a single loop - see comment for explanation here.

image

image

image

## Test plan

- [x] Open Toolbox → Grids → Column Filtering
- [x] Verify FilterChooser suggestions appear when typing a field name
- [x] Verify 'is blank' / 'is not blank' suggestions appear for fields with null values
- [x] Verify typing `Field is` shows 'blank' and 'not blank' completions
- [x] Verify the e.g. operator hint includes 'is' at the end of the operator list
- [x] Open a grid column filter and verify the [blank] checkbox option appears for columns with null data
- [x] Verify no console errors during any of the above

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/impl/QueryEngine.ts
+++ b/cmp/filter/impl/QueryEngine.ts
@@ -253,10 +253,8 @@ export class QueryEngine {
             }
         });
 
-        // Blank/not-blank options for fields with null values. Both ops are always
-        // included — the 'is' pseudo-operator handles blank/not-blank as a pair, and
-        // they're useful to surface regardless of whether the user typed '=' or '!='.
-        if (spec.hasBlankValues) {
+        // Blank/not-blank options for fields with null values.
+        if (values.some(v => v == null)) {
             const blankTestFn = queryStr ? testFn : null,
                 blankEntries: Array<{label: string; op: FieldFilterOperator}> = [
                     {label: 'blank', op: '='},
--- a/cmp/filter/FilterChooserFieldSpec.ts
+++ b/cmp/filter/FilterChooserFieldSpec.ts
@@ -128,10 +128,6 @@ export class FilterChooserFieldSpec extends BaseFilterFieldSpec {
         return valueParser;
     }
 
-    get hasBlankValues(): boolean {
-        return this.values?.some(v => v == null) ?? false;
-    }
-
     loadValuesFromSource() {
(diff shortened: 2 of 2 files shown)


cmp/filter/impl/QueryEngine.ts
```diff
@@ -93,16 +98,12 @@ export class QueryEngine {
         // Suggest matching *fields* for the user to select on their way to a more targeted query.
         let ret = this.getFieldOpts(q.field);
 
-        // If a single field matches, reasonable to assume user is looking to search on it.
```
jskupsik: This change explained by CC. This hurts my head. I'd be willing to revert it and accept the old bug.

> The difference: the old code called the single-match spec twice — once with ''
>   (all values) and once with q.field (filtered values). The filtered set is always a
>    subset of all values, so the second call only produced duplicates that
>   react-select would dedup by value (which is JSON.stringify(filter)).
> 
>   However, there is a subtle difference — the old code's second pass could mark
>   options as isExact: true when a field's value happened to match the query text
>   (e.g. a "City" field with a value literally named "City"). The new code doesn't
>   produce that isExact marking for the single-match spec since it compares against
>   ''. But the old code had the same effective bug — react-select keeps the first
>   occurrence during dedup, which was the pass-1 option with isExact: false. So the
>   pass-2 exact marking was already lost.
> 
>   Net result: the behavior is equivalent. The new code just removes the redundant
>   second call. For all non-single-match specs, the behavior is identical — they only
>    ran in pass 2 before and only run in the forEach now, both with q.field.
>

PR #4334, opened 2026-04-15
Shed obsolete/unused runtime dependencies

### LINKED ISSUE #4332: Shed obsolete/unused Tier 1-2 dependencies
Audit of `package.json` surfaced several dependencies that are either unused, obsolete, or redundant with modern browser support. Grouped below in suggested order of attack — each tier gets cheaper-to-verify as you go up.

## Tier 1 — Clearly removable (no code changes or near-zero)

- **`react-transition-group`** — Zero imports in source; appears only in `package.json` / `yarn.lock` / `CHANGELOG`. Just delete.
- **`@popperjs/core`** (direct dep) — Not imported directly anywhere. Already pulled in transitively by `react-popper` and `@blueprintjs/*`. Remove from direct deps.
- **`regenerator-runtime`** — Used only in `static/polyfills.js`. With a modern TS target and evergreen-browser support, async/generator transforms no longer need the runtime. Coordinate with `@xh/hoist-dev-utils` since that package references the polyfills entry.

## Tier 2 — Small, focused changes

- **`resize-observer-polyfill`** — Single import at `utils/js/DomUtils.ts:8`. `ResizeObserver` is native in every browser Hoist supports. Drop the import and use the global.
- **`@onsenui/fastclick`** — Single import at `cmp/pinpad/PinPadModel.ts:15`. The 300ms tap delay FastClick existed to fix hasn't been a concern since ~iOS 9.3 / modern Android. Remove and smoke-test PinPad on a touch device.
- **`core-js`** — Also only in `static/polyfills.js`. Same calculus as `regenerator-runtime`; larger bundle-size impact than any other item here. Revisit whether a general ES polyfill shim is still warranted for Hoist's supported browser matrix.

## Not included here

Tier 3+ items (`react-beautiful-dnd` → `@hello-pangea/dnd`, `react-dates`, `jquery`/`golden-layout` v2, `codemirror` 6) are real migrations and should get their own tickets. See audit discussion for context.

lbwexler:
Closes #4332.

Removes six direct dependencies that are no longer needed given Hoist's modern browser matrix and existing transitive deps.

## Reasoning per removal

| Dependency | Where it was used | Why it's safe to drop |
|---|---|---|
| `react-transition-group` | Nowhere — zero source imports. Only appeared in `package.json` / `yarn.lock`. | Dead entry. `react-select` pulls its own copy transitively for any code that actually needs CSS transition hooks. |
| `@popperjs/core` (direct dep) | Not imported directly anywhere. | Still pulled in transitively by `react-popper` and `@blueprintjs/*`, so Popper-using components keep working — we just stop double-declaring it. |
| `regenerator-runtime` | Only referenced by `static/polyfills.js`. | A TS target of ES2017+ and Babel's modern preset don't emit regenerator calls anymore, so the runtime shim is inert. Removing it saves a small bundle chunk and an import most apps never need. |
| `resize-observer-polyfill` | Single import at `utils/js/DomUtils.ts`. | `ResizeObserver` ships natively in every browser Hoist's matrix supports (Chrome/Edge/Firefox/Safari all shipped it by 2020). The polyfill is dead weight — swap in the global. |
| `@onsenui/fastclick` | Single import at `cmp/pinpad/PinPadModel.ts`, attached on mount to work around the legacy 300ms touch-click delay. | That delay was fixed at the platform level by iOS 9.3 / Chrome for Android ~2016 once viewports declared `width=device-width`. Modern mobile browsers fire `click` immediately; FastClick now does nothing useful and can actually interfere with synthetic events. The attach logic (and its ref plumbing) goes with it. |
| `core-js` | Only referenced by `static/polyfills.js` (and pinned in `resolutions`). | General ES polyfill shim for older browsers. Hoist's supported matrix is evergreen — the language features `core-js` ships are all native. Biggest bundle-size win in this set. |

## Knock-on cleanup

* `PinPadModel` no longer owns a ref to attach FastClick on mount. The ref + `composeRefs` machinery in both `desktop/cmp/pinpad/impl/PinPad.ts` and `mobile/cmp/pinpad/impl/PinPad.ts` is simplified accordingly.
* `static/polyfills.js` is kept as a no-op because `@xh/hoist-dev-utils` wires it in as a webpack entry point for every app bundle; removing the file would force a coordinated dev-utils release. Its README entry is updated to reflect the new state.
* `core-js` removed from `resolutions` as well.

## Test plan

- [ ] `yarn install` / `yarn build` cleanly across a consuming app (Toolbox)
- [ ] Smoke test Toolbox desktop + mobile, including the PinPad example on a touch device
- [ ] Verify `ResizeObserver`-driven components (e.g. grid autosize, layout utilities) still behave correctly
- [ ] Confirm Popper-using components (e.g. tooltips, popovers) still render — transitive `@popperjs/core` should resolve via `react-popper`


TomTirapani:
Looks good, but I have some concerns - my recollection is that despite being unnecessary for _modern_ browsers, `FastClick` and `ResizeObserver` are required for the (frankly ancient) webkit versions shipping in some of our clients' secure mobile browsers. I don't know offhand the mobile story at all of our clients... if they are all using modern browsers, this should be good to go 👍 :)

lbwexler:
Thanks for flagging these concerns tom.

- Re: Fast Click, its only a potential performance enhancement to pinpad -- which is only used in 1 client apreak p at the moment, and we think not running on any legacy mobile browsers at this point anyway.    Risk is low, because it would not even break a legacy browser, just slightly degrade it.

- Re: ResizeObserver.  We think we are fine there as well,  but will test carefully without it on "Blackberry Access" browser which is still used in a few of our client apps.   If neccessary we will add it back in on that app alone.  Our understanding from Claude,  however, is that ag-Grid already relies on native ResizeObserver.  This is good evidence that we should be fine.
-

PR #4337, opened 2026-04-17
Admin Servers tab auto-selects primary instance on first load

### LINKED ISSUE #4315: Admin console Servers tab should auto-select primary instance on first load
Currently it's unclear if there is any default sort on the admin console listing of server instances, but there is no guarantee that the primary instance will be first in the list, and it's the first record that's auto-selected when the grid loads.  We should auto-select the primary by default - it's the most useful and the best to immediately orient the admin viewer in multi-instance scenarios where the primary is doing unique work.

lbwexler:
## Summary

- Admin Console's Servers tab now auto-selects the primary instance on first load, instead of whichever record happens to sort first. The primary is the most useful default and best orients the admin viewer in multi-instance scenarios.
- Existing user selections are preserved across refreshes via a `!hasSelection` guard. Falls back to first record if no primary is present.

Closes #4315

## Test plan

- [ ] Load the Admin Console Servers tab in a multi-instance deployment - confirm the primary instance is auto-selected
- [ ] Select a non-primary instance and wait for auto-refresh - confirm selection is preserved
- [ ] Load in a single-instance deployment - confirm that instance is selected

### DIFF
--- a/admin/tabs/cluster/instances/InstancesTabModel.ts
+++ b/admin/tabs/cluster/instances/InstancesTabModel.ts
@@ -71,7 +71,10 @@ export class InstancesTabModel extends HoistModel {
             usedPctMax: row.memory?.usedPctMax
         }));
         gridModel.loadData(data);
-        await gridModel.preSelectFirstAsync();
+        if (!gridModel.hasSelection) {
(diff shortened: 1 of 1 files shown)


PeteDarinzo (approved):
Reviewed

PR #4340, opened 2026-04-17
Stop persisting per-view icon in DashContainer state (#4320)

### LINKED ISSUE #4320: Investigate (and likely remove) persistence of icon within dashboard widget state

amcclain:
Closes #4320.

## Summary

Removes per-view `icon` persistence from `DashContainerModel` layout state, and drops the `serializeIcon` / `deserializeIcon` helpers that existed to support it. Icons now come from the `DashViewSpec` on every load, matching `DashCanvasModel` and avoiding stale icons after a viewSpec is updated in code.

Marked breaking since the serialize/deserialize helpers are public exports, though the blast radius is small - no consumers in the framework or any of the sample apps.

## What changed

- `desktop/cmp/dash/container/impl/DashContainerUtils.ts` - removed the `icon` write in `convertGLToState` and the matching read in `convertStateToGL`.
- `desktop/cmp/dash/container/DashContainerModel.ts` - removed the `deserializeIcon` handling in the GoldenLayout `registerComponent` callback. The computed `viewState` getter still tracks `icon`, so an app-driven change to `DashViewModel.icon` still triggers the tab-header redraw at runtime - the override just no longer round-trips through persisted state.
- `icon/Icon.ts` - removed the `serializeIcon()` / `deserializeIcon()` exports. Apps that need the equivalent can use `pickBy(iconEl.props)` and `Icon.icon(json)`.
- `icon/README.md` and `docs/README.md` - removed references to the dropped helpers.
- `CHANGELOG.md` - breaking change entry under `85.0.0-SNAPSHOT` with 🟢 LOW upgrade difficulty.

## Why

Reviewing the code around #4320 surfaced an asymmetry: `DashContainerModel` persisted widget icons, while `DashCanvasModel` did not. Tracing the history, the icon persistence path came in alongside the tab-rename feature (#1558), which pulled `title` and `icon` out of the generic `viewState` bag so that user-driven title renames could be saved without colliding with app-defined keys. At the time, `title` and `icon` were treated as a pair - and the `icon` half of that pair carried forward by analogy.

In practice the two have different profiles:

- `title` has a first-class user action (double-click to rename), so persisting user-entered values is load-bearing.
- `icon` has no user-facing UI. It is part of the widget's identity and belongs on the `DashViewSpec`. Persisting an earlier snapshot means that if an app updates the spec icon in code, users see the old icon until they clear their layout.

`DashCanvasModel`, written later, landed without icon persistence and has served apps well that way. Removing the DashContainer half aligns the two dash types, reduces state surface area, and removes a quiet source of drift between code and persisted state. With that internal user gone, the `serializeIcon` / `deserializeIcon` helpers were stranded and could be dropped from the public API as well.

## Impact on apps

- Apps that set `DashViewModel.icon` imperatively at runtime can still do so - the override drives the tab-header render via the existing reactivity. What changes is that the override is not written to persisted state, so a reload will show the viewSpec icon again. The recommended pattern for dynamic icons (and the one documented for dynamic titles in the dash README) is to drive them from a reaction on observable app state.
- Apps that import `serializeIcon` / `deserializeIcon` from `@xh/hoist/icon` will need to inline the equivalent logic. A grep across Toolbox, Jobsite, DirectLend, and Veracity found no direct consumers.

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] Lint / prettier clean via pre-commit hooks
- [ ] Smoke-test a DashContainer layout in Toolbox: add/rename/remove views, reload, confirm icons render from viewSpec and titles still persist

Hoist P/R Checklist
-------------------

- [x] Caught up with \`develop\` branch as of last change.
- [x] Reviewed for breaking changes, added \`breaking-change\` label + CHANGELOG if so.
- [x] Reviewed and tested on Mobile, or determined not required. (Dash is desktop-only.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,12 @@
 ### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW)
 
 * `DashContainerModel` no longer persists per-view `icon` in its layout state, aligning with
-  `DashCanvasModel`. Icons are now sourced exclusively from the `DashViewSpec` on every load,
-  avoiding stale icons when a viewSpec is updated in code. Apps that set `DashViewModel.icon` at
-  runtime can still do so - the override still drives tab header rendering via the existing
-  reaction, it just no longer round-trips through persisted state. For dynamic icons, drive them
-  from a reaction on observable state (see the "Dynamic Titles" pattern in the dash README).
-* Removed the `serializeIcon()` and `deserializeIcon()` helpers from `@xh/hoist/icon`. These were
-  used internally by the previous `DashContainer` icon persistence path and had no other
-  framework consumers. Apps that reference them directly should inline the equivalent logic
-  (`pickBy(iconEl.props)` to serialize, `Icon.icon(json)` to restore).
+  `DashCanvasModel`. Icons now always come from the `DashViewSpec`. Apps that set
+  `DashViewModel.icon` at runtime still see it render, but the override is no longer saved -
+  drive dynamic icons from a reaction on observable state.
+* Removed the `serializeIcon()` / `deserializeIcon()` helpers from `@xh/hoist/icon`, which
+  existed only to support the above. Apps still needing this can use `pickBy(iconEl.props)` and
+  `Icon.icon(json)` respectively.
 
 ### 🎁 New Features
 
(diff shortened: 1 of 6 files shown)


amcclain:
@lbwexler / @TomTirapani / @cnrudd / @haynesjm42 - I'd like to pick this up for a v86 release - I think it cleans up a source of occasional errors and removes this idea of serializing icons which I _believe_ is overly brittle but also not necessary.  Tagging you all as a group that would be the most likely to know if we should *not* remove this - i.e. if we are using it in a way I'm not appreciating.  No rush on this today, for next week perhaps.

lbwexler:
I think this sounds like a good idea!

TomTirapani:
Sounds good to me 👍

PR #4341, opened 2026-04-18
Fix race condition in FilterChooserModel.updateSelectValueAndBind

Ryanseanlee:
## Summary
  - Moved the outbound `bind.setFilter()` call in
  `FilterChooserModel.updateSelectValueAndBind` to run synchronously instead of being
   deferred behind a `wait()` microtask
  - Fixes persisted filter values being silently dropped when a ViewManager restores
  a view alongside other providers that trigger reactions on the same bind target
  - The `wait()` is retained for the `selectValue` UI update (its actual purpose)

  ## Problem
  When `ViewManagerModel.loadViewAsync` restores a view, all persistence providers
  push state in a single MobX action. `FilterChooserModel` set `this.value`
  synchronously but deferred `bind.setFilter()` behind a `wait()`. Before that
  microtask resolved, the action completed, other reactions fired and changed
  `bind.filter`, the inbound sync reaction overwrote `this.value`, and the staleness
  guard in the `wait()` callback bailed out — permanently losing the restored filter.

Hoist P/R Checklist
-------------------

### DIFF
--- a/cmp/filter/FilterChooserModel.ts
+++ b/cmp/filter/FilterChooserModel.ts
@@ -610,11 +610,23 @@ export class FilterChooserModel extends HoistModel {
     }
 
     /**
-     * Update the select value and bind the filter to the bound model. Runs asynchronously after
-     * selectOptions are set to ensure the component is ready to render the tags correctly.
+     * Update the select value and bind the filter to the bound model. The outbound sync to the
+     * bind target runs synchronously to avoid race conditions during persistence restoration
+     * (e.g. ViewManager). The selectValue update is deferred to ensure selectOptions are set
+     * and the component is ready to render the tags correctly.
      */
     private updateSelectValueAndBind(displayFilters = this.toDisplayFilters(this.value)) {
         const {bind, value} = this;
+
+        // Outbound sync: replace the FieldFilter portion of the bind target's filter with
+        // this model's value, preserving any FunctionFilters installed by other components
+        // (e.g. StoreFilterField). Done synchronously to ensure the bind target is updated
+        // before any deferred reactions (e.g. inbound sync) can observe stale state.
+        if (bind) {
+            const filter = appendFilter(bind.filter?.removeFieldFilters(), value);
+            bind.setFilter(filter);
+        }
+
         wait()
             .thenAction(() => {
                 // No-op if we've already re-entered this method by the time this async routine runs.
@@ -629,14 +641,6 @@ export class FilterChooserModel extends HoistModel {
                         return isFinite(idx) && idx > -1 ? idx : displayFilters.length;
                     }
                 );
-
-                // Outbound sync: replace the FieldFilter portion of the bind target's
-                // filter with this model's value, preserving any FunctionFilters
-                // installed by other components (e.g. StoreFilterField).
-                if (bind) {
-                    const filter = appendFilter(bind.filter?.removeFieldFilters(), value);
-                    bind.setFilter(filter);
-                }
             })
             .linkTo(this.filterTask);
     }


Ryanseanlee:
# FilterChooserModel Race Condition During View Restoration

## The Setup

A common pattern: a `FilterChooserModel` with `bind` pointing at a Store/cube View, persisted via
`ViewManagerModel`. Another `@persist` property on the same model (e.g. a relative date offset) has
a reaction that also calls `bind.setFilter()` on the same target to manage function filters.
Function filters can't go through FilterChooserModel, so a separate reaction is the only option.

## The Bug, Step by Step

### Step 1: ViewManager restores a view

`ViewManagerModel.loadViewAsync` calls all providers in a single MobX action:

```typescript
// ViewManagerModel.ts:642-644
.thenAction(latest => {
    this.setAsView(latest, pendingValue?.token == token ? pendingValue : null);
    this.providers.forEach(it => it.pushStateToTarget());
})
```

### Step 2: FilterChooser provider pushes its state

`pushStateToTarget()` calls `setPersistableState`, which calls `setValueInternal(savedFilter, true)`.
This sets `this.value = savedFilter` synchronously, then calls `updateSelectValueAndBind()`.

### Step 3: The outbound sync is deferred behind wait()

The **old** `updateSelectValueAndBind` captured `value` and deferred everything behind a microtask:

```typescript
// OLD CODE - FilterChooserModel.ts:616-642
private updateSelectValueAndBind(displayFilters = this.toDisplayFilters(this.value)) {
    const {bind, value} = this;
    wait()                          // <-- microtask delay
        .thenAction(() => {
            if (this.value !== value) return;   // <-- staleness guard

            this.selectValue = sortBy(/* ... */);

            // Outbound sync - DEFERRED
            if (bind) {
                const filter = appendFilter(bind.filter?.removeFieldFilters(), value);
                bind.setFilter(filter);
            }
        })
        .linkTo(this.filterTask);
}
```

At this point `this.value = savedFilter`, but `bind.filter` has NOT been updated yet.

### Step 4: Other providers push their state (still in the same action)

Another provider restores a tracked property (e.g. a relative date offset) via `pushStateToTarget()`.

### Step 5: The MobX action completes, deferred reactions fire

The reaction managing function filters observes the property change and calls `bind.setFilter()`
on the same target to add/remove a function filter. This changes `bind.filter`.

### Step 6: The inbound sync reaction fires and overwrites this.value

```typescript
// FilterChooserModel.ts:231-236
this.addReaction({
    track: () => bind.filter,
    run: filter => {
        this.setValue(filter?.removeFunctionFilters());
    }
});
```

`bind.filter` now contains only the function filter (the field filters from the FilterChooser
haven't been applied yet - they're waiting on `wait()`). So `removeFunctionFilters()` returns
`null`, and `this.value` is overwritten to `null`.

### Step 7: The wait() resolves - but the guard bails out

```typescript
if (this.value !== value) return;  // null !== savedFilter --> true --> BAIL
```

`this.value` was changed by the inbound sync in step 6. The guard sees the mismatch and returns.
`bind.setFilter(savedFilter)` never executes. **The restored filter is permanently lost.**

### Timeline Diagram

```
Time -->

MobX Action (thenAction)                    | Reactions fire            | Microtask
--------------------------------------------|---------------------------|------------------
FilterChooser.pushStateToTarget():           |                           |
  this.value = savedFilter                   |                           |
  updateSelectValueAndBind():                |                           |
    captures value = savedFilter             |                           |
    wait() -- schedules microtask -----------|-------------------------->| wait() resolves
                                             |                           |
OtherProvider.pushStateToTarget():           |                           |
  sets dateOffset = restored value           |                           |
                                             |                           |
Action ends                                  |                           |
                                             | dateOffset reaction:      |
                                             |   bind.setFilter(fnFilter)|
                                             |                           |
                                             | inbound sync reaction:    |
                                             |   bind.filter changed     |
                                             |   this.value = null  !!   |
                                             |                           |
                                             |                           | this.value !== value
                                             |                           | null !== savedFilter
                                             |                           | --> BAIL, filter lost
```

## The Fix

Move `bind.setFilter()` out of the `wait()` so it runs synchronously:

```typescript
// NEW CODE - FilterChooserModel.ts:618-646
private updateSelectValueAndBind(displayFilters = this.toDisplayFilters(this.value)) {
    const {bind, value} = this;

    // Outbound sync - NOW SYNCHRONOUS
    if (bind) {
        const filter = appendFilter(bind.filter?.removeFieldFilters(), value);
        bind.setFilter(filter);
    }

    wait()
        .thenAction(() => {
            if (this.value !== value) return;

            this.selectValue = sortBy(/* ... */);
        })
        .linkTo(this.filterTask);
}
```

### Why this works

With the outbound sync running synchronously (inside the same MobX action as `pushStateToTarget`):

1. `this.value = savedFilter` and `bind.filter = savedFilter + existingFnFilters` happen in the
   same action
2. When the action ends and the inbound sync reaction fires, `bind.filter` already contains the
   restored field filters
3. `bind.filter.removeFunctionFilters()` equals `savedFilter`, which equals `this.value`
4. `setValueInternal` hits its `this.value?.equals(value)` early return - a no-op
5. `this.value` is never overwritten, so when `wait()` resolves the guard passes and `selectValue`
   is updated normally

## Alternatives Considered

### 1. Suppress the inbound sync reaction during persistence pushes

Add a flag like `_suppressInboundSync` that's set during `pushStateToTarget` and checked in the
reaction:

```typescript
// Would require something like:
this.addReaction({
    track: () => bind.filter,
    run: filter => {
        if (this._suppressInboundSync) return;  // skip during restoration
        this.setValue(filter?.removeFunctionFilters());
    }
});
```

**Rejected because:**
- Adds mutable flag state that must be carefully managed (set/unset around async boundaries)
- The flag would need to span the `wait()` microtask gap, meaning it would suppress legitimate
  inbound syncs during that window too
- `FilterChooserModel` has no direct knowledge of ViewManager - adding awareness of the
  restoration lifecycle leaks abstraction boundaries
- Fragile: any new code path that changes `bind.filter` during the suppression window would be
  silently ignored

### 2. Remove the wait() entirely

Make the entire method synchronous - set `selectValue` immediately alongside `bind.setFilter()`:

```typescript
// Would become:
private updateSelectValueAndBind(displayFilters = this.toDisplayFilters(this.value)) {
    const {bind, value} = this;
    this.selectValue = sortBy(displayFilters.map(f => JSON.stringify(f)), /* ... */);
    if (bind) {
        const filter = appendFilter(bind.filter?.removeFieldFilters(), value);
        bind.setFilter(filter);
    }
}
```

**Rejected because:**
- The `wait()` exists for a reason: `selectOptions` must be set and rendered before `selectValue`
  references them, otherwise React-Select can't match option values to render tags correctly
- Removing it risks visual glitches or missing tags in the filter chooser UI
- The original author added this deliberately (per the JSDoc), and the UI concern is legitimate

### 3. Make the staleness guard use semantic equality instead of reference equality

Change `this.value !== value` to `!this.value?.equals(value)`:

```typescript
wait().thenAction(() => {
    if (!this.value?.equals(value)) return;  // semantic check instead of reference
    // ...
});
```

**Rejected because:**
- This doesn't actually fix the bug: the inbound sync sets `this.value` to `null` (no field
  filters on bind yet), which is semantically different from `savedFilter`. The `equals` check
  would still fail
- Even if it worked, it would mask the real problem: `bind.setFilter()` is reading `bind.filter`
  at the wrong time (after reactions have modified it)
- The reference equality guard is intentional - it detects re-entry, not value drift

### Why the synchronous outbound sync is the best fix

- **Minimal change**: one code motion, no new flags or state
- **Fixes the root cause**: the race exists because `bind.setFilter()` was unnecessarily deferred.
  Moving it to be synchronous eliminates the timing gap entirely
- **No behavior change for non-race cases**: the outbound sync produces the same result whether it
  runs now or one microtask later in normal (non-restoration) usage
- **Preserves the wait() for its actual purpose**: the UI-only `selectValue` update still defers
  correctly
- **No abstraction leaks**: FilterChooserModel doesn't need to know about ViewManager or
  persistence restoration mechanics

## Why This Is a FilterChooserModel Bug, Not a ViewManager Bug

ViewManager is doing nothing wrong. Restoring all providers in a single MobX action is correct -
that's how you batch related state changes. The bug is that `updateSelectValueAndBind` made a
synchronous promise (`this.value = savedFilter`) and then deferred the fulfillment of that promise
(`bind.setFilter(savedFilter)`) behind a microtask. That's a contract violation within
FilterChooserModel itself - it tells the world "my value is X" but doesn't update the bind target
to match until a microtask later, leaving a window where any observer of `bind.filter` sees stale
state.

The reason it only surfaces via ViewManager is that ViewManager is the most common code path that:

1. Sets FilterChooserModel's value via persistence, AND
2. Changes something else in the same action whose reaction touches the same `bind.filter`

But any code that did those two things in a single MobX action would hit the same race. ViewManager
just happens to be the only place that pattern occurs in practice.

Issue #4085, opened 2025-08-29
Expose ability for `dateInput` to occupy the full width of its parent container


ghsolomon:
Currently, there is no way via props to force a `dateInput` to occupy the full width of its parent container.  (Developers currently need a CSS rule targeting `dateInput`'s internal elements.  It would be nice if specifying a width or flex had the same affect on `dateInput` as other input components which can be made to fill their parent container.

Issue #2421, opened 2021-05-04
DateInput <div> wrapper prevents setting className, styles on outermost element


amcclain:
The desktop DateInput component renders with an outer `<div>` element that does not take the `className` or `style` props passed to the component - https://github.com/xh/hoist-react/blob/d7a09dfb28b0b893565fe666f40af55aaf6f38f9/desktop/cmp/input/DateInput.js#L391

Often classNames or styles are used for layout concerns like padding or margins, which can't be controlled here. Specifically, I might wish to use a class or style to override the padding applied to a dateInput within a toolbar - that's something that happens at the outermost layer.

I would vote for applying at least `className` to the outer div. If needed, could support `inputClassName`, but not sure we really need that.

amcclain:
See example of nested DOM elements:

- 1 anonymous (no class specific) outer div
- 2x BP popover spans
- 1 anon div (again)
- 1 div with all of the expected classnames (including duped up xh-input, FWIW)
- 1 BP input-group div
- the actual text input + BP input-actions div for buttons

<img width="872" alt="Screen Shot 2021-06-30 at 6 51 11 AM" src="https://user-images.githubusercontent.com/225806/126516443-2136fbbe-86e6-43cd-846f-689caca025e7.png">

amcclain:
Reviving this as I just ran into it again.  How about having DateInput take a dedicated `wrapperClassName` prop?

amcclain:
We should confirm if this is still an issue with latest BP library

PR #4346, opened 2026-04-22
Start input-sizing audit: planning doc + opt-in CSS CodeMirror mode

### LINKED ISSUE #3300: Desktop inputs require `width: null` to flex
We have a pattern in Desktop inputs, where we specify a default width in the style props, e.g.
```
width: withDefault(width, 200),
flex: withDefault(flex, null)
```
Unfortunately, this means that a dev has to pass `width: null` in order for `flex: true` to take effect, since the set width takes precedence.

We could consider conditionally applying the default width only if flex is nil.

### LINKED ISSUE #4287: HoistInput default widths conflict with flex-based sizing
## Summary

Several HoistInput components apply a default `width` via `withDefault(width, N)`. This default is applied unconditionally, so when a developer sets a `flex` value to let the component fill available space (e.g. in a toolbar), the hardcoded pixel width wins and the developer must pass `width: null` explicitly to opt out — an unintuitive requirement.

```ts
// Doesn't work as expected — default width overrides flex
picker({flex: 1})

// Workaround — have to explicitly clear the default
picker({flex: 1, width: null})
```

## Deeper Issue: Layout Props on Wrong Element (Picker)

For `Picker` specifically, there is a more fundamental problem. The Picker renders as:

```
<span class="xh-input xh-picker bp6-popover-target">   ← popover wrapper (no styles)
    <button style="min-width: 140px; max-width: 300px; flex: 1 1 0%;">  ← trigger button
```

Layout props (`flex`, `minWidth`, `maxWidth`, `width`) are extracted via `getLayoutProps(props)` and applied to the **inner trigger button** (`Picker.ts:415-416`), but the actual flex child of a toolbar is the **outer popover `<span>` wrapper**. This means:

- The toolbar distributes flex space to the unstyled `<span>`, which shrinks to fit its content
- The button's flex/sizing styles have no effect on toolbar layout
- Even with `width: null`, flex-based sizing won't work because the props are on the wrong element

This likely needs layout props applied to the `popover()` call at line 375 (or its wrapper), not the inner `triggerButton`.

## Affected Components (Default Width Issue)

| Component | Default Width | Location |
|-----------|-------------|----------|
| Picker | 160px | `desktop/cmp/input/Picker.ts:416` |
| Select | 200px | `desktop/cmp/input/Select.ts:803` |
| TextInput | 200px | `desktop/cmp/input/TextInput.ts:158` |
| NumberInput | 200px | `desktop/cmp/input/NumberInput.ts:277` |
| TextArea | 300px | `desktop/cmp/input/TextArea.ts:100` |
| Slider | 200px | `desktop/cmp/input/Slider.ts:115` |

## Suggested Fixes

1. **Default width**: Skip applying the default `width` when a `flex` value is present:
   ```ts
   width: flex != null ? width : withDefault(width, 200)
   ```

2. **Picker layout**: Move layout props to the outer popover wrapper so the component participates correctly in flex layouts. The other inputs listed above should be checked for the same wrapper issue.

amcclain:
Phase 1 of a systematic fix for the long-standing desktop-input sizing issues tracked in #3300, #4287, #4085, #2421.

## What's in this PR

- `docs/planning/input-sizing-audit.md` — inventory of desktop + mobile inputs, root-cause mapping of all four tickets to two shared patterns ("default width wins over flex" and "layout props on wrong element"), draft sizing contract for later insertion into `cmp/input/README.md`, and a six-phase fix plan.
- `desktop/cmp/input/modes/cssMode.ts` — opt-in side-effect import that registers CodeMirror's CSS language mode on hoist-react's CodeMirror instance. Consumers import via `@xh/hoist/desktop/cmp/input/modes/cssMode` so webpack resolves `codemirror` through hoist-react's node_modules, sidestepping a duplicate-CodeMirror issue under `inlineHoist` dev mode (filed as xh/hoist-dev-utils#67). First consumer is the matching toolbox test page.

## What's not in this PR

No library code changes yet — this is intentionally audit + tooling only. The actual fixes follow in subsequent PRs per the plan:

- Phase 2: shared `resolveInputLayout` helper + flex-aware defaults (closes #3300, half of #4287).
- Phase 3: outer-wrapper refactor for Picker / NumberInput / DateInput / SegmentedControl (closes other half of #4287, plus #4085 and #2421).
- Phase 4: migrate pixel defaults from inline style to CSS.
- Phase 5: commit sizing contract into `cmp/input/README.md`.
- Phase 6: mobile parity check.

## Companion PR

Toolbox admin test harness: xh/toolbox#842. Use it to confirm baseline and regression-check each library phase as it lands.

## Decision to confirm before Phase 2

The audit doc surfaces three paths for how we deliver "nice out of the box":

- **A.** CSS defaults, overridable by cascade (closest to current; minimal breakage).
- **B.** Flex-aware default (suppress inline default when flex/minWidth/maxWidth/width:100%/null is present).
- **C.** No pixel default, match MUI/Ant/Chakra/Mantine (largest behavior change).

Recommended: A with B's guard as a safety net. Revisit C in a later milestone if we want full convergence with mobile + library norms.

### DIFF
--- a/desktop/cmp/input/modes/cssMode.ts
+++ b/desktop/cmp/input/modes/cssMode.ts
@@ -1,0 +1,21 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2026 Extremely Heavy Industries Inc.
+ */
+
+/**
+ * Side-effect import that registers the CSS language mode on hoist-react's CodeMirror instance.
+ * Consumers opt-in so the mode file (and its cost) is not bundled into every app.
+ *
+ * Usage:
+ * ```
+ * import '@xh/hoist/desktop/cmp/input/modes/cssMode';
+ * codeInput({mode: 'css', ...});
+ * ```
+ *
+ * Must be imported via `@xh/hoist` so webpack resolves `codemirror` through hoist-react's
+ * node_modules - registering on the same CodeMirror instance the codeInput uses.
+ */
+import 'codemirror/mode/css/css';
--- a/docs/planning/input-sizing-audit.md
+++ b/docs/planning/input-sizing-audit.md
@@ -1,0 +1,143 @@
+# Input Sizing Audit
+
+Planning document for addressing the long-standing "inputs don't size predictably" issues
+tracked across #3300, #4287, #4085, #2421.
+
+A running test page is available at **Admin → Tests → Input Sizing** in Toolbox
+(`toolbox/client-app/src/admin/tests/inputSizing/`). Use it to confirm current behavior,
+validate proposed fixes, and prevent regressions.
+
+## 1. Current state — desktop inputs
+
+Inputs in `/desktop/cmp/input/`, ranked by layout-prop participation:
+
+| Input | Hardcoded default | `flex` passthrough | Layout props land on | Fills under `flex:1`? |
+|---|---|---|---|---|
+| `textInput` | `width: 200` | yes (`style`) | `<input>` via BP | ✅ fills |
+| `textArea` | `width: 300`, `height: 100` | yes (`style`) | `<textarea>` via BP | ✅ fills |
+| `numberInput` | `width: 200` | yes (`style`) | inner `<input>` (BP `numericInput` buries it) | ❌ 200px |
+| `select` | `width: 200` | via style | react-select root | ✅ fills |
+| `slider` | `width: 200` | via style | outer `.xh-slider` | ✅ fills |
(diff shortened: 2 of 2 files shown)


amcclain:
@TomTirapani I am eager to pick this up in our next release - see the toolbox branch for a new admin-client based input testing tool that I think came out very well.  We will need to proceed with care in terms of implementing the plan here - there are no changes yet - but I think it's a good time to finally improve and fix our story around sizing desktop inputs.  

Would love your "input" if you have a chance to look at any point. 😄

TomTirapani:
Finally had time to read this, think it all makes sense and is certainly a worth endeavour! Agree with the assessment the approach A (move default widths to stylesheets rather than inline) makes the most sense for out-of-the-box defaults - think that has the lowest chance of introducing issues in client usages, and is easy to reason about all round.

PR #4350, opened 2026-04-28
Add `pathPrefix` to `PersistOptions` + new `persistOptions()` merge function

amcclain:
## Summary

Two related additions to the persistence layer:

**1. `pathPrefix` on `PersistOptions`** — an inheritable, dot-delimited prefix that prepends to the resolved `path` and concatenates (rather than replaces) when merged through parent `persistWith` chains. Resolves a long-standing gap where setting `path` on a parent model's `persistWith` would silently overwrite the natural per-property/per-aspect path used by `@persist`, `markPersist`, `GridModel`, `PanelModel`, and other consumers, causing collisions when multiple persisted targets shared the same parent provider.

**2. New exported `persistOptions()` function** — public replacement for the static `PersistenceProvider.mergePersistOptions` method. Same merge semantics (later arguments win, type-related keys swap together, `pathPrefix` concatenates), but lives next to the `PersistOptions` interface in `core/persist/PersistOptions.ts` and tolerates `null`/`undefined` arguments to skip them. The old static method now emits a runtime `apiDeprecated` warning and will be removed in v87.

Internal call sites in hoist-react (`HoistBase`, `@persist`, `GridModel`/`ZoneGridModel`/`TabContainerModel`/`GroupingChooserModel`/`FilterChooserModel`/`CardModel`/`FormModel` init-persist) have been migrated to the new function and the spread-into-`PersistOptions` idiom (`{...persistWith, path: 'foo'}`) eliminated in favor of `persistOptions(persistWith, {path: 'foo'})`. This also surfaced and fixed several spots where `pathPrefix` would have been silently dropped by plain spread.

See [`docs/persistence.md`](docs/persistence.md#configuration-inheritance) for the inheritance rules and recommended child-model patterns.

## Hoist P/R Checklist

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (No breaking changes — `mergePersistOptions` still works, just deprecated.)
- [x] Updated doc comments / prop-types, or determined not required. (Updated `docs/persistence.md` and field JSDoc.)
- [x] Reviewed and tested on Mobile, or determined not required. (Framework-level change in `core/persist`; no platform-specific behavior.)
- [x] Created Toolbox branch / PR, or determined not required. (Branch `feat/persistPathPrefix` migrates Toolbox usages to `persistOptions()` — no PR yet.)

## Test plan

- [x] Smoke-tested in Toolbox (inlineHoist mode): app boots, persisted state restores correctly, no console errors/warnings/deprecation messages across multiple pages.
- [x] `tsc --noEmit` clean in both hoist-react and Toolbox.
- [ ] Verify existing `markPersist` / `@persist` flows on a model without `persistWith` still work.
- [ ] Verify a parent model with `persistWith: {prefKey, pathPrefix: 'reports'}` correctly nests `@persist` properties under `reports.<propertyName>`.
- [ ] Verify a `GridModel` inheriting that `persistWith` resolves columns/sort/grouping to `reports.grid.columns` etc.
- [ ] Verify two sibling `GridModel`s with distinct `path` values do not collide.
- [ ] Verify a child custom `HoistModel` constructed via `persistOptions(this.persistWith, {pathPrefix: 'detail'})` resolves to `reports.detail.*`.


amcclain:
👋  Marked as draft as I have looked over the code carefully but have not run and tested it interactively in a running app.

Just hit this persistence footgun in a client app and propose this to alleviate. I believe we've talked about this several times - I hope I am not missing anything big here, it's always possible, but adding a dedicated property seemed like a clean way to let the developer express one of two intents: "use exactly this path" (`path`) vs. "append whatever path you would naturally use to this other path" (`pathPrefix`).

It's teed up in the CHANGELOG for the next release but I would be entirely happy to defer until next-next - we've lived with this for a while, it is not urgent to get it into v85 (and it's non-breaking).

amcclain:
@lbwexler I just got bit by this AGAIN - it's a real footgun - flagging so we can work to get this into the next release

lbwexler:
Thought about this --- looks like a very pragmatic change.  I feel like we should emphasize that apps should manaully concatenate pathPrefix when setting it in a spread (and continue to manually spread).  The introduction of the (optional) merge method to apps seems clunky to me, and less readable.

lbwexler (approved):
Am good with this change, but see my note about preferring to emphasize the spread + append method of setting pathPrefix

PR #4354, opened 2026-04-29
Skip span creation in FetchService when tracing is disabled

lbwexler:
## Summary

- `FetchService.fetch()` now only creates a span (and applies a trace ID to the request) when `XH.traceService.enabled` is true.
- `createSpanConfig()` returns `null` when tracing is off; the core fetch work runs directly without the span wrapper, avoiding span overhead and the `traceparent` header on every request.
- Extracted trace-ID application into a small `withTraceId()` helper.

## Test plan

- [ ] Verify requests include `traceparent` and span timing when traceService is enabled
- [ ] Verify requests skip `traceparent` and span overhead when traceService is disabled
- [ ] Confirm correlation ID and tracking still work in both modes

### DIFF
--- a/svc/FetchService.ts
+++ b/svc/FetchService.ts
@@ -389,7 +389,7 @@ export class FetchService extends HoistService {
 
         // 4) Await underlying fetch and post-process response.
         const ret = await fetch(url, fetchOpts);
-        span.setHttpStatus(ret.status);
+        span?.setHttpStatus(ret.status);
 
         if (!ret.ok) {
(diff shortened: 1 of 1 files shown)


svc/FetchService.ts
cnrudd: can span be null here?

cnrudd (changes requested):

svc/FetchService.ts
cnrudd: yes it can. fixed.

PR #4363, opened 2026-05-06
Client-side metrics service

lbwexler:
## API additions

* New `XH.metricsService` — lightweight client-side service recording named timers and counters with optional tags, debouncing batches to the server's `/xh/recordMetrics` endpoint where they flow into the Micrometer registry. Requires hoist-core >= 40.0 (xh/hoist-core#556).
* `Runner.timer(name, tags?)` and `Runner.counter(name, tags?)` — record on completion of the wrapped fn and automatically attach an `xh.outcome` tag (`success` | `failure`) based on whether the fn threw.

## Behavior changes (minor)

* `Runner.counter()` semantic: previously fired before the wrapped fn ran; now records on completion.


svc/MetricsService.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Lightweight client-side metrics service. Records timers and counters by name and optional tags,
+ * batching to the server where they are registered with the Micrometer registry on hoist-core.
+ *
+ * Apps should generally use the {@link Runner} `counter()` / `timer()` builder methods instead of
+ * calling this service directly.
+ */
+export class MetricsService extends HoistService {
+    override telemetryPrefix = 'xh.client.metrics';
```
amcclain, line 31: Did a catch-up merge from the base branch - this was `override spanPrefix` prior to merge, but looks like that changed to `telemetryPrefix` - hope that's correct, made the change to get the merge in but haven't absorbed the overall change yet.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@
 
 ### 🎁 New Features
 
+* New client-side `MetricsService` (`XH.metricsService`) records named timers and counters with
+  optional tags, debouncing batches to the server's `/xh/recordMetrics` endpoint where they
+  flow into the Micrometer registry. Requires hoist-core >= 40.0.
```
amcclain, line 17: We expecting HC 40+ to be a hard requirement (breaking change)?  I'm assuming so - we can get that into the CHANGELOG as we get closer to a release, just checking on baseline expectations here now.

core/runner/Runner.ts
```diff
@@ -97,6 +120,7 @@ export class Runner {
     // Implementation
     //--------------------------
     private executeWrapped<T>(fn: RunFunction<T>): Promise<T> {
+        fn = this.wrapMetrics(fn);
```
amcclain, line 123: This was also a merge conflict w/wrapLink I believe.  Just calling that out in case this routine is sensitive to ordering of these calls

PR #4364, opened 2026-05-06
Fix GridHScrollbar header/body desync with virtual columns (#4344)

### LINKED ISSUE #4344: enableFullWidthScroll: headers desync from body when using virtual columns
## Description

When `enableFullWidthScroll: true` and `useVirtualColumns: true` are both set on a `GridModel`, horizontally scrolling the grid can cause the column headers to fall out of sync with the body content. The headers either lag behind or stop updating entirely while the body continues to scroll.

## Root Cause

`GridHScrollbar.scrollViewport()` only sets `scrollLeft` on `.ag-center-cols-viewport` (the body):

```typescript
scrollViewport(left: number) {
    this.agViewport.scrollLeft = left;
}
```

It relies on AG Grid's internal mechanism to sync `.ag-header-viewport` with the body. When virtual columns are enabled, AG Grid re-renders columns as they enter/exit the viewport during horizontal scrolling, and this internal header sync can fail — causing the header scroll position to drift from the body.

The reverse direction (`agViewport.addEventListener('scroll', ...)` → `scrollScroller()`) also only syncs the custom scrollbar, without force-syncing the header.

## Suggested Fix

Explicitly set `scrollLeft` on both `.ag-center-cols-viewport` **and** `.ag-header-viewport` in both scroll directions:

```typescript
scrollViewport(left: number) {
    this.agViewport.scrollLeft = left;
    this.agHeaderViewport.scrollLeft = left;
}

agViewport.addEventListener('scroll', e => {
    this.scrollScroller(e.target.scrollLeft);
    this.agHeaderViewport.scrollLeft = e.target.scrollLeft;
});
```

Where `agHeaderViewport` is `viewRef.current.querySelector('.ag-header-viewport')`.

## Reproduction

1. Create a `GridModel` with `enableFullWidthScroll: true`, `useVirtualColumns: true`, and enough pinned + unpinned columns to require horizontal scrolling
2. Scroll horizontally (via the custom scrollbar or mouse/trackpad)
3. Observe that column headers intermittently fall out of alignment with body content

The issue is more likely to appear with rapid scrolling or when the grid has many columns (triggering frequent virtual column mount/unmount cycles).

lbwexler:
## Summary
- Explicitly sync `.ag-header-viewport` scrollLeft alongside `.ag-center-cols-viewport` in both scroll directions in `GridHScrollbar`.
- Fixes header drift seen with `enableFullWidthScroll: true` + `useVirtualColumns: true`.

Fixes #4344

## Test plan
- [ ] Toolbox sample grid with virtual columns enabled - scroll horizontally and confirm headers stay aligned with body.


lbwexler:
Unable to meaningfully reproduce on my mac with 40 columns ands 200 rows and flags in question.  Will try on windows.

lbwexler:
## Testing notes

Verified in Toolbox (admin **Tests → Grid Scrolling**) against a grid configured with `enableFullWidthScroll: true` + `useVirtualColumns: true`, 60 columns, scrolled well past the virtual-column buffer.

**Mechanism:** ag-Grid syncs `.ag-header-viewport` to the body on its *next animation frame*. With many virtual columns that leaves the header a frame behind the body during continuous horizontal scroll, so it visibly drifts. This PR sets the header `scrollLeft` synchronously in the same handler that moves the body, so they can't diverge.

**A/B** — header-vs-body `scrollLeft` measured right after the scroll handler, before ag-Grid's rAF re-sync:

| body scrollLeft | header (without fix) | drift | header (with fix) | drift |
|---|---|---|---|---|
| 631  | 0    | 631px   | 631  | 0 |
| 1262 | 631  | 631px   | 1262 | 0 |
| 1894 | 1262 | 632px   | 1894 | 0 |
| 421  | 1894 | -1473px | 421  | 0 |

Without the fix the header lagged the body by 631–1473px at every step (and a rendered frame showed headers reading 38–59 while the body's left edge still showed column 0). With the fix, zero drift at every position and headers stay aligned with body columns.

Note: the drift is intermittent on a settled grid because ag-Grid's rAF re-sync papers over it once scrolling stops — it's the *during-scroll* state that desyncs, which the explicit sync fixes.

PR #4368, opened 2026-05-13
Security hardening: pin actions, re-enable dep audits, harden claude.yml and MCP

amcclain:
## Summary

A set of low-risk, high-confidence supply-chain hardening changes drawn from
two reviews: a fresh look prompted by the [TanStack npm postmortem](https://tanstack.com/blog/npm-supply-chain-compromise-postmortem),
and a separate security audit pass over `@xh/hoist` + the MCP server. Each
item is independently revertable; commits are split so they can be ack'd one
at a time.

Deliberately **not** included here — these need either a decision or a test
plan and would block the easy wins above:

- `npm publish --provenance` (needs npm trusted-publishing setup + `id-token: write` scoping)
- Separating `yarn install` from `npm publish` so `NPM_TOKEN` is not co-located with `postinstall` execution
- `SECURITY.md` updates re: npm 2FA posture
- Wrapping MCP tool output in untrusted-content envelopes (touches every tool handler)

## Changes

- **Pin `xh/hoist-dev-utils` composite actions to commit SHA** — `deployRelease.yml`
  and `deploySnapshot.yml` previously referenced `@master`, which is mutable.
  Pinned to the current master SHA; Dependabot's existing `github-actions`
  config will propose future bumps as reviewable PRs.

- **Re-enable Dependabot npm PRs with grouping** — was capped at
  `open-pull-requests-limit: 0`. Raised to 5, grouped minor/patch bumps
  separately for prod vs. dev to keep noise manageable. Majors remain
  ungrouped.

- **Gate CI on high / critical production-dep advisories** — `yarn audit --groups dependencies`
  runs on every PR and on the existing weekly schedule. The step fails the
  build on high or critical advisories (exit-code bitmask check); low and
  moderate findings appear in the step log but do not block the build.
  Current baseline is clean.

- **Harden MCP server `instructions` against indirect prompt injection** —
  the hoist-react MCP server pipes doc and JSDoc text verbatim into the
  client's context window. Extended the top-level `instructions` field to
  explicitly tell the client that tool output is reference material, not
  directives. Defense-in-depth; no behavior change in tool handlers.

- **Harden `claude.yml`** — added an `author_association` gate (OWNER /
  MEMBER / COLLABORATOR only); dropped `id-token: write` since the workflow
  does not use OIDC; SHA-pinned `anthropics/claude-code-action@v1`.

## Test plan

- [ ] CI passes on this branch (audit step lands green; lint + CodeQL unchanged)
- [ ] Dependabot picks up the new npm config on its next run (visible in the Insights tab)
- [ ] No release / snapshot publish is exercised by this PR — those workflows are unchanged in behavior, only the action refs are pinned
- [ ] Smoke-test the `@claude` trigger after merge from an OWNER/MEMBER comment to confirm the action still runs without `id-token: write`

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -12,17 +12,28 @@ on:
 
 jobs:
   claude:
+    # Restrict to actors with a trusted relationship to the repo. Without this
+    # gate, any GitHub user who can comment on an issue or PR could invoke a
+    # job that holds contents/issues/pull-requests write permissions.
     if: |
-      (github.event_name == 'issue_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review_comment' && contains(github.event.comment.body, '@claude')) ||
-      (github.event_name == 'pull_request_review' && contains(github.event.review.body, '@claude')) ||
-      (github.event_name == 'issues' && (contains(github.event.issue.body, '@claude') || contains(github.event.issue.title, '@claude')))
+      (github.event.comment.author_association == 'OWNER' ||
+       github.event.comment.author_association == 'MEMBER' ||
+       github.event.comment.author_association == 'COLLABORATOR' ||
+       github.event.issue.author_association == 'OWNER' ||
+       github.event.issue.author_association == 'MEMBER' ||
+       github.event.issue.author_association == 'COLLABORATOR' ||
+       github.event.review.author_association == 'OWNER' ||
+       github.event.review.author_association == 'MEMBER' ||
+       github.event.review.author_association == 'COLLABORATOR') &&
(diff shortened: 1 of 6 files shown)


.github/workflows/claude.yml
```diff
@@ -32,7 +43,9 @@ jobs:
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@v1
+        # Pinned to the commit SHA of the v1 tag at time of pinning.
+        # Bumped by Dependabot when newer versions of the action are released.
```
lbwexler, line 47: 👍🏻

lbwexler (approved):

PR #4381, opened 2026-05-18
Upgrade react-dropzone 10.x → 15.x

lbwexler:
## Summary

Replaces #4378 (dependabot major bump) with the code changes required by react-dropzone's v11–v15 breaking API changes.

- `FileChooser.accept` now takes the v14 `Accept` object form keyed by MIME type (e.g. `{'application/pdf': ['.pdf']}`) — see CHANGELOG breaking-change note.
- Drops use of removed `draggedFiles` render-prop field; drag-active message no longer includes a count.
- Types `FileChooserModel.onDrop` rejected arg as `FileRejection[]`.
- Re-exports `Accept`, `FileRejection`, `FileWithPath` from `@xh/hoist/kit/react-dropzone`.

Closes #4378.

## Test plan

- [ ] `npx tsc --noEmit` passes
- [ ] `yarn lint` passes
- [ ] Toolbox FileChooser example: drag & drop single + multi, click-to-browse, oversize rejection, extension rejection

PR #4385, opened 2026-05-20
Enhancements to FileChooser

### LINKED ISSUE #3570: `FileChooser` should accept a `maxFiles` prop

lbwexler:
Comprehensive redesign of `FileChooser`, building on the react-dropzone `15.x` upgrade already on `develop` (via #4381).

**Companion Toolbox PR:** xh/toolbox#860

### Highlights
- **Configuration moved to `FileChooserModel`** — `accept`, file-size limits, and other options are now set on the model constructor rather than as component props. 
- **Empty & single-file states** — the empty target is click-anywhere-to-browse, with a prompt and an auto-derived hint summarizing accepted types / size limits. In single-file mode (`maxFiles: 1`) a selected file renders as a compact card (type-specific icon, name, size) that doubles as a replace target plus a remove action, rather than a one-row grid.
- **Multi-file state** — once files are selected, a persistent drop target sits alongside the grid so users can keep adding files; placement is set via the `dropTargetPlacement` prop (`left` / `top` / `hidden`), and it disables clearly at the `maxFiles` limit.
- **Customizable display API** — new `emptyDisplay` / `fileDisplay` content props replace the former `targetText` / `showFileGrid`. Default `fileDisplay` is a file grid, or the single-file card when `maxFiles` is 1.
- **New capabilities** — `maxFiles` limit (#3570), `onFileAccepted` / `onFileRejected` callbacks, configurable rejection toasts, `maskOnDrag` / `maskOnDisabled`, and a programmatic `openFileBrowser()`.
- **API changes** — renamed `maxSize`/`minSize` → `maxFileSize`/`minFileSize`; removed `enableMulti`/`enableAddMulti` (use `maxFiles`); 

Also caught the branch up with `develop` and fixed a `filesize` v11 named-import bug. See CHANGELOG for breaking-change details.


amcclain:
@claude can you catch this branch up with the latest changes on `develop`, ensure the CHANGELOG and other merge conflicts are reconciled appropriately, then update this pull request with a description of what changed?

amcclain:
Claude was sick today - it needed his token settings adjusted and will be feeling better soon.  Still, turns out it wouldn't have been able to do what I asked anyway.  Another Claude is working on it now.

amcclain:
Had to run out of the office and forgot to push my work, but I did get this updated locally and reconciled the changelog. Happy to push that tomorrow morning. 

@jskupsik these changes seem very relevant to your current project we should have a look tomorrow.

amcclain:
CleanShot 2026-06-03 at 15 10 06@2x

desktop/cmp/filechooser/FileChooser.scss
```diff
@@ -1,35 +1,21 @@
-/*
```
jskupsik, line 1: We removed the copyright blurb from this file -- do we care about that?

jskupsik:
Toolbox change: https://github.com/xh/toolbox/pull/860

lbwexler:
Looks really great.  

Minor nit on the toolbox display side -- think it would benefit from having the controls to modify it displayed as options on the right.  So just super obvious that they are not part of the control, which would not typically have a toolbar.  

Not a showstopper, but when I was interacting with the toolbar, I would click off of the picker or select controls to dismiss, and then the file chooser would capture that click and come up.  That seems like non-optimal behavior.   In most configurations that less likely to be an issue,  but with the toolbar above it certainly is.  I think we could make the clickable surface limited to just the placeholder icon/text area in center or provide an explicit link.

package.json
```diff
@@ -71,6 +71,7 @@
     "jwt-decode": "~4.0.0",
     "lodash": "~4.18.0",
     "lodash-inflection": "~1.5.0",
+    "mime": "^4.1.0",
```
jskupsik: We want this new library, correct?

desktop/cmp/filechooser/FileChooserModel.ts
```diff
@@ -4,66 +4,143 @@
...
+import {pluralize, withDefault} from '@xh/hoist/utils/js';
+import {createObservableRef} from '@xh/hoist/utils/react';
+import {castArray, concat, filter, isEmpty, keys, fromPairs, map, uniqBy, isFunction} from 'lodash';
+import mime from 'mime';
+import {ReactElement, ReactNode} from 'react';
+import {DropzoneRef} from 'react-dropzone';
+
+export interface FileChooserConfig {
+    /** File type(s) to accept (e.g. `['.doc', '.docx', '.pdf']`). */
+    accept?: Some<string>;
```
jskupsik, line 20: So the old version / api of this took a map of MIME type string -> file type extension. This new API now no longer accepts MIME type, instead it only supports file extensions and then uses a new import `mime` library to find the mime types. Is this intentional?

jskupsik (approved):

PR #4397, opened 2026-05-27
Drop unmaintained micro-deps in favor of Hoist-owned utils

lbwexler:
## Summary

Sweep to remove small, unmaintained transitive deps in favor of Hoist-owned utilities. Behavior preserved at every callsite; original packages credited in source comments under their MIT licenses.

- **`@seznam/compose-react-refs`** (last published 2022) → `composeRefs` in `utils/react/ComposeRefs.ts`. Faithful port: variadic, pairwise WeakMap-cached for ref-identity stability across renders. 17 callsites migrated. TODO left in source noting the React 19 upgrade path (forward cleanup returns + add a `useComposedRefs` hook variant).
- **`clipboard-copy`** (last published 2022) → `copyToClipboard` in `utils/js/ClipboardUtils.ts`. Async Clipboard API with `execCommand` fallback retained for non-secure contexts and secure-context permission/focus failures. 4 callsites migrated. TODO left noting the fallback can be dropped once Hoist requires a secure context.
- **`debounce-promise`** (last published 2022) → `debouncePromise` in `@xh/hoist/promise`. Trimmed to the only mode Hoist uses: trailing-edge debouncer where all calls in a quiet window share a single Promise that resolves with the result of one invocation using the most recent args. Leading-edge and accumulate modes omitted. 2 callsites migrated (desktop and mobile `Select`).

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `yarn lint` (code + styles) clean
- [ ] Manual smoke in Toolbox: composed-ref consumers still mount/unmount cleanly (charts, treemap, tile frame, grids, dash containers, tab switchers, form inputs)
- [ ] Manual smoke in Toolbox: copy-to-clipboard works (`ClipboardButton`, About dialog, grid copy-cell menu, admin badge double-click)
- [ ] Manual smoke in Toolbox: `Select` with `queryFn` still debounces queries correctly while typing


utils/js/ClipboardUtils.ts
```diff
@@ -0,0 +1,57 @@
...
+ * `navigator.clipboard` is unavailable, and for secure-context cases where `writeText()`
+ * itself rejects (no user activation, unfocused document, Permissions Policy denial).
+ * Throws if both paths fail.
+ *
+ * Adapted from the (unmaintained) `clipboard-copy` package by Feross Aboukhadijeh,
+ * https://github.com/feross/clipboard-copy - MIT licensed.
+ *
+ * TODO: consider dropping the execCommand fallback?
+ */
+export async function copyToClipboard(text: string): Promise<void> {
```
TomTirapani, line 23: Is this the only function that we import into apps? The original package exported `clipboardCopy` (this), but also the two `navigator` and `exec` paths

TomTirapani: Sorry disregard - I reread the original package more carefully, only clipboardCopy is added to module.exports 👍

utils/react/ComposeRefs.ts
```diff
@@ -0,0 +1,61 @@
...
+        return composePair(refs[0], refs[1]) ?? null;
+    }
+    return refs.slice(1).reduce<OptRef<T>>((acc, ref) => composePair(acc, ref), refs[0]) ?? null;
+}
+
+const cache = new WeakMap<object, WeakMap<object, Ref<unknown>>>();
+
+function composePair<T>(a: OptRef<T>, b: OptRef<T>): OptRef<T> {
+    if (!a) return b;
+    if (!b) return a;
```
TomTirapani, line 36: I doubt anyone is going to look at this impl, but just wanted to say the early returns here are an improvement over the original :)

TomTirapani:
Looks good! We should remember to add lines to the upgrade notes to help agents migrate the imports

PR #4116, opened 2025-10-20
CodeMirror v6

Ryanseanlee:
Hoist P/R Checklist
-------------------

**Pull request authors:** Review and check off the below. Items that do not apply can also be
checked off to indicate they have been considered. If unclear if a step is relevant, please leave
unchecked and note in comments.

- [x] Caught up with `develop` branch as of last change.
- [x] Added CHANGELOG entry, or determined not required.
- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.
- [x] Updated doc comments / prop-types, or determined not required.
- [x] Reviewed and tested on Mobile, or determined not required.
- [ ] Created Toolbox branch / PR, or determined not required.

**If your change is still a WIP**, please use the "Create draft pull request" option in the split
button below to indicate it is not ready yet for a final review.

> **Pull request reviewers:** when merging this P/R, please consider using a *squash commit* to
> collapse multiple intermediate commits into a single commit representing the overall feature
> change. This helps keep the commit log clean and easy to scan across releases. PRs containing a
> single commit should be *rebased* when possible.

PR #4398, opened 2026-05-27
Add JSON Schema validation to JsonInput via AJV

lbwexler:
## Summary

Adds optional schema validation to `JsonInput` via AJV. Split out of #4116 (CodeMirror v6 migration) so each PR has a focused scope.

- New props on `JsonInput`:
  - `ajvSchema` — JSON Schema object passed to AJV for content validation
  - `ajvOptions` — options forwarded to the AJV constructor
- When `ajvSchema` is provided, validation errors are surfaced as CodeMirror lint annotations alongside the existing jsonlint syntax checks. Error highlights point at the offending property in the source text.
- When `ajvSchema` is omitted, behavior is unchanged from the v6 baseline (jsonlint-only).
- New direct dep: `ajv ~8.17.1`.

## Known limitations / follow-ups

- The AJV validator is currently rebuilt on every `JsonInput` render (`new Ajv()` + `ajv.compile()`). Consumers passing a stable schema reference still pay this cost. Worth memoizing.
- Error positioning uses `text.indexOf(\`\"key\"\`)`, which finds the first occurrence of the property name. If the same key appears in multiple objects, the highlight can land on the wrong one.

Both are acceptable for a first cut but worth tracking.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `yarn lint` clean
- [ ] Manual smoke: JsonInput without `ajvSchema` lints syntax errors as before
- [ ] Manual smoke: JsonInput with an `ajvSchema` flags schema violations (missing required props, type mismatches, `additionalProperties: false`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -563,6 +563,9 @@ detailed, step-by-step upgrade instructions with before/after code examples.
   hamburger menu. Set to `true` to render the current user's initials instead or provide a function
   to render a custom element for the user.
 * Added `AggregationContext` as an additional argument to `CubeField.canAggregateFn`.
+* Added `ajvSchema` and `ajvOptions` configs to JsonInputProps.
+    * `ajvSchema` - Used to validate the input JSON
+    * `ajvOptions` - Options to be passed to Ajv constructor (JSON schema validator)
 * Added `filterMatchMode` option to `ColChooserModel`, allowing customizing match to `start`,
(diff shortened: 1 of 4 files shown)


lbwexler:
this is work done originally by Ryan and Pete --- I just split it off from the (orthogonal) CodeMirror upgrade it was bundled with

PR #4409, opened 2026-06-04
New column chooser

### LINKED ISSUE #4059: Replace Desktop ColumnChooser with new GridEditor
Would like to support, among other things

+ more modern, compact look.
+ drag-and-drop
+ editing pinned state.
+ editing sizing mode, (grouping?, sorting?)

Need to consider agGrid's native offering for reference.

haynesjm42:
Replaces the desktop grid column chooser with a new drag-and-drop implementation, and adds a second, docked side-panel presentation of it.

## What's new

**Drag-and-drop chooser** - columns are presented in true grid order across three zones (pinned-left, unpinned, pinned-right), replacing the old available/displayed two-list layout. Drag within or across zones to reorder, pin, and unpin; toggle visibility via checkbox, double-click, or space. Multiple selected rows - or an entire `ColumnGroup` row - move in a single gesture. Drops honor `hideable`, `movable`, and `lockColumnGroups`: a drop that would split a locked group clamps to the nearest legal position, and refused drags explain themselves in the drag ghost.

**Column Library** (opt-in via `ColChooserConfig.columnLibrary`) - a docked list of all hidden columns, grouped by `Column.chooserGroup`. Drag a column out to show and position it; drag one in to hide it. With the library disabled, hidden columns instead remain inline in their bucket, dimmed.

**Docked panel presentation** (`GridConfig.colChooserPanelModel`) - a resizable side panel that stays open alongside the grid and always auto-commits. Open via `GridModel.showColChooserPanel()`, the new `colChooserPanel` context-menu token, or `colChooserButton({target: 'panel'})`.

Also included: a single filter field spanning all chooser grids, on-demand column description tooltips, and toolbar toggles for column-group / Column Library display that persist browser-local and live-sync across every open chooser.

## API additions

* `GridConfig.colChooserPanelModel` + `ColChooserPanelConfig`
* `ColChooserConfig.columnLibrary` + `ColLibraryConfig` (`collapseGroups`, `libraryWidth`)
* `ColChooserButton.target` - `'popover'` (default) | `'dialog'` | `'panel'`
* `colChooserPanel` grid context-menu token
* `GridModel.showColChooserPanel()`, `GridModel.isColumnHideable()`, `GridModel.isColumnMovable()`
* `IColChooserModel.toggle()`

## Implementation notes

* Model hierarchy is an abstract `ColChooserModel` holding all presentation-agnostic state (the three bucket models, the optional library, drag wiring, and the commit chokepoint), with `ColChooserModalModel` (dialog + popover) and `ColChooserPanelModel` (docked) subclasses. All `@internal`.
* Drop resolution is factored out into `colChooserDropEngine` - a pure module with type-only Hoist imports, no ag-Grid/MobX deps - covered by `colChooserDropEngine.spec.ts` (locked-group clamping, multi-row selections, cross-bucket moves). Note hoist-react has no test runner configured, so the spec is not wired into CI; it runs under bare `tsx`. `.npmignore` now excludes `*.spec.ts` from the published package.
* Shared display options (`showGroups`, `showLibrary`) live on a new `AppContainerModel.colChooserOptionsModel` singleton, persisted browser-local so the toggles roam across grids and live-sync between choosers already on screen.
* `GridModel.getColumn()` and `updateColumnState()` now use indexed lookups rather than recursive tree walks and repeated linear scans.

## Upgrade notes for apps

Called out in the CHANGELOG's Breaking Changes section:

* **`RowDragModule` must be registered.** Hoist does not register ag-Grid modules itself, so apps that maintain an explicit module list (rather than the v76-recommended `AllCommunityModule`) have to add `RowDragModule` - without it the chooser renders but its drag-and-drop silently does nothing. Toolbox needed this; see the paired PR's `Bootstrap.ts`.
* `Column.chooserGroup` now groups columns only within the opt-in Column Library - apps that set it lose grouping until they enable `columnLibrary: true`.
* Custom app CSS that targeted the chooser's former `LeftRightChooser`-based DOM needs updating.

Toolbox: xh/toolbox#861

Hoist P/R Checklist
-------------------

- [x] Reviewed and tested on Mobile, or determined not required. Desktop-only change - the mobile chooser is untouched apart from adding `toggle()` to satisfy `IColChooserModel`.


cmp/grid/GridModel.ts
```diff
@@ -1805,9 +1805,14 @@ export class GridModel extends HoistModel {
             if (!col.resizable || !state.manuallySized) state = omit(state, 'width');
             state = {...state, manuallySized: state.manuallySized ?? false};
 
-            // Remove all metadata other than the id and the hidden state from hidden columns, to save
-            // on space when storing user configs with large amounts of hidden fields.
-            if (state.hidden) state = pick(state, 'colId', 'hidden');
+            // Strip metadata other than id, hidden, and pinned state from hidden columns to save
+            // space when storing user configs with large amounts of hidden fields. Pinned state is
+            // preserved so a column's pin survives being hidden and is restored when it is shown again.
+            if (state.hidden) {
```
lbwexler: I find the lack of symmetry between the three properties confusing.

desktop/cmp/grid/impl/colchooser/ColChooserBucketModel.ts
```diff
@@ -0,0 +1,906 @@
+/*
```
lbwexler, line 1: + should this be @internal, perhaps within `grid/impl/` ColumnChooserModel also @internal as you have it and could go there?
+ could then promote ColumnChooser to `/grid`

lbwexler:
Think it would be nice to call this new public component object 'ColChooser', to match with the existing public `ColChooserConfig` and `GridModel.colChooser`.  Could we just extend that existing API, naming scheme right into this new public component?

appcontainer/ColChooserOptionsModel.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Copyright © 2026 Extremely Heavy Industries Inc.
+ */
+import {HoistModel} from '@xh/hoist/core';
+import {bindable, makeObservable} from '@xh/hoist/mobx';
+
+/**
+ * Shared display options for the grid column chooser, held as a single {@link AppContainerModel}
+ * singleton so a toggle in one open chooser live-syncs to every other chooser on screen. Persisted
+ * browser-local.
```
lbwexler, line 13: 👍🏻

cmp/grid/columns/Column.ts
```diff
@@ -779,7 +780,10 @@ export class Column {
                 suppressMovable: !this.movable,
                 lockPinned: !gridModel.enableColumnPinning || XH.isMobileApp,
                 pinned: this.pinned,
-                lockVisible: !this.hideable || !gridModel.colChooserModel || XH.isMobileApp,
+                lockVisible:
+                    !this.hideable ||
+                    (!gridModel.colChooserModel && !gridModel.colChooserPanelModel) ||
```
lbwexler, line 785: seems. a bit of a shame for code like to have to check both of these

Issue #3918, opened 2025-02-07
Draggable library react-beautiful-dnd deprecated - to be archived on Apr 30, 2025


jskupsik:
A client developer expressed concerns over our drag-n-drop library `react-beautiful-dnd` being deprecated.

https://github.com/atlassian/react-beautiful-dnd/issues/2672 - it appears to have started on Oct 14, 2024 and will be fully archived/readonly on Apr 30, 2025.

It looks like https://github.com/atlassian/pragmatic-drag-and-drop is their newer library, with even an automatic migration: https://atlassian.design/components/pragmatic-drag-and-drop/optional-packages/react-beautiful-dnd-migration/about.

Thoughts?

amcclain:
Resolved on develop for v86: react-beautiful-dnd has been replaced by @hello-pangea/dnd 18.0, a maintained, React 19-ready, drop-in fork (#4410). No Hoist or app API changes - the kit/react-beautiful-dnd exports are unchanged.

PR #4410, opened 2026-06-05
Replace react-beautiful-dnd with @hello-pangea/dnd

lbwexler:
Swaps the now-archived `react-beautiful-dnd` for [`@hello-pangea/dnd`](https://github.com/hello-pangea/dnd), its maintained, API-compatible fork. Addresses #3918.

## Why

- **Retires an abandoned dependency.** `react-beautiful-dnd` was archived in April 2025 and will receive no further updates — security or otherwise. `@hello-pangea/dnd` is the actively-maintained continuation with an identical API.
- **Unblocks React 19.** `@hello-pangea/dnd` 18.0 officially declares support for `react@^18 || ^19`. The old package is capped at React 18, making it one of the hard blockers for an eventual React 19 upgrade.
- **Shrinks the dependency tree.** `@hello-pangea/dnd` drops the `react-redux`/`redux` subtree that `react-beautiful-dnd` carried (net ~30 fewer lines in `yarn.lock`).
- **Supply-chain hygiene, not a CVE fix.** `yarn audit` currently reports 0 vulnerabilities — this is preventative risk reduction (removing a frozen dependency before a future CVE lands in it), not an active-vulnerability patch. React 19 itself is *not* claimed as a security improvement here.

## Changes

- Replaced the `react-beautiful-dnd` dependency with `@hello-pangea/dnd` `^18.0.1`.
- Repointed the kit re-export (`kit/react-beautiful-dnd/index.ts`) at the new package. The runtime API is identical, so the **four consumers are untouched**: desktop & mobile `GroupingChooser`, desktop `DynamicTabSwitcher`, and mobile `ColChooser`.
- One small type adjustment: `@hello-pangea/dnd` types `DragDropContext`'s `children` as *required* (the old bundled `@types` had it optional). Our `dragDropContext` call sites pass content via the element factory's `item` alias rather than a literal `children` key, so the kit marks `children` optional on just that factory's spec — via `SetOptional<DragDropContextProps, 'children'>` (type-fest), the existing Hoist idiom for this. `droppable`/`draggable` are unchanged; they take a required `children` render-prop supplied directly at every call site.

## Note — possibly a temporary step

This is deliberately a low-churn, drop-in swap to remove the dead dependency and unblock React 19 **now**, independent of any larger migration. `@hello-pangea/dnd` is itself a community maintenance fork rather than a strategically-funded library. If we later decide to move off the `react-beautiful-dnd` lineage entirely (e.g. to Atlassian's [Pragmatic drag-and-drop](https://atlassian.design/components/pragmatic-drag-and-drop/)), the kit remains the single integration point to revisit. The kit directory name was intentionally left as-is to keep this diff minimal.

## Testing

- `npx tsc --noEmit` — clean.
- `yarn lint` — clean.
- Smoke-tested in Toolbox (via `startWithHoist` / `inlineHoist`, compiling Hoist from source):
    - **Desktop** — confirmed `@hello-pangea/dnd` is the live library (DOM uses its `data-rfd-*` namespace; the old `data-rbd-*` namespace is gone). `GroupingChooser` reorders correctly via `onDragEnd` (verified a drag flipping `City › Win/Lose` → `Win/Lose › City`, with the grid re-grouping to match). `DynamicTabSwitcher` mounts and renders its tabs as draggables. No console errors.
    - **Mobile** — drag-and-drop verified working as-is.


TomTirapani:
Looks good and seems like a pragmatic first step 👍

kit/react-beautiful-dnd/index.ts
```diff
@@ -5,10 +5,17 @@
...
 import {elementFactory} from '@xh/hoist/core';
-import {DragDropContext, Draggable, Droppable} from 'react-beautiful-dnd';
+import {DragDropContext, DragDropContextProps, Draggable, Droppable} from '@hello-pangea/dnd';
+import {SetOptional} from 'type-fest';
 
 export {DragDropContext, Droppable, Draggable};
 
-export const dragDropContext = elementFactory(DragDropContext),
-    droppable = elementFactory(Droppable),
+// Sourced from @hello-pangea/dnd, a maintained, React 19-ready, drop-in fork of the now-archived
```
cnrudd, line 11: I think this comment should be moved up to above line 8.

Issue #4205, opened 2026-01-15
Upgrade to React 19


lbwexler:
React 19 now out for about a year.  We wanted to let this one bake in, but seems appropriate to take a look now.

lbwexler:
Opened a draft PR for the React-18-safe prep work identified in this assessment: #4416 ("Final Prep for React 19").

It lands the three forward-compatible changes — `findDOMNode` removal, Blueprint `Popover` → `PopoverNext`, and the mobile `react-popper` → `@floating-ui/react` swap — without bumping React off 18, so it can merge to `develop` independently. The remaining items (`@types/react@19` / peer bump + `resolutions`, and the optional `forwardRef` cleanup) stay deferred to the actual React 19 migration.

PR #4416, opened 2026-06-07
Prep for React 19

lbwexler:
React-18-safe preparation for an eventual React 19 upgrade. Every change here is forward-compatible — it behaves identically on React 18 and just removes an API/component that React 19 drops or breaks. No runtime bump, so this can merge to `develop` independently of any React 19 timeline.

### Changes

* **`findDOMNode` removal** — `HoistInputModel.domEl` now resolves its element directly from `domRef` rather than falling back to the React-19-removed `findDOMNode`. A trace of all 25 desktop + mobile `HoistInput` implementations confirms each already roots `domRef` on a DOM element (host element, Blueprint `forwardRef` control, or react-onsenui's forwardRef-to-host chain), so the fallback was defensive legacy.
* **Blueprint `Popover` → `PopoverNext`** — the `kit/blueprint` `Popover` wrapper now renders Blueprint's Floating UI-based `PopoverNext` (React 19 compatible) instead of the legacy `findDOMNode`-based `Popover`. It converts our existing `PopoverProps` API via Blueprint's `popoverPropsToNextProps` helper (maps `position`/`modifiers`/`minimal`/`boundary` and preserves the legacy `shouldReturnFocusOnClose` default), so the public `popover` factory API is unchanged and no call sites need edits. (`Overlay` was already migrated to `Overlay2`.)
* **Mobile `Popover`: `react-popper` → `@floating-ui/react`** — swapped `usePopper` for `useFloating`, reusing the same Floating UI library Blueprint already pulls in (deduped to a single copy rather than adding a new dependency). Removed `react-popper` as a direct dependency and dropped the obsolete Popper.js-specific `popperOptions` prop.

### Deferred to the actual React 19 migration (not in this PR)

* `forwardRef` → ref-as-prop — deprecated but still works in React 19, so it's not a blocker; and ref-as-prop is React-19-only, so it can't be done while on 18.
* `@types/react@19` / peer-range bump and the Blueprint / react-onsenui peer-dep `resolutions` overrides.

### Testing

Verified in toolbox via `startWithHoist` (React 18.2.0 / Blueprint 6.15.0): the desktop app-menu, GroupingChooser, and FilterChooser popovers plus the Select input / filter lifecycle all work with zero console errors or warnings. The mobile app builds and loads cleanly on the Floating UI swap (no module/resolution errors), but the mobile Popover's open/positioning was not exercised through the toolbox UI (its only consumer is the app-menu, which isn't surfaced in the current mobile screens) — worth a manual mobile spot-check.

See `docs/planning/react-19-upgrade.md` for the full plan. Refs #4205.


lbwexler:
Tested the migrated mobile Popover (Blueprint `PopoverNext` + Floating UI) in Toolbox mobile on this branch — popovers open, position, and dismiss correctly. Working fine here on `react-19` (React 18).

Note: mobile popovers did *not* appear to work on the stacked `react-19-upgrade` branch (actual React 19 bump) — investigating that separately.

lbwexler:
closing this PR-- simpler to focus on the one that includes the actual upgrade

PR #4432, opened 2026-06-17
Preserve router5-plugin-browser augmentation in emitted .d.ts

### LINKED ISSUE #4431: Router augmentations from router5-plugin-browser not exposed in published .d.ts
## Problem

Consuming apps get a TypeScript error when using methods that `router5-plugin-browser` adds to the router via module augmentation:

```
error TS2339: Property 'buildUrl' does not exist on type 'Router'.

    const url = XH.router.buildUrl('reportSessionId', {...});
```

This also affects `matchUrl`, `replaceHistoryState`, and `lastKnownState`, all of which are merged onto the base `Router` interface by `router5-plugin-browser/types/index.d.ts`:

```ts
declare module 'router5/types/types/router' {
    interface Router {
        buildUrl(name: string, params?: {[key: string]: any}): string
        matchUrl(url: string): State | null
        replaceHistoryState(name: string, params?: {[key: string]: any}, title?: string): void
        lastKnownState: State
    }
}
```

The methods exist at runtime (the plugin is installed and used in `RouterModel`), but downstream apps can't see them in the type system.

## Root Cause

In `appcontainer/RouterModel.ts`:

```ts
import browserPlugin from 'router5-plugin-browser';
```

This is a default-import of a value. `browserPlugin` is never referenced in any exported type, so when `tsc` emits `RouterModel.d.ts` (`emitDeclarationOnly: true`) the import is stripped from the declaration output. Apps consuming the published `@xh/hoist` therefore never load the `.d.ts` that contains the `declare module` augmentation, and `XH.router.buildUrl(...)` shows as missing.

Apps running against local hoist-react source (e.g. via `yarn startWithHoist`) don't see the error, because the original `.ts` import is present and the augmentation is loaded. That's why this only surfaces in apps consuming the published package.

## Suggested Fix

Add a side-effect import that survives `.d.ts` emit, so downstream programs pick up the augmentation transitively. A small dedicated file at the hoist-react root is the cleanest option:

```ts
// router5-augmentations.d.ts
import 'router5-plugin-browser';
```

The current `tsconfig.json` already has `include: ["./**/*"]`, so this file is copied into `build/types/` on publish. Any app that imports anything from `@xh/hoist` will then have the augmentation in its type graph.

Alternatives considered:
- Add `import 'router5-plugin-browser';` alongside the existing default import in `RouterModel.ts` — works, but is easy to remove as "redundant" in a future cleanup.
- Inline the augmentation as a hand-written interface extension in hoist-react — duplicates the plugin's types and risks drift.

## Current Workaround

Affected apps can add a side-effect import to their own code (e.g. in `Bootstrap.ts`):

```ts
import 'router5-plugin-browser';
```

Or list it in `tsconfig.json`:

```json
{ "compilerOptions": { "types": ["router5-plugin-browser"] } }
```

cnrudd:
## Summary

- Downstream apps consuming the published `@xh/hoist` types got `TS2339: Property 'buildUrl' does not exist on type 'Router'` (and similarly for `matchUrl`, `replaceHistoryState`, `lastKnownState`).
- Root cause: `RouterModel.ts` imported `router5-plugin-browser` only as a default value, so `tsc` stripped it from the emitted `.d.ts` and the plugin's `declare module 'router5/types/types/router'` augmentation never reached consumers.
- Added a side-effect `import 'router5-plugin-browser';` (alongside the existing default import) so the augmentation flows through to the published declaration output.

Fixes #4431.

## Test plan

- [x] Build hoist-react (`rm -rf build/types tsconfig.tsbuildinfo && tsc --build`) and confirm `build/types/appcontainer/RouterModel.d.ts` now contains `import 'router5-plugin-browser';`.
- [x] Confirm a downstream `XH.router.buildUrl(...)` call type-checks against the rebuilt output.
- [ ] CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## 87.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Fixed downstream TypeScript error `Property 'buildUrl' does not exist on type 'Router'`
+  (and similarly for `matchUrl`, `replaceHistoryState`, `lastKnownState`)
+  when consuming the published `@xh/hoist` declaration output. The `router5-plugin-browser` module
+  augmentation was being stripped from `RouterModel.d.ts` because its only import was a value-only
+  default that `tsc` removes when emitting declarations; a side-effect import now preserves it.
+
 ## 86.0.1 - 2026-06-16
 
 ### 🐞 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,7 @@
 
 ### 🐞 Bug Fixes
 
-* Fixed downstream TypeScript error `Property 'buildUrl' does not exist on type 'Router'`
-  (and similarly for `matchUrl`, `replaceHistoryState`, `lastKnownState`)
-  when consuming the published `@xh/hoist` declaration output. The `router5-plugin-browser` module
-  augmentation was being stripped from `RouterModel.d.ts` because its only import was a value-only
-  default that `tsc` removes when emitting declarations; a side-effect import now preserves it.
+* Ensure publication of `router5-plugin-browser` TS module augmentation.
 
 ## 86.0.1 - 2026-06-16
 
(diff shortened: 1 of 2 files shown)


cnrudd:
Using this hoist-react snapshot of this branch: [87.0.0-SNAPSHOT.1781709683673](https://www.npmjs.com/package/@xh/hoist/v/87.0.0-SNAPSHOT.1781709683673) in a non-wrapper toolbox checkout, I confirmed that the fix in this PR resolves the issue.

Issue #4057, opened 2025-08-01
`Select.menuWidth` not respected when used in an editor


amcclain:
Will update with repro case - we saw this in client app that had long values in options list, unable to constrain via `menuWidth` prop

Might be a grid editor thing or maybe a general select issue

amcclain:
https://github.com/xh/hoist-react/blob/86f27b9a26412b98a34b9024495c30c42c76ff6b/desktop/cmp/grid/editors/SelectEditor.ts#L39

PR #4452, opened 2026-06-24
Fix windowed Select/SelectEditor menu constrained to control width

### LINKED ISSUE #4325: Windowed selectEditor menu constrained to cell width
When enableWindowed: true is set on a SelectEditor's inputProps, the dropdown menu is constrained to the width of the cell. Without enableWindowed, the menu auto-sizes to fit its content and can grow wider than the cell. The root cause is a width: "100%" hardcoded in the react-windowed-select library that creates a circular CSS dependency.

SelectEditor applies default menu styles at SelectEditor.ts:33-41:

```
 rsOptions: {
     styles: {
         menu: styles => ({...styles, whiteSpace: 'nowrap', width: 'auto', minWidth: '100%'})
     }
 }
```

With standard (non-windowed) react-select, this works correctly: each option element renders with its intrinsic text width, whiteSpace: 'nowrap' prevents wrapping, and width: 'auto' on the Menu div causes it to expand to fit the longest option. minWidth: '100%' ensures the menu is at least as wide as the cell.

Why windowed mode breaks this: The react-windowed-select library (v3.1.2) replaces the default MenuList with a virtualized list using react-window. In node_modules/react-windowed-select/dist/MenuList.js:94, the VariableSizeList component is created with: width: "100%"

This creates a circular CSS dependency:
 1. The Menu div has width: auto — its width is determined by its children
 2. The react-window List has width: 100% — its width is determined by its parent
 3. Neither defines an intrinsic width, so the menu collapses to minWidth: '100%' (= the control/cell width)

In the non-windowed case, option elements have intrinsic text widths that push the auto-width Menu wider. In the windowed case, the 100%-width List acts as an opaque container that doesn't propagate child widths upward.

Workaround: Setting a fixed pixel width via rsOptions.styles.menu in inputProps bypasses the circular dependency. Because inputProps is spread after SelectEditor's defaults (SelectEditor.ts:43), user-provided rsOptions replace the defaults:

```
 editorProps: {
     inputProps: {
         enableWindowed: true,
         rsOptions: {
             styles: {
                 menu: s => ({...s, whiteSpace: 'nowrap', width: 300, minWidth: '100%'})
             }
         }
     }
 }
```

 This is not ideal since it requires a hardcoded width rather than auto-sizing to content.

lbwexler:
## Summary

Windowed (`enableWindowed`) `Select` and `SelectEditor` dropdown menus were constrained to the width of the control/cell, truncating option text, whereas standard (non-windowed) menus auto-size to fit their content.

**Root cause:** virtualized menus can't auto-size via CSS. `react-window` absolutely-positions its rows at `width: 100%`, so option text never widens the menu and it collapses to the control width (the `width: auto` / `minWidth: 100%` menu style resolves to the control width). This is inherent to virtualization, not specific to `react-windowed-select` - the v5 upgrade in #4382 did not resolve it.

**Fix:** restore content-based auto-sizing by measuring the widest option label up front (canvas `measureText`, the same technique used by the grid's column autosize) and applying an explicit pixel width to the menu in windowed mode. No custom `MenuList` needed - `react-windowed-select` keeps handling the virtualization.

## Changes

- `Select.ts` - compute `windowedMenuWidth` from measured option labels (when options change) and apply it as the menu width in windowed mode. An explicit `menuWidth` prop and app-provided `rsOptions.styles` still take precedence; `minWidth: '100%'` floors the menu at the control width. Falls back to control width when labels aren't measurable strings (e.g. fully custom `optionRenderer` content).
- `SelectEditor.ts` - its `width: 'auto'` menu workaround now applies only in non-windowed mode; otherwise it would override the measured width and re-collapse the menu.

## Testing

Verified in Toolbox (running against this branch via `--env inlineHoist`): on the windowed Select demo with a narrow control, the menu now grows to fit long option labels while still virtualizing (scrollbar present). `tsc --noEmit` and `eslint` clean.

Fixes #4325
Related: #4057


haynesjm42 (commented):
I think this might need to account for a custom `optionRenderer` - at least to not apply the calculated width in that case. Looks like the calculation is specifically accounting for the default option renderer.

Is there an approach more like the column autosizing where we find the longest label and then render it (with `formatOptionLabel`) and measure that rendered content width? Or do we just want to leave it up to the developer if they are providing a custom `optionRenderer` to also handle the width?

lbwexler:
Good catch, thanks @haynesjm42. Reworked the sizing to use the same render-and-measure approach as the grid's column autosize (`ColumnWidthCalculator`), so a custom `optionRenderer` is handled correctly rather than measuring the label string:

- Each option is rendered to its actual menu markup (via the configured `optionRenderer` when provided, `renderToStaticMarkup`) and canvas-estimated, the widest sample is then measured for real `clientWidth` in a hidden probe. So icons / multi-element option content size the menu correctly.
- Falls back to the control width if an option can't be statically rendered (e.g. a renderer relying on context/hooks).

On your "render the longest and measure it" point - that's exactly what it does now. I considered ranking candidates cheaply by label text to bound the work, but per your instinct it's no cheaper-per-item than the grid already accepts (the grid renders every value too), and a Select only sizes one menu once per options-change vs. the grid autosizing many columns over thousands of records - so it now renders all options to rank, like the grid.

Pulled the logic out into `impl/CalcWindowedMenuWidth.ts` to keep `Select.ts` focused. Verified in Toolbox (windowed Select with long labels + a custom `optionRenderer`, narrow control): menu grows to fit content while still virtualizing.

haynesjm42 (approved):

PR #4454, opened 2026-06-24
Add `lossless` option to `fmtQuantity`

### LINKED ISSUE #4219: The `fmtQuantity` util should have a `lossless` option
Hoist ships with a `fmtQuantity` formatter that will, based on the scale of the number, swap to formatting in millions or billions. 

https://github.com/xh/hoist-react/blob/9a6d8ec1f7abab4bbbee9acf15d6427f4fc96984/format/FormatNumber.ts#L269

This is often desirable when displaying large or highly variable quantities.  The issue with this approach however is that it does not tell the user if there is extra precision they are missing, and in some cases losing that precision (quietly) could be very problematic.  E.g. `fmtQuantity(7_100_000)` and `fmtQuantity(7_100_100)` will both output `7.10m`.

In some display contexts this might be acceptable, but in others it is not. 

We should add support for a new option - `lossless` is one proposed name - that will apply the compact formatting in the first case, where `7.10m` is a full accurate representation of the number, but not in the second case, where the trailing precision + `lossless` should force it to be rendered as `7,100,100`.

Additionally, when `lossless` is not applied and the number rendered might not match the actual underlying value, we should default `tooltip: true` to ensure that, out of the box without relying on a dev to specify, the actual value is accessible in some form to the user.

lbwexler:
Adds a `lossless` option to `fmtQuantity` (issue #4219). When set, the formatter compacts to millions / billions units **only** when doing so loses no precision; otherwise it renders the full value at its natural precision.

- `7,100,000` → `7.10m` (exact, compacts)
- `7,100,100` → `7,100,100` (would lose the trailing `100`, so renders full)
- `123.456` with `lossless` → `123.456` (no longer rounded to `123`)

### Changes
- New `lossless` opt on `QuantityFormatOptions` (default false).
- Extracted the auto/fixed precision resolution out of `buildFormatConfig` into a shared `parsePrecision` helper, reused by the compaction-exactness check.
- Compaction decision uses the standard display precision as its threshold; `lossless` independently drives full-precision rendering when not compacting.

Closes #4219

### Not changing the default tooltip
The issue also floated defaulting `tooltip: true` for non-`lossless` lossy renders. Skipped: that would silently add hover tooltips to every existing `fmtQuantity` call site - a broad, app-wide behavior change disproportionate to an opt-in feature. `lossless` already surfaces the lost precision inline, and `tooltip: true` remains available per-call.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 87.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+
+* Added a `lossless` option to `fmtQuantity` to compact values to millions / billions units only
+  when doing so loses no precision, rendering the full value otherwise (e.g. `7,100,100` stays
+  `7,100,100` rather than collapsing to `7.10m`).
+
 ## 86.1.0 - 2026-06-22
 
 ### 🎁 New Features
--- a/format/FormatNumber.ts
+++ b/format/FormatNumber.ts
@@ -14,6 +14,7 @@ import {
     isNil,
     isNumber,
     isString,
+    isUndefined,
     round
 } from 'lodash';
 import Numbro from 'numbro';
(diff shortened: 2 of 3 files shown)


amcclain:
Thanks for landing this - Toolbox formatter demo updated @ https://github.com/xh/toolbox/commit/625e7279fd1a9be35d709eba1c93b928cb3e6ed4

lbwexler:
Had a pr for this with same name.   Will look to see if anything worth
adding

PR #4457, opened 2026-06-25
Add popover mode to FilterChooser; deprecate PopoverFilterChooser (#4084)

### LINKED ISSUE #4084: popoverFilterChooser target does not occupy space when open
When `popoverFilterChooser` is open, the underlying target `filterChooser` is omitted, causing other elements to shift.  Instead, it should always occupy the same space in the DOM.

Additionally, there is no way to clear an existing filter (by clicking on an "X") without first opening the popover. This requires 2 clicks in a bit of a confusing way. The same is true for opening the favorites menu. A user clicks the favorite icon, then the popover moves to its "open" state, and then the user must click again to bring up the favorites popover.

lbwexler:
Closes #4084.

Folds the popover behavior into `FilterChooser` as a first-class `popover` mode backed by a **single** real `select`, replacing the `PopoverFilterChooser` wrapper's duplicate-control design (two `filterChooser`s sharing one `inputRef`, an `omit` hack, and a `disabled` target) that was the root cause of all three reported symptoms.

### What changed
- **`filterChooser({popover: true})`** renders the control collapsed in-place and expands it into a popover when opened. New `popover?: boolean | FilterChooserPopoverOptions` prop (`position`, `width`).
- Only the interactive instance binds `model.inputRef`, so the collapsed trigger and expanded content mount together without contention - no more `omit`.
- The trigger is made non-interactive via react-select `menuIsOpen`/`isSearchable` instead of `disabled`, keeping its clear and favorites controls live. The expand popover skips opening on clicks that land on those controls (`elemWithin`), and favorites is gated so only one menu shows at a time.
- `PopoverFilterChooser` is now a thin `@deprecated` alias delegating to `filterChooser({popover: true})`; to be removed in a future major.

### Fixes
- Collapsed control no longer disappears on open - surrounding layout no longer shifts.
- Clear and favorites are now single-click from the collapsed state.

### Migration
- Admin `ActivityTrackingPanel` and `RolePanel` updated to the new mode.

Verifying behavior in the browser before marking ready for review.


desktop/cmp/filter/PopoverFilterChooser.ts
```diff
@@ -5,100 +5,23 @@
...
-                        if (open) {
-                            impl.open();
-                        } else {
-                            impl.close();
-                        }
-                    }
-                })
-            });
+        render(props) {
+            return filterChooser({popover: true, ...props});
```
TomTirapani, line 25: Great that we're keeping this (deprecated) alias around for now

TomTirapani (approved):
Looks good - great change 👍

amcclain:
> Clear and favorites are now single-click from the collapsed state.

Nice pickup on this one, thanks for the improvement!

PR #4458, opened 2026-06-25
Support generic record types across the data + grid layer (#3691)

### LINKED ISSUE #3691: Support `Store` generic types to spec types for `StoreRecord.data` and `StoreRecord.id`
It would be excellent if we could spec an interface for the data element of all records within a given `Store`, so you would have type-aware support when reading their data in code. 

Have not thought it through in detail, but I am assuming `StoreRecord` would take a generic and then `Store`, `GridModel`, etc. would have some way to pass it though. And of course it would need to come out the other side of a variety of `StoreRecord`-emitting APIs. But I believe it would be a big win, given how common it is to pass records around through an application.

Being able to tell TS that your `id`s will be either numbers or strings (as opposed to either) could also eliminate some common boilerplate casting.

amcclain:
Closes #3691.

Adds **opt-in** generic type parameters `<T, Id>` across the data and grid layer, so apps can declare a record's data shape once and get typed access wherever records surface.

### What you get
- `new Store<Person>({...})` / `new GridModel<Person>({...})` → typed `store.records[i].data`, `gridModel.selectedRecord.data`, and `record.id`.
- `record.id` is typed `Id` (auto-derived from the interface's `id` via `RecordId<T>`, or pass it explicitly: `Store<Trade, string>`), so ids flow into typed service APIs without casts.
- Parameterized: `Store`, `StoreRecord`, `StoreSelectionModel`, `GridModel`, `DataViewModel`, `ZoneGridModel`, `UrlStore`, `RestStore`, `RestGridModel`.

### Design
- **Fully opt-in / zero-breakage**: every param defaults (`T = PlainObject`, `Id = RecordId<T>`), so existing untyped code compiles and behaves exactly as before.
- `record.data` is typed as the **bare field shape `T`** - consistent with Hoist's "id is not a field" model. `id` is the top-level `record.id`, not a member of `data`.
- `T` appears in **output positions only** (the write API stays `PlainObject`), keeping `Store<Person>` covariantly assignable to plain `Store` - so all internal framework consumers compile unchanged.
- Internal impl (`RecordSet`, validators) stays monomorphic; the public read getters cast at the boundary.

### Demos
- Migrated admin `RoleModel` (`GridModel<RoleGridRow>`, drops `record.data as HoistRole` casts on the selection path) and `ClusterObjectsModel` (`GridModel<ClusterObjectRecord>`).
- A companion Toolbox PR migrates the Portfolio / Todo / Recalls examples.

### Not in scope (deliberate)
- Column renderer/editor callbacks and `RecordAction` callbacks still receive an untyped `StoreRecord` - typed callback context is a possible future follow-on. The demo migrations honestly retain casts there.

### Verification
- A compile-time type-assertion fixture (`data/__typetests__/dataGenerics.typetest.ts`) guards typed data, derived/explicit id, the zero-breakage cast, and covariance.
- `tsc --noEmit` and `yarn lint` clean; docs (data + grid READMEs, CHANGELOG, doc registry) updated.


amcclain:
Companion Toolbox PR exercising these generic data-layer types in the example apps: xh/toolbox#874

PR #4466, opened 2026-06-29
Fix grid column filter Custom tab handling of multi-value filters

### LINKED ISSUE #3522: Grid filter custom tab inputs are unable to read an array of values
The grid filter custom tab's inputs are breaking when they receive an array of values and enableValues is false. The custom tab currently reads a compound filter but should be able to read and display array filters and emit array filters where possible, such as an OR filter with multiple equals clauses or an AND filter with multiple not equals clauses. This would make it behave more closely to the other filtering methods: FilterChooser, context menu, and the grid filter's values tab. Steps to reproduce the problematic behavior plus screenshots are below. 

For dates, if an = or != filter is set via the FilterChooser or the context menu, opening the column header's custom tab shows an empty date input but with an X to clear the blank selection. Clicking on the calendar shows an error. 
DateInput_1
DateInput_2

For numbers, setting 2 or more = or != filters from the FilterChooser will display NaN in the column header's custom tab's number input. 
NumberInput

For strings, setting 2 or more = or != filters from the FilterChooser will display a comma-separated list of values in the column header's custom tab. 
String_1
String_2

lbwexler:
Fixes #3522

### Problem
The grid column filter's **Custom tab** mapped each `FieldFilter` to a single row, so a multi-value `=`/`!=` filter (as produced by `FilterChooser`, the column context menu, or the Values tab) dumped an array into a single-value number/date/text input - showing `NaN`, a broken date picker, or a comma-joined string. Most visible when `enableValues` is `false`, since the Custom tab is then the only tab.

### Fix
- Multi-value clauses are expanded into one row per value on read, and same-field/op rows are recombined into a single multi-value filter on commit - round-tripping to the representation used by the other filtering methods.
- Filters that genuinely can't be shown as flat rows under a single op (nested single-field trees, not producible via the header UI) are detected and left untouched with a notice, rather than silently corrupted.
- The input-selection decision is centralized in a shared `usesMultiValueInput` helper used by both `CustomRow` and `CustomTabModel`, closing a latent case where numeric/date fields with `enableValues` would still receive an array.

### Testing
Verified live in Toolbox (Grids › Column Filtering) running this branch via `inlineHoist`, plus a clean `tsc --noEmit`.

- **Multi-value `=` on a numeric column** (`trade_volume = [2942400000, 5445000000]`, no Values tab): Custom tab now expands the array into one `numberInput` per value showing the actual numbers (previously a single input → `NaN`), with the **OR** join toggle correctly derived. Commit collapses the rows back to a single `{op: '=', value: [...]}` - lossless round-trip confirmed via "View Grid Filters".
- **Multi-value `!=` on the same column**: expands to one "Not equals" row per value, join toggle correctly derived as **AND**, and commits back to a single `{op: '!=', value: [...]}`.
- **Multi-value on an enum column** (`City`, has a Values tab): the shared `usesMultiValueInput` helper keeps this on a single multi-select row rather than expanding it - no spurious per-value rows.
- No console errors during any interaction; the "too complex to edit" notice path was not reachable through the header UI (as expected - such filters aren't producible there).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## 87.0.0-SNAPSHOT - unreleased
 
+### 🐞 Bug Fixes
+
+* Hardened the grid column filter's Custom tab against filters it previously mishandled - multi-value
+  clauses are now expanded into editable rows and recombined on commit, and filters it cannot
+  represent are left untouched rather than corrupted.
+
 ### ⚙️ Typescript API Adjustments
 
 * Retyped `GridModel.colChooserModel` as the new cross-platform `IColChooserModel` interface,
--- a/desktop/cmp/grid/impl/filter/headerfilter/custom/CustomRow.ts
+++ b/desktop/cmp/grid/impl/filter/headerfilter/custom/CustomRow.ts
@@ -12,7 +12,7 @@ import {dateInput, numberInput, select, textInput} from '@xh/hoist/desktop/cmp/i
 import {Icon} from '@xh/hoist/icon';
 import {kebabCase} from 'lodash';
 
-import {CustomRowModel} from './CustomRowModel';
+import {CustomRowModel, usesMultiValueInput} from './CustomRowModel';
 
 /**
  * Row with operator and value combination for CustomTab.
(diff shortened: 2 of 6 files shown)


TomTirapani:
Looks good 👍

PR #4472, opened 2026-07-02
Fix quick-filter/Find returning different results when grouping changes (#4070)

### LINKED ISSUE #4070: Changing grouping (groupBy) when a filter is active can result in incorrect filter results
Best explained by this screen recording:

https://github.com/user-attachments/assets/ac11949a-49b0-4241-b07a-8728e191f8a4

lbwexler:
Fixes #4070

### Problem
`StoreFilterField` and grid Find derived their searched-field set partly from `GridModel.groupBy`, so grouping by a **hidden** column silently added that field to the search. An active filter/query then returned different records across regroupings.

Reproduced on `develop` (Toolbox Standard grid, `winLose` hidden by default) — filtering `loser`:
- Ungrouped → **0** rows (`winLose` hidden, not searched)
- Group by Win/Lose → **438** rows (`winLose` now grouped, so it's searched)
- Ungrouped again → **0** rows

### Fix
Removed the `groupBy`-based inclusion from `getActiveFields()` in both `StoreFilterFieldImplModel` and `GridFindFieldImplModel`. The searched field set now derives from visible columns (plus explicit include/exclude) only, so it stays stable regardless of grouping. `GridFindField` still tracks `groupBy` in its reaction to keep result ordering aligned with the grid's visual row order.

### Note
Trade-off: quick-filter/Find no longer matches on a *hidden* dimension's values while grouped by it (its values are visible as group headers). This is the intended, stable behavior per discussion.

Separately noticed (not addressed here, likely warrants its own issue): ag-Grid's default un-group behavior re-shows a column that was configured `hidden`, and Hoist's `columnState` isn't reconciled — so e.g. `winLose` becomes visible on the grid after a group→ungroup cycle while `GridModel.isColumnVisible()` still reports it hidden.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### 🐞 Bug Fixes
 
+* Fixed `StoreFilterField` and grid Find so an active quick-filter or find query no longer returns
+  different results when the grid's `groupBy` changes.
 * Fixed `Select` to correctly handle non-primitive (object) values: selected-option matching and
   async query de-duplication now use deep equality, so object values no longer render as
   `[object Object]` or collide with one another.
--- a/cmp/store/impl/StoreFilterFieldImplModel.ts
+++ b/cmp/store/impl/StoreFilterFieldImplModel.ts
@@ -66,7 +66,7 @@ export class StoreFilterFieldImplModel extends HoistModel {
 
         this.addReaction(
             {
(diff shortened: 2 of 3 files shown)


cnrudd (approved):
Seems pretty clean.   But what was the original motivation for including groupBy in find and filter?  what are we losing by removing it?  Were we trying to insure that groups whose field we were searching on that had no matching children would still end up in the results because the value of the grouping field matched?

PR #4487, opened 2026-07-10
Title admin User Data tab as "User Data & Roles" when roles are enabled

amcclain:
The admin console's **User Data** tab hosts the Roles editor, but its title gave no hint of that — creating recurring friction around where roles are managed.

### Changes

- Title the tab **"User Data & Roles"** and show its **Roles** sub-tab only when the app runs the Hoist-provided `DefaultRoleService` (i.e. role management is enabled). When disabled, the title stays **"User Data"** and the previously-dead Roles sub-tab — which only rendered a *"Default Role Module not enabled"* message — is omitted entirely.
- Determine this via a single `roleAdmin/config` query, loaded once at admin init by `AppModel` and cached on a new `AppModel.roleModuleConfig` property (10s timeout, swallowed on failure so a transient error defaults the tab to showing Roles rather than hiding it).
- `RoleModel` now reads the config from `AppModel` (single load point) instead of fetching its own copy, asking `AppModel` to reload only if the init load didn't complete. `RolePanel` gates on `gridModel` for the brief post-load window where the config is known but the grid is still being built.

Route and in-code ids are unchanged.

### DIFF
--- a/admin/AppModel.ts
+++ b/admin/AppModel.ts
@@ -7,9 +7,8 @@
 import {GridModel} from '@xh/hoist/cmp/grid';
 import {TabConfig, TabContainerModel} from '@xh/hoist/cmp/tab';
 import {ViewManagerModel} from '@xh/hoist/cmp/viewmanager';
-import {CallContextLike, HoistAppModel, HoistRoute, InitContext, XH} from '@xh/hoist/core';
+import {HoistAppModel, HoistRoute, InitContext, XH} from '@xh/hoist/core';
 import {Icon} from '@xh/hoist/icon';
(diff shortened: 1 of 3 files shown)


amcclain:
I know this is kinda overkill but I always have this persistent mental friction when I go to find Roles (which for many apps  is an absolutely core admin function) in our top-level tabset.

lbwexler:
Great change -- I struggle myself to find it.  

I was able to simplify the handling of the (hopefully very rare) unhappy path that the config won't load.  Hope that makes sense to you.  Still will never bring down the admin app.

PR #4499, opened 2026-07-22
Typed-config admin editor + CodeInput line styling

lbwexler:
Enhances the Admin Console config editor to show more than one view of a config's value when relevant. JSON configs backed by a typed class and/or an active instance-config override now render a tabbed editor — **Resolved** (effective value with typed defaults applied), **Instance Value** (the override), and the editable **DB Value** — with the explicitly-set keys highlighted against defaults. Plain configs are unchanged.

Also includes two small framework additions used by the feature:
- `CodeInput.lineStyles` — apply custom CSS class(es) to specific lines.
- `RestForm` — a `Revert` button (enabled when dirty); `Save Changes` renamed to `Save`.

Requires hoist-core v41+ for the resolved/override data; falls back to the standard editor against older servers.


lbwexler:
Screenshot of the tabbed editor in action - `slackAlertConfig` has both a typed class and an active instance override, so all four views apply. The Resolved tab shows the effective value with non-default keys highlighted against the muted code defaults; the overridden DB Value tab label is struck through:

config-editor-tabs-v2

lbwexler:
CR: PETE

Issue #4500, opened 2026-07-22
Store: record `data` objects fall into V8 dictionary mode — per-store codegen'd literal factory measures 6-8× smaller records at typical widths


amcclain:
## Summary

`Store.parseRaw()` constructs each record's `data` object via `Object.create(this._dataDefaults)` followed by per-property assignment of non-default values. Empirical testing against current V8 shows this construction route causes record `data` objects to be **demoted to dictionary ("slow") properties mode once a record has more than ~15–20 populated fields** — which describes most real-world grid records. Dictionary-mode objects cost ~30–33 bytes per populated field and are excluded from inline-cache optimization on every downstream `data[field]` read.

A per-store **codegen'd object-literal factory** — buildable at Store construction time, since the field configuration is fully known then — measured **6–8× smaller** per record at typical store widths, with fast-properties mode preserved (shared hidden class, linear property slots, working ICs). For a store holding tens of thousands of records this is a triple-digit-MB difference, plus a CPU win for grids, sorting, filtering, and Cube aggregation.

The win has a hard V8 boundary at ~100 fields (details below), so the strategy should be selected per-store at construction: factory for stores with ≤ ~100 fields, current sparse-prototype approach for wider/sparser stores where it genuinely wins.

## Current implementation (v86.2.0 refs)

- `data/Store.ts` — `parseRaw()`: `const ret = Object.create(this._dataDefaults)` + `forIn(...)` assigning each non-default parsed value as an own property.
- `data/Store.ts` — `createDataDefaults()`: shared prototype holding every field's `defaultValue`.
- `data/Store.ts` — `parseUpdate()`: uses **`delete ret[name]`** when a value returns to its default.
- `data/StoreRecord.ts` — ctor adds `data.id = id`; `finalize()` optionally `Object.freeze`s `data`; doc comment on `data` describes the own-props-for-non-defaults / defaults-via-prototype design.

The sparse-prototype design is genuinely good for stores whose records populate few of many defined fields (defaults cost 0 B/record). The problem is purely the **construction route**: growing an object by assignment crosses V8's dynamic-add limit almost immediately.

## V8 mechanics, with citations

[...]

PR #4501, opened 2026-07-22
Store: add `useFixedDataShape` config - opt-in memory reduction for records populating 20+ fields

amcclain:
Supersedes #4500.

Adds `Store.useFixedDataShape`, an opt-in memory optimization for stores whose records populate 20 or more fields. Off by default; no change to any existing app on upgrade.

```typescript
const store = new Store({
    fields: [...],
    useFixedDataShape: true
});
```

Also settable app-wide via `Store.defaults.useFixedDataShape`.

### Why

By default `parseRaw()` builds each record's `data` onto `Object.create(defaults)`, creating an own property only for fields holding a non-default value. Two consequences: records **differ in shape** from one another depending on which fields they populate, and each is built by a run of *keyed* property assignments. Past ~20 of those, V8 stops tracking the shape and demotes the object to a per-object hashtable ("dictionary mode") — roughly 24 bytes per property instead of 8, plus slack.

When enabled, `Store` clones each `data` from a shared template carrying every Field, so every record has one identical, fixed shape. Writing a property that already exists is not an *add*, so the demotion never happens. That shape uniformity is the mechanism, and it is what the config name refers to.

### When to enable

**The criterion is a single number: do records populate 20 or more fields?** That is where V8 demotes, and it holds at any field count — there is no upper limit past which it stops working.

"Populated" means fields carrying a non-default value on a typical record, *not* fields declared on the Store. For dense data the two are the same; for sparse data the distinction is the whole story:

| Store shape | Effect of enabling |
|---|---|
| 10 fields, 10 populated | No change — the representations are equivalent |
| 25 fields, 25 populated | Substantially less memory |
| 150 fields, 150 populated | Substantially less memory |
| 150 fields, 25 populated | Modestly less memory |
| 150 fields, **8 populated** | **More memory — do not enable** |

Narrow records are a no-op, not a regression. The one losing case is wide-and-sparse: the template pays for every *declared* field on every record, while the default representation costs nothing for unpopulated ones.

`Store` checks the setting against its data once after first load and logs guidance — a warning if enabled where records populate too few fields, a debug hint if disabled where the data would qualify. **Advisory only; the setting is always honored.** `Store.recordDataMode` reports the representation in use, and is known at construction.

### Measured

Toolbox admin grid tester (xh/toolbox#879), Chrome with `--js-flags="--expose-gc"`. 50,000 records, 157 declared fields, **144 populated** (measured, not assumed), mixed value types, streaming load, all other Store flags at Hoist defaults.

Cold load, N=1 on a fresh page per row — V8 decides property storage per *isolate*, so measuring both representations in one session contaminates whichever runs second. Two replicates per side:

| `useFixedDataShape` | Heap Δ | Bytes/record |
|---|---|---|
| off | 453.4 MB / 453.2 MB | 9,069 / 9,064 |
| on | 322.1 MB / 327.4 MB | 6,441 / 6,548 |

**128.6 MB saved — 1.40x, a 28.4% reduction in whole-app heap growth for one load.** Run-to-run spread is 0.2 MB with the flag off and 5.3 MB with it on, both far inside the effect.

This whole-app figure *understates* the effect on record data itself, which shrinks several-fold — record `data` is only part of what a load retains, alongside raw source objects, string payload and the grid row model. Figures are from a dev build and are valid as a comparison between rows, not as an absolute "your app will save X".

**The saving is independent of what the fields contain.** Re-measured across four value distributions at the same record shape:

| Value mix | off | on | Saved |
|---|---|---|---|
| numeric | 387.1 MB | 254.8 MB | 132.3 MB |
| mixed | 453.3 MB | 324.7 MB | 128.6 MB |
| categorical | 511.3 MB | 386.5 MB | 124.8 MB |
| unique strings | 566.4 MB | 441.7 MB | 124.7 MB |

The absolute saving holds within a 7.6 MB band while the baselines themselves span 179 MB. That is what the mechanism predicts: the flag removes dictionary-mode *property* overhead, which is per-property and independent of the value stored in it. Practically, the saving can be estimated from populated fields × record count alone — roughly 18 bytes per populated field per record — without knowing anything about the data.

### Relationship to the other memory flags

Measured at the same shape, `useFixedDataShape` and `retainRaw` are **independent and additive** — they shrink different things, so neither depends on the other. Values are Chrome heap deltas for one cold load of 50k records, `valueMix: mixed`:

| Config | Heap Δ | vs baseline |
|---|---|---|
| baseline (all defaults) | 453.3 MB | — |
| `retainRaw: false` | 296.3 MB | −157.0 MB |
| `useFixedDataShape: true` | 324.7 MB | −128.6 MB |
| both | 171.5 MB | −281.8 MB |

The solo savings sum to 285.6 MB against a measured combined 281.8 MB — a 3.8 MB difference, inside the 5.3 MB spread, so there is no interaction to report either way. Worth stating plainly that **`retainRaw: false` saves more on its own** at this shape; the two are separate levers chosen on different grounds (`retainRaw` on whether the app reads `StoreRecord.raw`, this flag on how densely records are populated).

`freezeData` is memory-neutral: 451.3 MB vs a 453.3 MB baseline.

**`useRawAsData` is mutually exclusive with this flag, and Store now throws if both are set.** That config adopts each raw object as its record's `data` and returns early from `createRecord()` without calling `parseRaw()` — exactly where this optimization does its work — so with both enabled it would silently do nothing while still allocating a per-Store template. The two are alternatives chosen on the shape of the incoming data: `useRawAsData` when objects already arrive in their final form, `useFixedDataShape` when Store must parse and build them. `Store.recordDataMode` reports `'raw'` for the former, and the advisory check stays quiet in that mode rather than recommending a config that would throw.

### Trade-offs to weigh before enabling

- **`data` enumeration changes.** Records carry an own property for every field, so `Object.keys()`, spread and `JSON.stringify()` of `data` include default-valued fields. Reads return the same values either way — use `record.getValues()` or `record.getModifiedValues()` rather than enumerating `data` directly.
- **Memory only.** Field reads are not measurably faster in practice: much of the framework's data access runs through call sites shared across every Store in an app, and those behave the same regardless of how record data was built. No read or construction speed benefit is claimed.
- **V8 only.** Safari and Firefox do not demote objects at these field counts — no benefit there, and no harm.

<details>
<summary>Implementation notes for reviewers</summary>

- **The representation is fixed at construction** (`useFixedDataShape` is `readonly`), so records within a Store can never mix — which matters both for keeping their hidden class shared and for keeping the deep-equal comparisons in `modifyRecords()` and `RecordSet` comparing like with like.
- **The template is deliberately a separate object from `_dataDefaults`.** It is built as `{..._dataDefaults}`, which both converts it to fast-properties mode (the defaults object, built by keyed assignment, is itself in dictionary mode for wider stores) and keeps it distinct from the object used as a prototype on the sparse route — V8 declines its fast clone path for anything that has been used as a prototype.
- **`StoreRecord` assigns `data.id` after the clone**, which *is* a genuine property add. Every record takes that same transition in the same order, so the hidden class stays shared and the optimization holds — the bytes-per-record figures above are measured with it in place.
- **`parseUpdate` on the optimized route needs no `delete`**, since defaults are real own properties there. The default route still uses `delete`: its `parseRaw` creates own properties only for non-default values, so a value reverting to its default would otherwise leave an own-key mismatch against committed data and break revert-to-clean detection in `modifyRecords()`.
- **The advisory check** counts own non-default properties on up to 100 already-parsed records, so it is exact rather than an estimate from raw data, and needs no second `processRawData` pass. It runs once, after first load, and never overrides the app's setting.
- **A field named `__proto__` is now rejected** in `parseFields`, alongside `id`. Assigning it would replace the data object's prototype rather than set a value, on either route. This throws at construction where it previously failed silently.

</details>

<details>
<summary>Validation</summary>

- `npx tsc --noEmit` and `yarn lint` clean.
- In-app A/B in Toolbox at 50k records, per the figures above, exercising cold load, both representations and the advisory check.
- The repo has no test infrastructure, so behavioral parity between the two routes — value equality, default handling, and modify → revert → clean normalization on both routes — has been verified by reading and in-app exercise rather than by an automated harness. That is the area most worth a reviewer's attention.

</details>

### Toolbox

Companion PR xh/toolbox#879 adds an admin grid tester with configurable record width, populated-field density and value distribution, plus a repeatable heap/timing harness. It forces a page reload when either record-data representation is toggled, so each side of an A/B is measured in a fresh isolate. Merge only after this PR merges and a snapshot publishes.

---

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (Opt-in and off by default, so no app-visible change on upgrade. The `data` enumeration difference is documented on the feature entry.)
- [x] Created Toolbox branch / PR, or determined not required. (xh/toolbox#879)


lbwexler:
Great stuff. Two thoughts:

- Let's gate this via `xhStoreExperimental` for initial release, so it can be toggled via soft-config without a build (vs. the static kill switch).
- On CSP: the silent fallback means the same app gets different `data` shapes across environments. Not right away, but in a future major we should consider requiring `unsafe-eval` and failing fast at startup. For now maybe just a one-time warn when the factory declines due to CSP.

lbwexler:
One more consideration: sparse data.

- Narrow stores (≲15 fields) are fine either way — legacy objects never demote, both representations stay fast-mode and similarly sized.
- The concern is stores *under* the 100-field cutoff whose records populate only a handful of fields. The true crossover variable is *populated* props per record, not store field count: below V8's ~20-own-prop dictionary demotion, the legacy sparse object stays fast-mode at ~10B per populated field, and the factory's slot-per-field becomes a memory regression (probed: 100-field store / 10 populated → 128 B/rec legacy vs 833 factory). Field count is only a proxy — population is a data property, unknowable at Store construction.
- The PR's sparse row (60/25) is already past the demotion cliff, so the table never shows this regime. Suggest benchmarking sub-20-populated widths, documenting the density assumption, and possibly a per-store opt-out for known-sparse stores.

amcclain:
Reworked — description rewritten to match. Summary of what changed since review:

@lbwexler both of your points drove this.

**Sparse data** — you were right that the crossover variable is populated props per record, not store field count, and that it can't be known at construction. The field-count cutoff is gone. Store now samples up to 200 raw records on first load and declines below a mean of 20 populated fields, logging the decision; `Store.recordDataMode` reports the outcome. Confirmed your regime measures as you expected: at 80 fields / 6 populated the old cutoff let it through at ~52 → ~350 B/rec.

**Gating via `xhStoreExperimental`** — done, and it's now the sole way in. Off by default, per-store via `experimental: {optimizeRecordData: true}` or app-wide via the soft-config.

**CSP** — resolved by removing the codegen rather than warning around it. `{...template}` measures byte-identical to the compiled factory (604 B/rec at 58 fields, both), so the codegen was only buying construction time. With it gone there's no `unsafe-eval` dependency, no environment-dependent fallback, and so no possibility of an app seeing different `data` shapes across environments — which removes the need for the fail-fast decision you floated for a future major.

One correction to figures I posted earlier in this thread: I quoted a 23x sparse regression, which was a contaminated measurement. Real figure is ~6.7x, and some read numbers I gave were measured on Node and did not hold up in Chrome. Everything in the description is Chrome, fresh-page-per-route.

amcclain:
Superseded by #4521, which carries this work forward on the `store-fixed-data-shape` branch.

Reopening as a fresh PR rather than continuing here: the implementation, the config name and the baseline it merges into have all changed enough that the history on this PR is noise rather than context. The codegen'd factory approach discussed above is gone, `optimizeRecordData` is now `useFixedDataShape`, and `develop` has since gained `useRawAsData`, which interacts with it.

@lbwexler your two review points here both landed, and are worth restating since they drove the shape of what is now in #4521 - the sparse-data concern is documented as an explicit "do not enable" case with the criterion stated as *populated* fields rather than declared, and the CSP problem is gone entirely with the codegen removed. #4521 also notes what changed since, including that the config is now a first-class `StoreConfig` rather than gated behind `xhStoreExperimental`.

PR #4503, opened 2026-07-22
Cube: build ViewRowData row objects via per-View codegen'd assigner - ~5-6x smaller rows

### LINKED ISSUE #4502: Cube: ViewRowData row objects fall into V8 dictionary mode - per-View codegen'd assigner measures ~5-6x smaller rows at typical view widths
## Summary

Follow-up to #4500, extending the same analysis to the Cube/View subsystem. Every row a `View` produces gets a `ViewRowData` whose field values are installed via **dynamic keyed-store loops** - the same construction route that demotes objects to V8 dictionary ("slow") mode. Class instances get slightly more headroom than plain objects: demotion sets in between **20 and 30 view fields** (F=20 stays fast, F=30 is dictionary-mode for every instance in steady state). Wide views in data-intensive apps - exactly the ones built over large cubes - pay full price.

Measured on the same harness as #4500: a per-View **codegen'd named-property assigner** keeps `ViewRowData` instances in fast-properties mode with shared hidden classes - **~5-6x smaller rows** and order-of-magnitude faster row construction, with `instanceof`, prototype getters, and the entire updates machinery untouched.

These rows are heavily retained - `View._rowCache`, `_leafMap`, `result.rows`, and as `StoreRecord.raw` in every View-bound store - so for a 50k-leaf, 30-field view this is ~83MB → ~17MB of `ViewRowData` per view, before counting aggregate rows. Dashboard-style apps run several views. Note #4500 already covers the Cube's internal `Store` and the bound stores' own `data` objects; the `ViewRowData` population is what remains.

## Current implementation

- `data/cube/row/LeafRow.ts` - ctor: `view.fields.forEach(({name}) => this.data[name] = rawRecord.data[name])` - one leaf per cube record in the view.
- `data/cube/row/BaseRow.ts` - `initAggregate()` (Aggregate + Bucket rows): `view.fields.forEach(({name}) => data[name] = null)` + `Object.assign(data, appliedDimensions)`, then `computeAggregates()` rewrites values in place.
- `data/cube/row/BaseRow.ts` - `initAggregate()` also builds `canAggregate` per aggregate row via keyed `reduce` into `{}` - same dictionary fate at ≥~30 fields (smaller population, worth fixing in passing).
- Per-row shape: `id` + `cubeRowType`/`cubeLabel`/`cubeDimension` + F fields + `children` (+ `cubeBuckets` when bucketed, `_cubeLeafChildren` with `provideLeaves`).
- Value-rewrite paths (`computeAggregates`, `applyLeafDataUpdate`, `applyDataUpdate`) only overwrite existing properties - rewrites never demote, so the updates machinery needs no changes.

## V8 mechanics

Background and citations in #4500. The specific mechanism here, verified by probe: V8's fast-properties limit applies to keyed stores that **create new map transitions**; constant-key (named) stores are exempt. A `new Function`-compiled assigner with literal property keys (`d["fieldA"]=v[0];...`) therefore keeps class instances fast with a shared hidden class through ~120 total properties.

A notable probe artifact worth preserving: keyed stores that **follow existing transitions** do not demote - once one object completes the shape fast, subsequent identically-shaped objects can stay fast even via keyed loops. Real Views never get this benefit (no instance ever completes the chain fast), but it cleanly confirms the mechanism.

## Measurements

Node 24 / V8 13.6, 50k leaf-style rows, forced-GC heap deltas, fresh field-name chains per experiment (mode verified via `%HasFastProperties` on first/mid/last instance):

| view fields | today (dict) B/row | assigner (fast) B/row | ratio | build 50k rows |
|---|---|---|---|---|
| 30 | 1,658 | 337 | **4.9x** | 76 → 5ms |
| 57 | 3,192 | 553 | **5.8x** | 107 → 18ms |

Aggregation-style reads (per-field sweeps over rows, the `SumAggregator` pattern) run ~1.55x faster against fast-mode rows (800 → 516ms, 10 sweeps x 30 fields x 50k rows). Read gains are smaller than #4500's grid case because cube reads are keyed-by-variable-name (megamorphic) rather than fixed-name monomorphic.

Steady-state mode by construction route (all 50k instances):

| route | F=20 | F=30 | F=50 | F=80 |
|---|---|---|---|---|
| `ViewRowData` + keyed loop (today) | FAST | dict | dict | dict |
| + codegen'd named assigner | FAST | FAST | FAST | FAST (to ~120) |

## Proposed change

Build on #4500's `RecordDataFactory` (same module, shared guards):

1. Add `RecordDataFactory.createAssigner(fieldNames)` - compiles `(target, values) => void` with literal keys; returns null under the same guards (field count over threshold, `__proto__` field name, no `unsafe-eval`, internal kill switch).
2. `View` compiles one assigner per query (recompiled on `updateQuery` when fields change) and exposes it internally to its rows.
3. `LeafRow` ctor and `BaseRow.initAggregate()` use the assigner when present - each keeps its current loop as the fallback; `initAggregate` seeds values as `appliedDimensions.hasOwnProperty(name) ? appliedDimensions[name] : null`, folding away the separate `Object.assign`.
4. `canAggregate` built via the same assigner (boolean values).
5. Late named adds (`children`, `cubeBuckets`, `_cubeLeafChildren`) unchanged - probe-verified shape-safe.

No public API changes. `ViewRowData` remains a class with its getters and index signature; updates flow untouched.

## Validation plan

1. Headless: extend #4500's probe suite - assigner-built rows FAST + shared maps at 30-120 fields (incl. post-`children`/`cubeBuckets` adds), value parity vs the legacy loops, guard behavior (threshold/CSP/`__proto__`).
2. In-app (Toolbox cube tester): query round-trips with dimensions and buckets, leaves included/excluded, `updateDataAsync` deltas flowing up through aggregates, bound-store loads, `getDimensionValues`.
3. Regression: `omitFn`/`lockFn` paths, `provideLeaves`, row-cache reuse across repeated queries, `updateQuery` with changed field sets (assigner recompile).

amcclain:
Fixes #4502. **Stacked on #4501** (`record-data-factory`) - review just this PR's single commit; merge after #4501 lands (GitHub will retarget to `develop` when its branch is deleted).

**TLDR:** Cube `View`s build one `ViewRowData` per leaf and aggregate row via dynamic keyed-store loops - the same V8 dictionary-mode trap #4501 fixed for `Store` records, with the cliff at ~20-30 view fields. These rows are heavily retained (`_rowCache`, `_leafMap`, `result.rows`, and as `StoreRecord.raw` in every bound store). Populating them through a per-View compiled **named-property assigner** keeps them fast: **~5-6x smaller rows** (1,658 → 337 B/row at 30 fields; 3,192 → 553 at 57), ~10x faster row construction, ~1.55x faster aggregation-style reads. For a 50k-leaf, 30-field view: ~83MB → ~17MB, per view. No API changes; wide-field and strict-CSP environments transparently keep the current route.

### Why a different mechanism than #4501

`ViewRowData` is a real class (getters like `isCubeLeaf`, `instanceof` identity) that takes its field properties post-construction - it can't use the literal-object factory. The assigner exploits a narrower V8 rule, verified by probe: the fast-properties limit applies only to keyed stores that *create new map transitions*; constant-key (named) stores are exempt. A compiled `t["fieldA"]=v[0];...` keeps instances fast with a shared hidden class through ~120 properties, and the late `children`/`cubeBuckets` adds and all value-rewrite paths (`computeAggregates`, `applyLeafDataUpdate`/`applyDataUpdate`) are shape-safe - the updates machinery needed no changes.

### What changed

- **`RecordDataFactory.createAssigner(fieldNames)`** - sibling to `create()`, sharing all its guards (≤100 fields, `__proto__` name, CSP feature-detection, internal kill switch).
- **`View`** - compiles one assigner per query, recompiles on `updateQuery`.
- **`LeafRow` ctor / `BaseRow.initAggregate`** - use the assigner when available; each keeps its original loop as the fallback. The per-aggregate-row `canAggregate` map is built via the same assigner, fixing its parallel dictionary-mode fate in passing.

Measurement detail, probe methodology, and the transition-following artifact that initially masked the problem are written up in #4502.

<details>
<summary>Validation</summary>

- 20-assertion headless suite: fast mode + shared maps at scale (50k rows, first/last instance), post-`children`/`cubeBuckets`/`_cubeLeafChildren` adds, keyed value rewrites never demote, plain-object `canAggregate` targets, guards (threshold/CSP/`__proto__`/kill switch), exotic field names, exact leaf and aggregate value parity vs the legacy loops.
- In-app against the live Toolbox Cube tester (20k facts): aggregate sum parity before/after streaming updates, `canAggregateFn` semantics preserved (`quantity` null at fund level, aggregated at symbol level), dimension switching with assigner recompile, leaves mode with exact leaf-vs-cube-record parity, legacy-path A/B in the same session (identical behavior incl. row-identity on updates), clean console.
- Note: in real (Babel) builds, `ViewRowData`'s bare class-field declarations emit as `undefined` own props, so row shapes carry 7 base props rather than the 5 in TS-stripped simulations - confirmed compatible; doesn't change the cliff or the fix.

</details>

---

Hoist P/R Checklist
-------------------

- [x] Caught up with `develop` branch as of last change. (Stacked on #4501, which is.)
- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (None - `ViewRowData` shape/API unchanged; row field props simply present for all view fields, as before.)
- [x] Created Toolbox branch / PR, or determined not required. (Validated against existing Cube tester - no toolbox changes needed; xh/toolbox#879 covers the grid-side testbed.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@
   (defaults included) on the optimized path - apps relying on `Object.keys(record.data)` or
   spread/`JSON.stringify` of `data` returning only non-default fields should use
   `record.getModifiedValues()` or `record.getValues()` as appropriate instead.
+* Applied the same optimization to Cube `View` row data - `ViewRowData` objects (leaf and
+  aggregate rows) are now populated via a per-View compiled assigner, measured ~5-6x smaller at
+  typical view widths (more than ~20 fields), with faster row generation and aggregation reads.
+  Views with very many fields and strict-CSP environments continue to use the prior route.
 
 ### ⚙️ Technical
 * Moved both desktop and mobile popover implementations off the deprecated, React-18-capped Popper.js
--- a/data/cube/View.ts
+++ b/data/cube/View.ts
@@ -21,6 +21,7 @@ import {
     StoreRecord,
     StoreRecordId
 } from '@xh/hoist/data';
+import {RecordDataAssigner, RecordDataFactory} from '@xh/hoist/data/impl/RecordDataFactory';
 import {ViewRowData} from '@xh/hoist/data/cube/ViewRowData';
 import {action, makeObservable, observable} from '@xh/hoist/mobx';
 import {shallowEqualArrays} from '@xh/hoist/utils/impl';
@@ -131,6 +132,9 @@ export class View
     private _bucketDependentFields = new Set<string>();
     _aggContext: AggregationContext = null;
     _rowCache: Map<string, BaseRow> = null;
+    // Compiled per-query assigner keeping ViewRowData objects in V8 fast-properties mode, or
+    // null to use legacy keyed-assignment loops - see RecordDataFactory.createAssigner.
+    _rowDataAssigner: RecordDataAssigner = null;
 
     /** @internal - applications should use {@link Cube.createView} */
     constructor(config: ViewConfig) {
@@ -140,6 +144,7 @@ export class View
         const {query, stores = [], connect = false} = config;
 
         this.query = query;
+        this._rowDataAssigner = this.createRowDataAssigner();
         this.stores = this.parseStores(stores);
         this._rowCache = new Map();
         this.fullUpdate();
@@ -201,6 +206,7 @@ export class View
         if (oldQuery.equals(newQuery)) return;
 
         this.query = newQuery;
+        this._rowDataAssigner = this.createRowDataAssigner();
 
         // If the cube is changing then we need to clear the row cache, and potentially disconnect
         // from the old cube and connect to the new one
@@ -516,6 +522,10 @@ export class View
         }
 
         return false;
+    }
+
+    private createRowDataAssigner(): RecordDataAssigner {
+        return RecordDataFactory.createAssigner(this.query.fields.map(it => it.name));
     }
 
     private cachedRow<T extends BaseRow>(id: string, children: BaseRow[], fn: () => T): T {
--- a/data/cube/row/BaseRow.ts
+++ b/data/cube/row/BaseRow.ts
@@ -142,28 +142,60 @@ export abstract class BaseRow {
         this.children = children;
         children.forEach(it => (it.parent = this));
 
-        view.fields.forEach(({name}) => (data[name] = null));
-        Object.assign(data, appliedDimensions);
-
-        this.canAggregate = reduce(
-            view.fields,
-            (ret, field) => {
-                const {name} = field;
+        // Install field values (and the parallel canAggregate flags) via compiled assigner when
+        // available, keeping these objects in V8 fast-properties mode - see
+        // RecordDataFactory.createAssigner.
+        const {_rowDataAssigner} = view;
+        if (_rowDataAssigner) {
+            const {fields} = view,
+                ctx = view._aggContext,
+                len = fields.length,
+                dataVals = new Array(len),
+                canAggVals = new Array(len);
+            for (let i = 0; i < len; i++) {
+                const field = fields[i],
+                    {name} = field;
                 if (appliedDimensions.hasOwnProperty(name)) {
-                    ret[name] = false;
+                    dataVals[i] = appliedDimensions[name];
+                    canAggVals[i] = false;
                 } else {
-                    const {aggregator, canAggregateFn} = field,
-                        ctx = view._aggContext;
-
-                    ret[name] =
+                    const {aggregator, canAggregateFn} = field;
+                    dataVals[i] = null;
+                    canAggVals[i] = !!(
                         aggregator &&
                         (!canAggregateFn ||
-                            canAggregateFn(dimOrBucketName, val, appliedDimensions, ctx));
+                            canAggregateFn(dimOrBucketName, val, appliedDimensions, ctx))
+                    );
                 }
-                return ret;
-            },
-            {}
-        );
+            }
+            _rowDataAssigner(data, dataVals);
+            const canAggregate = {};
+            _rowDataAssigner(canAggregate, canAggVals);
+            this.canAggregate = canAggregate;
+        } else {
+            view.fields.forEach(({name}) => (data[name] = null));
+            Object.assign(data, appliedDimensions);
+
+            this.canAggregate = reduce(
+                view.fields,
+                (ret, field) => {
+                    const {name} = field;
+                    if (appliedDimensions.hasOwnProperty(name)) {
+                        ret[name] = false;
+                    } else {
+                        const {aggregator, canAggregateFn} = field,
+                            ctx = view._aggContext;
+
+                        ret[name] =
+                            aggregator &&
+                            (!canAggregateFn ||
+                                canAggregateFn(dimOrBucketName, val, appliedDimensions, ctx));
+                    }
+                    return ret;
+                },
+                {}
+            );
+        }
 
         this.computeAggregates();
     }
--- a/data/cube/row/LeafRow.ts
+++ b/data/cube/row/LeafRow.ts
@@ -38,9 +38,23 @@ export class LeafRow extends BaseRow {
         this.data.cubeLabel = rawRecord.id.toString();
         this.data.cubeDimension = null;
 
-        view.fields.forEach(({name}) => {
-            this.data[name] = rawRecord.data[name];
-        });
+        // Install field values via compiled assigner when available, keeping this data object in
+        // V8 fast-properties mode - see RecordDataFactory.createAssigner.
+        const {_rowDataAssigner} = view;
+        if (_rowDataAssigner) {
+            const {fields} = view,
+                len = fields.length,
+                vals = new Array(len),
+                recData = rawRecord.data;
+            for (let i = 0; i < len; i++) {
+                vals[i] = recData[fields[i].name];
+            }
+            _rowDataAssigner(this.data, vals);
+        } else {
+            view.fields.forEach(({name}) => {
+                this.data[name] = rawRecord.data[name];
+            });
+        }
     }
 
     applyLeafDataUpdate(newRec: StoreRecord, updatedRowDatas: Set<PlainObject>) {
--- a/data/impl/RecordDataFactory.ts
+++ b/data/impl/RecordDataFactory.ts
@@ -86,6 +86,35 @@ export class RecordDataFactory {
         return new RecordDataFactory(fields);
     }
 
+    /**
+     * Compile an assigner that installs the given field names onto a target object via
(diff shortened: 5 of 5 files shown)


haynesjm42:
This looks great - excited to see the effect of the gains in some client apps that have started struggling recently as row counts increased.

I'd like to propose we also handle a use-case this doesn't yet reach (drawn from a real client app - ~175 cube fields, ~35 view query fields for an average connected grid store):

- Cube with **>100 fields**
- View querying **>20 and <100** of them
- Store constructed with **all** cube fields, plus fields for the `ViewRowData` `cube…` properties (`cubeLabel`, etc.)

Here the **View** gets the assigner optimization from this PR, but the connected **Store** never does: at >100 registered fields it's past `RecordDataFactory`'s cap, so it builds no factory and falls back to the legacy `Object.create(defaults)` + keyed-assignment route — i.e. dictionary mode — even though only ~30 fields are ever populated. The store spends nearly all its life in the steady state (cube ticks → `updateData`, no query change), so this is the common case, not an edge.

The fix: since the View knows exactly which fields it populates on every row, it can hand that set to each connected store right before `loadData` during `fullUpdate`. The Store filters the set against its full field list and rebuilds `_dataFactory` as a **projection** — fast-mode own props for just the covered subset — while keeping its full field registry intact for grid/validation/lookup. Because the covered subset is well under the cap and dense, the factory applies and the store's records go fast-mode too. (If the projected subset is itself >100, or under strict CSP, it declines and falls back to the legacy route exactly as today.)

Since a projection factory only installs own props for the covered fields, the remaining fields' defaults are resolved via a **shared prototype on the record object** — the same defaults strategy the legacy sparse route already uses (`Object.create(_dataDefaults)`), just applied to the codegen'd literal via a deliberate `__proto__:` entry. I verified on V8 13.x (Node 25) that a literal with a set `__proto__` keeps its own props in fast-properties mode **with a cross-instance shared hidden class**, that the post-literal `id` add stays shape-safe, and that uncovered fields resolve to their defaults through the prototype. Measured on the described shape (150 registered / 30 populated / 100k records): **~1,632 → ~312 B/rec (5.2x), 163 MB → 31 MB**, defaults preserved.

Three changes:

**1. A projection factory method on `RecordDataFactory`** (defaults via shared prototype):

```ts
/**
 * Create a projection factory: installs `ownFields` as fast-properties own props, with every
 * other field's default resolved via `defaultsProto` - set as the record's prototype. Lets a
 * store whose records populate only a subset of its fields (e.g. one fed by a Cube View whose
 * query selects a subset of cube fields) build records in fast-properties mode, sized to the
 * populated subset. Returns null under the same conditions as {@link create}.
 */
static createProjection(ownFields: Field[], defaultsProto: PlainObject): RecordDataFactory {
    if (
        !RecordDataFactory.enabled ||
        ownFields.length > RecordDataFactory.MAX_FACTORY_FIELDS ||
        ownFields.some(f => f.name === '__proto__') ||
        !isCodegenSupported()
    ) {
        return null;
    }
    return new RecordDataFactory(ownFields, defaultsProto);
}
```

```ts
private constructor(fields: Field[], defaultsProto: PlainObject = null) {
    this.fields = fields;
    this.fieldIndices = new Map(fields.map((f, idx) => [f.name, idx]));
    this.defaults = fields.map(f => f.defaultValue);

    const slots = fields.map(({name}, idx) => `${JSON.stringify(name)}:v[${idx}]`).join(',');
    if (defaultsProto != null) {
        // Projection: a deliberate `__proto__:` literal entry installs the shared defaults
        // prototype, so fields absent from `fields` resolve to their default at zero per-record
        // cost - the same defaults strategy as the legacy `Object.create(_dataDefaults)` route,
        // but with the covered fields kept in fast-properties mode. A field literally named
        // `__proto__` is rejected above, so this is the only `__proto__` key emitted.
        const fn = new Function('p', 'v', `return {__proto__:p,${slots}};`) as (
            p: PlainObject,
            v: any[]
        ) => PlainObject;
        this.createFn = v => fn(defaultsProto, v);
    } else {
        // Base path (unchanged): plain literal, own prop for every field.
        this.createFn = new Function('v', `return {${slots}};`) as (values: any[]) => PlainObject;
    }
}
```

**2. `View.loadStores` provides the query fields + special cube fields** to each connected store before reloading:

```ts
// Framework-level fields set on every ViewRowData, in addition to the query fields.
const CUBE_META_FIELDS = ['cubeLabel', 'cubeRowType', 'cubeDimension', 'cubeBuckets'];
```

```ts
private loadStores() {
    const {_leafMap, _rowDatas} = this;
    if (!_leafMap || !_rowDatas) return;

    // Skip degenerate root in stores/grids, but preserve in object api.
    const storeRows = _leafMap.size !== 0 ? _rowDatas : [],
        projectionFields = [...this.fieldNames, ...CUBE_META_FIELDS];
    this.stores.forEach(s => {
        s.setProjectionFields(projectionFields);
        s.loadData(storeRows);
    });
}
```

**3. `Store.setProjectionFields`** rebuilds `_dataFactory` for the covered subset — no separate field-set state, it reads the covered set back off the current factory to no-op when unchanged:

```ts
/**
 * Rebuild this store's data factory to cover only the given subset of its fields as fast-mode
 * own properties, with all remaining fields' defaults resolved via a shared prototype. Lets a
 * store with more registered fields than {@link RecordDataFactory.MAX_FACTORY_FIELDS} - or one
 * that is otherwise sparsely populated - still build records in V8 fast-properties mode,
 * to the populated subset.
 *
 * Called by a connected Cube {@link View}, which knows the fields its current query populates.
 * Must be followed immediately by a `loadData()`; existing records keep their prior data
 * until rebuilt (the View reloads right after calling this).
 *
 * @internal
 */
setProjectionFields(fieldNames: string[]) {
    const populated = new Set(fieldNames),
        ownFields = this.fields.filter(f => populated.has(f.name)),
        current = this._dataFactory?.fields;

    // No-op if the current factory already covers exactly this set - avoids recompiling on
    // every fullUpdate when the query fields are unchanged.
    if (current?.length === ownFields.length && current.every((f, i) => f === ownFields[i])) {
        return;
    }

    this._dataFactory = RecordDataFactory.createProjection(ownFields, this._dataDefaults);
}
```

Note both the load and cube-tick paths run through `parseRaw` (`createRecord` → `parseRaw`), which already consumes `_dataFactory` uniformly — so no parse-logic changes are needed, only *which* factory is installed. I'd gate this behind the same experimental config we land the base factory und-store `data` from the legacy sparse shape to the projected one (own props for the coveredset; `getValues()`/`getModifiedValues()` remain the right accessors).

haynesjm42:
Filed #4506 which supersedes the proposed changes in the above comment, and further optimizes the Cube->View->Store use-case

PR #4505, opened 2026-07-23
Store: add `loadDataAsync()` - streaming load without buffering the full raw dataset

lbwexler:
**TLDR:** New `Store.loadDataAsync(rawData)` - the streaming counterpart to `loadData()` - plus `XH.fetchNdjson()` - a FetchService convenience that fetches and decodes an NDJSON response incrementally, the natural source for the new load method (`store.loadDataAsync(XH.fetchNdjson({url}))`) and directly iterable via `for await` for non-Store streaming.

### Why

Today `loadData()` requires the complete raw dataset as one array, so a large load's peak heap includes every parsed raw row at once - and chunked `updateData({add})` calls are not a substitute (per-transaction RecordSet rebuilds and observable churn). With a streaming source, rows can instead be consumed as they arrive: parsing overlaps download, and each raw row becomes garbage as soon as its record is built.

Measured motivation: NDJSON parsed line-at-a-time via native `JSON.parse` costs only ~12% over one monolithic parse (179ms vs 159ms, 100k rows x 57 fields) while never holding more than one row of raw input. Pairs with #4504 (`retainRaw: false`) to eliminate raw-row retention entirely - together, peak heap during load approaches the steady-state record set. Also complements #4501.

### `Store.loadDataAsync(rawData)`

- Accepts a sync or async iterable yielding raw records (or arrays/chunks of records), creating records incrementally as the source yields. The Store is untouched until the source completes, then all records install in a single observable transaction - identical downstream behavior to `loadData()`. If the source throws, the Store remains unchanged.
- Mirrors `loadData()` semantics for row data: duplicate-ID checks (including cross-chunk) and tree data via nested `children`. Deliberately does NOT accept summary data - a summary is an aggregate, unavailable until a stream completes; install via `updateData({rawSummaryData})` after loading. `loadRootAsSummary` stores throw (such payloads nest all rows in a single root node and cannot be streamed).
- `Cube.loadDataAsync()` likewise accepts a streaming source for its rawData, delegating to the new Store method - `cube.loadDataAsync(XH.fetchNdjson({url}))`. Array inputs take the existing synchronous path unchanged.
- Purely additive - `loadData()` is untouched.

### `XH.fetchNdjson(opts)`

- Issues the request eagerly through the standard managed fetch pipeline (correlation IDs, traceparent, interceptors, default headers) and returns an `AsyncGenerator` yielding arrays of parsed records as chunks arrive - no more than one network chunk of raw text is ever buffered.
- **Telemetry scopes to the full stream lifetime**: a single client-kind span (per OTel semconv, ending when the response is fully read) and a `track` entry that fires on stream completion/failure with true elapsed time. The server-side span nests under it via the propagated traceparent, so traces decompose generation vs. download/consumption. A failed request closes the span even if the returned generator is never consumed.
- **Enriched exceptions across both phases**: request-phase failures throw the usual FetchExceptions; failures while reading/parsing the stream throw a new `Fetch Stream Failed` FetchException carrying the response status, URL, correlation ID, and cause - and mid-stream aborts map to the routine `abortedException`, keeping `catchDefault()` quiet.
- Known limitation, documented in-method: `timeout` covers the request phase only - no timeout applies while the stream is being read. In-flight stream abort via `autoAbortKey` is a noted follow-up.

### Notes

- Compiles and lints clean; not yet exercised at runtime - planning to validate against a live Toolbox app (streaming NDJSON + equivalence vs `loadData` on the same dataset) before merge.

---

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (None - additive API.)
- [x] Created Toolbox branch / PR, or determined not required. (xh/toolbox#880 - draft pending the runtime validation note above.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,10 @@
   async iterable yielding raw records or chunks. Creates records incrementally without buffering
   the complete raw dataset in memory, then installs them in a single transaction once the source
   completes.
-* Added `ndjsonChunks()` util (`@xh/hoist/utils/async`) to read an NDJSON `Response` body
-  incrementally - the natural streaming source for `Store.loadDataAsync()`.
+* Added `XH.fetchNdjson()` to consume an NDJSON (newline-delimited JSON) response incrementally
+  as an async iterable - the natural streaming source for `Store.loadDataAsync()`, and usable
+  directly via `for await` for any streamed endpoint. Backed by a new lower-level
+  `ndjsonChunks()` util in `@xh/hoist/utils/async` for use with an existing `Response`.
 
 ### ⚙️ Technical
 * Moved both desktop and mobile popover implementations off the deprecated, React-18-capped Popper.js
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,14 @@
     * The `popperOptions` escape-hatch prop has been removed from the mobile `Popover`.
 
 ### 🎁 New Features
-* Added `Store.loadDataAsync()` to load a complete dataset from a streaming source - a sync or
-  async iterable yielding raw records or chunks. Creates records incrementally without buffering
-  the complete raw dataset in memory, then installs them in a single transaction once the source
+
+* Added `Store.loadDataAsync()` to load a complete dataset from a streaming source - a sync or async
+  iterable yielding raw records or chunks. Creates records incrementally without buffering the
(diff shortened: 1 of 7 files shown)


lbwexler:
Closing in favor of new PR, that places this on FetchService and combines it with String Interning

PR #4507, opened 2026-07-24
Store: opt-in adoptRawData zero-copy mode for connected Cube Views

### LINKED ISSUE #4506: Store: zero-copy `adoptRawData` mode for connected Cube Views
# Store: zero-copy "adopt raw data" mode for connected Cube Views

## Summary

Add an opt-in `Store` mode that treats each record's pre-parsed raw object **as** its `data` object, shared by reference, instead of re-parsing and copying it. Targets the dominant Cube use-case: a Cube feeding one or more connected `View`s, each driving a `Store` behind a (tree) grid.

Today every row is two retained objects: the `ViewRowData` the View builds and caches, and a separate per-record `data` object the Store re-parses from it via `parseRaw`. The Store's copy is redundant here - it re-runs `parseVal` (already applied by the Cube) and allocates a second object holding the same values, while the `ViewRowData` stays alive anyway as `StoreRecord.raw`.

This mode makes `StoreRecord.data === the ViewRowData`, collapsing 2N row objects to N and skipping the per-row re-parse on every load and tick. As a bonus, the framework-level `cube*` fields (`cubeLabel`, `cubeRowType`, `cubeDimension`, `cubeBuckets`) and the `isCubeLeaf` / `cubeLeaves` getters are present on records for free - no field registration required.

Builds on #4500/#4502 (record `data` factory, View row assigner). `retainRaw` (#4504) is moot on this path - `raw === data`.

## Motivation

Target pattern (real client apps): non-editable, ticking/streaming data loaded into a Cube, fanned out to several connected Views, each driving a tree grid. Rows are display-only; the grid never edits store records locally. At 10k-100k rows these apps are memory- and GC-sensitive, and #4501/#4502 already showed this is where V8 object cost dominates.

In that pattern the Store's re-parse buys nothing: leaf values arrive already typed (they came out of the Cube's own internal store, parsed by the same `CubeField` instances the Store uses), aggregates are computed by aggregators, and the Store is a read-only projection - the committed/dirty/revert machinery the copy exists to support is never exercised. Yet we pay for it twice per row, for the life of the grid.

Headline win is **memory** (one fewer ~30-field object per row, plus no second retained copy). Performance is a modest secondary win: skips one object allocation + the `parseRaw` field loop per row per tick, reducing GC churn on high-frequency updates.

## Why sharing is safe (verified against current code)

Hoist grid updates are driven at the **record-identity** level, not on `data`:

- `StoreRecord.data` is a plain, non-observable `readonly` prop (`StoreRecord.ts`).
- `Grid.genTransaction` diffs by StoreRecord **reference identity** - a record lands in the ag-Grid `update` list iff `existing !== rec` (`Grid.ts`), never by inspecting `.data`. (Note: Hoist does **not** use ag-Grid's immutable-data / delta-row mode; it computes transactions itself and calls `agApi.applyTransaction()`, with `updateGridOptions({rowData})` only on the initial load.)
- The tick path (`updateData` -> `RecordSet.withTransaction`) **unconditionally** installs the new record for each update (`newRecords.set(id, rec)`) - no data-equality dedup. So every tick mints a new identity and the grid always sees an update, regardless of whether `data` was shared/mutated.

The two mutation contracts that would otherwise collide (Store freezes + snapshots; View mutates `ViewRowData` in place) are reconciled by scoping this to read-only projections with `freezeData: false`, and by the Store never touching `data` internals in this mode.

Checked the one path that could wrongly reuse a record under shared+mutated data - the `loadData` reuse path (`RecordSet.areRecordsEqual`, deep-equals `.data`). It cannot misfire: the only in-place mutation happens in `View.dataOnlyUpdate` -> `updateData` (no dedup). The `loadData` path only ever sees unchanged leaves (query-only change) or freshly-built rows (row cache cleared on `noteCubeLoaded`).

## Proposal

Add a first-class boolean flag to `StoreConfig` (proposed name: **`adoptRawData`**; default `false`). When set:

- **Zero-copy record creation** - adopt the incoming raw object as `data` directly, by reference; skip `processRawData` and `parseRaw`. `record.data === the raw object` (typically the `ViewRowData`), so `cube*` fields and the `isCubeLeaf` / `cubeLeaves` getters are present with no field registration.
- **Never touch `data`** - do not `parseRaw`, do not assign `data.id`, do not freeze. The Store does not own this object in this mode.
- **`freezeData` forced to `false`** - the View mutates rows in place on tick updates; a frozen shared object would throw on the next mutation.
- **Read-only** - editing/commit/revert APIs are unsupported (throw), since committed === current always. Field registry becomes metadata for sort/filter/columns/export, not the definition of `data`'s shape.
- Records are still recreated on load and `updateData`, reusing the (in-place-updated) `data` refs -> new identities -> correct ag-Grid transactions. This already happens; no new View->Store contract is needed (View already passes `ViewRowData` into `loadData`/`updateData`).

### Mode contract (document prominently on the flag)

This is effectively an **"unsafe"/"turbo" mode** - it puts substantial trust in the data provider.

1. **Raw data MUST already be parsed.** Because `parseRaw` is skipped, store field `type`/`parseVal` are **not applied** to `data`. Fields are pure metadata (sort/filter/columns/export); values are exactly what the provider supplied. This is always true for Cube/View data - the Cube already parsed it, typically using the same `CubeField` instances the Store uses. A dev defining a store field `type` expecting coercion in this mode would be surprised.
2. **The Store does not own `data`.** It never reads or writes `data` internals. The provider (the View) is authoritative and may mutate rows in place.
3. **Rows may be shared across connected stores.** `View.loadStores` feeds the same `_rowDatas` array to every connected store; with by-reference adoption, their records point at one object set. Do not mutate these objects from app code.

### Intended use / generality

Primary use-case is `ViewRowData` fed by a connected `View`. The mechanism is general enough to work with any already-parsed data whose provider follows the same contract (pre-parsed, read-only, in-place updates mint new records). Decision to make when finalizing: expose it as generally usable (documented caveats) vs. document it as **intended only for `ViewRowData` from a connected View**. Lean general-but-clearly-caveated unless the `data.id` verification (below) surfaces a reason to restrict.

### `idSpec` / `data.id`

`StoreRecord`'s constructor normally does `data.id = id`. In this mode we skip that assignment (we never touch `data`). `ViewRowData.id` already matches the default `idSpec: 'id'` read, so identity is consistent with zero writes. Default `idSpec: 'id'` for this mode.

**Verification item (must do before finalizing):** grep/confirm nothing *internal* reads `record.data.id` on the read-only path. Known writers/consumers of `data.id` (ctor assignment, `parseUpdate`, `getModifiedValues`) are all on the edit machinery this mode disables - confirm there are no others. Document the caveat that `data.id` is not maintained by the Store in this mode.

### Guardrails

**Construction-time (throw):**

- `processRawData` present - the skipped parse would make it a silent no-op.
- `reuseRecords: true` - already rejected by `View.parseStores` for connected stores; also assert at the Store level for this flag.
- `freezeData: true` (if explicitly set) - or silently force to `false`.

**Runtime (throw / no-op with clear error):**

- `modifyRecords`, `addRecords`, `removeRecords`, `revertRecords`, `revert` - unsupported in this read-only mode.

## Scope of changes

- **`Store`** - add `adoptRawData` to `StoreConfig` and the class; branch in `createRecord` to adopt the raw object as `data` (skip `processRawData`/`parseRaw`/`data.id`/freeze); force `freezeData: false`; add construction guardrails; guard the edit APIs.
- **`View`** - no new data contract required for the zero-copy path (already passes `ViewRowData` into `loadData`/`updateData`). Optionally: validate/encourage that connected stores using this flag are configured compatibly.
- **`GridModel` / `Grid`** - no code changes expected (transaction generation is already identity-based). Covered by the acceptance test below.

## ag-Grid cell change-detection - required acceptance test (primary risk)

This is the one behavior not guaranteed by Hoist code alone; it depends on ag-Grid internals and **must be demonstrated, not assumed.**

Under this mode the View mutates the `ViewRowData` *before* the new record is built, so when ag-Grid processes the update transaction, `newRecord.data === oldRecord.data ===` the already-mutated object. Whether a simple-renderer cell repaints then depends on ag-Grid recomputing its `valueGetter` and diffing against its own per-cell cached last-rendered value (expected), rather than diffing old-vs-new data-object references. Hoist force-refreshes only complex-renderer columns (`Grid.ts`); simple columns ride entirely on ag-Grid's change detection. Today's copy-based design masks this (the prior frozen copy genuinely differs); zero-copy removes that cushion.

**Setup:** Cube -> connected View -> `Store` with `adoptRawData: true` -> tree grid. Include at least one column with a plain/simple renderer (plain text or `numberRenderer`, not `rendererIsComplex`) plus one complex-renderer column for contrast.

**Action:** `Cube.updateDataAsync` with a leaf value change that does **not** alter dimensions or row population, forcing the `dataOnlyUpdate` (tick) path, not `fullUpdate`. Cover a changed leaf value, a changed aggregate/parent value, and a value that does not change.

**Assert:**

- Simple-renderer cell repaints to the new value.
- Aggregate/parent-row cells repaint (mutated via `parent.applyDataUpdate`).
- Unchanged cells do not flicker.
- `includeRoot` "Total" row renders and updates.
- Pinned summary row (if any) stays correct.

If simple columns go stale, fall back to force-refreshing all visible columns for updated rows in this mode (or a targeted refresh), and note the added cost.

## Non-goals

- No change to default Store behavior (opt-in flag, default `false`) - **non-breaking.**
- Not making View row updates non-mutating/immutable (larger change, not needed here).
- No editing support in this mode.

haynesjm42:
Adds an opt-in `StoreConfig.adoptRawData` mode (default `false`) for read-only projections of already-parsed data - the dominant Cube use case: a Cube feeding connected `View`(s), each driving a (tree) grid. Resolves #4506.

When enabled, each record adopts the provider's raw object **as** its `data`, by reference, instead of re-parsing and copying it. This collapses the usual two per-row objects (the `ViewRowData` plus the Store's parsed copy) down to one and skips the per-row `parseRaw` on every load and tick.

**Stacked on #4500 / #4502 - this PR targets `cube-view-row-data`, not `develop`.**

### Behavior

- Zero-copy `createRecord`: adopts raw as `data` (skips `processRawData` / `parseRaw` / `data.id` write / freeze), minting a fresh record identity each load/update so ag-Grid transactions still fire.
- Read-only: `addRecords` / `removeRecords` / `modifyRecords` / `revertRecords` / `revert` throw; data flows in via `loadData` / `updateData`. Summary records (including the `loadRootAsSummary` / `includeRoot` "Total" row) are adopted the same way.
- Construction fails fast on explicitly-incompatible config (`processRawData`, `reuseRecords`, `freezeData: true`); `freezeData` is otherwise normalized to `false`.
- Non-breaking: opt-in, default `false`, no change to existing behavior.

### Measured gains

Toolbox cube tester, 30,320 rows, real Cube -> View, Chrome with precise heap + forced GC. Isolated on standalone stores loading identical `ViewRowData`, so this measures the Store's own cost - and both modes already benefit from #4500's fast-properties factory / #4502's assigner:

| Fields | Store retained (legacy -> adopt) | Mem saved | Record build (legacy -> adopt) |
|---:|---|---:|---|
| 14 | 5.9 -> 3.3 MB | 43% | 47.9 -> 16.3 ms (2.9x) |
| 30 | 7.8 -> 3.3 MB | 57% | 83.9 -> 17.4 ms (4.8x) |
| 50 | 10.1 -> 3.4 MB | 67% | 140.6 -> 16.3 ms (8.6x) |

adoptRawData is **constant-cost** (~116 B/record, ~16 ms) regardless of field width - it adds no per-field work and no data object. Legacy scales linearly with width, so the win grows with record width. Memory is the headline; faster record build is a secondary win (end-to-end tick latency gains less, since Cube recompute + grid render dominate that path and are mode-independent).

### Verification

Cube -> View -> tree grid acceptance test (companion Toolbox PR): under a leaf tick that mutates `ViewRowData` in place, both simple- and complex-renderer cells repaint, aggregates and the `includeRoot` Total update, and unchanged rows stay stable. Edit-API and construction guards confirmed to throw at runtime. `tsc` and `eslint` clean.

Hoist P/R Checklist
-------------------

- [x] Caught up with base branch as of last change. _(Stacked on `cube-view-row-data`; caught up with that base, not `develop`.)_
- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. _(Non-breaking: opt-in, default `false`.)_
- [x] Updated doc comments / prop-types, or determined not required. _(Extensive JSDoc on the `adoptRawData` config.)_
- [x] Reviewed and tested on Mobile, or determined not required. _(Data-layer change, platform-agnostic.)_
- [x] Created Toolbox branch / PR, or determined not required. _(xh/toolbox#881.)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,14 @@
   aggregate rows) are now populated via a per-View compiled assigner, measured ~5-6x smaller at
   typical view widths (more than ~20 fields), with faster row generation and aggregation reads.
   Views with very many fields and strict-CSP environments continue to use the prior route.
+* Added an opt-in `Store.adoptRawData` mode for read-only projections of already-parsed data - most
+  notably a connected Cube `View` feeding a (tree) grid. Records adopt the provider's row object as
+  their `data` by reference rather than re-parsing and copying it, collapsing the usual two per-row
+  objects to one and skipping the per-row parse on every load and update - a memory-focused
+  optimization for large, high-frequency Cube-backed grids. This is a read-only mode - the local
+  edit/commit/revert APIs throw, `freezeData` is forced off, and the data provider must supply
+  pre-parsed values. See the `adoptRawData` config docs for the full contract. Default off,
+  non-breaking.
 
 ### ⚙️ Technical
 * Moved both desktop and mobile popover implementations off the deprecated, React-18-capped Popper.js
--- a/data/Store.ts
+++ b/data/Store.ts
@@ -148,6 +148,44 @@ export interface StoreConfig {
     reuseRecords?: boolean;
 
     /**
+     * Zero-copy "turbo" mode for read-only projections of already-parsed data - most notably a
+     * connected Cube {@link View} feeding a (tree) grid.
+     *
+     * When true, the Store adopts each incoming raw object **as** its record's `data`, by reference,
+     * rather than re-parsing and copying it into a dedicated object. This collapses the usual two
+     * per-row objects (the provider's row object + the Store's parsed copy) down to one, skipping
+     * the per-row `parseRaw` field loop on every load and update. The headline win is memory; a
+     * modest reduction in allocation/GC churn on high-frequency updates is a secondary benefit.
+     *
+     * This mode puts substantial trust in the data provider and comes with a strict contract:
+     *
+     * 1. **Raw data MUST already be parsed.** `parseRaw` is skipped, so Field `type`/`parseVal` are
+     *    NOT applied - Fields become pure metadata (sort/filter/columns/export) and `data` holds
+     *    exactly what the provider supplied. Always true for Cube/View data, which the Cube has
+     *    already parsed (typically via the same field instances).
+     * 2. **The Store does not own `data`.** It never reads or writes `data` internals (including
+     *    `data.id`) and never freezes it. The provider is authoritative and may mutate rows in place.
+     * 3. **Rows may be shared across connected stores.** A View feeds the same row objects to every
+     *    connected store; with by-reference adoption their records point at one object set. Do not
+     *    mutate these objects from app code.
+     *
+     * Consequences:
+     * - **Read-only.** The local edit/commit/revert APIs (`addRecords`, `removeRecords`,
+     *   `modifyRecords`, `revertRecords`, `revert`) throw - committed always equals current.
+     *   Data still flows in via `loadData`/`updateData` (the provider path).
+     * - **Summary records are adopted too.** The dedicated summary record(s) - whether supplied via
+     *   `rawSummaryData` or extracted from the root row under `loadRootAsSummary` (e.g. a Cube
+     *   View's `includeRoot` "Total" row) - adopt their raw object as `data` by the same rules:
+     *   shared by reference, not owned, not frozen, recreated on update.
+     * - `freezeData` is forced to `false` (a frozen shared object would throw on the provider's
+     *   next in-place mutation). Explicitly passing `freezeData: true` is rejected.
+     * - `processRawData` and `reuseRecords` are incompatible and rejected at construction.
+     *
+     * Default false.
+     */
+    adoptRawData?: boolean;
+
+    /**
      * Set to true to always validate all uncommitted records on every change to
      * uncommitted records (add, modify, or remove). Default false.
      */
@@ -270,6 +308,7 @@ export class Store
     idEncodesTreePath: boolean;
     freezeData: boolean;
     reuseRecords: boolean;
+    adoptRawData: boolean;
     validationIsComplex: boolean;
 
     @observable.ref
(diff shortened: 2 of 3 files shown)


lbwexler:
Superseded by #4516, which rebases this commit (authorship preserved) directly onto `develop`, decoupling it from the #4501/#4503 stack so it can land independently. Thanks @haynesjm42!

PR #4514, opened 2026-07-26
Add "View Surrounding Lines" action to Admin Console log viewer

RyanMG:
Adds a "View Surrounding Lines" right-click action to the Admin Console log viewer. After filtering for a term, right-clicking a matching line clears the filter and reloads the log positioned around that line, then re-selects it and centers it in the viewport - letting you read a hit in the context of the entries around it without manually clearing the filter and hunting for the line number.


lbwexler:
Works great!  Just tweaked the overly verbose comments and changed a variable name.

PR #4523, opened 2026-07-29
NdjsonResult API for fetchNdjson + StringInternSpec.retainMode

lbwexler:
Pre-release adjustments to the v87 NDJSON streaming and string-interning APIs.

* `XH.fetchNdjson()` now returns an `NdjsonResult` - `lines` (the chunked record stream, e.g. `store.loadDataAsync(XH.fetchNdjson({url}).lines)`) plus `meta`, a promise for an optional leading metadata record requested via the new `firstLineIsMeta` option. `meta` resolves eagerly as soon as the first record arrives, so apps can use it before consuming `lines`.
* Implementation: NDJSON stream mechanics extracted from `FetchService` into `svc/impl/NdjsonResultImpl`, which implements `NdjsonResult` and is returned to callers directly. Service concerns (interning, exception enrichment) are injected via constructor args; the runner tracks the full stream lifetime via `whenCompleteAsync()`.
* Replaced `StringInternSpec.retainAcrossFetches` with `retainMode: 'never' | 'nextCall' | 'always'` - the new `'always'` mode suits successive responses covering different slices of a dataset (paging, alternating filters). Specs are adopted per call and interners persist per key until `clearInternCaches()` - never reset by spec changes.

### DIFF
--- a/svc/FetchService.ts
+++ b/svc/FetchService.ts
@@ -23,20 +23,10 @@ import {PromiseTimeoutSpec} from '@xh/hoist/promise';
 import {isLocalDate, SECONDS} from '@xh/hoist/utils/datetime';
 import {apiDeprecated, warnIf} from '@xh/hoist/utils/js';
 import {StatusCodes} from 'http-status-codes';
-import {
-    isDate,
-    isEqual,
-    isFunction,
-    isNil,
-    isObject,
-    isString,
-    noop,
-    omit,
-    omitBy,
-    truncate
-} from 'lodash';
+import {isDate, isFunction, isNil, isObject, isString, noop, omit, omitBy, truncate} from 'lodash';
 import {IStringifyOptions, stringify} from 'qs';
 import ShortUniqueId from 'short-unique-id';
+import {NdjsonResultImpl} from './impl/NdjsonResultImpl';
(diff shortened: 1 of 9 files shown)


svc/FetchService.ts
```diff
@@ -1050,6 +963,28 @@ export interface FetchOptions {
...
+     * True to treat the first record in the stream as metadata, delivered via
+     * {@link NdjsonResult.meta} rather than yielded with the data records. Default false.
+     */
+    firstLineIsMeta?: boolean;
+}
+
+/** Streamed result returned by {@link FetchService.fetchNdjson}. */
+export interface NdjsonResult {
+    /** Parsed data records, yielded in chunks (arrays) as they arrive off the network. */
+    lines: AsyncGenerator<PlainObject[]>;
```
jskupsik: This parameter is called "lines" - which is an unnecessary extra synonym to "chunks" and "rows". If I had to guess as to its meaning, I would incorrectly guess it meant rows ( AsyncGenerator<PlainObject> ) rather than chunks. I'm wondering if we could find a better term for this property. (edited)

jskupsik (changes requested):

Issue #4520, opened 2026-07-28
Add built-in support for `Badge.icon`


amcclain:
We often want to have a badge with a leading icon that is well aligned and properly sized out of the box

PR #4524, opened 2026-07-30
Store.skipDataCopy + skipDataParsing configs (replace useRawAsData)

lbwexler:
Replaces the not-yet-released `Store.useRawAsData` config with two orthogonal, opt-in performance configs for large datasets:

- `skipDataCopy` - use each raw object *as* its record's `data`, by reference and parsed in place. The Store takes ownership of adopted objects, installing its shared Field-defaults prototype on plain objects so missing Fields still resolve to their `defaultValue`.
- `skipDataParsing` - skip Field-level value parsing (type coercion + XSS sanitization) for data already in its final client-side form. Honored on load and update paths.

Also included:

- Improved `Store.processRawData` semantics - implementations are now handed a Store-owned shallow clone of each raw object and can mutate and return it directly, no defensive cloning needed. Now compatible with `skipDataCopy` (the raw object itself is passed).
- Cube `View` guards the unsafe combination (`skipDataCopy` without `skipDataParsing` would parse in place on View-owned rows) and logs a recommendation to set both flags on connected stores, which use View rows as record data directly.
- Changelog + data/cube README updates, including the recommended zero-copy flags in the connected-store example.

### DIFF
--- a/data/Store.ts
+++ b/data/Store.ts
@@ -88,8 +88,14 @@ export interface StoreConfig {
 
     /**
      * Function to run on each individual data object presented to `loadData()` prior to creating
-     * a `StoreRecord` from that object. This function must return an object, cloning the original
-     * object if edits are necessary.
+     * a `StoreRecord` from that object. Must return an object - either the object it was passed,
+     * edited in place, or a replacement.
+     *
+     * The Store passes this function its own shallow clone of each raw object, so top-level edits
+     * are safe and efficient - mutate and return it, no defensive cloning needed. (Nested values
+     * are still shared with the source object and should be replaced rather than mutated.) With
+     * `skipDataCopy` the raw object itself is passed and adopted - see that config for the
+     * ownership contract.
      */
     processRawData?: (data: PlainObject) => PlainObject;
 
@@ -159,18 +165,30 @@ export interface StoreConfig {
 
     /**
      * True to use each incoming raw object *as* its record's `data`, by reference, rather than
-     * parsing and copying it. A zero-copy mode for already-parsed data - e.g. a connected Cube
-     * {@link View}, or a server endpoint returning data in its final client-side form.
+     * copying it into a Store-owned object. A zero-copy mode for large datasets - e.g. a
+     * connected Cube {@link View}, or a server endpoint returning data in its final client-side
+     * form.
      *
-     * Raw data must already match what the Store's Fields would parse - `type`, `parseVal`, and
-     * `defaultValue` are not applied. The Store never modifies or freezes raw objects (regardless
-     * of `freezeData`); the provider may mutate rows in place but must then publish via
-     * `updateData()`, as `loadData()` would skip reference-equal objects as unchanged.
+     * Setting this flag hands ownership of raw objects to the Store: it installs its shared
+     * Field-defaults prototype on each adopted plain object (so missing Fields resolve to their
+     * `defaultValue` - objects with a custom prototype, e.g. class instances, are adopted with
+     * their prototype intact) and, unless `skipDataParsing` is also set, parses Field values in
+     * place, directly on the raw object. The Store never freezes adopted objects (regardless of
+     * `freezeData`) or writes `data.id`; the provider may mutate rows in place but must then
+     * publish via `updateData()`, as `loadData()` would skip reference-equal objects as unchanged.
      *
      * `data` will carry every key on the raw object, not just declared Fields. Not compatible
-     * with `processRawData` or `reuseRecords`. Default false.
+     * with `reuseRecords`. Default false.
      */
-    useRawAsData?: boolean;
+    skipDataCopy?: boolean;
+
+    /**
+     * True to skip Field-level parsing when creating or updating record data - values are taken
+     * as-is, with no {@link FieldType} coercion and no XSS sanitization. For data already in its
+     * final client-side form, where the per-value parse is unwanted overhead. Missing Fields
+     * still resolve to their `defaultValue`. Default false.
+     */
+    skipDataParsing?: boolean;
 
     /**
      * Set to true to always validate all uncommitted records on every change to
@@ -296,7 +314,8 @@ export class Store
     freezeData: boolean;
     reuseRecords: boolean;
     retainRaw: boolean;
-    useRawAsData: boolean;
+    skipDataCopy: boolean;
+    skipDataParsing: boolean;
     validationIsComplex: boolean;
 
     @observable.ref
@@ -354,7 +373,8 @@ export class Store
         idEncodesTreePath = false,
         reuseRecords = false,
         retainRaw = true,
-        useRawAsData = false,
+        skipDataCopy = false,
+        skipDataParsing = false,
         validationIsComplex = false,
         experimental,
         data
@@ -366,12 +386,8 @@ export class Store
             'Store cannot be configured with both `reuseRecords` and `retainRaw: false` - record reuse requires retained raw data references.'
         );
         throwIf(
-            useRawAsData && processRawData,
-            'Store.useRawAsData cannot be used with processRawData.'
-        );
-        throwIf(
-            useRawAsData && reuseRecords,
-            'Store.useRawAsData cannot be used with reuseRecords.'
+            skipDataCopy && reuseRecords,
+            'Store.skipDataCopy cannot be used with reuseRecords.'
         );
 
         this.experimental = this.parseExperimental(experimental);
@@ -387,7 +403,8 @@ export class Store
         this.idEncodesTreePath = idEncodesTreePath;
         this.reuseRecords = reuseRecords;
         this.retainRaw = retainRaw;
-        this.useRawAsData = useRawAsData;
+        this.skipDataCopy = skipDataCopy;
+        this.skipDataParsing = skipDataParsing;
         this.validationIsComplex = validationIsComplex;
         this.lastUpdated = Date.now();
 
@@ -673,7 +690,7 @@ export class Store
             throwIf(isNil(id), `Must provide 'id' property for new records.`);
             throwIf(this.getById(id), `Duplicate id '${id}' provided for new record.`);
 
-            const parsedData = this.parseRaw(it),
+            const parsedData = this.parseData(it),
                 parent = this.getById(parentId);
 
             return new StoreRecord({
@@ -1228,20 +1245,6 @@ export class Store
     ): StoreRecord {
         const id = this.idSpec(raw);
 
-        // Zero-copy - use the raw object as `data` by reference, with no parsing, no data.id
-        // write, and no freeze. The Store does not own this object - see StoreConfig.useRawAsData.
-        if (this.useRawAsData) {
-            return new StoreRecord({
-                id,
-                store: this,
-                raw,
-                data: raw,
-                committedData: raw,
-                parent,
-                isSummary
-            });
-        }
-
         // Potentially re-use existing record if raw data is reference equal and tree path identical
         if (this.reuseRecords) {
             const cached = this._committed?.recordMap.get(id);
@@ -1250,17 +1253,15 @@ export class Store
             }
         }
 
-        const {processRawData, retainRaw} = this;
+        const {skipDataCopy, processRawData, retainRaw} = this;
         let data = raw;
         if (processRawData) {
-            data = processRawData(raw);
-            throwIf(
-                !data,
-                'Store.processRawData should return an object. If writing/editing, be sure to return a clone!'
-            );
+            // For efficiency hand processRawData the object it can typically mutate
+            data = processRawData(skipDataCopy ? raw : {...raw});
+            throwIf(!data, 'Store.processRawData should return an object.');
         }
 
-        data = this.parseRaw(data);
+        data = this.parseData(data);
         const ret = new StoreRecord({
             id,
             store: this,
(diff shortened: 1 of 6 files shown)


data/Store.ts
```diff
@@ -1246,17 +1249,15 @@ export class Store
...
+        const {skipDataCopy, processRawData, retainRaw} = this;
         let data = raw;
         if (processRawData) {
-            data = processRawData(raw);
-            throwIf(
-                !data,
-                'Store.processRawData should return an object. If writing/editing, be sure to return a clone!'
-            );
+            // For efficiency hand processRawData the object it can typically mutate
+            data = processRawData(skipDataCopy ? raw : {...raw});
```
haynesjm42, line 1256: It's possible that the app code doesn't want to mutate `raw` and then return it - it might be returning it's own new object - in which case we are creating this copy unnecessarily. Not sure how likely that is in practice, the standard pattern is certainly to make a copy of `raw`, add/mutate properties, and return it. Still, I'm not sure this change in behavior is necessary - the developer will know if they are in `skipDataCopy` mode and should just mutate `raw`

Concretely: a store with `processRawData` now allocates three objects per row - this clone, whatever the app returns, and `parseRaw`'s output - where two sufficed before. That cost lands on every existing app on upgrade, while the benefit only arrives for those that rewrite `processRawData` to mutate.

data/StoreRecord.ts
```diff
@@ -303,11 +303,11 @@ export class StoreRecord {
     // --------------------------
     /**
      * True if this record's `data` object belongs to it alone and may be written to and frozen.
-     * False only for records holding a provider-owned raw object under `useRawAsData`.
+     * False for all records of a Store with `skipDataCopy`, which adopts provider-owned objects.
      * @internal
      */
     get ownsData(): boolean {
-        return this.data !== this.raw;
+        return !this.store.skipDataCopy;
```
haynesjm42, line 310: It's possible that we "own" the data even if we skipped the data copy. `modifyRecords` gets a fresh object from `parseUpdate`, and if we keep the allocation in `addRecords` (see my comment there) that path does too. The original `data !== raw` check handles both - for added records `raw` is null, so it's trivially true. Right answer, slightly different reason.

Worth being concrete about what `ownsData: false` costs on a Store-owned object: `data.id` is never written (line 264) and the object is never frozen despite `freezeData: true` (line 323). We want both in that case.

data/Store.ts
```diff
@@ -159,18 +165,30 @@ export interface StoreConfig {
...
      * `data` will carry every key on the raw object, not just declared Fields. Not compatible
-     * with `processRawData` or `reuseRecords`. Default false.
+     * with `reuseRecords`. Default false.
+     */
+    skipDataCopy?: boolean;
+
+    /**
+     * True to skip Field-level parsing when creating or updating record data - values are taken
+     * as-is, with no {@link FieldType} coercion and no XSS sanitization. For data already in its
+     * final client-side form, where the per-value parse is unwanted overhead. Missing Fields
```
haynesjm42, line 188: The note about "Missing Fields still resolve to their `defaultValue`" is only true if raw data is provided as plain objects - `parseData` installs the defaults prototype only when the prototype is still `Object.prototype`, so a `ViewRowData` never gets it. That means the claim fails precisely for the connected-store case the cube README now recommends, where any Store field the View doesn't populate reads `undefined` rather than its `defaultValue`.

Worth noting that distinction here at minimum. It also ties into my comment on the `setPrototypeOf` call - if we drop the defaults install for not paying its way, this caveat disappears because the behavior becomes uniform. One decision, not two.

amcclain:
real anselm: was working to test and validate the below but ran out of time - stashed by cc in comment here below, did not read yet myself, can revisit this but not today, Lee said he was aware of issue i will not opine on severity but did seem potentially very meaningful.  we should be able to measure

-------

Reviewed the branch. One substantive concern plus two smaller notes — none blocking, but the first is worth a look given the cube README now recommends these flags for all connected stores.

## Deep-equal cost on reload for `skipDataCopy` connected stores

`RecordSet.areRecordsEqual` compares whole `data` objects via `fast-deep-equal` (`data/impl/RecordSet.ts:242`). Under `skipDataCopy` a connected store's `data` *is* the `ViewRowData`, which carries `children` (`BaseRow.ts:92`) and, under `provideLeaves`, `_cubeLeafChildren` (`BaseRow.ts:68`). So each record comparison recurses through its subtree rather than scanning a flat map of field values.

This only bites when row objects aren't reference-equal — i.e. after any cache-clearing update (`noteCubeLoaded`, structural `noteCubeUpdated`), so every reload after the first. It's cheap for aggregate-only views, but with `includeLeaves` or `provideLeaves` set the leaf population enters the comparison graph and the cost scales with leaf count rather than aggregate count. Sampling cube query sites across our apps and Toolbox, effectively all of them set one or both — so that's the mainline shape here, not an edge case.

Possible fix, contained to `RecordSet` — when `skipDataCopy` is set, compare declared Fields rather than the whole object:

```ts
private dataEqual(d1: PlainObject, d2: PlainObject): boolean {
    if (d1 === d2) return true;
    const {store} = this;
    if (!store.skipDataCopy) return equal(d1, d2);
    return store.fields.every(f => f.isEqual(d1[f.name], d2[f.name]));
}
```

fast-deep-equal still does all the comparing, via `Field.isEqual` (which nothing in the framework currently calls) — just rooted per-field instead of at the object, so `children` and undeclared provider keys drop out. Gating on `skipDataCopy` avoids regressing wide-and-sparse stores, where the existing whole-object scan is cheaper. One semantic difference worth noting: key-set changes would no longer register as a change.

Caveat: this is read from source, not measured. The mechanism is verified; the magnitude isn't.

## Cross-store prototype adoption

`parseData` installs the Store's defaults prototype only when the object's prototype is still `Object.prototype` (`data/Store.ts:1322`). Loading the same plain objects into a second `skipDataCopy` Store silently skips the install, and that Store's records then resolve missing Fields against the *first* Store's defaults — no error, no warning. Objects with their own prototype (e.g. `ViewRowData`) are unaffected, since they take no defaults either way.

A symbol marker on `_dataDefaults` lets `parseData` distinguish "another Store's defaults" from "a legitimate foreign class" and throw. Have that written locally if it's wanted.

## The new advisory warn in `View.parseStores`

Fires for every existing connected store on every View construction and `setStores()`, for an opt-in perf config. Worth considering `logDebug` instead — and given the above, it currently pushes the config hardest at the views where it may cost most.

amcclain:
`parseData` installs the defaults prototype only when the prototype is still `Object.prototype` (`data/Store.ts:1322`). Load the same plain objects into a second `skipDataCopy` Store and the install is silently skipped - that Store's records then resolve missing Fields against the first Store's defaults. No error, no warning. Objects with their own prototype (`ViewRowData`) are unaffected.

Marking the defaults object lets `parseData` distinguish another Store's defaults from a foreign class:

```diff
diff --git a/data/Store.ts b/data/Store.ts
index 51c10723b..5090b8506 100644
--- a/data/Store.ts
+++ b/data/Store.ts
@@ -177,6 +177,11 @@ export interface StoreConfig {
      * `freezeData`) or writes `data.id`; the provider may mutate rows in place but must then
      * publish via `updateData()`, as `loadData()` would skip reference-equal objects as unchanged.
      *
+     * A given plain object may therefore be adopted by only one Store - the defaults prototype
+     * installed on it is Store-specific, and loading the same objects into a second `skipDataCopy`
+     * Store will throw. (Objects with their own prototype take no defaults and may be shared, as
+     * with the rows of a Cube `View` feeding multiple connected Stores.)
+     *
      * `data` will carry every key on the raw object, not just declared Fields. Not compatible
      * with `reuseRecords`. Default false.
      */
@@ -1318,9 +1323,18 @@ export class Store
             // Install the shared defaults prototype on plain objects so missing Fields resolve
             // to their defaultValue, as in copy mode. A single shared prototype keeps this cheap
             // (V8 caches the transition per shape) and a no-op for re-adopted objects. Objects
-            // with a custom prototype (e.g. ViewRowData) are adopted with their prototype intact.
-            if (Object.getPrototypeOf(data) === Object.prototype) {
+            // with a custom prototype (e.g. ViewRowData) are adopted with their prototype intact,
+            // but must not already carry *another* Store's defaults - they would silently resolve
+            // missing Fields against that Store's definitions.
+            const proto = Object.getPrototypeOf(data);
+            if (proto === Object.prototype) {
                 Object.setPrototypeOf(data, _dataDefaults);
+            } else {
+                const owner = proto?.[DEFAULTS_OWNER];
+                throwIf(
+                    owner && owner !== this,
+                    `Object has already been adopted by another Store with 'skipDataCopy'. An adopted object carries that Store's Field defaults on its prototype and cannot be shared - drop 'skipDataCopy' on one of the Stores, or load a copy.`
+                );
             }
             if (!skipDataParsing) this.parseInPlace(data);
             return data;
@@ -1384,7 +1398,7 @@ export class Store
     }
 
     private createDataDefaults() {
-        const ret = {};
+        const ret: PlainObject = {[DEFAULTS_OWNER]: this};
         this.fields.forEach(({name, defaultValue}) => (ret[name] = defaultValue));
         return ret;
     }
@@ -1433,6 +1447,10 @@ export class Store
     }
 }
 
+// Identifies a Store's shared Field-defaults object, installed as the prototype of all record
+// data. Symbol-keyed, so invisible to for..in, Object.keys, JSON.stringify and deep-equal.
+const DEFAULTS_OWNER = Symbol('xhDataDefaultsOwner');
+
 function isChildRawDataObject(obj): boolean {
     return obj.hasOwnProperty('rawData') && obj.hasOwnProperty('parentId');
 }
```

amcclain:
The two warns in `parseStores` contradict each other. `View.ts:604` tells any store without the flags to set them; `View.ts:592` then tells a `skipDataCopy` store with `processRawData` that it is now handling View-owned rows. Comply with the first and you earn the second. The first never mentions `processRawData` as a reason to hesitate.

Both also use `ret.some(...)` and name no store, so with multiple connected stores you cannot tell which one triggered it.

Suggest excluding `processRawData` stores from the recommendation and identifying the offenders:

```ts
const needsFlags = ret.filter(s => (!s.skipDataCopy || !s.skipDataParsing) && !s.processRawData);
if (!isEmpty(needsFlags)) {
    this.logWarn(
        'Store(s) connected to a Cube View should set `skipDataCopy` and `skipDataParsing` for optimal performance - View rows are already parsed and can be used as record data directly.',
        needsFlags
    );
}
```

Same treatment for the `processRawData` warn at `View.ts:592`.

data/Store.ts
```diff
@@ -669,7 +686,7 @@ export class Store
             throwIf(isNil(id), `Must provide 'id' property for new records.`);
             throwIf(this.getById(id), `Duplicate id '${id}' provided for new record.`);
 
-            const parsedData = this.parseRaw(it),
+            const parsedData = this.parseData(it),
```
haynesjm42, line 689: Are we sure we want to call `parseData` here? Under `useRawAsData` we treated added and modified record data as fully Store-owned - adoption applied only to raw data arriving via the load paths.

Adopting here means two `addRecords` calls with the same object (an app reusing a template, setting a fresh `id` each time) produce two live records sharing one `data` object, so both report the last id and later template edits rewrite both. That works today because `parseRaw` allocates per call.

Suggest keeping the allocation here and confining adoption to the load path - which also lets `ownsData` stay `data !== raw` (see my comment on `StoreRecord`). Note `modifyRecords` is already fine: `parseUpdate` allocates a new object regardless.

data/Store.ts
```diff
@@ -1310,16 +1311,34 @@ export class Store
...
 
+    /** Produce a StoreRecord's data object from source data, per this Store's copy/parse configs. */
+    private parseData(data: PlainObject): PlainObject {
+        const {skipDataCopy, skipDataParsing, _dataDefaults} = this;
+        if (skipDataCopy) {
+            // Install the shared defaults prototype on plain objects so missing Fields resolve
+            // to their defaultValue, as in copy mode. A single shared prototype keeps this cheap
+            // (V8 caches the transition per shape) and a no-op for re-adopted objects. Objects
+            // with a custom prototype (e.g. ViewRowData) are adopted with their prototype intact.
+            if (Object.getPrototypeOf(data) === Object.prototype) {
```
haynesjm42, line 1322: The comment above makes a specific claim - "V8 caches the transition per shape" - that I don't think we've verified. My understanding runs the other way: `setPrototypeOf` after object creation is a known deopt, invalidating prototype-chain inline caches and potentially moving the object off its hidden-class transition chain. Here we're doing it per row on every load, to buy Field defaults.

Can we measure this against the `Object.create(_dataDefaults)` allocation it replaces, at the row counts that motivate the flag? If the install doesn't pay, I'd drop it - `skipDataCopy` then just doesn't provide Field defaults, which is already what happens for `ViewRowData` (see my comment on the `skipDataParsing` docs), so we'd end up with one consistent rule instead of an exception. I'd rather not add a third flag to control the second one.

Minor, awareness only: `setPrototypeOf` throws `TypeError` on a non-extensible object, so a frozen raw payload fails hard at load rather than degrading. Not a case I think we need to support - just worth knowing it's an error and not a fallback.

data/Store.ts
```diff
@@ -1310,16 +1311,34 @@ export class Store
...
 
         // b) apply parsed data as needed.
-        const {_fieldMap} = this;
+        const {_fieldMap, skipDataParsing} = this;
         for (const name in data) {
             const field = _fieldMap.get(name);
             if (field) {
-                const val = field.parseVal(data[name]);
+                const raw = data[name],
+                    val = skipDataParsing ? (raw ?? field.defaultValue) : field.parseVal(raw);
```
haynesjm42, line 1341: `skipDataParsing` still resolves null/undefined to `defaultValue` here, so it isn't quite "take values as-is" - it drops coercion and XSS sanitization but keeps the one piece of `parseVal` that substitutes defaults. That makes the two flags disagree: `{qty: null}` with `defaultValue: 0` lands as `0` under `skipDataParsing` alone, but stays `null` under both flags, since `parseData` skips `parseInPlace` entirely and never rewrites the adopted object.

Sharper version of the same split - `parseUpdate` applies `raw ?? defaultValue` regardless of `skipDataCopy`, so in a store with both flags set a field loaded as `null` reads `null`, but the same field set to `null` via `modifyRecords` reads `0`. Same store, same field, different value depending on how it got there.

The zero-copy side can't really change - rewriting nulls means writing to the adopted object, which is what the flag exists to avoid and what we forbid for View rows. So either take values fully verbatim in both paths (nulls stay null, which is what "skip parsing" implies), or document that `defaultValue` substitution survives `skipDataParsing` only in copy mode. All unreleased, so we're free to pick.

haynesjm42 (commented):
I think this is definitely the right path - I was just looking for this in a client app where I was integrating `useRawAsData`, but couldn't optimize the `Cube` in the way I wanted to.  

The case I care about most is `skipDataCopy` *without* `skipDataParsing` - we have a number of Cubes loading straight from an API where we still need Field parsing (mostly `Date`/`LocalDate`) and computed fields via `processRawData`. The changelog already frames this as a primary use ("parsed in place - most useful for data freshly fetched via `FetchService`, with no other owner"), so the semantics are right - the gap is just that `Cube` can't be configured this way.

We should also expose these on `Cube` - `CubeConfig` only picks a subset today, and a Cube is really just a Store wrapper. `Omit<StoreConfig, ...>` plus redeclared `fields`/`fieldDefaults` (both narrowed to Cube types) would do it, but `freezeData` and `idEncodesTreePath` need to stay Cube-controlled - it hardcodes `freezeData: false` and `idEncodesTreePath: true` and depends on both. Happy for the full type refactor to be a follow-up if we just add the two new flags to `CubeConfig` here.

lbwexler:
Superceded by #4520 4534

PR #4526, opened 2026-07-31
Fix CodeInput leaking a new CodeMirror editor on every re-render

### LINKED ISSUE #4525: Additional detail text bleeds through the component
Reproducible on https://toolbox-dev.xh.io/admin/activity , NOT on https://toolbox.xh.io/admin/activity -> it is a new regression / bug.

Only happens in dark mode.

If you select a row with Additional Data in the toolbox admin user activity, and then select another row with additional data, you will see the spooky ghostly remains of the previous selections' text. Scrolling the additional data may make this more obvious in the case of the two rows having the same / similarly structured data.

Image

jskupsik:
Fixes #4525

`CodeInput` was accumulating CodeMirror editors in the DOM — one per re-render. Since `.xh-code-input__inner-wrapper > .cm-editor` is absolutely positioned at `inset: 0`, the orphans stacked exactly on top of each other, each retaining the content it was created with. On `/admin/activity` the editor count grew by one per record selection (measured 5 → 6 → 7).

### Root cause

The React 19 upgrade (2c0a91b31) changed the container ref from a stable function reference to an inline arrow, to avoid returning a Promise (React 19 treats a ref callback's return value as a cleanup fn):

```ts
ref: el => { model.createCodeEditor(el); }   // new identity every render
```

React re-invokes callback refs whose identity changes, so every re-render detached (a no-op on `null`) and re-attached, appending a new `EditorView` without disposing the previous one.

### Why it looked like a dark-mode-only bug

The defect is theme-independent; only its visibility isn't. Verified by measuring both themes on the unfixed code:

| Theme | Editors stacked | Distinct stale docs | `.cm-editor` background |
|---|---|---|---|
| Light | 5 | 5 | `rgb(255, 255, 255)` — opaque |
| Dark | 5 | 5 | `rgba(0, 0, 0, 0.3)` — 30% opaque |

In light theme the top editor's opaque background completely paints over the stale layers, so the panel looks clean despite five editors behind it.

### Changes

* Restored a stable ref (`editorContainerRef`) that returns void.
* `createCodeEditor` now disposes any prior editor before creating a new one — matching `Chart`, which always destroys before re-creating.
* Added a guard for the container being swapped while awaiting async extension loading.

### Verification

`tsc --noEmit`, `eslint`, and `prettier --check` clean. Live in Toolbox, editor count stays at 1 across record switches, light/dark toggles, fullscreen enter/exit (content preserved), and typing in an editable JSON config.

### No changelog entry

The regression is limited to unreleased `87.0.0-SNAPSHOT` — released v86.x used a stable ref and is unaffected, consistent with the issue reproducing on toolbox-dev but not production.

### DIFF
--- a/desktop/cmp/input/CodeInput.ts
+++ b/desktop/cmp/input/CodeInput.ts
@@ -178,6 +178,7 @@ class CodeInputModel extends HoistInputModel {
 
     private themeCompartment = new Compartment();
     private editableCompartment = new Compartment();
+    private editorContainer: HTMLElement = null;
 
     get fullScreen(): boolean {
         return this.modalSupportModel.isModal;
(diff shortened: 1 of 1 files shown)


lbwexler:
Great Catch Jakub -- tweaked a bit to minimize Claude verbosity and merged.

amcclain:
Really glad we got this fix in

PR #4533, opened 2026-08-03
Add `badge.icon` prop

RyanMG:
https://github.com/xh/hoist-react/issues/4520


lbwexler:
@RyanMG heads up - pushed a small tweak: `items: icon ? [icon, ...castArray(children)] : children`, avoids passing React a nested children array (key warnings) and leaves the no-icon path untouched.

amcclain (commented):
Great thanks for getting this in.

Issue #4522, opened 2026-07-29
Store memory/load-time tuning: evidence base and decision rubric


amcclain:
Working state for the Store memory/load-time tuning effort — the evidence base behind
[#4521](https://github.com/xh/hoist-react/pull/4521) and the raw material for a tuning guide that
does not exist yet.

**This issue is the durable home for these notes.** They were a local scratch file; parking them here
so the work can be picked up on any machine. Numbers stay here rather than in shipped docs — they are
environment-specific and go stale, and the harness can re-run them.

All six levers are now measured. Jump to **Initial pass: complete** for the summary table, or **The
organising idea** / **The rubric** for the conclusions.

## The two goals

1. **Decide whether the hoist-react change delivers a real, measurable, releasable win.**
   **Resolved: yes.** 128.6 MB / 28.4% at the baseline shape, replicated, and the saving holds within
   a 7.6 MB band across four value distributions. See the anchor and data-character sections.
2. **Build an evidence-backed rubric for the interrelated Store memory/load-time flags**, useful
   independent of that one change. This is now the live work. Initial pass done; see "Initial pass:
   complete" for what is measured and what is not.

---

## Repo state

Both repos on branch **`store-fixed-data-shape`**, both fully pushed, both 0 behind `origin/develop`.

- **hoist-react** — PR [xh/hoist-react#4521](https://github.com/xh/hoist-react/pull/4521).
  Net diff vs develop: `data/Store.ts`, `data/StoreRecord.ts`, `data/README.md`, `CHANGELOG.md`,
  `yarn.lock`.
- **toolbox** — PR [xh/toolbox#884](https://github.com/xh/toolbox/pull/884). Its
  `client-app/tsconfig.json` deliberately points `@xh/hoist` at the local `../../hoist-react`
  checkout, because the tester sets a config not yet in a published snapshot. **Re-comment before
  merge** — flagged in the PR as a merge blocker.

Superseded and closed: hoist-react #4501 and #4500, toolbox #879. Their branches are deleted. The
config was renamed from `optimizeRecordData` to `useFixedDataShape` along the way, so anything
referring to the old name predates this state.

**Running locally:** Grails on :8080 (`./gradlew bootRun`), webpack on :3000 from `client-app/`
via `yarn startWithHoist` (inline hoist-react). Admin → Tests → Grid. Chrome **must** be launched
with `--js-flags="--expose-gc"` — the harness reports which mode it used on every row, check it.

---

## What the change is

[...]

PR #4534, opened 2026-08-03
Store: automatic record-data memory optimization, `projectionOnly` + digest-based `reuseRecords` - a simplified synthesis of the v87 store work

### LINKED ISSUE #4502: Cube: ViewRowData row objects fall into V8 dictionary mode - per-View codegen'd assigner measures ~5-6x smaller rows at typical view widths

### LINKED ISSUE #4517: Grid: ag-Grid `gridOptions.rowData` pins the record array from the last load into an empty grid
`GridLocalModel.syncData()` sets ag-Grid's `gridOptions.rowData` only when syncing into an
empty grid:

https://github.com/xh/hoist-react/blob/develop/cmp/grid/Grid.ts#L696-L705

Every subsequent sync goes through `applyTransaction()`, which updates ag-Grid's internal row
model but leaves `gridOptions.rowData` holding the `StoreRecord[]` it was handed on that first
load. ag-Grid retains that array for the life of the grid, so it pins:

- the record array itself,
- every `StoreRecord` in it,
- each record's `data` object, and its `raw` reference if `retainRaw` is true (the default).

Net effect is that a grid retains one extra generation of records indefinitely - the generation
from the last load into an empty grid. It does not grow without bound, but it does not shrink
either, and clearing the store does not release it: a store load that empties the grid takes the
`applyTransaction` branch (a `remove` transaction), so `rowData` still references the old array.

Observed while benchmarking large stores in Toolbox's grid tester - on a 50k-record grid with wide
records the retained generation showed up as tens of MB, and the retainer chain in a Chrome heap
snapshot ran through ag-Grid's stored grid options. Reproducible by loading a large grid, reloading
it with fresh data, and taking a snapshot: the first generation's records are still reachable.

### Notes toward a fix

The obvious move - re-set `rowData` after `applyTransaction()` - needs checking before it is
adopted. In ag-Grid, `updateGridOptions({rowData})` is a data-setting operation, so calling it on
every sync may discard the transaction's incremental semantics (row identity, selection, sort
state, animation) that the transaction path exists to preserve. Verify against ag-Grid 35 behavior
rather than assuming.

Narrower options worth evaluating:

- Set `rowData: []` when the new `RecordSet` is empty. Fixes only the cleared-grid case, but that
  is the case where the retention is most obviously wrong.
- Set `rowData: []` once immediately after the initial load, so ag-Grid's row model owns the rows
  and grid options hold nothing. Needs confirmation that ag-Grid does not re-read `rowData` later.

Filing for triage rather than proposing a specific patch - the retention is confirmed, the safe
remedy is not.

### LINKED ISSUE #4519: Cube View: bucket dependentFields outside Query.fields force a full update on every tick
A Cube `View` forces a **full update on every tick** - never taking the incremental `dataOnlyUpdate` path - if any `BucketSpec.dependentFields` entry is not also present in the View's `Query.fields`.

Silent: no warning, no error, correct results. The only symptom is that every update re-aggregates and re-loads the whole View.

### Mechanism

`View.hasDimOrBucketUpdates()` compares two objects with different shapes:

```ts
// View.ts:512-516
const fieldNames = uniq([...dimensions.map(it => it.name), ...bucketDependentFields]);
for (const rec of update) {
    const curRec = this._leafMap.get(rec.id);
    if (fieldNames.some(name => rec.data[name] !== curRec.data[name])) return true;
}
```

- `rec` is a `StoreRecord` from the **Cube's own store** - its `data` carries every field declared on the Cube.
- `curRec` is a `LeafRow` from `_leafMap` - and `LeafRow` projects **only `view.fields`** onto its data:

```ts
// row/LeafRow.ts:41-43
view.fields.forEach(({name}) => {
    this.data[name] = rawRecord.data[name];
});
```

So for any `fieldNames` entry outside the query, `curRec.data[name]` is permanently `undefined` while `rec.data[name]` holds a real value. The comparison is unconditionally `true`, `getSimpleUpdates()` returns `false` (View.ts:477-479), and `noteCubeUpdated()` takes the `fullUpdate()` branch (View.ts:273-275) on every tick, forever.

Dimensions are safe - `Query` guarantees dimension fields are query fields. Bucket `dependentFields` carry no such guarantee: they are collected independently into `_bucketDependentFields` (View.ts:441) from whatever the `BucketSpecFn` returns.

Note this is the inverse of what `dependentFields` is designed to prevent. Its documented purpose (`BucketSpec.dependentFields`) is avoiding a *false negative* - missing a re-bucket that should have happened. This is a permanent *false positive*.

### Reproduction

A Cube with a `bucketSpecFn` whose spec declares `dependentFields: ['someComputedField']`, where `someComputedField` is a Cube field but is **not** in the connected View's `Query.fields`. Then `cube.updateDataAsync(rows)` with rows whose dimension and bucket values are all unchanged.

Observed by instrumenting `View.prototype.fullUpdate` / `dataOnlyUpdate` / `getSimpleUpdates` and driving one tick of 1,973 updated rows against a 31k-record / 174-field Cube feeding a View of 35 fields:

```
{ full: 1, dataOnly: 0, simpleFalse: 1, aggregatorsAreSimple: true }
```

Every tick cost ~3,400ms - effectively constant regardless of how many rows were updated (1,412 through 5,089), matching the View's full-update cost rather than any incremental one. Adding the dependent field to `Query.fields` fixed it: ticks dropped to 272-727ms and scaled with the number of rows updated, as expected.

### Impact

For a streaming Cube app this is potentially an order-of-magnitude difference in per-tick cost, and it scales with total View size rather than with the size of the update. It is easy to hit, since it is natural to bucket on a computed field that the grid never displays and therefore never made it into the query.

### Suggested fixes

Preference for the second - it fixes the comparison rather than constraining the caller:

1. Have `Query` validate that all bucket `dependentFields` appear in `fields` (throw or warn), or auto-include them.
2. Compare like against like in `hasDimOrBucketUpdates` - resolve the previous value from the Cube's prior `StoreRecord` (`_recordMap`) rather than from the query-projected `LeafRow`. This also makes bucketing on non-queried fields work as an app would reasonably expect.

Either way, worth a warning when a `dependentField` is absent from the projection, since the current failure is silent.

Found while benchmarking #4516 - unrelated to that change.

lbwexler:
Supersedes #4521. Synthesizes the ongoing v87 store memory-optimization work (#4516, #4521, #4522) with a focus on **simplifying the API**: the framework now makes the representation decisions those PRs asked app developers to make, and the new-config surface drops to zero.

### Record data: automatic per-record representation

`Store` now chooses each record `data` object's internal representation automatically, counting the fields holding non-default values in a single parse pass:

- **Sparse** (own properties for non-default values only, defaults via a shared prototype) for records below the threshold - the pre-v87 form, cheapest for lightly-populated records.
- **Fixed shape** (clone of a shared per-Store template carrying every Field) for records populating 17+ fields - avoiding the ~3x V8 "dictionary mode" demotion that previously hit wide records, per the measurements in #4521 / #4522.

This replaces #4521's opt-in `useFixedDataShape`, its 20-field guidance table, and its advisory warnings - there is nothing left for apps to configure or be warned about. Wide-and-sparse stores, the one shape the opt-in flag could not serve, automatically keep the cheap sparse form.

`parseUpdate()` now rebuilds modified records through the same path, so locally-modified records no longer drop into dictionary mode (they always did pre-v87), and a record's representation is a pure function of its content - as the deep-equal revert-to-clean checks in `modifyRecords()` require.

**New contract (breaking):** `StoreRecord.data` should be read by field name only, never enumerated - `Object.keys()`, spread and `JSON.stringify()` do not reliably see every field. Use `getValues()` / `getModifiedValues()`. Sparse records enumerate exactly as pre-v87; only densely-populated records change.

An experimental `denseRecordThreshold` config overrides the cutover for testing/tuning - e.g. `999` restores the pre-v87 all-sparse behavior.

### `projectionOnly` replaces `useRawAsData`

Renamed and reframed around intent: a read-only projection of data parsed and owned elsewhere - recommended for stores connected to a Cube `View`, where re-parsing already-final rows is pure overhead. Local modification APIs now throw in this mode, restoring the original read-only contract and eliminating the dirty-tracking edge cases raw adoption had under local edits. `View` logs a debug hint when connected stores don't set it. Both `useRawAsData` and `useFixedDataShape` are unreleased v87 work, so no deprecations are needed.

### Cube `View` rows: fixed-shape templates (merged from #4535)

`ViewRowData` converted from a class to an interface, with all View rows cloned from a shared per-View template so they hold one compact V8 shape - the row-level counterpart of the record-data work above. Aggregate-only views (no `includeLeaves`/`provideLeaves`) now read leaf data zero-copy from Cube records, eliminating per-View leaf copies; such views no longer publish `result.leafMap` (breaking - see CHANGELOG). See #4535 for full details and measurements.

### `reuseRecords`: digest-based record reuse

`Store.reuseRecords` extended from `boolean` to `boolean | string | ((raw) => RecordDigest)`:

- `true` - unchanged: reuse on raw object identity. For sources with stable, immutable row objects.
- string - the digest is the named raw property, e.g. a server-provided timestamp or sequence number.
- function - a derived digest for composite keys, built as a string - e.g. `raw => raw.type + '|' + raw.seq`.

Digests are primitives - a new exported `RecordDigest` type (`string | number`) - snapshotted on the record when built (exposed as new public `StoreRecord.digest`) and compared with a single `===`. No deep equality and no object digests, so comparisons are cheap, digests log/inspect cleanly, and the mutate-in-place aliasing hazards of object-valued stamps are unrepresentable. A raw yielding an unchanged digest for its id reuses the existing record outright, skipping raw processing, parsing, and construction; nil digests never match, guarding misconfigured fields.

Applies to `loadData()` and `updateData()` alike - updates yielding an unchanged digest are dropped from the transaction as no-ops, so bulk update batches containing equivalent rows shrink to their genuine delta. Load-path misses still fall back to the standard fieldwise comparison. The string/function forms also compose with `retainRaw: false` and `projectionOnly` (previously blocked pairings).

**Cube `View` stores are configured automatically.** A View installs its own digest on every connected store (via an internal `Store.setDigestFn()` hook), reading `ViewRowData.cubeRowDigest` - the monotonic stamp the View maintains on every row it mints or mutates in place. There is nothing for apps to configure and no magic string in app code: `reuseRecords` must be left unset on connected stores, and any explicit value throws at connection, since neither reference identity nor a custom digest can stay correct against the View's mutate-in-place update channel.

Also hardened the projection-store equality check in `RecordSet`: record comparison now considers declared Field values only, rather than deep-equaling provider row objects whose nested `children`/leaf graphs made the old check O(subtree) per record on every cube reload.

### Cube `View` row reuse across structural updates and reloads

Previously any add/remove or full reload cleared the View's row cache outright, re-minting every row (and every row digest), so connected stores rebuilt every record even with digest reuse enabled. Now:

- New internal `RowCache` validates every cached row on lookup - leaves by source `StoreRecord` identity (records are immutable, so identity proves data unchanged), aggregates/buckets by child-row identity. A change rebuilds only the changed rows and their ancestor chains; everything else carries over with digests intact, skipping re-aggregation of untouched subtrees. Views with complex (non-`dependsOnChildrenOnly`) aggregators now reuse rows throughout - aggregate/bucket values (which may read the per-generation `AggregationContext`) are re-derived in place on reused rows each generation, with digests bumped and store records rebuilt only for values that actually changed. Previously such views got no reuse at all.
- Entries untouched by a generation are retained - a filter narrow → widen round-trip reuses the filtered-out subtrees verbatim. When the cache has outgrown the generation's live rows by 50%+ (and cube data has changed since the last sweep), a reachability-based sweep drops exactly the rows whose underlying records were removed or replaced - bounding dead entries, and the replaced records they pin, to a fraction of the view's working set. Per-generation cache stats (reused/rebuilt/created/removed/sweepTime) log at debug level.
- New `CubeConfig.reuseRecords` passes through to the Cube's internal store, preserving record identity across full `loadDataAsync()` reloads and extending reuse to wholesale refreshes.
- `ViewRowData.cubeBuckets` is now derived from each row's current ancestor `BucketRow`s on every generation (`BaseRow.syncBuckets`), replacing construction-time stamping. Fixes stale bucket values on reused rows (e.g. stale `Closed` tags on trades under a reopened position) with no purity constraints on `bucketSpecFn`.

Measured (30k leaves, 2-dim view, connected `projectionOnly` store with the View-installed digest): structural add/remove of 10 rows ~60ms → ~28ms, with store record construction dropping from ~30k to ~31 per update; 1% reload 52ms → 31ms (30k → 614 builds); no-op reload 57ms → 31ms (0 builds). A 28-assertion node smoke suite covers structural/data-only updates, reloads, filter round-trips, bucket-exit self-healing, raw-identity reuse, and the connected-store config contract; Toolbox Cube-tester validation of this increment to follow.

### Also included

- Fixed `StoreRecord.getModifiedValues()` omitting fields locally modified back to their default value - a latent bug in all released versions. Now iterates Fields rather than `data`'s own keys, so it behaves identically under every representation.
- A codegen'd per-Store constructor was built and measured as an alternative to template cloning - equivalent on memory, so removed in favor of the simpler clone.

### Validation

- `npx tsc --noEmit` and `yarn lint` clean.
- Runtime invariants verified: routing determinism between load and update paths, modify → revert deep-equal on both forms and across a threshold flip, the sparse invariant (no own default-valued keys), and `getModifiedValues()` on sparse records.
- The companion Toolbox PR (xh/toolbox#886) exposes the threshold knob for continued in-app A/B against the #4521 / #4522 measurement protocol, and its Cube tester adds a `Reuse Records` toggle (default on, matching the automatic View-installed digest) with a live record-instance-survival readout - toggling off restores a no-reuse baseline for A/B comparison. Verified in-app: streaming ticks render correctly with reuse enabled (no swallowed updates - every value change takes a fresh record instance), unchanged rows retain record identity across updates and filter changes, and the fieldwise fallback preserves reuse across full cube reloads.

### Known trade-off: row-cache retention across disjoint filter flips

One scenario where this work can consume *more* memory over time: the row cache deliberately retains entries a generation didn't touch (so filter narrow → widen round-trips reuse subtrees verbatim), and the sweep drops only entries that can no longer validate - so an app cycling a filter across *disjoint* subsets (e.g. a fund selector) accumulates still-valid dormant rows for every subset visited. It's memory-only (a disjoint flip gets no cache hits either way, so rebuild cost is unchanged) and bounded - within one grouping shape, ids are unique against the full unfiltered tree, so worst case ≈ one unfiltered view build; shape changes already clear outright. But for cubes only ever viewed filtered *because* the full view is too big, that bound is the thing being avoided. Options considered: an explicit `View.resetRowCache()` (or a `resetRowCache` option on `updateQuery`) as a sanctioned context-switch hint - cheap and impossible to hold wrong, since dropping the cache never costs correctness, only one full-cost regeneration; automatic staleness detection was rejected, as it reintroduces the arbitrary age constants the sweep design deliberately avoids. Deferring until a real app hits it - and note a workaround exists today: `updateQuery` clears the cache whenever anything beyond the filter changes, and its function props compare by reference, so passing a fresh-but-identical `lockFn` alongside the new filter forces the drop.

---

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (The `data` enumeration contract and `leafMap` availability - see above and CHANGELOG.)
- [x] Created Toolbox branch / PR, or determined not required. (xh/toolbox#886)


data/Store.ts
```diff
@@ -128,49 +145,70 @@ export interface StoreConfig {
...
      *
      * By default, Store reuses existing StoreRecord instances when new data is loaded with
-     * matching IDs and identical field values (determined via deep equality comparison). This
+     * matching IDs and identical field values (determined via equality comparison). This
      * preserves row state in grids for unchanged records.
      *
-     * When `reuseRecords` is true, the Store skips the fieldwise comparison and instead reuses
-     * records when the raw data object itself is **reference-identical** to the previously loaded
-     * object. This avoids equality checks, record creation, and raw data processing overhead.
+     * Set to instead derive a *version* from each incoming raw object, snapshotted on the record
```
amcclain: Set to function or string.... 

Questioning term "version" as it carries some baggage - might imply an actual record versioning / audit trail kind of system in an unwanted way.  Is this more like a fingerprint?

PR #4537, opened 2026-08-04
Do not add newly-released columns to user-curated grid views

amcclain:
Columns added to the code after a user saved grid state are always merged into that state. Whether they arrive **visible** was governed solely by their in-code `hidden` config. That is right for state recording a user's last-used layout, but wrong for a named view the user curated deliberately, often to capture a specific report.

New columns are now initially **hidden** when column state is persisted to a `ViewManagerModel` or `DashViewModel`, and unchanged everywhere else.

| Provider | New column visibility |
|---|---|
| `viewManager`, `dashView` | Hidden |
| All others | Follows in-code `hidden` config (unchanged) |

## Changes

* Added `GridModelPersistOptions.hideNewColumns`, defaulted from the provider resolved for `persistColumns` (falling back to the root options). Set `false` to restore the prior behavior.
* New columns remain available in the column chooser, and do not mark a view dirty - the existing `PersistableColumnState.equals` override already excludes hidden columns from that comparison.
* Columns with `hideable: false` (including tree columns) or `excludeFromChooser: true` are never force-hidden, as the app requires them shown and/or a user could not restore them.
* A ViewManager's in-code default view still follows the code. Apps that disable it in favor of a globally-shared default get the curated behavior throughout, with new columns surfacing only once an admin adds and saves them.
* `GridModel.setColumnState` accepts an optional `ColumnStateOptions`. Documented its wholesale-replace contract, and documented that `setColumns` resets column state.
* Exposed `PersistenceProvider.parseProviderClass` as `@internal`, so callers can resolve the provider a set of options would produce without duplicating that precedence.

## Verification

Run against Toolbox with hoist inline, driving the Portfolio example app: positions grid on a ViewManager, orders grid on a `DashViewModel` nested in that same ViewManager.

* Both grids hide new columns on a saved view, leave the stored view untouched, and stay clean (`isValueDirty: false`) - including with auto-save enabled and on a runtime view switch, the path that does write.
* New columns appear in the chooser's available list.
* The in-code default view surfaces new visible columns, including when it carries unsaved changes from before the release.
* `hideable: false` and `excludeFromChooser: true` columns arrive visible.
* A localStorage-backed grid surfaces new visible columns while preserving saved widths and hidden state.
* `hideNewColumns: false` on a ViewManager restores the prior behavior.

## Not in scope

Testing surfaced a pre-existing inconsistency in when a newly visible column marks a view dirty: a runtime view switch writes and dirties, while a cold load directly onto a saved view does not, because `PersistenceProvider.bindToTarget` installs its reaction after applying persisted state. Unrelated to this change, and moot under the new default.

### DIFF
--- a/cmp/grid/GridModel.ts
+++ b/cmp/grid/GridModel.ts
@@ -114,6 +114,7 @@ import {managedRenderer} from './impl/Utils';
 import {
     ColChooserConfig,
     ColumnState,
+    ColumnStateOptions,
     GridModelPersistOptions,
     GroupRowRenderer,
     RowClassFn,
@@ -1244,6 +1245,15 @@ export class GridModel extends HoistModel {
         this.store.clear();
     }
 
+    /**
+     * Replace the columns for this grid, rebuilding all `Column` instances from the configs
+     * provided.
+     *
+     * Note this resets all column state - visibility, width, order, and pinning - to the defaults
+     * specified by the new configs. For a grid with persistence enabled, that reset is itself
+     * persisted, discarding any state the user had saved. Use {@link setColumnState} or
+     * {@link updateColumnState} to change how the *existing* columns are displayed.
+     */
     @action
     setColumns(colConfigs: ColumnOrGroupSpec[]) {
         colConfigs = this.enhanceColConfigsFromStore(colConfigs);
@@ -1255,8 +1265,15 @@ export class GridModel extends HoistModel {
         this.columnState = this.getLeafColumns().map(it => this.getDefaultStateForColumn(it));
     }
 
-    setColumnState(colState: ColumnState[]) {
-        this.columnState = this.cleanColumnState(colState);
+    /**
+     * Replace the current column state wholesale with the state provided.
+     *
+     * Note that any columns missing from `colState` will be restored to their in-code default
+     * state, or hidden if `opts.hideNewColumns` is set - this method does not patch the existing
+     * state. Use {@link updateColumnState} to apply targeted changes to particular columns.
+     */
+    setColumnState(colState: ColumnState[], opts?: ColumnStateOptions) {
+        this.columnState = this.cleanColumnState(colState, opts);
     }
 
     showColChooser() {
@@ -1794,7 +1811,7 @@ export class GridModel extends HoistModel {
         );
     }
 
-    private cleanColumnState(columnState) {
+    private cleanColumnState(columnState, opts?: ColumnStateOptions) {
         const gridCols = this.getLeafColumns();
 
         // REMOVE any state columns that are no longer found in the grid. These were likely saved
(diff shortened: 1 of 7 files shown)


cmp/viewmanager/README.md
```diff
@@ -391,16 +391,21 @@ class PortfolioGridModel extends HoistModel {
             persistWith: {
                 viewManagerModel: XH.appModel.portfolioGridViewManager,
                 path: 'grid',
+                persistColumns: true,
                 persistSort: true,
-                persistGrouping: true,
-                persistFilter: true,
```
RyanMG, line 396: ~~Was persistFilter outdated (and thus removed)?~~

This appears to be the case - I see no other references in the code to "persistFilter"

RyanMG (approved):

cmp/viewmanager/README.md
amcclain: Yes we caught a few docs / commenting bugs as part of this change, this appears to be one of them.

cmp/grid/GridModel.ts
```diff
@@ -1805,7 +1822,15 @@ export class GridModel extends HoistModel {
         // Insert these columns in position based on the index at which they are defined.
         gridCols.forEach((col, idx) => {
             if (!find(ret, {colId: col.colId})) {
-                ret.splice(idx, 0, this.getDefaultStateForColumn(col));
+                const state = this.getDefaultStateForColumn(col);
+
+                // Hide new columns if so requested - but never those the app requires to be shown,
+                // or that a user could not restore for themselves via the column chooser.
+                if (opts?.hideNewColumns && col.hideable && !col.excludeFromChooser) {
```
lbwexler, line 1829: why the last clause?  Seems like the app could theoretically provide another way for adding it -- why should the idea that it should not be added automatically no longer make sense here.

cmp/grid/Types.ts
```diff
@@ -41,6 +41,19 @@ export interface ColumnState {
...
 }
 
+/** Options for {@link GridModel.setColumnState}. */
+export interface ColumnStateOptions {
+    /**
+     * True to hide any columns missing from the provided state, regardless of their in-code
+     * `hidden` config. Default false, respecting that config.
+     *
+     * Columns configured with `hideable: false` or `excludeFromChooser: true` are never hidden by
+     * this option, as the app has indicated that they must be displayed and/or cannot be restored
```
lbwexler, line 51: see above re: excludeFromChooser  and whether it should factor in to this.

cmp/grid/GridModel.ts
amcclain: I thought about that for a while, as the `hideable:false` signal is clearly stronger and this one seemed optional.

The expectation driving that last clause is that the app is taking away the ability of the user to manage that col's viz via the known-available route, and also leaving it visible.  This is common for e.g. status and action cols. Ideally the dev would also set hideable:false but that might be forgotten / assumed not needed (it's visible and col chooser control to hide is disabled). Ideally dev would set hideable:false but if you looked across our code I wonder how often that's done.

The failure case (new col added, set to viz, excluded, never appears in saved views and no way for user to add or even know it's there) seemed bad enough to warrant the additional automatic switching here.

amcclain: But I should say I am fine with taking this out if it unblocks the PR - devs just need to be aware of importance of setting `hideable:false` explicitly for always-on cols and we should prob doc comment accordingly

lbwexler (approved):

cmp/grid/GridModel.ts
lbwexler: all set -- argument makes sense

PR #4550, opened 2026-08-07
Sort fields by name to ensure canonical order for equality checks

### LINKED ISSUE #4547: `View.updateQuery` may unnecessarily rebuild for a functionally identical config
When constructing a `Query` we augment the `fields` with the `dimensions`, but we spread them in _after_ we call `this.parseFields` on the provided `fields`. `parseFields` is effectively giving us a stable sort of the fields, but we then spread the `dimensions` at the end.

The case where this breaks is if you update a query by including a field in `fields` that is also in `dimensions` - this will result in a different order for `Query.fields` due to the `uniq` that is applied, which will result in `Query.equalsExcludingFilter` failing since it does an `isEqual` on the `fields` of the two `Query` objects.

Pretty unlikely to happen, but if the app code is building the field list from currently visible grid columns, and a user happens to add a column for a field that is already in their dimensions, then it would trigger a full rebuild (and potentially a row cache clear) unnecessarily since the field was already included in the current result.

The fix is very simple - we just need to ensure that `Query.fields` are sorted by name in the constructor.

haynesjm42:
Fixes #4547 

Example repro - logs false before the change, true after.

```typescript
const cube = new Cube({
    fields: [
        {name: 'region', isDimension: true, aggregator: 'UNIQUE'},
        {name: 'trader', isDimension: true, aggregator: 'UNIQUE'},
        {name: 'pnl', aggregator: 'SUM'}
    ]
});

const view = cube.createView({query: {dimensions: ['region'], fields: ['pnl']}}),
    before = view.result;

// User adds a column for an existing dimension
view.updateQuery({fields: ['region', 'pnl']});   

// Expect no full rebuild due to functionally equivalent queries
console.log(view.result === before);
```

Hoist P/R Checklist
-------------------


lbwexler:
Great catch @haynesjm42  -- this one was totally defeating some of our new optimizations!

PR #4551, opened 2026-08-07
Use raw fields when cloning to drop stale dimensions

### LINKED ISSUE #4548: `Query.cloneConfig` returns augmented `fields` which may include stale dimensions
We augment `Query.fields` with the `dimensions` to ensure they are included in the query and the caller doesn't need to include them manually in the `fields` list, but we then include the augmented `fields` when creating a clone with overrides from an updated query, which may be a dimension-only update. This will result in the cloned `Query.fields` including `dimensions` from the previous config that were not included in the overrides.

Since dimension fields use the `UNIQUE` aggregator this means we are aggregating the stale dimension fields for every aggregate row even though it wasn't requested and likely isn't displayed in the UI.

We need to maintain the raw `fields` separately from the augmented ones so that `cloneConfig` can accurately create a clone on a dimension-only query update.

haynesjm42:
Fixes #4548 

```typescript
import {Cube, Query} from '@xh/hoist/data';

const cube = new Cube({
    idSpec: 'id',
    fields: [
        {name: 'region', isDimension: true, aggregator: 'UNIQUE'},
        {name: 'trader', isDimension: true, aggregator: 'UNIQUE'},
        {name: 'pnl', aggregator: 'SUM'}
    ],
    data: [
        {id: 1, region: 'EMEA', trader: 'ann', pnl: 100},
        {id: 2, region: 'EMEA', trader: 'ann', pnl: 200},
        {id: 3, region: 'APAC', trader: 'bob', pnl: 50}
    ]
});

// 1. The stranded field.
const q = new Query({cube, dimensions: ['region'], fields: ['pnl']});
console.log(q.fields.map(f => f.name));                       // ['pnl', 'region']

const q2 = q.clone({dimensions: ['trader']});
console.log(q2.fields.map(f => f.name));
//   BEFORE: ['pnl', 'region', 'trader']   <- 'region' is no longer a dimension
//   AFTER:  ['pnl', 'trader']

// 2. It isn't inert - the stale field is aggregated on every row.
const view = cube.createView({query: {dimensions: ['region'], fields: ['pnl']}});
view.updateQuery({dimensions: ['trader']});

const annRow = view.result.rows.find(it => it.cubeLabel === 'ann');
console.log('region' in annRow, annRow.region);
//   BEFORE: true 'EMEA'    <- UNIQUE ran across every child, for a field nothing requested
//   AFTER:  false undefined
```

Hoist P/R Checklist
-------------------

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,9 @@
 * Fixed `Grid` retaining an extra generation of records in memory indefinitely - ag-Grid's stored
   `rowData` pinned the record array from the last load into an empty grid, along with every
   `StoreRecord`, `data`, and retained `raw` object in it.
+* Fixed `Query.clone()` retaining dimensions dropped by a dimension-only update within its `fields`.
+  Cube `View`s changing dimensions via `updateQuery()` accumulated these stale fields, aggregating
+  each one on every aggregate row despite nothing having requested or displayed them.
 
 ### ⚙️ Technical
 
--- a/data/cube/Query.ts
+++ b/data/cube/Query.ts
@@ -166,7 +166,6 @@ export class Query {
         omitFn = cube.omitFn
     }: QueryConfig) {
         this.cube = cube;
-        // Retained for the life of this Query and its clones - copy to insulate from the caller.
         this._rawFields = fields?.slice();
         this.dimensions = this.parseDimensions(dimensions);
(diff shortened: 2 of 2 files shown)


lbwexler:
Great catch -- not sure how we lived with this one as long as we did.  Tweaked slightly to avoid holding references to the inputs.

PR #4556, opened 2026-08-09
Migrate repo to pnpm 11 (from yarn classic)

amcclain:
## Summary

Migrates this repo's own package management from yarn classic to pnpm 11, matching hoist-dev-utils and the Toolbox pnpm branch. Apps consuming the published `@xh/hoist` package are unaffected - pnpm, yarn classic, and npm all remain fully supported.

- `pnpm-lock.yaml` replaces `yarn.lock`; pnpm 11.21.0 pinned via `packageManager`; vendored yarn binary and `.yarnrc` removed.
- New `pnpm-workspace.yaml` holds reviewed dependency build scripts, an `@xh/*` minimum-release-age exemption, and React 19 peer acknowledgements.
- Declared packages this library imports directly that had resolved only via yarn's flat hoisting - notably `sass-material-colors`, which broke `inlineHoist` under pnpm. Each newly-declared package takes the version we want rather than the one hoisting happened to supply, which moves `globals` to 17.x.
- CI and both deploy workflows run on pnpm with frozen-lockfile installs, and `pnpm audit --prod --audit-level high` replaces the yarn audit exit-code bitmask.
- Added a `typecheck` script and CI step. Linting never reported compiler diagnostics, so the only `tsc` gate was `prepack` at publish time - a type error reached `develop` and surfaced as a failed deploy rather than a failed PR check.
- Docs, README, CLAUDE.md, husky pre-commit, `.npmignore`, and skills updated - pnpm is now the reference package manager throughout, with yarn/npm acknowledged.

---

## Detail

- **Lockfile + pin** - fresh resolve at the latest versions allowed by existing specs; `@xh/hoist-dev-utils` devDep now `^14.0.0`.
- **No `overrides`** - the three yarn `resolutions` (`@types/react`, `@types/react-dom`, `core-js`) were each verified to be no-ops here and dropped. Removing them changes no resolved version; the only lockfile churn is peer-range strings reverting to their published values.
- **Phantom deps** - `sass-material-colors` (imported by shipped SCSS), `@azure/msal-common` and `unified` (type-only), plus dev-time `@types/lodash`, `@xh/eslint-config`, `globals`, and `stylelint-config-recommended-scss` (replacing the declared-but-unused `standard-scss` variant).
- **Versions for those** - `globals` is imported by our own `mcp/eslint.config.mjs` for `globals.node`; nothing constrains it, so it takes 17.x (v17's only breaking change splits `audioWorklet` out of `browser`, leaving `node` untouched). The remaining newly-declared packages are already at latest, except `@azure/msal-common`, deliberately held at `~16.11.3` because `@azure/msal-browser` pins that exact version and widening it would duplicate a runtime dependency.
- **Two source fixes surfaced by the stricter graph** - removed a bogus `Dictionary` import from `express-serve-static-core` in `ViewMenu.ts`, and reworked `FetchService`'s URL redaction onto `URLSearchParams.forEach` so shipped source type-checks in app programs that lack `dom.iterable` or `@types/node`.
- **Type-check gate** - `typecheck` is deliberately outside the `lint:*` namespace and outside `lint:all`, keeping linting and type-checking as separate concerns; every `lint:*` script stays a linter and `lint:all` runs all of them. The deploy workflows need no equivalent step, since `npm publish` type-checks via `prepack` / `tsc --build`.

## Verification

- **CI green** on the tip: frozen-lockfile install, `pnpm lint`, `pnpm typecheck`, and the `--audit-level high` production audit gate.
- **Snapshot deploy from this branch** end-to-end: [run 31432829388](https://github.com/xh/hoist-react/actions/runs/31432829388) published `87.0.0-SNAPSHOT.1786396460928` to the `next` tag and fired the downstream Toolbox dispatch. Log confirms `prepack` / `tsc --build` ran during publish.
- **`inlineHoist`** verified from the Toolbox pnpm branch: clean webpack compile against this checkout, with `/app`, `/admin`, and `/mobile` all served.
- **New gate proven to gate**: a probe file with `const n: number = s` (where `s: string`) passes `pnpm lint` with exit 0 and fails `pnpm typecheck` with `error TS2322`.


amcclain:
Paired with xh/toolbox#887, which migrates Toolbox's `client-app` to pnpm 11. The two should land together: Toolbox's `startWithHoist` now runs `pnpm install` in the sibling hoist-react checkout, so inline development needs both.

Inline mode (pnpm Toolbox against this branch) is verified working - clean compile, and desktop/admin/mobile with grids and charts all rendering.

amcclain:
Heads-up for reviewers: this PR includes one small source fix that is independent of the pnpm migration and should not get stuck behind it.

`desktop/cmp/viewmanager/ViewMenu.ts` was importing `Dictionary` from `express-serve-static-core` - an IDE auto-import mishap dating to #3850. Express typings have no business in hoist code; the import only resolved because yarn's flat layout hoisted `@types/express-serve-static-core` in from dev-utils' webpack-dev-server chain. Now typed as the structurally-identical `Record<string, ViewInfo[]>` (no behavior or type-shape change).

If this PR is delayed in review/merge, that one-line fix is worth cherry-picking to `develop` on its own - any consumer type-checking against a stricter or isolated layout would hit it as a hard compile error.

PR #4565, opened 2026-08-11
Fix GridModel.leafColumnMap staleness

lbwexler:
One fix from an API review of #4409. Branched from that PR's head (`dba1613d`) and targeted here, since `new-column-chooser` itself is no longer on origin - retarget if that's wrong.

**`GridModel.leafColumnMap` is now a `@computed` over `columns`**, rather than a field rebuilt alongside it in `setColumns`. `columns` is public and writable, so any assignment outside `setColumns` left the map stale and `getColumn()` returning `null` for every column - silently breaking `setGroupBy`, `setSortBy`, `noteAgColumnStateChanged`, and `cleanColumnState`.

Marked `keepAlive` because `getColumn()` has hot callers outside any reaction (notably `cleanColumnState`, which calls it once per column); a plain `@computed` would not cache there and would undo the indexed-lookup speedup. That's a new pattern in this repo - happy to drop it if you'd rather keep the plain field.


lbwexler:
Superseded by #4566, which targets develop now that #4409 has merged.

PR #4573, opened 2026-08-14
Roles Admin: display names + search for directory groups (Entra ID support)

### LINKED ISSUE #4572: Roles Admin: display names + search for directory groups (Entra ID support)
Hoist Core v41 (in development, branch `entra-id-service`) adds `EntraIdService` - an opt-in alternative to `LdapService` for resolving directory-group role memberships from Microsoft Entra ID via the Graph API. Both services now implement a common `DirectoryService` interface consumed by `DefaultRoleService`.

A key difference with Entra ID: directory groups are identified by their **object ID (a GUID)** rather than an LDAP DN. GUIDs are stable through group renames but unreadable, so the Roles admin UI needs support for resolving and displaying friendly names. Two new server endpoints back this:

- `roleAdmin/directoryGroupsInfo?names=...` - resolves display info for assigned directory groups. Returns a map of group identifier to either `{id, displayName, description, ...}` or a string error description (e.g. group not found).
- `roleAdmin/searchDirectoryGroups?query=...` - searches the directory by partial name. Returns a list of `{id, displayName, ...}` objects. Gated at `HOIST_ROLE_MANAGER` (vs `HOIST_ADMIN_READER` for the info endpoint), matching the edit-flow it supports.

## Desired changes

1. **Show display names for assigned directory groups.** Wherever the Roles admin UI lists a role's directory groups (grid, detail panel, form), resolve and show the `displayName` alongside or in place of the raw stored identifier, via `directoryGroupsInfo`. Error strings (group not found, lookup failure) should surface visibly on the affected group. This also improves the LDAP case, which today shows truncated DNs - the endpoint returns `displayName` = CN for LDAP groups.

2. **Search-based group picker when adding directory groups to a role.** In the role editor, offer a typeahead/select backed by `searchDirectoryGroups` so admins can find a group by name and have its identifier stored. Free-text entry must remain supported (paste a known GUID or DN directly).

3. **Keep using the existing `directoryGroupsDescription` hint** - it is now provider-specific server-side ("enter a DN" for LDAP vs "enter an object ID" for Entra ID), so no client change needed beyond continuing to display it.

## Notes

- Both endpoints throw (400, routine) if the app is not using `DefaultRoleService` - but the Roles module is already gated on `config().enabled`, so this should not occur in practice.
- Availability/behavior when no directory service is enabled: `directoryGroupsInfo` returns per-group error strings, `searchDirectoryGroups` returns an empty list. The UI may want to hide/disable the search affordance when directory groups are unsupported (`directoryGroupsSupported` already exposed via module config).
- Requires hoist-core >= 41.0 (entries under 41.0-SNAPSHOT in its CHANGELOG). Server-side design/decision doc: `docs/planning/entra-id-service-design.md` in hoist-core.

amcclain:
Adds client-side support for directory group display names and directory search in the Roles admin module, the counterpart to xh/hoist-core#590, which introduces `EntraIdService`, `DirectoryService`, and the `roleAdmin/directoryGroupsInfo` + `roleAdmin/searchDirectoryGroups` endpoints consumed here. Friendly names now replace the opaque GUIDs that Entra ID uses as group identifiers wherever groups are displayed, which also improves LDAP DN display, and the role editor gains a search-based group picker. Also fixes a framework-level bug that left inline grid editors without keyboard focus when an edit was started programmatically.

### Roles Admin

- `RoleModel` resolves and caches display info for directory groups after each role list load. Friendly names render in the Dir. Groups detail grid, the Users tab source tags, and the role editor's assignments grid. Raw identifiers and any lookup error move to grid column tooltips, with errors also flagged by warning icons.
- The role editor's Directory Groups column gains a search-based picker (min 2 chars, options show description/mail to disambiguate same-named groups). Free-text entry of a known GUID/DN remains supported and resolves its display name on commit.
- Assignment editors now open their dropdown on focus, so the empty state doubles as a prompt for what to type. For directory groups that surfaces the server's `directoryGroupsDescription`, which is more discoverable there than on the panel info icon.
- Degrades gracefully against earlier hoist-core versions: one warning logged per 404'd endpoint, no re-requests, falling back to identifier display and client-side filtering of already-assigned groups. Verified interactively against both the core feature branch (live Entra ID lookups) and the current published core snapshot.
- Fixes a latent bug where `BaseMembersModel` never wired up its `nameRenderer`, leaving the CN formatting and error icons in the details grids unused since #3661.

### Grid

- Fixes `GridModel.beginEditAsync()` opening an inline editor that never took keyboard focus, so the user had to click the cell before typing. `InlineEditorModel` gated its focus-on-render reaction on `agParams.cellStartedEdit`, which ag-Grid reports as `false` for edits started through its `startEditingCell` API; that guard is only needed under full row editing. `beginEditAsync` now also sets the focused cell, which ag-Grid's own edit paths do and `startEditingCell` does not.

Requires hoist-core >= 41.0.0 for full functionality. Closes #4572.

---

Hoist P/R Checklist
-------------------

- [x] Reviewed and tested on Mobile, or determined not required. *(Admin Console module is desktop-only; the grid fix touches desktop inline editors.)*
- [x] Created Toolbox branch / PR, or determined not required. *(No Toolbox changes needed - tested interactively in Toolbox against the hoist-core branch.)*


admin/tabs/userData/roles/details/members/DirectoryMembers.ts
```diff
@@ -42,23 +42,27 @@ class DirectoryMembersModel extends BaseMembersModel {
...
 
     override nameRenderer: ColumnRenderer = (name, {record}) => {
-        const {error} = record.data;
+        const {displayName, error} = record.data;
         return hbox({
             alignItems: 'center',
             items: [
                 box({
-                    item: RoleModel.fmtDirectoryGroup(name),
+                    item: displayName ?? RoleModel.fmtDirectoryGroup(name),
```
lbwexler: why not do this fallback on line 65 in the raw data on line 49?

lbwexler (approved):

PR #4588, opened 2026-08-18
Tiered ag-Grid transaction handling for large-grid update performance

lbwexler:
## Overview

Grid data updates now run only as much of ag-Grid's client-side model refresh as each transaction actually requires. A new `GridTransactionManager` (delegated to by Grid's local model) picks one of three modes per transaction:

- **Suppress** - update-only transactions provably unable to affect row order, grouping, or tree structure skip the sort/filter/group/flatten stages entirely (via `suppressModelUpdateAfterUpdateTransaction`): just a row data swap + cell refresh.
- **Delta** - transactions changing fewer than `experimental.deltaSortRatio` percent of rows (default 50 - see appendix) run a full refresh but sort only their changed rows, merging into the existing order (ag-Grid `deltaSort`, now managed dynamically - `experimental.deltaSort` removed).
- **Full** - everything else, unchanged behavior.

Order-stability is proven two ways: a new `StoreTransaction.changedFields` (a producer's assertion of exactly which fields a value-only update touched - Cube `View`s supply it automatically from the per-field diff their dataOnly updates already compute, carried through `RecordSet` and surfaced on the grid's delta), or a per-record comparison of sorted field values against the prior RecordSet. Sorts on custom comparators, custom `getValueFn`s, or `sortValue` functions are never suppressed. Cached sort-path resolution invalidates on `sortBy`/`columns` changes.

Order-affecting updates also apply suppressed (cells tick live), with a managed, idle-scheduled re-sort paced adaptively off its own measured cost - `experimental.deferredSortFactor` tunes the pacing, 0 restoring synchronous re-sorting. New records (adds) always sort into place immediately - ag-Grid ignores suppression for transactions with adds/removes.

## Results

Toolbox cube tester, tree grid (Fund > Trader, leaves), ~550-row update transactions, avg `applyTransaction` time:

| Rows | Suppress | Delta (values changing) | Full |
|---|---|---|---|
| 400k | **10.5ms** | 592ms | 2,237ms |
| 40k | **15.8ms** | 419ms | 1,254ms |

- At 400k rows the previous per-transaction full refresh (~1.4-2.2s) could not keep up with a 1/sec update stream; suppressed applies are ~100x faster and size-independent (dominated by the fixed cell-refresh cost).
- The win holds well down at 40k rows: ~80x vs full for order-stable updates, ~3x for genuinely reordering ones.
- On a commission-sorted 400k grid: all updates applied at ~11ms, with re-sorts amortized off the transaction path and paced to a bounded fraction of main-thread time.
- Verified mode selection live at both scales: order-stable streams suppress on 100% of transactions via `changedFields`; sorting by a field the stream changes correctly drops to delta.

## Notes for review

- `RecordSet` carries `changedFields` via single-step ordinal provenance (no predecessor references retained). `diffFrom` surfaces it only when the diff window is exactly one value-only step - multi-step windows conservatively report unknown.
- Row-reuse stores mutate row data objects in place, so the per-record fallback detects shared data (`prev.data === rec.data`) and declines to suppress - `changedFields` is what enables the fast path for view-connected stores.
- Companion Toolbox commit (local, not yet pushed) removes the cube tester's now-obsolete Delta Sort toggle.

## Appendix: delta vs. full re-sort - measured crossover

The delta/full threshold (`experimental.deltaSortRatio`) was raised from an initial 20% to a default of 50%, from live measurement in the cube tester (fund > trader tree with leaves, n = 20,055, sorted on a streaming field; per-flush timings via the new `sortFlush` diagnostics):

| changed rows (t/n) | delta | full |
|---|---|---|
| 1% | 22ms | ~83ms |
| 18% | 37ms | ~83ms |
| 33% | 51ms | ~83ms |
| 51% | 73ms | ~83ms |
| 98% | 98-141ms | ~83ms |

- Full-sort cost is flat in changed-row count (it re-sorts everything regardless); delta is linear: ~21ms + ~5µs per changed row at this scale. At full coverage delta is ~1.7-3x *slower* than full - it pays a full-sized sort of the changed set plus its own index/merge bookkeeping.
- Measured break-even at t/n ≈ 0.55 matches a comparator-count model (`t·log₂t` vs `n·(log₂n_g − a)`, a ≈ 5, with n_g ≈ 440-row leaf groups here). The crossover rises with leaf-group size - flatter grids favor delta longer - and falls for deeply nested, small-group shapes.
- 0.5 sits just below break-even for this shape, and erring toward delta is the safe side of wrong: modestly more total sort time, but in smaller, smoother chunks - vs. full's single larger block.

### DIFF
--- a/cmp/grid/impl/GridTransactionManager.ts
+++ b/cmp/grid/impl/GridTransactionManager.ts
@@ -43,17 +43,15 @@ const DELTA_SORT_MAX_RATIO = 0.2;
  */
 export class GridTransactionManager extends HoistBase {
     private model: GridModel;
-    private currentDeltaSort: boolean;
     private sortDirty = false;
     @managed private sortTimer: Timer;
 
     // Cached provable sort paths - undefined = stale, null = sort not provably value-based.
     private _sortPaths: Array<Some<string>> | null | undefined;
 
-    constructor(model: GridModel, initialDeltaSort: boolean) {
+    constructor(model: GridModel) {
         super();
         this.model = model;
-        this.currentDeltaSort = initialDeltaSort;
 
         this.addReaction({
             track: () => [model.sortBy, model.columns],
@@ -68,19 +66,23 @@ export class GridTransactionManager extends HoistBase {
 
     apply(transaction: RecordSetDelta, prevRs: RecordSet, newRs: RecordSet) {
         const {agApi} = this.model,
-            mode = this.getRefreshMode(transaction, prevRs, newRs);
+            mode = this.getRefreshMode(transaction, prevRs, newRs),
+            suppress = mode === 'suppress';
 
-        if (mode === 'suppress') {
-            agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', true);
-        } else {
-            this.syncDeltaSort(mode === 'delta');
-        }
-        agApi.applyTransaction(transaction);
-        if (mode === 'suppress') {
-            agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', false);
-        } else {
+        // Set both options explicitly per transaction - each apply is self-contained, with no
+        // dependence on state left by a prior pass. Same-value writes are no-ops within ag-Grid.
+        agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', suppress);
+        agApi.setGridOption('deltaSort', mode === 'delta');
+        try {
+            agApi.applyTransaction(transaction);
             // Any un-suppressed pass re-sorted (delta implies a clean starting order - see below).
-            this.sortDirty = false;
+            if (!suppress) this.sortDirty = false;
+        } finally {
+            // Never leave suppression on - transactions applied outside this manager (or after an
+            // exception above) must get ag-Grid's full default refresh behavior.
+            if (suppress) {
+                agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', false);
+            }
         }
     }
 
@@ -174,15 +176,6 @@ export class GridTransactionManager extends HoistBase {
             ret.push(path);
         }
         return ret;
-    }
-
-    // Delta sort is managed per-transaction. Only touch the ag option when the desired state
-    // actually changes.
-    private syncDeltaSort(desired: boolean) {
-        if (desired !== this.currentDeltaSort) {
-            this.model.agApi.setGridOption('deltaSort', desired);
-            this.currentDeltaSort = desired;
-        }
     }
 
     private flushPendingSort() {
--- a/cmp/grid/impl/GridTransactionManager.ts
+++ b/cmp/grid/impl/GridTransactionManager.ts
@@ -69,20 +69,17 @@ export class GridTransactionManager extends HoistBase {
             mode = this.getRefreshMode(transaction, prevRs, newRs),
             suppress = mode === 'suppress';
 
-        // Set both options explicitly per transaction - each apply is self-contained, with no
-        // dependence on state left by a prior pass. Same-value writes are no-ops within ag-Grid.
-        agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', suppress);
-        agApi.setGridOption('deltaSort', mode === 'delta');
+        agApi.updateGridOptions({
+            suppressModelUpdateAfterUpdateTransaction: suppress,
+            deltaSort: mode === 'delta'
+        });
         try {
             agApi.applyTransaction(transaction);
-            // Any un-suppressed pass re-sorted (delta implies a clean starting order - see below).
             if (!suppress) this.sortDirty = false;
         } finally {
-            // Never leave suppression on - transactions applied outside this manager (or after an
-            // exception above) must get ag-Grid's full default refresh behavior.
-            if (suppress) {
-                agApi.setGridOption('suppressModelUpdateAfterUpdateTransaction', false);
-            }
+            // Never leave suppression on - transactions applied outside this manager must get
+            // ag-Grid's full default refresh behavior.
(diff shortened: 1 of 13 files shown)


cmp/grid/impl/GridTransactionManager.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    private computeSortPaths(): Array<Some<string>> | null {
+        const {model} = this,
+            ret = [];
+        for (const sorter of model.sortBy) {
+            const col = model.getColumn(sorter.colId);
+            if (!col || col.comparator || col.getValueFn !== col.defaultGetValueFn) return null;
+            const {sortValue} = col;
+            if (isFunction(sortValue)) return null;
+            const path = sortValue ?? col.fieldPath;
```
haynesjm42: A string `sortValue` doesn't fully replace the column's own value - `Column.getSortValue()` is `record?.data[sortValue] ?? v`, so when the `sortValue` field is nullish the sort falls back to the `getValueFn`/`fieldPath` value. Dropping `fieldPath` here makes `sortUnchanged()` claim more than it can prove.

Given `column({field: 'price', sortValue: 'priceSort'})` where `priceSort` is null on some records, an update that changes `price` (leaving `priceSort` null) satisfies `sortUnchanged()` -> `getRefreshMode` returns `'suppress'` **without** calling `notePendingSort`, so no deferred flush is ever scheduled. Row order stays wrong indefinitely, until some unrelated add/remove forces a full refresh. Affects both the `changedFields` fast path and the value-comparison path.

Suggest pushing both paths when `sortValue` is a string (or just bailing to `null`).

cmp/grid/impl/GridTransactionManager.ts
```diff
@@ -0,0 +1,247 @@
...
+            floor = this.nextFlushAllowed - performance.now();
+        if (floor > 0) await wait(floor);
+        runWhenIdle(
+            () => {
+                this.flushQueued = false;
+                this.flushPendingSort();
+            },
+            {timeout: Math.max(1, deadline - performance.now())}
+        );
+    }
```
haynesjm42, line 219: No `isDestroyed` guard across an await that can span up to `MAX_DEFERRED_SORT`. Neither the post-`wait` continuation, the `runWhenIdle` callback, nor `flushPendingSort` checks it.

If the grid unmounts with `pendingSortIds` set and remounts inside the window, the new mount resets `_syncedRs` and builds a fresh manager - but the old manager's idle callback still sees `pendingSortIds` non-null, `model._syncedRs` non-null and `model.isReady` true, and fires `updateGridOptions({deltaSort: true})` + `applyTransaction` against the *new* grid. Spurious transaction (extra `rowDataUpdated` + pipeline refresh, can disturb an in-progress inline edit) plus `deltaSort` flipped on behind the new manager's back.

`if (this.isDestroyed) return;` after the await and at the top of `flushPendingSort` covers it.

cmp/grid/impl/GridTransactionManager.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    private flushPendingSort() {
+        const {model, pendingSortIds} = this,
+            latestRs = model._syncedRs;
+        if (!pendingSortIds || !latestRs || !model.isReady) return;
```
haynesjm42, line 226: This early return abandons the flush rather than rescheduling it: the caller already cleared `flushQueued` (line 211), and `pendingSortIds` is left populated with nothing re-arming the timer - only a later `notePendingSort` will schedule again. If the update stream goes quiet right after the guard trips, the class doc's guarantee that "row order can never go stale for longer than this" doesn't hold.

Minor in practice, since `!model.isReady` is the only realistic way to trip it, but the documented invariant isn't actually enforced.

cmp/grid/impl/GridTransactionManager.ts
```diff
@@ -0,0 +1,247 @@
...
+        const start = performance.now(),
+            {agApi} = model,
+            update = [];
+        pendingSortIds.forEach(id => {
+            const rec = latestRs.getById(id);
+            if (rec) update.push(rec);
+        });
+
+        if (update.length && update.length / latestRs.count < this.deltaSortRatio) {
+            agApi.updateGridOptions({deltaSort: true});
```
haynesjm42, line 238: `deltaSort` is a global grid option that's turned on here and never restored - `apply()` resets `suppressModelUpdateAfterUpdateTransaction` in its `finally` but leaves `deltaSort` at whatever the last transaction set.

Benign inside hoist today (no other `applyTransaction` call sites, and `onSortChanged`/`refreshClientSideRowModel` pass no `changedRowNodes`, so ag-Grid falls back to a full sort). But it leaves a trap for app code calling `gridModel.agApi.applyTransaction(...)` directly - that transaction would delta-merge into an order the manager may have deliberately left stale. Worth resetting symmetrically.

haynesjm42 (commented):
Code review of the transaction-manager changes. Four findings, all in `GridTransactionManager` - see inline comments. (1) is the only one I'd consider blocking.

cmp/grid/impl/GridTransactionManager.ts
lbwexler: Fixed!

cmp/grid/impl/GridTransactionManager.ts
lbwexler: Fixed it

cmp/grid/impl/GridTransactionManager.ts
lbwexler: cleaned up, good catch

cmp/grid/impl/GridTransactionManager.ts
lbwexler: letting this go to avoid the complexity

PR #4592, opened 2026-08-19
Inspector: Cube/View tracking and data-pipeline Diagnostics panel

amcclain:
Surfaces the new data-pipeline `diagnostics` (#4585) in the Hoist Inspector, and extends the Inspector's reach to Cubes and cube Views.

- `Cube` and `View` now register with `InstanceManager` (mirroring `Store`), with new `XH.getCubes()` / `XH.getViews()` accessors and a `Cube.isCube()` type guard. Both appear in the Inspector's instances grid under new groups, and property values referencing them render as navigation links.
- New **Diagnostics** tab (paired with Properties in a tab container) shows a live readout for any selected Stores, Cube Views, and GridModels - one row per op kind with the last op's type, work detail, timing, and cumulative count/avg. Rows are discovered from each diagnostics object's shape rather than per-class schemas, insulating the UI from that deliberately unstable internal API. A selected `Cube` reports via its internal `Store`, where its data ops actually land.
- **Log operations to console** switch reflects and toggles the per-instance `diagnostics.logLevel` - sticky per instance, so logging continues as the selection moves and any number of instances can stream at once. `BaseDiagnostics.logLevel` is now observable to support this.
- All four Inspector grids gain structured column filters (`filterModel` + filterable columns, hover-only header affordances), and the instances grid gets a fixed group order (Models, Services, Cubes, Views, Stores) plus headerTooltips explaining the Sync columns.

Verified live against Toolbox (full local stack), primarily via the admin Cube Data test panel - selecting the cube store / view / grid pipeline together shows each stage's ops ticking during loads and updates, in both docked and popped-out Inspector modes.

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (Additive only - new registry tracking, XH accessors, and xhImpl Inspector UI.)
- [x] Reviewed and tested on Mobile, or determined not required. (Inspector is desktop-only; grid filterModel is a no-op on mobile.)
- [x] Created Toolbox branch / PR, or determined not required. (No Toolbox changes needed - verified against unmodified Toolbox.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,6 +222,12 @@ columns.
 * `TextInput` now accepts a `leftElement` prop, rendered inline at the left of the input.
 * Hoist Inspector now pops out in a separate browser window, leaving the app's viewport to the app
   while ensuring Inspector itself is not covered by masks or other modal content.
+* Hoist Inspector now tracks `Cube` and cube `View` instances alongside models, services, and
+  stores, and adds a Diagnostics panel - a live readout of the data-pipeline `diagnostics`
+  published by selected Stores, Cube Views, and GridModels, with controls to reset counters and
+  stream ops to the devtools console.
+* Added `XH.getCubes()` and `XH.getViews()` to enumerate all active `Cube` and `View` instances,
+  now registered with Hoist's instance registry.
 
 ### 🐞 Bug Fixes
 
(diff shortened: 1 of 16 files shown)


lbwexler (approved):
Very surgical --- seems like it slotted in quite cleanly!

PR #4600, opened 2026-08-21
Hot-Loop performance improvements in data/grid

lbwexler:
Targeted allocation and complexity fixes on per-comparison, per-cell, and per-record hot paths, found in a follow-up perf audit after the v87 wave.

* **Sort path**: Column comparators resolve their `GridSorter` via a new allocation-free `GridModel.getSorter()` instead of a lodash `{colId}` matcher per comparison (which allocated matcher internals on every call - ~15M objects for one 400k-row sort). The default path resolves once and passes it into `defaultComparator`; app comparators calling via params get the cheap lookup automatically.
* **Value reads**: new `Column.buildFastValueGetter()` - when a column uses the default `getValueFn`, the ag-Grid `valueGetter` reads `record.data` directly, skipping the per-call params object (paid per cell render and twice per sort comparison). Custom `getValueFn`s keep the fully-featured path.
* **Tree-grid transactions**: `Grid` now skips its post-transaction expand-state sync - an ag-Grid walk over *all* rows plus deep-equal/freeze of the state tree - for updates that change no `parentId`. Adds/removes and grouped-by-value grids sync exactly as before.
* **FieldFilter**: `=`/`!=`/`includes`/`excludes` test via a Set instead of a linear candidate scan per record, capping the O(records × candidates) worst case (e.g. header-filter value selections on high-cardinality columns). `=`/`!=` use the Set only when all candidates are primitives, preserving `isEqual` semantics for Dates; `includes`/`excludes` were already SameValueZero, so the Set is semantically exact.

Validated in Toolbox: header sorts incl. abs-value cycling, tree grid reload with expansion preserved, grouped rows, and column filtering (`=` and tags `includes`) all clean.

Stacked follow-up with a second batch (StoreValidator, GridFindField, selection sync): #4602.

### DIFF
--- a/cmp/grid/Grid.ts
+++ b/cmp/grid/Grid.ts
@@ -26,6 +26,7 @@ import {
     XH
 } from '@xh/hoist/core';
 import type {StoreRecord} from '@xh/hoist/data';
+import type {RecordSet, RecordSetDelta} from '@xh/hoist/data/impl/RecordSet';
 import {GridTransactionManager} from '@xh/hoist/cmp/grid/impl/GridTransactionManager';
 import {
     colChooser as desktopColChooser,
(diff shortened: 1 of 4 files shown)


PeteDarinzo (approved):
Reviewed

Issue #4486, opened 2026-07-10
Poor Validation Performance for Bulk Edit


lbwexler:
A client developer reported the following issues when trying to use HoistValidation while editing/importing ~100s/1000s of records at once

> 
> 1. No-op overhead when validation is unused — a Store with no field rules still spins up a RecordValidator per uncommitted record on every change to the uncommitted set.
> 
> 2. No "validate-in-isolation" mode — even with validationIsComplex: false, a bulk change re-validates every uncommitted record, and the driving reaction re-scans all records on every change.
> 
> 3. O(N²) reactive churn — each record's validation completion invalidates O(N) store-level aggregates (isValid, errors, errorCount); if anything observes them, they recompute per-record → N × O(N). 
> 
> In practice, adding/editing on the order of 10³–10⁴ uncommitted records produces multi-second main-thread freezes (and, with wide records, memory pressure). We worked around it at the app layer, but the root cause is in hoist-react.

lbwexler:
Here is the assesment from the client's Claude:

[store.txt](https://github.com/user-attachments/files/29905113/store.txt)

lbwexler:
Splitting this up. **Item 1 is fixed** on `hot-loop-perf-5`, along with some incidental churn in the same reaction; items 2 and 3 are both invasive enough to want their own PR, so sketching the shape here.

### Item 1 - no-op overhead when validation is unused (done)

`StoreValidator` now computes a `hasRules` flag at construction (`store.fields.some(f => !isEmpty(f.rules))`) and skips creating its sync reaction entirely when false. No `RecordValidator` instances, no chunked validate pass, no observable writes. Worth noting how much was being paid per record: `new RecordValidator` is two `makeObservable` calls, since it eagerly constructs a `TaskObserver.trackLast()` (a `CompoundObserver`, itself 2 observables + 3 computeds) on top of its own 1 observable + 7 computeds.

Two related fixes went in alongside, which help stores that *do* have rules:

* `StoreValidator.uncommittedRecords` is now a `@computed({equals: comparer.shallow})`, so a transaction that leaves the uncommitted set unchanged no longer produces a new array identity and no longer fires the sync reaction.
[...]

PR #4602, opened 2026-08-21
Hot-loop performance: StoreValidator, GridFindField, and selection sync

lbwexler:
Stacked on #4600 - second batch of hot-path fixes, split out for reviewability. Will re-target to `develop` automatically when #4600 merges.

* **StoreValidator**: a Store with no validation `Rules` on any `Field` now skips validation entirely - `hasRules` is computed once at construction and no sync reaction is created when false, so no `RecordValidator` per uncommitted record and no chunked no-op validate pass. This is item 1 of #4486, where a client reported multi-second freezes editing 10^3-10^4 records. Also makes `uncommittedRecords` a `@computed({equals: comparer.shallow})`, so it holds a stable identity across transactions that leave the uncommitted set untouched. `StoreRecord.validationState` now reports `Valid` rather than a permanent `Unknown` for records in a Store with no rules.

* **GridFindField**: the data reaction is debounced with the existing `queryBuffer`, coalescing a burst of store transactions into a single results recompute. The sorting rework from this model is reviewed separately in #4608.

* **Selection sync**: `AgGridModel.setSelectedRowNodeIds()` applies the selection delta in at most two bulk `setNodesSelected()` calls - unchanged rows untouched, and select-all no longer fires an event per row. `Grid.selectionReaction` now tracks `selModel.selectedIds` rather than `selectedRecords`: everything below `syncSelection` is keyed by row node id, so a change confined to a selected record's data cannot affect the outcome, but under `.struct` on `selectedRecords` it fired the reaction anyway - burning n lazy `agId` materializations plus an `agApi.getSelectedNodes()` traversal per tick before the `isEqual` guard no-oped it.

Validated in Toolbox: GridFindField query/cycle/reload, select-random + scroll-to-selection, and inline-edit validation incl. the dirty->clean transition. Zero console errors. The `StoreValidator` gate is newer than that pass and has been verified by typecheck and review only - worth re-running the inline-edit cases against a Store that does have rules before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -306,7 +306,10 @@ columns.
     * Developers who build against a local hoist-react checkout (`inlineHoist`) now need pnpm to
       install this repo's dependencies (`corepack enable pnpm`, then `pnpm install`). The app itself
       can remain on yarn or npm. Update any `startWithHoist`-style app scripts accordingly.
-
+* `Store`s with no validation `Rules` on any `Field` now skip record validation entirely, no longer
+  creating a `RecordValidator` per uncommitted record. Note that `StoreRecord.validationState` now
+  reports `Valid` (rather than `Unknown`) for records in such a Store.
+*
 ### ⚙️ Typescript API Adjustments
 
 * Retyped `BaseRow.data` from `ViewRowData` to `PlainObject`. When reading row data, custom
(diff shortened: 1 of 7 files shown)


PeteDarinzo (approved):
Reviewed

PR #4608, opened 2026-08-21
Add GridModel.getSortedRecords(); fix GridFindField mutating Store records

lbwexler:
### Fixes

**`GridFindField` was mutating Store state.** Its recursive sort sorted `rec.children` in place, and that resolves to `RecordSet`'s memoized children array, not a copy - so using a find field on a tree grid permanently reordered the array returned by `Store.getChildrenById`, for every other consumer of it. Export sorted the same way but copied at each level of its recursion, so it was unaffected.

**Grid export ignored `GridModel.groupBy`.** A grouped grid exported in an order unrelated to what was on screen. Exported leaf order now matches the rendered order. Group header rows are still not emitted - those are synthetic ag-Grid nodes with no backing `StoreRecord`.

### Faster sorting

Both call sites sort the Store's full record set - all roots, plus every child on a tree grid - to reproduce rendered row order without going through ag-Grid. Export does this once per export, with the user waiting on it. `GridFindField` re-sorts whenever `store.records`, `columns`, `sortBy`, or `groupBy` changes while a query is active, debounced at `queryBuffer` (200ms default) - so on a streaming grid, up to five full re-sorts a second. The result is cached in `_records` until the next such change, so typing in the field does not itself re-sort.

Three changes:

* Sorter setup - column lookup, `getAgSpec().comparator`, direction - resolves once per sort rather than once per tree node. `getAgSpec()` builds a full `ColDef` with several closures, so on a deep tree that was the dominant cost.
* Decorate-sort-undecorate: each record's value and ag-Grid row node resolve once per sort instead of on every comparison - n lookups per key rather than ~2n log n. Worth noting ag-Grid's own `compareRowNodes` does not do this, re-resolving both per comparison.
* One composite comparator over the sort keys in priority order, replacing a separate stable pass per key. Equivalent for consistent comparators, and k passes become one.

### One implementation, exposed publicly

`GridExportService` and `GridFindFieldImplModel` held near-verbatim copies of this sort, which had drifted apart - hence both fixes above. Both now go through a new public `GridModel.getSortedRecords()` and retain no sorting code of their own. It delegates to `computeSortedRecords()` in the new `@internal` `cmp/grid/impl/RecordSortUtils.ts`.

Exposing it seemed generally useful rather than incidental: "the grid's records in the order it renders them" is what any app-level operation wants when it has to follow the on-screen order but cannot go through ag-Grid - custom exports, reports, copy-to-clipboard, print views. It works on an unrendered `GridModel`, which `agApi`-based approaches do not.

### Behavior notes

* `record.children` on a tree grid with an active find field now reads back in the Store's natural order rather than the find field's sort order. Grid row order comes from ag-Grid, so no in-framework consumer should care.
* `groupBy` is now ignored on tree grids, matching ag-Grid, which ignores row grouping in tree data mode. Find previously applied it to the flattened list, producing an order that did not match the grid.


CHANGELOG.md
```diff
@@ -314,10 +318,14 @@ columns.
     * Developers who build against a local hoist-react checkout (`inlineHoist`) now need pnpm to
       install this repo's dependencies (`corepack enable pnpm`, then `pnpm install`). The app itself
       can remain on yarn or npm. Update any `startWithHoist`-style app scripts accordingly.
+* Improved sort performance for grid export and `GridFindField`. Both now share a single
+  decorate-sort-undecorate implementation that resolves each record's value, comparator, and
```
amcclain: "decorate-sort-undecorate implementation" 🤔

cmp/grid/impl/RecordSortUtils.ts
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Return the GridModel's root records in the order the grid renders them - sorted per its groupBy
+ * and sortBy and, for tree grids, flattened depth-first through each record's children.
+ *
+ * Used to reproduce rendered row order for consumers that work off the Store rather than ag-Grid,
+ * e.g. export and find. Always returns a newly allocated array, and never mutates the Store's own.
+ * Sorts via decorate-sort-undecorate, resolving each record's value and row node once rather than
+ * on every comparison. Note this yields leaf records only - the group rows rendered for a grouped
+ * grid are synthetic ag-Grid nodes with no backing StoreRecord.
+ * @internal
```
amcclain, line 20: I am not pushing for this, but would this be useful for apps in any context we know of or could very plausible imagine? Just wondering about marking `@internal` and put within `impl` since it feels like a maybe useful general util

amcclain (approved):

CHANGELOG.md
lbwexler: Yeah -- "DSU"

cmp/grid/impl/RecordSortUtils.ts
lbwexler: I thought about that -- sure -- could be a single public method on GridModel.  I'll do it.

lbwexler: Done!

PR #4609, opened 2026-08-21
Stop Blueprint JS + CSS leaking into mobile bundles via the utils/impl barrel

### LINKED ISSUE #4593: Blueprint JS + CSS leak into mobile bundles via the utils/impl barrel
All of Blueprint's JS and CSS is currently pulled into **mobile** app bundles, where it should not be needed at all.

**Import chain:**

- `utils/impl/MenuItems.ts` imports `menuDivider`/`menuItem` from `@xh/hoist/kit/blueprint`
- `kit/blueprint/index.ts` imports all of `@blueprintjs/core` + `@blueprintjs/datetime`, including `blueprint.css` and `blueprint-datetime.css`
- `utils/impl/index.ts` re-exports `* from './MenuItems'`, and that barrel is imported by many mobile-reachable modules (`mobile/cmp/toolbar/Toolbar.ts`, `mobile/cmp/menu/impl/Menu.ts`, `appcontainer/*`, `svc/TrackService.ts`, `data/cube/row/*`, etc.)

**Impact:**

- Significant bundle bloat for mobile apps (Blueprint JS + two global stylesheets).
- Global CSS side effects: `blueprint-datetime.css` embeds react-day-picker **v8**'s full stylesheet with unscoped `.rdp-*` rules. Discovered while prototyping a mobile `DateInput` built on react-day-picker v9 - v8's `.rdp-day { display: flex }` collides with v9's table markup and breaks the calendar layout, requiring a scoped CSS reset as a workaround.

**Suggested fix:** move `parseMenuItems` out of the shared `utils/impl` barrel (e.g. into a desktop-only impl location, per the menu standardization TBD in #2400), so `kit/blueprint` is only reachable from desktop code. The workaround reset in the mobile DateInput prototype can be removed once resolved.

lbwexler:
Fixes #4593

- Moved the internal `parseMenuItems` util from `utils/impl` to `desktop/cmp/menu/impl`, severing the only shared-code import of `kit/blueprint`. The `utils/impl` barrel is imported by many mobile-reachable modules, so all of Blueprint's JS plus its two global stylesheets were landing in mobile app bundles.
- Updated the three (desktop-only) consumers to the new import path. New home mirrors `mobile/cmp/menu/impl` and gives the eventual menu standardization (#2400) a natural landing spot.
- Removed the react-day-picker v8 CSS reset from mobile `DateInput` - the clashing v8 rules came from `blueprint-datetime.css`, which no longer reaches mobile bundles.

Note: `parseMenuItems` is `@internal`, but Toolbox does import it (`BasicPanel`) - matching Toolbox update to follow after merge.


lbwexler:
CR: Don

PR #4618, opened 2026-08-22
Store: rescue unchanged digest-less records before creation

lbwexler:
Reworks the classic (digest-less) load path to detect unchanged records *during* the raw-data parse and reuse the committed record before allocating anything - replacing the old create-then-discard flow through `RecordSet.areRecordsEqual`.

- `Store.parseOrRescue` compares parsed values against the cached committed record in the same single pass that buffers them, returning the cached instance on a field-for-field match. Soundness comes from an identity test proving each matched cached value non-default, plus equal non-default counts (snapshotted at build as internal `StoreRecord.nonDefaultCount`) certifying omitted fields are default on both sides.
- `RecordSet.withNewRecords` now compares by instance identity only - `areRecordsEqual` and its duplicated `positionUnchanged` are deleted.
- `reuseRecords: true` is now literally the identity digest `raw => raw`, unifying the reference mode with the digest forms (a digest miss always signals change). The `RecordDigest` alias is removed - digests are `unknown`, compared via `===`.
- Digest-less `updateData()` drops value-equal updates as no-ops, preserving uncommitted local modifications - matching digest-store behavior.
- Projection stores get an equivalent pre-creation compare over declared fields.

Bench (200k rows x 8 fields, node): unchanged reload 111ms -> 70ms, 1%-changed 104ms -> 67ms, transient heap roughly halved. Changed records also gain by dropping the post-hoc whole-object deep compare.

Validated with a 50-assertion smoke suite covering the rescue-soundness edges (omitted non-default fields, explicit defaults, object-valued defaults, tree moves, `processRawData`, `denseRecordThreshold` crossings, refMode/digest/projection semantics, updateData no-ops). Run against the pre-change code, the suite passes except exactly the intended `updateData` behavior change.


data/Store.ts
```diff
@@ -1386,13 +1383,19 @@ export class Store
...
+            cached = this.getCachedRecord(id, parent);
+
+        // 1) A digest rescues or disqualifies a cached record immediately
+        if (digest != null) {
+            if (cached?.digest === digest) return cached;
+            cached = null;
+        }
 
-        // Projections can re-use raw data with no reparsing.
+        // 2) Projections can (re)use raw data with no reparsing.
```
haynesjm42: I don't think this is right - with a projection `raw === data` so unless we have a digest, we need to always mint a new record. Since we are calling `rawMatchesData` here we would never mint the new record for updates so the Grid would never know that anything changed, unless I am missing something?

data/Store.ts
```diff
@@ -1405,14 +1408,15 @@ export class Store
             });
         }
 
+        // 3) Otherwise parse (app + field parsing), comparing to the cached record in the same
+        // pass and reusing it on an exact match.  We really want to reuse!
         const {processRawData, retainRaw} = this;
         let data = raw;
-        if (processRawData) {
-            data = processRawData(raw);
-            throwIf(!data, 'Store.processRawData must return an object.');
```
haynesjm42, line 1412: Was this throw removed just to optimize the hot path here?

lbwexler: + Typescript already requiring an explicit return.  Thought it had outlived its usefulness in a typescript/AI world
+ This would not fail in a subtle way.
+ We don't really do this kind of runtime check elsewhere in the toolkit

data/Store.ts
lbwexler: So I think the issue is that if you have a projectionOnly data source its on you not to mutate it without having a digest to indicate the mutation.

Our view handles that contract by always providing a digest.  in the rare (if ever) case that you were building your own data source for a `projectionOnly` you would have to do the same.

lbwexler: Great catch -- I tweaked with the following reasoning
If you want it to assume an identical reference passed back is the same, set the reuseFn: true.  
Otherwise, we will assume an identical reference must be mutated to be safe.

This is all a bit theoretical until we actually get someone writing one of these (versus just using View, which provides the digest).

PR #4623, opened 2026-08-26
Upgrade react-dropzone 15 -> 20.1

lbwexler:
Replaces #4582, the Dependabot bump that went stale - its recreate failed on 8/18 and 20.1.1 has since shipped. Closed in favor of this.

Five majors, but our surface is two files (`kit/react-dropzone` and `desktop/cmp/filechooser`) and **no source changes were needed for the upgrade itself**. Everything we consume survives in v20: the default `Dropzone` component (still `forwardRef`), render-prop `children`, `DropzoneRef.open()`, `ErrorCode`, `Accept`, `FileRejection`, `FileWithPath`. `pnpm typecheck` and `pnpm lint` are clean.

I also confirmed the `_/_` dummy-MIME trick in `FileChooser.ts` still matches by extension under the new `attr-accept` v4 - v18 unbundled the extension-to-MIME table, but our accept lists are extension-only so it doesn't bite.

### The one behavior change worth reviewing

v19 made react-dropzone accept files **up to** `maxFiles` and reject only the surplus, instead of rejecting the whole batch. `FileChooserModel.addFilesInternal()` had its own all-or-nothing guard on top of that, so an overflowing drop would fire a rejection toast *and* a "File limit exceeded" warning and add nothing.

Rewrote it to match the new semantics: retain the current selection, replace any same-named file, take incoming files until full, and warn about the surplus. `onFileAccepted` now receives the files actually added rather than everything the dropzone handed us.

### Notes for the reviewer

- v20 declares `engines: {node: ">= 22"}` - worth a sanity check against any app pipeline still on 20.
- v16 dropped the UMD build and `dist/es/`; the package is now `"type": "module"` with an `exports` map. Bare imports are unaffected.
- Dependabot flagged a new `prepare` script - it's `prek install || exit 0`, which only runs for git/local installs, not registry tarballs.
- Free wins we could pick up separately: `getErrorMessage` (would replace the manual `FileInvalidType` message rewrite in `FileChooserModel`), async `validator`, `isProcessing`, paste-to-upload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,22 @@
 
 ## 88.0.0-SNAPSHOT - unreleased
 
+### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - one behavior change to `FileChooser`)
+
+* `FileChooser` now accepts as many dropped files as its `maxFiles` limit allows, rejecting only
+  the surplus. Previously a drop that would exceed the limit was discarded in full. Apps relying on
+  the all-or-nothing behavior should check `onFileAccepted` - it now reports only the files actually
+  added.
+
+### ⚙️ Technical
+
+* Upgraded `react-dropzone` to v20, which drops its UMD build and ships as an ESM + CJS package with
+  an `exports` map. Requires Node >= 22 to install.
+
+### 📚 Libraries
+
+* react-dropzone `15.0 → 20.1`
+
 ## 87.0.0 - 2026-08-25
 
 ### 💥 Breaking Changes (upgrade difficulty: 🟠 MEDIUM - React 19, data layer, column chooser)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,10 @@
 
 ## 88.0.0-SNAPSHOT - unreleased
 
-### 💥 Breaking Changes (upgrade difficulty: 🟢 LOW - one behavior change to `FileChooser`)
+### 🎁 New Features
 
-* `FileChooser` now accepts as many dropped files as its `maxFiles` limit allows, rejecting only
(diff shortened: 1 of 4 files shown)


lbwexler:
Verified live in Toolbox against this branch (inline Hoist, real Grails backend). Toolbox `pnpm typecheck` is clean, and the FileChooser demo plus the FileManager example app both behave correctly under react-dropzone v20.

Exercised via `Other → FileChooser`:

| Scenario | Result |
|---|---|
| Drop 5 files, `maxFiles: 3`, empty selection | 3 added; only the 2 surplus rejected ("Too many files"). Pre-v19 this rejected all 5. |
| 2 already selected, drop 3 more, `maxFiles: 3` | Fills to 3, single toast: `File limit of 3 reached - 2 files not added.` |
| Drop `.pdf` + `.gif` against `.png`/`.txt` | Both rejected with `File type must be one of .png, .txt` - dummy MIME correctly hidden |
| Drop untyped `.txt` (no MIME at all) | Accepted - confirms the `_/_` extension-only accept path survives v18 unbundling the extension-to-MIME table |
| Single-file chooser, drop twice | Second file replaces the first, silently |
| FileManager example app | Both files queued with correct icons, sync enabled |

No console errors or react-dropzone warnings on load.

The second row is the case this PR's `addFilesInternal` rewrite targets - previously it produced a rejection toast *and* a "File limit exceeded" warning while adding nothing.

One gap worth naming: the drops above were driven with synthetic `DataTransfer`/`DragEvent`s, so everything downstream of `onDrop` is covered but the native file-picker path (`Browse` → OS dialog) and real OS drags are not. Worth a quick manual pass on those during review.

PeteDarinzo (approved):
Reviewed

PR #4054, opened 2025-07-29
Suppress Options for Messages

lbwexler:
Hoist P/R Checklist
-------------------

**Pull request authors:** Review and check off the below. Items that do not apply can also be
checked off to indicate they have been considered. If unclear if a step is relevant, please leave
unchecked and note in comments.

- [X] Caught up with `develop` branch as of last change.
- [X] Added CHANGELOG entry, or determined not required.
- [X] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so.
- [X] Updated doc comments / prop-types, or determined not required.
- [] Reviewed and tested on Mobile, or determined not required.
- [] Created Toolbox branch / PR, or determined not required.

**If your change is still a WIP**, please use the "Create draft pull request" option in the split
button below to indicate it is not ready yet for a final review.

> **Pull request reviewers:** when merging this P/R, please consider using a *squash commit* to
> collapse multiple intermediate commits into a single commit representing the overall feature
> change. This helps keep the commit log clean and easy to scan across releases. PRs containing a
> single commit should be *rebased* when possible.

PR #4624, opened 2026-08-26
Add message suppression feature with browser storage

### LINKED ISSUE #3890: Support `suppressOptions` argument to XH.confirm() and XH.alert()
FROM ATM

WDYT about adding an option to XH.confirm() that would include a standardized “don’t ask me again for this session” checkbox?
I’m imagining that arg could take a dev-defined key (arbitrary string) - something like suppressForSessionKey or snoozeForSessionKey as a cuter option.
Another option would be to take a PersistOptions object that would also support a more permanent snooze option via prefs or localStorage, but that gets a bit more complicated with a need to spec a suitable prompt and all that.

FROM LEE

maybe a `suppress: SuppressOpts` with 

key: string 
type?: 'session'|'localStorage'    (default session)
label?: string                                 (default something standard, sensitive to type)

lbwexler:
## Description

Adds a new `suppress` configuration option to `MessageSpec` that allows users to opt out of future copies of a message. When enabled, users see a "Don't show again" checkbox. If they confirm with the checkbox checked, their response is saved to browser storage and returned immediately by future calls with the same `messageKey`, bypassing the dialog entirely.

Closes #3890. Completes the earlier attempt in #4054.

### Core Features

- **`MessageSpec.suppress`**: New optional config (boolean or `MessageSuppressSpec` object) to enable message suppression.
- **`MessageSuppressSpec` interface**: Configuration object supporting:
    - `expiry`: Time (in ms) for the saved response to remain in effect, e.g. `30 * DAYS`. Default null suppresses indefinitely.
    - `storage`: `'local'` (default) or `'session'`. Specify `'session'` to suppress only for the lifetime of the current browser tab.
    - `label`: Custom checkbox label, auto-generated if omitted.
    - `initialValue`: Initial checkbox state (default false).
- **`MessageModel.getSuppressedResult()`**: Static check for a previously saved response, called before any dialog model is constructed.
- **Auto-generated labels**: Formatted from the configured scope and expiry, e.g. "Don't show again for 7 days", or "Don't show again in this tab" for session-scoped suppression.

### Implementation Details

- Suppressed responses are stored with an optional expiry timestamp and ignored once past it.
- `messageKey` is required whenever `suppress` is enabled - validated in the `MessageModel` constructor.
- `MessageSourceModel.message()` resolves immediately to a saved response, so no `MessageModel` is constructed and no already-open message sharing the key is disturbed.
- Desktop and mobile `Message` components both render the suppress checkbox via a form field.

### Documentation

- Updated JSDoc for `MessageSpec.messageKey` to note the suppression requirement.
- Added JSDoc for the `MessageSuppressSpec` interface.
- Updated `XH.message()` JSDoc to document suppression behavior.
- Added CHANGELOG entry describing the feature.

## Testing

Existing message dialog tests continue to pass. The feature is backward compatible - `suppress` is optional and defaults to disabled.

Manual testing should verify:

- Checkbox appears when `suppress` is enabled, and is absent otherwise.
- Confirmed responses are saved and returned by later calls with the same `messageKey`.
- Expiry timestamps prevent stale responses from being used.
- Session vs. local storage behavior works as configured.
- Auto-generated labels format correctly across expiry durations and both storage scopes.

A companion Toolbox change exercises the feature in the desktop Popups example (Other > Popups > Confirm > "with suppress").

## Checklist

- [x] Added CHANGELOG entry
- [x] Updated doc comments and type definitions
- [x] Reviewed for breaking changes (none - fully backward compatible)
- [x] Updated both Desktop and Mobile Message components
- [x] Validation added for required `messageKey` when suppression enabled


lbwexler:
@TomTirapani @amcclain  -- this seemed like a nice little piece of candy that dropped in relatively cleanly

CHANGELOG.md
```diff
@@ -16,8 +16,15 @@
 
 ### 🎁 New Features
 
+* Added new `MessageSpec.suppress` config for `XH.message()` and its `alert`, `confirm`, and
+  `prompt` variants. Set to `true` (or a config object) to offer users a "Don't show this message
```
amcclain, line 20: I don't mind it just being consistent and working across all messages, but it does seem a bit funny to allow it for `prompt` types.  Of course a dev would need to opt-in, so in the end I think we just leave it be and allow it.  Just talking it through here....

appcontainer/MessageModel.ts
```diff
@@ -168,4 +210,38 @@ export class MessageModel extends HoistModel {
...
+        const {suppress} = this,
+            {expiry} = suppress;
+        getSuppressStore(suppress).set(getSuppressKey(this.messageKey), {
+            value,
+            expiry: expiry ? Date.now() + expiry : null
+        });
+    }
+
+    // Humanized suppress expiry duration, expressed in the largest unit that divides it evenly.
+    private get suppressExpiryLabel(): string {
```
amcclain, line 224: Could consider `moment.duration().humanize()` but don't feel strongly

amcclain (approved):
Great - thanks for looping back on this and finishing it off!

Issue #4620, opened 2026-08-24
Data: unified calculated fields for Store and Cube View


lbwexler:
Captures the co-design of two converging efforts: the standing "calculated columns" project for Stores, and the lazy derived-fields alternative proposed in #4539 for Cube Views with global (pct-of-total style) aggregators. Evaluating them together showed ~80% shared machinery — one concept, shipped at two layers. Targeted for v88.

## Motivation

- **Stores**: applications want columns computed from a record's other fields without requiring them in source data or paying server round-trips.
- **Cube Views** (#4539): a single non-`dependsOnChildrenOnly` aggregator disqualifies an entire View from the incremental data-only update path — every tick pays a full generation, plus a near-full-breadth store/grid transaction when a shared denominator moves every row's value.

Both reduce to the same design: values computed lazily at read time via prototype getters on generated data classes, excluded from digest/equality contracts, with a repaint story for values whose inputs change outside their own row.

## The concept

`FieldSpec.calculatedFn` — one name at both layers, each declaring its own signature:

- **Store**: `(data, store) => any`
- **Cube** (`CubeFieldSpec`): `(row, aggContext) => any`

Values are read through prototype getters on a generated per-Store / per-View data class — the exposed-leaf getter-class technique from `RowDataGenerator` (#4586), generalized into `data/impl` and consumed by both layers.

Lazy getters rather than eager parse-time computation, for two reasons:

- `projectionOnly` stores never parse — adopted data can only carry computed values via getters. Cube-connected stores are exactly the ones pushed toward `projectionOnly`.
- Freshness by construction: a value whose inputs live outside its own row (a summary total, an aggregation-context memo) can change without the row itself changing. A getter is always current; a parse-time snapshot goes stale.

This is also how mature BI systems model such values — MDX/DAX calculated measures and Tableau table calculations are read-time expressions over eagerly-maintained aggregates; none implement pct-of-total as an aggregator.

## Store layer

Row-local computation plus store context. Notably, `store.summaryRecords` makes pct-of-total expressible with no Cube API at all when the View sets `includeRoot` + `loadRootAsSummary` — the denominator is O(1) at read and incrementally maintained upstream by the cube's existing `replace()` path:

```ts
{
    name: 'pctCommission',
    calculatedFn: (data, store) =>
        (data.commission / store.summaryRecords[0]?.data.commission) * 100
}
```

[...]

PR #4636, opened 2026-08-31
Data: unified calculated fields for Store and Cube View (#4620)

lbwexler:
Implements #4620 - `calculatedFn` fields at both data layers - plus the follow-on hardening that grew out of in-PR review, including a breaking change to connected-store configuration.

## Key changes

* **`FieldSpec.calculatedFn: (data, store) => any`** - Store fields computed on the client at read time via lazy prototype getters on record `data`. Never stored, parsed, or included in equality/digest comparisons; always current, even when inputs live outside the row (e.g. a `Store.summaryRecords` denominator). Read-only: bound grid columns are never editable and `modifyRecords()` throws.
* **`CubeFieldSpec.calculatedFn: (row, ctx) => any`** - the Cube-layer form, computed on View rows with the `AggregationContext`. Carries no aggregator, so pct-of-total-style Views retain the incremental data-only update path. `ctx.filteredRecords` is readable from these fns and always current (sourced lazily from `_records` or the patched leaf map, per update type).
* **Declarative freshness** - grids repaint calculated columns after each transaction; a `FieldFilter` on a calculated field upgrades incremental refilters to full; sorts on calculated columns are restored through the paced deferred-sort flush, including on empty (summary-only) syncs.
* 💥 **Connected stores are now always `projectionOnly`** - new `Store.connectView(view)` / `disconnectView()` encapsulate connected-store adoption (row digest, projection mode, calculated field name marking), called by the View on connect/`setStores`/query change/destroy. An explicit `projectionOnly: false` or `processRawData` on a connected store now throws - parse mode offered connected stores nothing that worked (`parseVal` is a no-op on typed view rows; `updateData` never applied `processRawData`). See the CHANGELOG entry and new `docs/upgrade-notes/v88-upgrade-notes.md`.
* **Generator symmetry** - Store record data generation (shared defaults/template plus the generated dense and projection classes carrying calculated getters) is extracted to `data/impl/RecordDataGenerator.ts`, mirroring the Cube's `RowDataGenerator` - parent `ViewRowData` rows are themselves now generated-class instances. Both generators consume the shared getter util `data/impl/FieldGetterSupport.ts` (the #4586 technique, generalized).

Follow-on filed: #4637 - reconcile connected store fields with the View's query fields, deleting the calc-name marking shim inside `connectView`.

Companion Toolbox branch/PR exercises the feature in the cube tester: `PctTotalAggregator` rewritten as a calculated field, no `editable`/`projectionOnly` config needed, and the projection A/B switch removed.

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (Connected-store `projectionOnly` enforcement - CHANGELOG breaking section + v88 upgrade notes.)
- [x] Updated doc comments / prop-types, or determined not required. (JSDoc throughout, plus `data/README.md` and `data/cube/README.md` sections.)
- [x] Reviewed and tested on Mobile, or determined not required. (Shared `GridModel`/`Store` machinery - no platform-specific code touched.)


lbwexler:
Superseded by #4638, which contains this PR's full history plus the #4637 follow-on work, now retargeted to develop as the single reviewable unit.

PR #4647, opened 2026-09-02
Inspector overhaul: standalone window, Watchlist, persistent memory stats, instance labels

lbwexler:
Overhauls the Hoist Inspector into a standalone developer tool that stays out of the app's way and keeps its context across reloads.

**Window**
- Activating the Inspector opens it straight into its own browser window, one per app tab (named and titled by `XH.tabId`), with top-level `Objects` and `Memory` tabs switched from the header. Closing the window deactivates the service; a blocked popup on reload deactivates with a toast.
- Tab button brings the owning app tab to the front (`InspectorModel.focusApp()`).

**Objects**
- Left-hand `All` grid of live instances plus a new `Watchlist` tab, replacing Favorites. Star any instance or property to add it to the Watchlist, where starred instances (top) and properties (bottom) stay in view regardless of selection or filters. Selecting a watched instance drives the Properties and Diagnostics detail tabs just like the All grid.
- Watchlist entries for instances with an `xhName` persist across reloads (placeholder rows until alive again); unnamed entries are keyed by `xhId` and last the page load.
- New `Anon` quick filter - unnamed instances hidden by default. Grid sorted by label, Sync column hidden by default.
- Properties: `Getters` toggle to auto-evaluate getters; `Log to console` and `Load all getters` move to the context menu. Throwing getters render their error rather than breaking the grid.
- `InstancesModel` split into `AllPanel`, `details/` (Properties, Diagnostics, shared `BasePropsModel`) and `watchlist/`.

**Memory**
- Last hour of stats persisted per browser tab, so the trend leading up to a reload survives it.
- Stats grid becomes a fixed-width side panel so the chart takes the remaining width; y-axis bands rebalanced.
- Force GC button, enabled when Chrome is launched with `--js-flags=--expose-gc`.

**Instance labels (framework)**
- `parseNameSource` labels HoistBase instances as `ClassName [xhName]`, or `ClassName [id]` when unnamed. An `xhName` matching the class name (services, `appModel`) marks a singleton and collapses to the bare class name. Inspector uses the same label via `formatInstanceLabel`.
- Dropped the bare id/name `xhName` fallback on `TabModel`, `FormModel` fields, `DashViewModel`, `DockViewModel` and `PageModel` - children of unnamed parents stay unnamed.

README and CHANGELOG updated. Verified live in Toolbox.


lbwexler:
@RyanMG @amcclain  -- think this solves a few of the problems we were running into when trying to troubleshoot with the inspector last week!   Hopefully the combined "watchlist" persistence and extra room make sense to you guys.

amcclain:
Still on my shortlist - I have a design brief I'm working on and will revert back here as soon as I have a chance to action it, should be a nice little side project for the weekend.

lbwexler:
@Anselm McClain ***@***.***> -- thanks for the update

PR #4652, opened 2026-09-02
Admin Console detail panels for Config, User Preferences, and JSON Blobs

amcclain:
Adds a read-only detail panel beside the Config, User Preferences, and JSON Blobs admin grids. The panel tracks grid selection and shows the full record, so the RestGrid editor dialog is now for editing only.

**What changed**

- **Shared `restDetailPanel`** (`admin/detail/`) - finds the enclosing `RestGridModel` from context, mirrors the selected record into a read-only `FormModel`, and provides the docked panel, persistence, and Edit button. Each tab supplies a `renderForm` callback of `formFieldSet`s.
- **Config** - panel shows all views of the value (Resolved / Instance / Database / Defaults) via the existing `configValue` tabs, opening on the most-derived. Notes render as Markdown. The editor dialog now opens JSON on the Database tab, drops the save-then-reopen workaround, and double-click is a no-op for read-only admins.
- **User Preferences / JSON Blobs** - new panels. Blob Value and Meta editors split 3:2; Meta is omitted when empty.
- **Admin read-only form style** - `xh-admin-readonly-form` class gives stacked, muted, uppercase labels via the form-field CSS vars. Also applied to the Roles detail view.
- **Type badges** - `valueTypeRenderer` shows config value types as icon + colored badges in the grid and panel.
- **Helpers** - `valOrNa`, `naSpan`, `yesNoRenderer`, `markdownRenderer` in `AdminUtils`.

**Where to look**

- `admin/detail/RestDetailPanel.ts`, `RestDetailModel.ts` - the reusable piece.
- `admin/tabs/general/config/ConfigDetailPanel.ts`, `ConfigValue.ts` - config specifics and the default-tab / fill-sizing changes.
- `admin/App.scss` - the read-only form and badge styling.

**Testing**

Verified in Toolbox against JSON, string, int, and pwd configs, prefs, and blobs, including a save round-trip refreshing the panel, in light and dark themes. No local config has an instance override, so that tab was not exercised live; its logic is unchanged apart from the default tab.

**Screenshots**

CleanShot 2026-09-02 at 16 21 07@2x

CleanShot 2026-09-02 at 16 30 35@2x

Hoist P/R Checklist
-------------------

**Related:** xh/hoist-core#599 reworks the notes on the bootstrapped `xh*` configs to render well in the new detail panel. Recommended together, not required in either direction.


amcclain:
Hi @lbwexler  and @PeteDarinzo 

This change was motivated by the awesome updates to our config editor to show the database versus resolved values. I really love that change, but it has one drawback, which is that I open up the config dialog both to view configs and to edit them. When viewing them, showing the resolved tab is exactly what we want, but I'm more often open to edit the tab. I find myself briefly being confused as I attempt to modify the read-only resolved values. Then I remember to switch to the database tab, and I'm ready to go. 

Adding a dedicated reader pane seemed to improve the situation all around. First of all, we've got a reader pane. No need to pop in and out of things or try and read values in the grid column. And then the editor dialog can be optimized for that task and open up to the database sub-tab for JSON configs so it's ready-to-change.

Since this change introduced the general pattern of the docked readonly detail panel, it seemed worth bringing the same to the user prefs and JSON Blob tabs.

Finally, a bit of extra flair with e.g. config notes rendering markdown, the little type tags, and the built-in demo of a nicer looking set of form field styles.

Hope you think it looks solid and we can take it, happy to discuss any aspect of it, of course.

lbwexler:
this is really great -- would like to do a quick follow-on where we no longer cause the server to serialize/output the entire blob database -- that's a huge liability for us right now.  Am going to look at that today -- have been wanting to close that ticket for a while

Issue #4559, opened 2026-08-10
NavigatorModel: back-navigation (slidePrev) silently no-ops after Swiper 12→14 bump


febbraiod:
Observed in Alpha testing of Flagship app at client after Hoist upgrade.

Symptom: After navigating forward at least once, using the app menu or the in-app back-caret to go back does nothing — no error, router/URL state updates fine, but the visible page never changes. Swiping back manually works fine; only the programmatic back-navigation path is affected. Only reproduces on-device (iOS); not visible in Chrome's emulator.

Root cause: onRouteChange()'s backOnePage branch calls this._swiper.slidePrev(transitionMs) directly, with no this._swiper.update() first — unlike every other branch, which calls update() before slideNext()/slideTo():

if (backOnePage) {

    this._swiper.slidePrev(transitionMs);          // no update() first

} else {

    this.stack = stack;

    this._swiper.update();                          // update() called here

    ...
}

allowSlidePrev/allowSlideNext are passed to Swiper as reactive React props (Navigator.ts), but the live Swiper instance only appears to pick up the current value on update(). Since slidePrev() is only ever called from the one branch that skips update(), allowSlidePrev stays frozen at its initial-mount value (false, since the stack starts with just one page) for the rest of the session — so Swiper silently rejects every subsequent slidePrev() call.

Confirmed live on device: navigatorModel.allowSlidePrev (the model getter) reads true, but navigatorModel._swiper.params.allowSlidePrev reads false, at the exact moment back-navigation fails.

Suggested fix: add this._swiper.update() before this._swiper.slidePrev(transitionMs) in the backOnePage branch, matching the other branch.

febbraiod:
Traced this to an upstream Swiper v14 regression. Confirmed on device. The observation in the original post was right about *what* is stuck, but wrong about the mechanism, and the fix suggested there would not have worked -- details below.

## Upstream root cause

`shared/utils.mjs`, `elementTransitionEnd()`. v12:

```js
function elementTransitionEnd(el, callback) {
    function fireCallBack(e) {
        if (e.target !== el) return;
        callback.call(el, e);
        el.removeEventListener('transitionend', fireCallBack);  // removed only on a MATCHING event
    }
    if (callback) el.addEventListener('transitionend', fireCallBack);
}
```

v14 rewrote this as a one-shot listener while keeping the target guard:

```js
function elementTransitionEnd(el, callback) {
    if (!callback) return;
[...]

PR #4658, opened 2026-09-04
Navigator: make back-navigation resilient to a missed transitionEnd

lbwexler:
Addresses the two latent items called out at the end of #4559. The Swiper 14 regression itself was handled by the revert in `97fa16d27`, and is now filed upstream as nolimits4web/swiper#8229 (still present in 14.2.0). This PR fixes the two things in `NavigatorModel` that turned that upstream timing bug into a silent, unrecoverable dead end - both of which survive the revert.

### `activePageIdx` is now observable state

It read `this._swiper.activeIndex`, a plain non-observable property, and `activePage`, `activePageId`, `allowSlideNext` and `allowSlidePrev` all derive from it. `allowSlidePrev` had no observable input at all - it was refreshed purely as a side effect of `onPageChange` reassigning `stack` (via `slice()`, so always a new identity) on every transition end. Correctness also depended on React deferring the init-branch render past an imperative `activeIndex` write.

It is now a `@bindable` synced from Swiper's `activeIndexChange` event, which fires from `updateActiveIndex` independently of the transition machinery.

### `backOnePage` has a backstop

It is the only branch that defers its state change to the animation - the router has already moved to the parent route while `stack` still holds the child. If `transitionEnd` never lands, the two disagree permanently: `onRouteChange` fires only on route *changes*, so a second back-tap invokes nothing at all and the app needs a reload. The Swiper regression was one way to lose the event; an interrupted transition or iOS suspending the app mid-slide are others.

The branch now reconciles the stack on a `transitionMs + 100` timeout if the event has not been processed. `onPageChange` is already effectively idempotent, so the guard only avoids churn on the happy path.

It also sets `allowSlidePrev` on the instance immediately before `slidePrev`, where render timing cannot defeat it - the call-site lock from the #4559 plan, which was never applied.

### To eyeball on device

The active page now flips at the *start* of a transition rather than at its end, since `activeIndexChange` fires when `slideTo`/`slidePrev` begins. A `renderMode: 'lazy'` page therefore renders as the slide begins instead of popping in when it lands. That looks like the better behavior, but it does mean a heavy page renders during the animation rather than after it - worth a look on-device before merge. Happy to split live index from settled active page if you would rather keep the old timing.

One case the backstop deliberately does not cover: if `slidePrev` is rejected outright, the slide never moves and trimming the stack would produce a visual jump. The call-site lock is what prevents that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@
 * `Store` and `Cube` now throw a clear error at construction when given fields with duplicate
   names. Previously such a `Cube` failed later with a cryptic `Cannot redefine property` error when
   creating a `View` that exposes leaves.
+* Fixed mobile `Navigator` back-navigation leaving the page stack and the route permanently out of
+  sync if a page transition failed to report its completion. The stack is now reconciled on a
+  timeout, `NavigatorModel.activePageIdx` is observable state synced from Swiper rather than a read
+  of its non-observable `activeIndex`, and back-navigation unlocks Swiper at the call site.
 
 ### ⚙️ Technical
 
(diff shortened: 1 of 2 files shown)


lbwexler:
@febbraiod -- let me know if you have 15 minutes to consider this one?

PR #4659, opened 2026-09-04
Cube: compositional aggregation for AVG / AVG_STRICT

### LINKED ISSUE #4601: Cube: compositional (stateful) aggregation for AVG / AVG_STRICT
## Problem

`AverageAggregator` and `AverageStrictAggregator` implement only `aggregate`, which walks the **entire subtree's leaves** via `forEachLeaf` - averages can't be composed from children's published averages the way sums compose from sums. Two costs follow:

* **Full builds / regroups**: `ParentRow.init` re-aggregates at every level, so AVG fields cost O(N × depth) per build instead of O(N).
* **Incremental ticks**: AVG is `dependsOnChildrenOnly`, so the fast `dataOnlyUpdate` path is taken - and then the inherited `Aggregator.replace` (full re-aggregate) walks all leaves under **every ancestor** of each changed leaf. One changed leaf in a 200k-leaf cube with `includeRoot` re-reads every leaf at the root level.

These are the last leaf-walking built-in aggregators - SUM, MIN, MAX, and LeafCount all compose from direct children.

## Proposed design

Make `aggregate` itself compositional via per-parent-row state; the inherited `replace` (= re-aggregate) then automatically becomes O(direct children). No `replace` override needed.

* **State slot on `ParentRow`** - lazily-created `aggStates` object keyed by field name, on the row instance (not `ViewRowData`, which is published; not `AggregationContext.appData`, which is rebuilt per tick). Parents are few relative to leaves. Leaves need no state - their contribution derives from `data[fieldName]`.
* **Plumbing** - `AggregationContext.aggregate/replace` already wrap calls with `activeField` try/finally; add the owning row the same way and expose state write + child-state reads. Internal signature change only.
* **`AverageAggregator.aggregate`** - one pass over direct children: leaf → `(val, 1)` if non-null, parent → its stored `{total, count}`; store own state, return `count ? total / count : null`. **AVG_STRICT** adds a composed `hasNull` flag, publishing null when set.
* Bottom-up guarantees already hold on both paths: `groupAndInsertRecords` builds children before constructing parents, and `RowCache.getOrCreate` resolves children before `reuse()`.

Precedent in-tree: `LeafCountAggregator` already composes from direct children because its published value *is* its state - AVG just needs two numbers instead of one.

## Care points

* **Row-reuse staleness** (per the v87 row-reuse audit rule - every per-generation output on a reused row must be assigned every visit): state must be rewritten wherever `data[name]` is (`init`, `recomputeAggregate`) and **cleared when `canAggregateFn` flips to false** - today `recomputeAggregate` nulls the value without invoking the aggregator, which would leave stale state for ancestors to read.
* **`canAggregateFn` semantics** - today the two paths disagree about canAgg-false mid-tree rows: full builds walk through them (`forEachLeaf`), while `ParentRow.applyDataUpdate` stops propagation at them. Compositional AVG makes both paths behave like the incremental one (canAgg-false rows contribute nothing upward). No observable change for the standard monotonic pattern (false above some depth, ancestors also false/null); non-monotonic configs would see full-build results change *to match what incremental updates already produce*. Confirm acceptable before building.
* Incremental FP drift is the same trade `SumAggregator.replace` already accepts.

## Follow-ons

* The state protocol is exactly what app-side **weighted average** aggregators need - decide whether the state read/write API is public or stays `@internal` initially.

## Impact

Build/regroup for AVG fields: O(N × depth) → O(N). Single-leaf tick: O(all leaves × ancestors) → O(children along the ancestor chain) - hundreds of reads instead of hundreds of thousands.

Deferred from the 2026-08-20 perf review (see PR #4600 for the items taken immediately) - not being built now to avoid further API changes at this stage of the release.

lbwexler:
Fixes #4601.

`AVG` and `AVG_STRICT` were the last built-in aggregators to walk their whole subtree of leaves - every sibling composes from its direct children. That made them diverge from the rest by a factor of tree depth on builds, and by the size of each ancestor's subtree on incremental ticks, where `dependsOnChildrenOnly` picks the fast `dataOnlyUpdate` path and the inherited `replace` then re-aggregates from scratch under every ancestor. Nothing at the call site (`aggregator: 'AVG'`) hinted at it.

Both now compose from their direct children, keeping the running total and count they need as per-row aggregator state, and both implement `replace` in terms of the originating leaf's change - which, total and count being plain sums over leaves, applies unchanged at every level of the ancestor chain.

### Key changes

* `AverageAggregator` / `AverageStrictAggregator` compose from direct children; `AVG_STRICT` carries a composed `hasNull` flag and falls back to a re-scan only when a leaf value crosses null.
* State held in a lazily-created `aggStates` map on `ParentRow` (leaves need none), written via new **public** `AggregationContext.setAggState()` / `getAggState()` and cleared when a row stops aggregating a field.
* `RowUpdate` retains the originating leaf's values, since each ancestor rewrites `oldValue`/`newValue` on the way up.
* Added the **`ViewRow`** interface, documenting the row-level API passed to aggregators and to the `lockFn` / `omitFn` / `bucketSpecFn` hooks. Those are long-standing public config hooks that typed their rows with unexported internal classes (`AggregateRow | BucketRow`, `BaseRow`), so apps could never annotate them - and two internal call sites needed `as any` casts, now gone.
* Cube README gains a **Custom Aggregators** section with a worked `WeightedAverageAggregator`, since the state protocol is exactly what app-side weighted averages need.

### Performance

200k rows, four dimensions, six averaged fields:

| | before | after | `SUM` floor |
|---|---|---|---|
| build | 448ms | **325ms** | 329ms |
| regroup | 277ms | **169ms** | 172ms |
| 100 single-record updates | 8812ms | **1439ms** | 1427ms |
| 20 full-row updates | 2211ms | **788ms** | 834ms |

AVG now costs what SUM costs on every path. A field-level probe confirms the mechanism: one build went from 750,000 leaf reads to zero, one tick from 276,927 to zero. With a single averaged field the win is smaller (2247 -> 1423ms on ticks) - the multiplier is the field count.

### Behavior change worth a look

A row whose `canAggregateFn` returns false now contributes nothing to its ancestors' averages. This matches how incremental updates have always treated such rows (`ParentRow.applyDataUpdate` stops propagating at them) and how every other aggregator already behaves - `SUM` composes from published values, so a non-aggregating row has always contributed `null` upward. Previously a full rebuild would compute an ancestor's average from all descendant leaves, including those under non-aggregating rows, giving a value that was correct at build time and then frozen under live updates.

Predicates written over `appliedDims` - the idiomatic "this measure is only meaningful once grouped by symbol" form, and inherently monotonic since `appliedDims` only grows downward - see no change at all. Only a predicate that turns false *below* a row it leaves true is affected. `CubeField.canAggregateFn` now documents the contribution rule.

### Validation

No test infra in this repo, so this was validated with a node harness (in the untracked `.planning/bench`): 1012 assertions comparing every aggregate row's published value against a brute-force walk of its own leaves, across full builds, builds with nulls interspersed, `modifyRecords` ticks (the `replace` path, cycling values through null), full-row updates, five successive regroups, ticks after regrouping, and both `canAggregateFn` shapes. The README's `WeightedAverageAggregator` is run verbatim as part of that suite, so the documented example is covered too.


data/cube/row/ParentRow.ts
```diff
@@ -173,12 +177,15 @@ export abstract class ParentRow extends BaseRow {
...
+            {name} = field;
+
+        let val = null;
+        if (this.canAggResults?.[name] !== false) {
+            val = view._aggContext.aggregate(children, field, this);
+        } else if (aggStates) {
+            aggStates[name] = null;
+        }
 
         if (data[name] === val) return false;
```
haynesjm42: This early return is no longer safe, and I think it's the one blocking issue on the PR.

`cubeRowDigest` now carries two meanings that have diverged:

1. "my published data changed" - consumed by connected stores for record reuse (`View` installs `digestSpec: 'cubeRowDigest'`). Value-change-only is correct here, and is the optimization `RowCache`'s header advertises.
2. "my ancestors must re-aggregate" - consumed by `reuse()` above: `!children.some(it => it.data.cubeRowDigest > genStartDigest)`.

Before this PR those were the same thing, because an ancestor only ever consumed `child.data[name]` - aggregators walked to the leaves themselves. Now a row also publishes `aggStates`, and returning `false` here suppresses the `assignDigest` in `reuse()` even though the state we just wrote via `setAggState` may be completely different. Meaning 1 stays correct; meaning 2 breaks.

Two concrete failures:

**a) `canAggregateFn` flip with a coincidentally-unchanged value.** Field `F` = `AVG_STRICT` with a `canAggregateFn`. Row `A` leaves `[null, 10, 20]` -> `aggStates.F = {total: 30, count: 2, hasNull: true}`, `data.F = null`. Sibling `A2` -> `{40, 4, false}`, `data.F = 10`. Parent `B` -> `{70, 6, true}`, `data.F = null`. Next generation `A`'s `canAggregateFn` flips false: we take the `else if (aggStates)` branch, set `aggStates.F = null`, `val` stays `null`, `data[name] === val` -> return `false`, so A's digest is not bumped. At `B`: children equal, not a bucket, no child digest moved -> `simpleAggsAreCurrent`. The view does have `canAggregateFn` fields so `hasContextDependentFields` is true and `B` enters the `else if`, but neither sub-case fires - `AVG_STRICT` is `dependsOnChildrenOnly` so it's absent from `_complexAggFieldsByDepth`, and `force` is `B`'s *own* changed `canAggResults`, which didn't change. `B` never re-aggregates and publishes `null` where the new documented semantics ("contributes nothing to the aggregations of its ancestors") require `10`.

**b) Corrupted state that poisons the next incremental update.** Plain `AVG`, no `canAggregateFn`, so `B` skips outright. `A` leaves `4, 6` -> `{10, 2}`, `data.F = 5`; `A2` -> `{5, 1}`, `5`; `B` -> `{15, 3}`, `5`. Add a leaf with `F = 5` under `A` (structural, so `fullUpdate` -> `generateRows`): `A` re-aggregates to `{20, 4}`, value still `5`, digest not bumped; `B` skips and keeps `{15, 3}` against children now totalling `{20, 4}`. B's published value is coincidentally still right, so nothing looks wrong - until the next data-only update to that leaf (`5 -> 9`) runs `AverageAggregator.replace` on the stale state and publishes `19/3 = 6.33` where the truth is `(4+6+9+5)/4 = 6.0`. Adding identical duplicate leaves is the easy real-world trigger: every AVG value is unchanged while every state doubles.

Suggested fix - don't widen the digest to cover state changes, that costs the store record reuse `RowCache` documents. Split the two meanings instead:

- Add an internal aggregation stamp on `BaseRow`; leaves can mirror their data digest, since for a leaf "value changed" *is* the right ancestor signal.
- On `ParentRow`, bump it whenever the row actually re-aggregated a stateful field or its `canAggResults` changed, regardless of published value. `recomputeCanAggregate` already returns the changed names, so that half is nearly free.
- Point `simpleAggsAreCurrent` at that stamp, and leave `cubeRowDigest` value-change-only.

Reusing `view._rowDigest` as the counter keeps the `> genStartDigest` comparison working as-is. Ordering is fine either way - `groupAndInsertRecords` recurses children-first, so a parent always sees this generation's child stamps.

data/cube/aggregate/AverageAggregator.ts
```diff
@@ -7,20 +7,52 @@
...
-                total += val;
-                count++;
+        for (const row of rows) {
+            if (row.isLeaf) {
+                const val = row.data[fieldName];
+                if (val != null) {
+                    total += val;
+                    count++;
+                }
+            } else {
```
haynesjm42: A non-leaf child with no aggregator state is silently dropped here rather than descended into, which regresses a case the old `forEachLeaf` implementation handled.

`View.buildIndices` sets `_aggFieldsByDepth[d] = aggFields.filter(it => !appliedDimNames[d].has(it.name))`, so a field is excluded from aggregation at exactly the depths where it is applied as a dimension. Nothing in `CubeField` forbids a dimension field from also carrying an aggregator. So with `dimensions: [year]` and `year: {aggregator: 'AVG'}`, the depth-0 total row aggregates `year` over depth-1 rows that never ran `ctx.aggregate` for it - `getAggState(row)` is `undefined`, the child is skipped, `count` ends at 0, and the row publishes `null`. The old leaf walk produced the real average.

Either fall back to a leaf walk when a non-leaf child has no state, or read the child's `data[fieldName]`. Same at `AverageStrictAggregator` line 31.

data/cube/aggregate/AverageAggregator.ts
```diff
@@ -7,20 +7,52 @@
...
+                }
             }
-        });
+        }
+
+        context.setAggState({total, count});
+        return count ? total / count : null;
+    }
+
+    override replace(rows, currAgg, update, context) {
```
haynesjm42, line 46: Minor, but worth a deliberate decision rather than leaving it accidental: floating-point error in the running total is now permanent.

`replace` mutates `state.total` by leaf deltas, and `aggregate` composes parents from their children's already-drifted `state.total`. Nothing ever re-derives a total from leaf values, so on a long-lived streaming view the drift accumulates monotonically - e.g. an AVG over values that net to zero eventually reporting `~1e-16`. Previously every re-aggregation walked the leaves and reset the error to zero.

`AverageStrictAggregator.replace` has the same property. If we're accepting it, a note in the class doc would help; otherwise some periodic exact re-derivation.

data/cube/aggregate/AggregationContext.ts
```diff
@@ -62,12 +69,13 @@ export class AggregationContext {
...
      */
-    aggregate(rows: BaseRow[], field: CubeField): any {
+    aggregate(rows: ViewRow[], field: CubeField, row: ParentRow): any {
         this.activeField = field;
+        this.activeRow = row;
         try {
             return field.aggregator.aggregate(rows, field.name, this);
         } finally {
-            this.activeField = null;
+            this.activeField = this.activeRow = null;
```
haynesjm42: This should save and restore the previous values rather than nulling them.

The base class explicitly contemplates delegation ("Aggregators that delegate to other aggregators must all agree on this setting"), and an aggregator that delegates via `ctx.aggregate(...)` now returns to its outer frame with `activeRow` and `activeField` both `null`. A subsequent `setAggState()` then throws `TypeError: Cannot set properties of null` on `(activeRow.aggStates ??= {})`, and `getAggState()` throws on `this.activeField.name`. Before this PR the same reset only degraded the `filteredRecords` guard's warning; now it's a hard crash.

Same in `replace` below.

data/cube/aggregate/Aggregator.ts
```diff
@@ -66,10 +65,10 @@ export abstract class Aggregator {
      * @param rows - array of child rows
      * @param fn - the function to call on each leaf.
      */
-    protected forEachLeaf(rows: BaseRow[], fn: (leaf: LeafRow) => boolean | void): boolean {
+    protected forEachLeaf(rows: ViewRow[], fn: (leaf: ViewRow) => boolean | void): boolean {
```
haynesjm42: Narrowing the callback param from `LeafRow` to `ViewRow` is a breaking change for application aggregators that isn't called out in the changelog.

`ViewRow` has no `cubeRecord` / `cubeRecordId`, so any app aggregator reaching the source record from a leaf - `this.forEachLeaf(rows, leaf => leaf.cubeRecord.data...)` - no longer compiles, and `LeafRow` isn't exported from `data/index.ts`, so there's no replacement type to annotate with either.

Either expose the leaf-specific surface on `ViewRow` (or add a `ViewLeafRow`), or add a bullet under "⚙️ Typescript API Adjustments".

data/cube/README.md
```diff
@@ -77,6 +78,72 @@ const cube = new Cube({
...
+arrive as the row's direct children - a mix of leaf rows and already-aggregated parent rows, typed
+as `ViewRow` - so most aggregations compose naturally from `row.data[fieldName]`.
+
+Aggregations that cannot be derived from their children's published values alone (a weighted
+average, a standard deviation) can keep the extra terms they need as **aggregator state**, via
+`AggregationContext.setAggState()` / `getAggState()`. This keeps each row's work proportional to
+its child count rather than to its entire subtree of leaves:
+
+```typescript
+export class WeightedAverageAggregator extends Aggregator {
```
haynesjm42, line 93: This example goes stale on incremental updates as written - it needs `dependsOnChildrenOnly = false`, or the section needs to state the constraint.

It reads a *second* field (`this.weightField`, e.g. `qty`) off its children. `ParentRow.applyDataUpdate` only re-aggregates fields present in `childUpdates` - i.e. fields whose own value changed on the leaf - and `View.getSimpleUpdates` admits the `dataOnlyUpdate` path whenever `aggregatorsAreSimple` (all aggregators `dependsOnChildrenOnly`). So a leaf update that changes `qty` but not `price` takes the fast path, emits a `RowUpdate` for `qty` only, and the weighted-average `price` aggregate is never recomputed - it silently reports the pre-update weighting until the next full regroup.

Since this is the headline example for the new API, it's the one place we don't want a subtly wrong pattern to get copied.

data/index.ts
```diff
@@ -37,7 +37,9 @@ export * from './cube/Cube';
 export * from './cube/CubeField';
 export * from './cube/Query';
 export * from './cube/View';
+export * from './cube/ViewRow';
 export * from './cube/ViewRowData';
+export * from './cube/row/RowUpdate';
```
haynesjm42, line 43: While we're here: `BucketSpec` is the one public hook the de-internalization pass missed.

`BucketSpec.bucketFn` is still typed `(row: BaseRow) => string`, naming the `@internal` class, while its sibling `BucketSpecFn` in `Cube.ts` moved to `ViewRow[]` in this PR. Compounding it, `BucketSpec` itself is exported from nowhere - `data/index.ts` omits it and only `Cube.ts` imports it - so an application implementing `bucketSpecFn` can't name its own return type.

Switching `bucketFn` to `ViewRow` and adding `export * from './cube/BucketSpec'` alongside these new exports would finish the job.

data/cube/row/ParentRow.ts
lbwexler: new fix will be to make sure we check aggregation state every generation and update the digest.  Great catch John

data/cube/aggregate/AggregationContext.ts
lbwexler: Reverted to clearing these on exit in bd34ab81b. Aggregators delegate by calling another aggregator directly with the same context - `AggregationContext.aggregate()` is internal and takes a `CubeField` and `ParentRow` an app aggregator never has, so there is no nested call through it to restore for.

data/cube/aggregate/AverageAggregator.ts
lbwexler: Resolving with the second option kept, and the semantics made deliberate rather than accidental.

We considered going further and giving leaf-weighted results for this case (having rows at or below the grouping level run the aggregator for its state alone), and had it working - but it meant running aggregators whose values are then discarded, for a configuration that is already questionable. The rule we settled on instead, now documented on `CubeFieldSpec.aggregator` and in the Cube README:

- A field is never aggregated at or below the level at which it is applied as a dimension - rows there publish the dimension value.
- Rows above aggregate over those published values, one term per grouped row.

So `MIN`, `MAX` and `UNIQUE` remain meaningful for a field that is also a query dimension, while population-dependent aggregators are not - and `AVG` now simply behaves the way `SUM` always has here. This also matches how BI tools frame it: a dimension and a measure over the same column are two separate fields, and the docs point an app wanting to average a grouping field at declaring a second measure field over the same value.

No throw or warning, since `isDimension` means "can be grouped by" - the same field is a perfectly good measure whenever it is not in the current query's dimensions, and a dimension-chooser toggle should not blow up.

PR #4680, opened 2026-09-10
Center checkbox + radio indicators in compact Toolbars

### LINKED ISSUE #4671: RadioInput options overflow a compact Toolbar
# Summary

`RadioInput` does not fit inside a compact `Toolbar`. In `toolbar({compact: true})` the radio options keep their standard size, so the indicators and labels sit above the toolbar's vertical center and are clipped at the top edge. The same options render correctly in a standard toolbar.

# Reproduction

1. Open Toolbox's RadioInput page, "In a Toolbar" section (`/app/forms/radioInput`, xh/toolbox#900): the Standard frame is fine, the Compact frame shows the circles pushed up against and cut off by the toolbar's top edge.
2. Or, minimally: `toolbar({compact: true, items: [radioInput({inline: true, options: ['Equity', 'Fixed Income']})]})`.

Compare with `checkbox` and `switchInput` in the same compact toolbar (Toolbox's Checkbox & Switch page), which fit.

# Likely cause

`desktop/cmp/toolbar/Toolbar.scss` `&--compact` shrinks buttons, `.xh-input` / `.bp6-input-group` and select-style controls to `--xh-tbar-compact-item-height-px` (20px) and `--xh-tbar-compact-font-size-px`, but has no rule for Blueprint `.bp6-control` radio options: the 16px `.bp6-control-indicator` plus Blueprint's control line-height and the option margins from `RadioInput.scss` exceed the 24px compact toolbar (`--xh-tbar-compact-min-size`). The only toolbar-specific rule in `RadioInput.scss` (`.xh-toolbar .xh-radio-input { display: flex; ... }`) lays the options out horizontally and does not vary with compact.

# Suggested fix

Add a compact variant for radio options in toolbars, e.g. under `.xh-toolbar--compact .xh-radio-input .xh-radio-input-option`: reduce the indicator (`.bp6-control-indicator` font-size / size to ~12-14px), set `line-height` / `min-height` to `--xh-tbar-compact-item-height-px`, drop the vertical margin, and pick up `--xh-tbar-compact-font-size-px` for the label - mirroring what the compact rules already do for inputs. Worth checking `Checkbox` / `SwitchInput` alignment against the same token at the same time so all three controls share one compact-toolbar treatment.

Image

lbwexler:
Fixes #4671

### Cause

Blueprint aligns `.bp6-control-indicator` to the text *baseline* (16px inline-block with `margin-top: -3px`). In a compact `Toolbar`, `.xh-input` is pinned to `--xh-tbar-compact-item-height-px` (20px) and the type is smaller, so the indicator hangs above its container and overhangs the 24px toolbar - measured indicator top `137.7` vs toolbar top `137.4`, with 7.7px of unused space below it.

`Checkbox` has the same defect as `RadioInput` (it just reads as a less obvious overhang). `SwitchInput` already carried a `display: flex; align-items: center` rule in the compact block, which is why it looked right.

### Fix

In the compact block of `Toolbar.scss`, alongside the existing switch/select rules: lay checkbox and radio options out with flex and center them, replacing BP's baseline hang and its `padding-inline-start` / negative-margin indent with plain margins. The `bp6-align-right` variant (`labelSide: 'left'`) gets the row-reverse treatment the switch already uses, plus `float: none`, inert in flex.

Indicators stay at 16px rather than shrinking to 12-14px as the issue suggests - 16px is what the compact switch already renders, so all three controls now share one size and one alignment. Happy to add a size reduction if we'd rather have the indicators track the smaller type.

### Verification

The rule is scoped under `--compact`, so standard toolbars are untouched. Measured with a static harness built from the compiled SCSS (`vars` + `Toolbar` + `RadioInput`) and Blueprint's own DOM order:

| | before | after |
| --- | --- | --- |
| compact radio indicator | top 137.7 in a 137.4-161.4 toolbar (2.2px *above* its container) | top 141.9, bottom 157.9 - centered in the 139.9-159.9 input |
| compact checkbox indicator | top 212.4 vs container 214.6 | top 216.6, bottom 232.6 - centered |

Also checked `labelSide: 'left'`. A follow-up commit folds `SwitchInput` into the same rule (a `margin-top: 1px` fudge left it 1px low), sizes `SegmentedControl` to the compact item height when hosted by a compact toolbar, and centers an unlabeled `Slider` (`labelRenderer: false`, 16px in BP) - see the audit comment below. Labeled sliders are 40px and still overhang; that needs a call on whether a compact toolbar should suppress their labels. Worth an eyeball on Toolbox's RadioInput and Checkbox & Switch pages before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@
   may now also be changed after the first render.
 * Fixed mobile `Navigator` back-navigation leaving the page stack and the route permanently out of
   sync when a stale `allowSlidePrev` caused Swiper to silently skip the transition.
+* Fixed `RadioInput` and `Checkbox` options being clipped by top of the `Toolbar` - they
+  are now centered within the compact item height.
 * Fixed `dateEditor` crashing when opening its picker on a column backed by a `localDate` field -
   the editor now defaults its `valueType` from the Store field type.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,9 @@
   may now also be changed after the first render.
 * Fixed mobile `Navigator` back-navigation leaving the page stack and the route permanently out of
   sync when a stale `allowSlidePrev` caused Swiper to silently skip the transition.
-* Fixed `RadioInput` and `Checkbox` options being clipped by top of the `Toolbar` - they
+* Fixed `RadioInput` and `Checkbox` options overhanging the top of the `Toolbar` - they
   are now centered within the compact item height, sharing one rule with `SwitchInput`.
+* Fixed an unlabeled `Slider` (`labelRenderer: false`) sitting high in a compact `Toolbar`.
 * Fixed `SegmentedControl` overflowing a compact `Toolbar` - it now takes its compact sizing from
   the toolbar, without needing its own `compact: true`.
 * Fixed `dateEditor` crashing when opening its picker on a column backed by a `localDate` field -
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,9 @@
 * Fixed mobile `Navigator` back-navigation leaving the page stack and the route permanently out of
   sync when a stale `allowSlidePrev` caused Swiper to silently skip the transition.
 * Fixed `RadioInput` and `Checkbox` options being clipped by top of the `Toolbar` - they
-  are now centered within the compact item height.
+  are now centered within the compact item height, sharing one rule with `SwitchInput`.
+* Fixed `SegmentedControl` overflowing a compact `Toolbar` - it now takes its compact sizing from
+  the toolbar, without needing its own `compact: true`.
 * Fixed `dateEditor` crashing when opening its picker on a column backed by a `localDate` field -
   the editor now defaults its `valueType` from the Store field type.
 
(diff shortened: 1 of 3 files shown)


lbwexler:
Audited the other controls commonly dropped into a compact `Toolbar`, per the ticket's note on `SwitchInput`. Measured each item's painted content box against the 24px bar in the same harness:

| control | verdict |
| --- | --- |
| `SwitchInput` (both `labelSide`s) | was already flexed + centered, but a `margin-top: 1px` fudge left its indicator 1px below the newly centered checkbox/radio - folded into the same rule, all three now measure dead center |
| `SegmentedControl` | **overflowed** - tray rendered 31.3px inside the 20px compact item |
| `Button`, `ButtonGroupInput`, `Picker` | centered, within the bar |
| `TextInput` / `NumberInput` / `DateInput` (BP input-group), `Select` | centered, within the bar |
| `Slider` | overflows - BP renders it 40px tall with the handle label below. Pre-existing and not really a sizing tweak, so left alone; happy to open a separate issue. |

`SegmentedControl` sizes its buttons off `--xh-tbar-min-size-px` - a *standard* toolbar - so in a compact one the tray spills out of the bar unless you also pass `compact: true` to the control. Its own `--compact` styles now apply when it is hosted by a compact toolbar too, so that prop is no longer needed there. Measured 19.3px after the change, identical to passing `compact: true`.

TomTirapani (approved):

amcclain:
Thanks @lbwexler and @TomTirapani for grabbing this one and fixing - looking good again on new toolbox demo 👍

PR #4684, opened 2026-09-10
Add `Column.cellFlag` for Excel-style corner flags on grid cells

### LINKED ISSUE #4673: Grid: support Excel-style corner flags on cells via a new `Column.cellFlag` config
# Summary

Add a public `Column.cellFlag` config so apps can render a small Excel-style triangular flag in the
corner of a grid cell, colored by a standard Hoist `Intent` - e.g. to mark a stale price, an
overridden value, or a figure outside tolerance.

Hoist already draws exactly this glyph, but only for itself. `Grid.scss:107-183` renders a
`::before` triangle in the top-right corner for `.xh-cell--invalid` / `.xh-cell--warning` /
`.xh-cell--info`, sized off the grid's size mode (tiny 3px through large 6px). Those classes are
installed at `Column.ts:1060` inside `if (editor)`, keyed off `record.validationResults[field]`
severity. So the capability is unreachable unless a column is editable, its only input is validation
severity, it is locked to the top-right corner, and `success` has no flag at all.

This proposes promoting that private mechanism to a first-class column config and re-expressing the
validation flags on top of it, so there is one implementation rather than two.

# Motivation

Corner flags are a dense, low-cost way to mark individual cells without spending a column, changing
the cell's text, or fighting the row-level striping and selection colors that a full cell background
disturbs. Apps reach for them to annotate exactly the cases grid columns are bad at: a value that is
technically present but should be distrusted.

Today an app that wants one has to reimplement the glyph in its own SCSS - the transparent-border
triangle, the four size-mode breakpoints, and a light/dark color pair per intent - and wire up its
own `cellClassRules`. That is a fair amount of CSS to get right for something the framework already
ships internally, and app copies will not track changes to Hoist's grid sizing or intent palette.

# Proposal

## Config

Reuse the existing `Intent` and `Corner` types from `core/types/Types.ts` - `Corner` is already
`'tl' | 'tr' | 'bl' | 'br'` and already consumed by `LoadingIndicator.corner`.

```typescript
// cmp/grid/Types.ts

/** Specification for a corner flag rendered on a grid cell. */
export interface CellFlagSpec {
    /** Intent, driving the flag's color. */
    intent: Intent;

    /** Corner in which to render the flag. Default 'tr'. */
    corner?: Corner;
}

/** Function to derive a cell flag for a given record. Return null for no flag. */
export type ColumnCellFlagFn = (params: {
    value: any;
    record: StoreRecord;
    column: Column;
    gridModel: GridModel;
}) => Intent | CellFlagSpec | null;
```

```typescript
// cmp/grid/columns/Column.ts - ColumnSpec

/**
 * Render a small triangular flag in a corner of each cell in this column - a compact marker for
 * values that warrant attention without consuming a column or altering the cell's contents.
 *
 * Specify an Intent for the simple case, a CellFlagSpec to also control the corner, or a function
 * to derive either per record. Functions returning null render no flag.
 *
 * At most one flag is rendered per cell - see "Deliberately out of scope" below.
 */
cellFlag?: Intent | CellFlagSpec | ColumnCellFlagFn;
```

The union return keeps the common case a bare string while leaving the object form as the place any
future per-flag option would land.

```typescript
// Static - every cell flagged the same way
{field: 'price', cellFlag: 'warning'}
{field: 'price', cellFlag: {intent: 'warning', corner: 'tl'}}

// Per record
{
    field: 'price',
    cellFlag: ({record}) => record.data.isStale ? 'warning' : null
}

// Per record, varying corner
{
    field: 'price',
    cellFlag: ({record}) => {
        if (record.data.isOverridden) return {intent: 'primary', corner: 'tl'};
        if (record.data.isStale) return {intent: 'warning', corner: 'tr'};
        return null;
    }
}
```

## Implementation

**Emit via `cellClassRules`, not `cellClass`.** This is load-bearing: `cellClass` never *removes* a
class when data changes, as its own JSDoc at `Column.ts:150` warns. A flag built on `cellClass`
would appear correctly once and then stick to the cell permanently.

Resolve in `Column.getAgSpec()` along two paths:

- **Static `cellFlag`** (an `Intent` or `CellFlagSpec`, not a function) - append the two classes to
  the generated `cellClass` function's output. No rules, no per-cell rule evaluation.
- **Function `cellFlag`** - compose eight entries into the ag colDef's `cellClassRules`: four intent
  classes (`xh-cell--flag-danger` and friends) and four corner classes (`xh-cell--flag-tr` and
  friends). Ag-Grid evaluates all rules for a single cell consecutively, so a one-deep memo on the
  resolved spec (comparing `params.data` and `params.value` identity) makes seven of the eight
  evaluations free and keeps the app's function called once per cell per refresh.

**Collapse the SCSS.** `Grid.scss:107-183` currently repeats the size ramp four times, once per size
mode, listing all three classes each time. Replace with a `--xh-grid-cell-flag-size` token set once
per size mode, one base `::before` rule, four corner-position rules, and four intent-color rules.
Each corner sets the two border sides that meet there - `tr` sets top and right, `tl` sets top and
left, and so on. Colors come from the existing `--xh-intent-*` vars, which already carry light and
dark theme values in `styles/vars.scss`, so both themes work without new definitions.

**Re-express the validation flags on top of it.** The editor path in `Column.ts` becomes a caller
rather than a special case, mapping severity to intent - `error` to `danger`, `warning` to
`warning`, `info` to `primary`. That mapping preserves today's colors exactly. Where a column has
both an `editor` and an app `cellFlag`, validation state wins, since it reports an error condition
rather than an annotation.

## Autosize: no calculator changes needed

Worth recording explicitly, since `ColumnWidthCalculator` takes a materially more expensive path for
columns carrying cell class rules - it abandons the single-measurement fast path and instead walks
every record behind each of the 10 sampled values, computing row and cell classes per record, then
runs a hidden-cell layout measurement per unique class combination.

Cell flags are exempt from that cost for two independent reasons:

1. **Architecturally, the calculator never sees these rules.** There are two separate stores.
   `Column.cellClassRules` (`Column.ts:649`) holds only the app's own config, and that is what the
   calculator reads (in `calcLevelWidthAsync()` and `getCellClass()`). The ag colDef's `cellClassRules` is
   seeded from it at `Column.ts:787` and then overwritten at `Column.ts:1061` with the merged
   validation rules. So today's validation flag rules already never reach autosize. `cellFlag`
   inherits that exemption for free by composing into the colDef inside `getAgSpec()` and never
   assigning to `this.cellClassRules`.
2. **On the merits, flags cannot change measured width.** The flag is a `position: absolute`
   pseudo-element, so it is out of flow and contributes nothing to the cell's content width, and no
   padding is added to accommodate it.

The one requirement this places on the implementation is a negative one: do not "helpfully" merge
flag rules into `Column.cellClassRules`, or every flagged column silently takes the slow autosize
path for no measurement benefit.

## Deprecation of the existing classes

`.xh-cell--invalid`, `.xh-cell--warning` and `.xh-cell--info` stay in `Grid.scss` as deprecated
aliases, since an app may be applying them itself today via its own `cellClassRules`. They keep
rendering a top-right flag in the same colors as now.

The deprecation is strictly one-directional: an app that applies an old class still gets a flag, but
**Hoist itself must never emit the old classes again**. The validation path emits only the new
intent-keyed classes. Aliases are comment-marked for removal in a future major.

## Deliberately out of scope

- **Flag tooltips.** No one has asked, and it is not free: the existing validation tooltip override
  at `Column.ts:913` is editor-gated too, so a flag-supplied tooltip means un-gating that override
  and defining precedence between app flag text, `column.tooltip`, and validation messages. The
  `CellFlagSpec` object form leaves room to add it later without a breaking change.
- **Multiple flags per cell.** One flag per cell, by design. Two or more corners flagged at once
  reads as noise rather than signal, and supporting it would mean an array-valued config plus a
  second pseudo-element for marginal benefit.

## Files

| File | Change |
|---|---|
| `core/types/Types.ts` | none - `Intent` and `Corner` already exported |
| `cmp/grid/Types.ts` | add `CellFlagSpec`, `ColumnCellFlagFn` |
| `cmp/grid/columns/Column.ts` | `ColumnSpec.cellFlag`; resolution in `getAgSpec()`; validation path rewritten onto the primitive |
| `cmp/grid/Grid.scss` | intent- and corner-keyed flag rules; size ramp reduced to a token; deprecated aliases |
| `cmp/grid/README.md` | document `cellFlag` alongside `cellClass` / `cellClassRules` |

# Notes

- **Cross-platform.** `cmp/grid` and `Grid.scss` are shared, so this lands on desktop and mobile
  together.
- **Rule fn params under autosize.** Not applicable given the exemption above, but for the record:
  `ColumnWidthCalculator.getCellClass()` invokes rule fns with only `{value, data}` - no `column`
  or `api`. Any rule that does reach it must guard, as the existing group-border rules do.
- **Exports.** Flags are CSS pseudo-elements and will not appear in grid Excel or CSV exports. Apps
  needing the distinction in an export should carry it in a field.
- Toolbox should pick up a demo column once this lands.

amcclain:
Closes #4673.

`cellFlag` renders a small triangle in a grid cell's top-right corner, in the color of a standard Hoist `Intent`. Called per record, returning an `Intent` or null.

Hoist already drew this glyph, but only for itself - the styles were reachable only via the classes installed for cell validation state on editable columns. This promotes that private mechanism to a public column config and re-expresses the validation flags on top of it, leaving one implementation rather than two.

```typescript
{field: 'volume', cellFlag: volume => (volume >= 9_000_000_000 ? 'warning' : null)}
```

### Notes for reviewers

Three decisions are load-bearing and easy to "tidy" into bugs later, so each carries a comment in the code:

- **`cellClassRules`, never `cellClass`** - a `cellClass` is never removed once applied, so a flag built on it would stick permanently.
- **Deliberately uncached** - `validationResults` is populated asynchronously and can change while a record and its value stay identical. A record-keyed memo (which this originally had) serves a stale flag once an async rule settles; it was reproducible in the Toolbox inline-editing demo with async validation on. Holding no record reference also keeps the closure, which lives as long as the ag colDef, from pinning a `StoreRecord`. The function runs once per candidate `Intent` per rendered cell, so it must stay cheap - documented on the config.
- **Composed into the ag colDef, not `Column.cellClassRules`** - the latter is what `ColumnWidthCalculator` reads. Flags are absolutely-positioned pseudo-elements that cannot affect measured width, so this keeps flagged columns off the calculator's class-permutation path. Do not merge them into `Column.cellClassRules`.

Scope was deliberately trimmed during review: no static (non-function) config form, and no configurable corner - both were speculative, and the corner doubled the per-cell call count for no requested use case. Either can be added later without a breaking change.

### Deprecation

`.xh-cell--invalid`, `.xh-cell--warning` and `.xh-cell--info` remain styled, so an app applying them directly still renders a flag. Hoist itself no longer emits them.

### Toolbox

Matching branch `grid-cell-flag` swaps the `SampleGrid` Volume column's hand-rolled `cellClassRules` + SCSS for a `cellFlag`. It cannot be committed until a SNAPSHOT carrying this change is published, since its pre-commit `tsc` resolves `@xh/hoist` from npm.

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. *(Deprecated classes stay styled, so apps applying them keep working.)*
- [x] Reviewed and tested on Mobile, or determined not required. *(`cmp/grid` and `Grid.scss` are shared, so this lands on both platforms.)*
- [x] Created Toolbox branch / PR, or determined not required. *(Branch ready; see above.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,20 @@
 
 ## 88.0.0-SNAPSHOT - unreleased
 
+### 🎁 New Features
+
+* Added `Column.cellFlag`, rendering a small triangular flag in a grid cell's top-right corner in
+  the color of a Hoist `Intent` - a compact marker for values warranting attention. Called per
+  record, returning the `Intent` to draw, or null for no flag.
+
+### ✨ Styles
+
+* Grid cell flag styles are now keyed by `Intent` (`.xh-cell--flag-{intent}`), with size driven by
+  the new `--xh-grid-cell-flag-size` custom property. The classes previously emitted for cell
+  validation state - `.xh-cell--invalid`, `.xh-cell--warning`, and `.xh-cell--info` - are
+  deprecated but still styled, so apps applying them directly continue to render a flag. Retarget
+  any CSS overriding these at the new class names.
+
 
 ## 87.3.0 - 2026-09-10
 
--- a/cmp/grid/Grid.scss
+++ b/cmp/grid/Grid.scss
@@ -104,40 +104,63 @@
     padding-right: 0;
   }
 
-  // Render badge on cells with validation issues
-  .ag-cell.xh-cell--invalid,
-  .ag-cell.xh-cell--warning,
+  // Cell corner flags - small triangular markers in a cell's top-right corner, colored by Intent.
+  // Driven by `Column.cellFlag`, and by the framework to surface cell validation state.
+  .ag-cell.xh-cell--flag-primary {
+    --xh-grid-cell-flag-color: var(--xh-intent-primary);
+  }
+
+  .ag-cell.xh-cell--flag-success {
+    --xh-grid-cell-flag-color: var(--xh-intent-success);
+  }
+
+  .ag-cell.xh-cell--flag-warning {
+    --xh-grid-cell-flag-color: var(--xh-intent-warning);
+  }
+
+  .ag-cell.xh-cell--flag-danger {
+    --xh-grid-cell-flag-color: var(--xh-intent-danger);
+  }
+
+  // DEPRECATED - legacy validation flag classes, superseded by the intent-keyed classes above.
+  // Hoist no longer emits these; they are retained only so that an app applying them directly
+  // (e.g. via its own `cellClassRules`) continues to render a flag. Remove in a future major.
   .ag-cell.xh-cell--info {
+    --xh-grid-cell-flag-color: var(--xh-intent-primary);
+  }
+
+  .ag-cell.xh-cell--warning {
+    --xh-grid-cell-flag-color: var(--xh-intent-warning);
+  }
+
+  .ag-cell.xh-cell--invalid {
+    --xh-grid-cell-flag-color: var(--xh-intent-danger);
+  }
+
+  .ag-cell[class*='xh-cell--flag-'],
+  .ag-cell.xh-cell--info,
+  .ag-cell.xh-cell--warning,
+  .ag-cell.xh-cell--invalid {
+    // Drawn as a triangle via two colored borders meeting at the cell's top-right corner.
     &::before {
       content: '';
       position: absolute;
       top: 0;
       right: 0;
       border-color: transparent;
+      border-top-color: var(--xh-grid-cell-flag-color, transparent);
+      border-right-color: var(--xh-grid-cell-flag-color, transparent);
       border-style: solid;
+      // Sized per grid sizing mode - see the size-mode blocks below.
+      border-width: var(--xh-grid-cell-flag-size, 5px);
     }
 
-    // Render badge on top of nested editor, and move to account for padding while editing
+    // Render flag on top of a nested editor, nudged to clear the editor's padding.
     &.ag-cell-inline-editing::before {
       z-index: 1;
       margin-top: 1px;
       margin-right: 3px;
     }
-  }
-
-  .ag-cell.xh-cell--invalid::before {
-    border-right-color: var(--xh-intent-danger);
-    border-top-color: var(--xh-intent-danger);
-  }
-
-  .ag-cell.xh-cell--warning::before {
-    border-right-color: var(--xh-intent-warning);
-    border-top-color: var(--xh-intent-warning);
-  }
-
-  .ag-cell.xh-cell--info::before {
-    border-right-color: var(--xh-intent-primary);
-    border-top-color: var(--xh-intent-primary);
   }
 
   // Render left / right group borders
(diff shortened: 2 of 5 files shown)


amcclain:
Toolbox demo: xh/toolbox#903 (draft - blocked on a SNAPSHOT carrying this).

lbwexler (approved):
Approved.  Will merge after putting the comments/tsdoc on a diet!

PR #4686, opened 2026-09-11
Split grid tooltip styling from tooltip state

amcclain:
**Stacked on #4684** - please merge that first. Base is `grid-cell-flag`, so this diff shows only the tooltip work.

`.xh-grid-tooltip--default` carried two unrelated things at once: the *state* "this tooltip's content was a string" and the *appearance* "draw a frame and pre-wrap". That conflation is the root of several problems this PR fixes.

### Appearance now composes

Applied by Hoist to the tooltip content it renders itself:

- **`xh-grid-tooltip-frame`** - background, border, radius, padding, max-width. A standalone **utility** class, not an element or modifier, because apps apply it to their own custom tooltip root where no `.xh-grid-tooltip` ancestor exists (header tooltips get `ag-tooltip` only). This is the one app-facing class.
- **`xh-grid-tooltip--prewrap`** - honors `\n` breaks while wrapping at max-width. An ordinary **modifier**: Hoist only ever applies it to the wrapper, which always carries the block.

Previously an app wanting a custom tooltip to match had to copy the five `--xh-grid-tooltip-*` tokens, or co-opt `--default` and assert something untrue about its content.

### Bug fix: validation tooltips

`isElement` was computed from a column's own tooltip - but validation messages *replace* that tooltip entirely, so the wrapper was labelled from content that was never rendered. A column with both an `editor` and a custom element `tooltip` therefore got an unframed wrapper, leaving the validation list to rebuild the frame via a parallel `--custom > div >` rule that supplied background and border but **neither border-radius nor max-width**. Two columns with identical validation errors rendered differently based on unrelated config.

Validation state is now resolved *before* the classes are assigned, so validation always uses the standard frame and the two `> div >` rules collapse into one.

### Class changes

- `--default` / `--custom` **removed**. Once styling moved out nothing read them - no rule in hoist-react, no usage across our sample apps - and a class that looks meaningful but is never consumed invites exactly the co-opting this PR ends. `.xh-grid-tooltip` remains on every grid tooltip as the stable hook.
- `--validation` **moved to the tooltip** from the message list inside it, making it a true modifier; `--validation--single` → `--validation-single`. The list now carries no class.

Net result is six flat rules with no structural `> div >` selectors, and every class correctly categorised as block, modifier or utility.

### Testing

Verified in Toolbox against a local hoist-react: string tooltips framed with pre-wrap; custom element tooltips unstyled (`bg: transparent`, `border: 0px`); validation tooltips framed with `radius: 4px` / `max-width: 400px` in both the string-tooltip and custom-tooltip cases - the latter being the bug above. Final `--validation` placement verified by @amcclain.

A matching Toolbox branch demos `cellFlag` plus a framed custom tooltip; it will go up once both this and #4684 are merged and a SNAPSHOT is published.

Hoist P/R Checklist
-------------------

- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. *(Class removals/renames called out with ⚠️ sub-bullets; no app in our sample referenced any of them.)*
- [x] Updated doc comments / prop-types, or determined not required. *(New "Tooltips" section in the grid README.)*
- [x] Reviewed and tested on Mobile, or determined not required. *(`cmp/grid` and `Grid.scss` are shared.)*
- [x] Created Toolbox branch / PR, or determined not required. *(Branch ready; see above.)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,20 @@
 
 ### ✨ Styles
 
+* Added `.xh-grid-tooltip-frame`, a standalone utility class carrying Hoist's standard tooltip
+  chrome - background, border, radius, padding and max-width. Hoist applies it to the tooltip
+  content it renders itself, and apps can add it to a custom (element) tooltip's own root to match.
+  Line-break handling moves alongside it to a `.xh-grid-tooltip--prewrap` modifier.
+    * ⚠️The `.xh-grid-tooltip--default` and `--custom` classes have been removed. They carried
+      the styling that now lives in the utility classes above, and nothing consumed them once it
+      moved out. Apps with CSS targeting either should retarget `.xh-grid-tooltip`, still applied
+      to every grid tooltip, or the new utility classes.
+* Fixed validation tooltips on an editable column rendering without rounded corners or a max-width
+  when that column also defined a custom (element) `tooltip`. Validation messages now always use
+  the standard frame, since they supersede the column's own tooltip entirely.
+    * ⚠️`.xh-grid-tooltip--validation` now sits on the tooltip itself rather than the message
+      list inside it, making it a true modifier of `.xh-grid-tooltip`, and `--validation--single`
+      is renamed `--validation-single` to match. The list carries no class of its own.
 * Grid cell flag styles are now keyed by `Intent` (`.xh-cell--flag-{intent}`), with size driven by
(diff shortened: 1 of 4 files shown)


amcclain:
Toolbox demo: xh/toolbox#903 (draft - blocked on a SNAPSHOT carrying this). It uses `xh-grid-tooltip-frame` as the reference example for the class added here.

PR #4696, opened 2026-09-16
Menu tune-up: headings, Menu and MenuButton, closed token union, menu fixes

amcclain:
Adds menu headings, gives desktop a public menu component and menu button, closes a type hole in the grid context menu API, and fixes a handful of menu behaviors on both platforms.

Demos and app-side usage in xh/toolbox#906 and xh/toolbox#907.

### Headings

- `MenuHeading` labels and groups the items below it - e.g. `{heading: 'This Row'}`. Every menu that takes a `GridContextMenuItemLike` or `MenuItemLike` accepts it: grid context menus, desktop menus and context menus, and mobile menus.
- A heading draws its own rule, so it needs no adjacent `'-'`. Hoist drops a heading with no items below it, which matters most in grid context menus, where items hide themselves based on the clicked row.
- A heading's `displayFn` receives the same context as the items beside it - `ActionFnData` in a grid, the chart or tab context in those menus - so it can name the clicked record or hide itself.
- Four `--xh-menu-heading-*` properties style all three renderers together.

### Menu components

- `Menu` renders a menu from `MenuItem` configs, `'-'` tokens, and headings. `MenuButton` adds a trigger button and takes `ButtonProps` directly. Both replace the popover and Blueprint menu that apps assembled by hand. `ContextMenu` shares the same parsing.
- `MenuButton` accepts `menuItems` as a function of the menu context, and `context` as a function. Both are called on the click that opens the menu, mirroring how `useContextMenu` materializes a `ContextMenuSpec`. Desktop and mobile alike; the mobile menu now passes context to `prepareFn`, `actionFn`, and heading `displayFn`.
- `MenuItem` gained `active`, to mark the current selection. Honored on desktop and mobile.
- `clipboardMenuItem()` returns a `MenuItem` config rather than a component, so it works in any menu on either platform. It models the pattern for reusable items: a function returning a config. It lives in a shared `cmp/clipboard` package; the desktop import path still works.

### Types

- `GridContextMenuItemLike` no longer accepts an open `string`, which had collapsed the union and left `GridContextMenuToken` with no completions or typo-checking. It now takes actions, headings, the Hoist tokens, and ag-Grid's `DefaultMenuItem` tokens.
- `RecordActionLike` is `RecordAction | RecordActionSpec`. Menu entries that may also be a separator, heading, or token - `RecordAction.items`, `RestGridConfig.menuActions` - are typed `GridContextMenuItemLike`.

### Fixes

- Desktop submenus no longer close the instant the pointer leaves the parent item. Blueprint hardcodes a zero delay; Hoist restores its 300ms default, tunable via `Menu.defaults.submenuHoverCloseDelay`. Grid context menus already behave this way under ag-Grid.
- `clipboardMenuItem()` aligns with the items around it. It rendered a styled `ClipboardButton` before.
- Right-click outside an open `ContextMenu` no longer shows the browser menu. Right-click on a `MenuButton` or its open menu no longer falls through to a context menu beneath.
- Mobile menus render `'-'` as a divider rather than a literal hyphen.

### Review notes

- `ContextMenu` keeps `usePortal: true` for its submenus, which fixes submenu position for a menu anchored at the cursor (#3724). It is now an explicit option rather than a difference hidden in duplicated code.
- `MenuButton` covers the desktop menu button that #2400 describes. The trigger, placement, and styling questions in that issue remain open.
- The shared menu heading helpers live in `cmp/menu/impl`, so `utils` does not depend on `core`. `Separators` stays in `utils` for the toolbars.


cmp/grid/GridContextMenu.ts
```diff
@@ -4,19 +4,25 @@
...
+ *
+ * A string is a token. Use either a Hoist {@link GridContextMenuToken}, which includes '-' for a
+ * separator, or an ag-Grid `DefaultMenuItem` token.
  * {@link https://www.ag-grid.com/javascript-grid-context-menu/#built-in-menu-items}
+ *
+ * Note that `RecordActionLike` also accepts a {@link MenuHeading} - i.e. `{heading: 'Section'}` -
+ * which labels and groups the items below it.
  */
-export type GridContextMenuItemLike = RecordActionLike | GridContextMenuToken | string;
+export type GridContextMenuItemLike = RecordActionLike | GridContextMenuToken | DefaultMenuItem;
```
lbwexler: Nice pickup

core/types/Interfaces.ts
```diff
@@ -332,21 +332,60 @@ export interface MenuItem<T = MenuToken, C = MenuContext> {
     /** True to disable this item. */
     disabled?: boolean;
 
+    /** True to render this item as active - e.g. to mark the current selection. */
+    active?: boolean;
```
lbwexler, line 336: This confused me a bit ---  I now see it  applies a highlight, and I see its the same name as the underlying blueprint property and distinct from e.g. 'selected' (which blueprint also offers).

We could call it 'highlight', but not sure that's better and it just introduces more indirection.  Just something to think about.

desktop/cmp/rest/RestGridModel.ts
```diff
@@ -118,7 +118,7 @@ export class RestGridModel extends HoistModel {
     readonly: boolean;
     editors: RestGridEditor[];
     toolbarActions: Array<RecordAction | RecordActionSpec>;
```
lbwexler: Based on these three lines, it seems more correct that RecordActionLike should be RecordAction|RecordActionSpec, and the '-' should be kept out of it.

lbwexler (changes requested):
Have reviewed everything here and made a few initial changes.  More substantial changes (but not huge) are in stacked PR #4698

desktop/cmp/rest/RestGridModel.ts
lbwexler: took care of this in stacked PR

amcclain:
Thanks happy with whatever else you think we need to merge - appreciate it!

